PR #10, opened 2024-07-28
Perlmutter profiling

IamShubhamGupto:
profiling using nsight


.python-perlmutter
suryadheeshjith, line 1: * Need to change LOGDIR. 
* Why are we using two files i.e .python-perlmutter and .python-perlmutter-launch. Lets just use one.

configs/compute/greene/node/a100_1hrs.yaml
suryadheeshjith, line 1: call it a100_45mins

configs/conf.yaml
suryadheeshjith, line 1: Need to figure out separation of our conf files

scripts/train.sh
suryadheeshjith, line 1: Create a new .sh file for profiling. Do not touch train and local_train.sh

src/train_3D.py
suryadheeshjith, line 1: Can we modularize the code for profiling out of the train_3D.py file? If not, just create some functions for these lines of code. Also, it would be good if the profiling code runs only if a profiling flag is set in the configuration.

submitit_hydra.py
suryadheeshjith, line 1: I want to test a few things with this code. What are the steps for me to run this code with profiling (other than path changes)? Would this work - ./.python-perlmutter submitit_hydra.py compute/greene=2x2 compute/greene/node=a100_3hrs wandb.mode=online exp=train_unet_global_3D_all epochs=4 name="$(date +%F)-train_convnextunet_global_3D_hist1_5levels" region=global_3D batch_size=4 scheduler=True rand_seed=10 unet.ch_width=[80,100,150,300,400] hist=1 --qos=preempt

Do I need to set the ENABLE_NSYS_PROFILING flag somewhere?

suryadheeshjith (changes requested):
Ive added comments to the code

configs/compute/greene/node/a100_1hrs.yaml
IamShubhamGupto: This file will be deleted, debug max is only 30mins

.python-perlmutter
IamShubhamGupto: what changes you want in LOGDIR? want to send to m4331 project dir? 
`.python-perlmutter-launch` is used by submit_hydra to begin the experiment, I don't want profiling code to capture any of that

having a separate `.python-perlmutter` lets us launch nsys and cprofile directly on Trainer. Letting nsys observe submit_hydra launch Trainer does not seem to profile, same with cprofile. 

I can run another experiment on preempt to confirm this

configs/conf.yaml
IamShubhamGupto: I can add it to gitignore, we use discord to observe changes in conf.yaml

suryadheeshjith: I got a fix. I will push it

.python-perlmutter
suryadheeshjith: No worries, lets keep both files and rename them. I want the all .sh fles to run .python-perlmutter so either we could exchange the file names or have something else like .python-perlmutter-worker or something.

IamShubhamGupto: with the latest commit, all .sh files can use either `.python-perlmutter-launch` or `.python-perlmutter`. For profiling, all .sh files must strictly use `.python-perlmutter-launch` for nsight to not break

we can rename it as long as `submitit_hydra.py` is calling the one with the Nsight modifications

suryadheeshjith: Yes, we will do that.

.python-perlmutter-worker
```diff
@@ -0,0 +1,19 @@
...
+LOGDIR=${EXPT_DIR}/profile-logs
+mkdir -p ${LOGDIR}
+
+module load conda
+conda activate emulator
+
+# Profiling
+if [ "${ENABLE_NSYS_PROFILING:-0}" -eq 1 ]; then
+    module load cudatoolkit
+    NSYS_ARGS="profile --trace=cuda,nvtx,osrt,cudnn,cublas --backtrace=dwarf --gpu-metrics-device=all"
```
IamShubhamGupto: Instead of `all`, set this to 0 for profiling on a single gpu

suryadheeshjith: Will this disable profiling on all gpus instead? We want it to profile on all gpus but just prevent all of them to create an nsys file.

Issue #21, opened 2024-08-20
[Major Refactoring Required] Tests


suryadheeshjith:
- [x] Running 'pytest' does not work
- [x] Add github workflow
- [x] Prevent tests like weight tests to run on github
- [x]  Additional tests using pytests

alxmrs:
Mind if I start working on this issue? My guess is that a blocker may be either a) getting the code to run on CPU or b) running a GPU in CI. 

@jder has just figured out the ops related to doing (b), for example.

suryadheeshjith:
Yes please!

alxmrs:
I'm going to mark this issue as fixed via #64. We have tests set up. For the near future, we'll add more tests (like #71 and #73).

PR #55, opened 2025-02-11
Improving project test setup.

alxmrs:
This change runs tests inside CI via GitHub Actions. In addition, it introduces a test marking scheme to filter out what can run in CI and locally. Two pytest markers have been added: `cuda` and `manual`. Manual tests, like the model weights test, and GPU-powered tests have been excluded in the default test Action. Addresses https://github.com/suryadheeshjith/Ocean_Emulator/issues/21.

I've added the beginnings of a contributors guide to document how tests should be run.

In the (near) future, I plan to add significantly more tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import pytest
+import torch
 
 
 def pytest_addoption(parser):
@@ -18,3 +19,9 @@ def model1_path(request):
 @pytest.fixture
 def model2_path(request):
     return request.config.getoption("--model2")
+
+
+# Used to automatically filter out CPU- or GPU- only tests.
+@pytest.fixture(params=["cpu", pytest.param("cuda", marks=pytest.mark.cuda)])
+def device(request):
+    return torch.device(request.param)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,7 +21,7 @@ def model2_path(request):
     return request.config.getoption("--model2")
 
 
-# Used to automatically filter out CPU- or GPU- only tests.
+# Run a test for both CPU and GPU, and allows selecting or skipping CUDA tests.
 @pytest.fixture(params=["cpu", pytest.param("cuda", marks=pytest.mark.cuda)])
 def device(request):
     return torch.device(request.param)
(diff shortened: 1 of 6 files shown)


suryadheeshjith (approved):
LGTM

tests/test_placeholder.py
```diff
@@ -0,0 +1,7 @@
+# TODO(#21): Replace with real project tests.
+import torch
+
+
+def test_placeholder(device):
+    """A placeholder test that should fail when run on CPU."""
+    assert device != torch.device("cuda:0")
```
jder, line 7: Is this test (or comment?) inverted? Wouldn't this fail on GPU, not CPU?

tests/test_weights.py
```diff
@@ -18,6 +19,7 @@ def compare_model_weights(model1_state_dict, model2_state_dict):
     return are_equal
 
 
+@pytest.mark.manual
 def test_models_have_same_weights(model1_path, model2_path):
```
jder, line 25: Huh, I have never seen pytest used this way where it takes options from the command line. Maybe it would be good to document how to call this in your readme example? `pytest -k "models_have_same_weights"` will presumably fail because of missing args unless I'm missing something.

tests/conftest.py
```diff
@@ -18,3 +19,9 @@ def model1_path(request):
 @pytest.fixture
 def model2_path(request):
     return request.config.getoption("--model2")
+
+
+# Used to automatically filter out CPU- or GPU- only tests.
```
jder:
```suggestion
# Run a test for both CPU and GPU, and allows selecting or skipping CUDA tests.
```

jder (approved):
Thanks! Just a couple questions from me. I don't think I've seen such fancy pytesting before; looks very nice!

tests/test_weights.py
suryadheeshjith: To be honest, we could remove this test, since we had a very specific usecase before.

tests/test_placeholder.py
alxmrs: Haha, you're right! The comment is the opposite of true. I'll correct it!

tests/test_weights.py
alxmrs: I'll keep it for now. Who knows? Maybe it will be relevant later. Or, this kind of "spot check" will be relevant later.

PR #56, opened 2025-02-11
Standalone inference

suryadheeshjith:
Implementation of standalone inference and windowed inference / wandb logging


src/eval.py
```diff
@@ -0,0 +1,256 @@
...
+from utils.model import get_model_summary
+from utils.wandb import WandBLogger
+
+
+class Eval:
+    def __init__(self, cfg) -> None:
+        self.device = get_device()
+
+        # Adjust workers and memory pinning based on device
+        if not using_gpu():
```
alxmrs, line 29: I appreciate the care taken in this project to ensure that everything works on CPU.

PR #62, opened 2025-02-13
Specifying minimum version of cftime.

alxmrs:
ARM support was only added in cftime==1.5.2. Thus, I'm updating the requirements.txt such that 1.5.2 is the minimum allowed version. This allows me to open Zarr datasets with Xarray on M1+ Macs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,7 +31,7 @@ certifi==2024.2.2
 certipy==0.1.3
 cf_xarray==0.9.5
 cffi==1.16.0
-cftime==1.5.1.1
+cftime>=1.5.2
 charset-normalizer==3.3.2
 click==8.1.7
 click-default-group==1.2.4


jder (approved):
Seems like a pretty safe release.

jder:
Should we make this an == constraint like everything else in this file?

alxmrs:
> Should we make this an == constraint like everything else in this file?

That's a good question. I personally think that we should allow for a wider range of versions and then check in a `uv.lock` file. However, I didn't want to make such an extreme change before meeting face-to-face.

suryadheeshjith:
[Reference](https://github.com/Unidata/cftime/blob/master/Changelog#L45-L50) to the changelog just in case

suryadheeshjith (approved):

jder:
We can discuss live about the goal, I just think that as committed this change will actually upgrade to 1.6.4 which is more risky (but probably still fine)

PR #64, opened 2025-02-14
Created test-bench for data loaders.

alxmrs:
This PR creates an in-memory dataset and test config to test the shape of training and validation data loaders. These tests abstract away the details of the data loader setup. Thus, the tests should work even if the codebase is refactored.

### DIFF
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -36,7 +36,7 @@ distributed:
 
 experiment:
   name: train
-  sub_name: local_cm4_samudra
+  sub_name: test_cm4_samudra
   rand_seed: 15
   base_output_dir: .LOCAL
   gantry: false
@@ -44,15 +44,15 @@ experiment:
   wandb:
     mode: disabled
     project: 3D_ocean_emu_CM4
-    entity: suryadheeshjith
+    entity: testrobot
   network: convnextunet
   exp_num_in: '3D_5'
   exp_num_extra: '3D_5'
   exp_num_out: '3D_5'
 data:
-  data_path: cm4_piControl_ocean_200yr_full.zarr
-  data_means_path: 2024-11-12-cm4-piControl-ocean-200yr-dataset-stats/centering.nc
-  data_stds_path: 2024-11-12-cm4-piControl-ocean-200yr-dataset-stats/scaling-full-field.nc
+  data_path: fake/data/path/cm4_piControl_ocean_200yr_full.zarr
+  data_means_path: fake/data/path/cm4-piControl-ocean-200yr-dataset-stats/centering.nc
+  data_stds_path: fake/data/path/piControl-ocean-200yr-dataset-stats/scaling-full-field.nc
   scaling_residuals_file: null
   depth_mode: all
   time_delta: 5
@@ -61,10 +61,10 @@ data:
 
 
 unet:
-  ch_width: [157, 200, 250, 300, 400]
-  n_out: 77
-  dilation: [1, 2, 4, 8]
-  n_layers: [1, 1, 1, 1]
+  ch_width: [2, 3]
+  n_out: 4
+  dilation: [1, 2]
+  n_layers: [1, 1]
   pred_residuals: false
   last_kernel_size: 3
   pad: "circular"
@@ -73,7 +73,7 @@ unet:
     block_type: "conv_next_block"
     kernel_size: 3
     activation: "capped_gelu"
-    upscale_factor: 4
+    upscale_factor: 2
     norm: "batch"
 
   corrector:
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -15,13 +15,13 @@ inference_epochs: [-1]
 data_percent: 1.0
 train:
   start_time: '1969-08-15'
-  end_time: '1969-08-20'
+  end_time: '1969-10-20'
 val:
-  start_time: '1969-09-04'
-  end_time: '1969-09-09'
+  start_time: '1969-10-20'
+  end_time: '1969-11-20'
 inference:
-  - start_time: '1969-09-06'
-    end_time: '1969-09-14'
+  - start_time: '1969-11-20'
+    end_time: '1969-12-15'
 data_stride: [1]
 steps: [1]
 step_transition: []
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -1,0 +1,84 @@
+# A config file to test training.
+# Used in automated tests.
+debug: true
+disk_mode: true
+pin_mem: false
+save_freq: 5
+epochs: 120
+batch_size: 2 # 4
+learning_rate: 0.0001
+scheduler: false
+loss: mse
+finetune: false
+resume_ckpt_path: null
+inference_epochs: [-1]
+data_percent: 1.0
+train:
+  start_time: '1969-08-15'
+  end_time: '1969-08-30'
+val:
+  start_time: '1969-09-01'
+  end_time: '1969-09-05'
+inference:
+  - start_time: '1969-09-06'
+    end_time: '1969-09-12'
+data_stride: [1]
+steps: [4]
+step_transition: []
+
+distributed:
+  enabled: true
+  dist_url: null
+  world_size: null
+  rank: null
+  gpu: null
+  dist_backend: null
+
+experiment:
+  name: train
+  sub_name: local_cm4_samudra
+  rand_seed: 15
+  base_output_dir: .LOCAL
+  gantry: false
+  cluster_data_dir: will_be_replaced
+  wandb:
+    mode: disabled
+    project: 3D_ocean_emu_CM4
+    entity: suryadheeshjith
+  network: convnextunet
+  exp_num_in: '3D_5'
+  exp_num_extra: '3D_5'
+  exp_num_out: '3D_5'
+data:
+  data_path: cm4_piControl_ocean_200yr_full.zarr
+  data_means_path: 2024-11-12-cm4-piControl-ocean-200yr-dataset-stats/centering.nc
+  data_stds_path: 2024-11-12-cm4-piControl-ocean-200yr-dataset-stats/scaling-full-field.nc
+  scaling_residuals_file: null
+  depth_mode: all
+  time_delta: 5
+  num_workers: 4
+  hist: 0
+
+
+unet:
+  ch_width: [157, 200, 250, 300, 400]
+  n_out: 77
+  dilation: [1, 2, 4, 8]
+  n_layers: [1, 1, 1, 1]
+  pred_residuals: false
+  last_kernel_size: 3
+  pad: "circular"
+
+  core_block:
+    block_type: "conv_next_block"
+    kernel_size: 3
+    activation: "capped_gelu"
+    upscale_factor: 4
+    norm: "batch"
+
+  corrector:
+    non_negative_corrector_names:
+      - "so"
+
+  down_sampling_block: "avg_pool"
+  up_sampling_block: "bilinear_upsample"
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -5,7 +5,7 @@ disk_mode: true
 pin_mem: false
 save_freq: 5
 epochs: 120
-batch_size: 2 # 4
+batch_size: 1
 learning_rate: 0.0001
 scheduler: false
 loss: mse
@@ -15,15 +15,15 @@ inference_epochs: [-1]
 data_percent: 1.0
 train:
   start_time: '1969-08-15'
-  end_time: '1969-08-30'
+  end_time: '1969-08-20'
 val:
-  start_time: '1969-09-01'
-  end_time: '1969-09-05'
+  start_time: '1969-09-04'
+  end_time: '1969-09-09'
 inference:
   - start_time: '1969-09-06'
-    end_time: '1969-09-12'
+    end_time: '1969-09-14'
 data_stride: [1]
-steps: [4]
+steps: [1]
 step_transition: []
 
 distributed:
@@ -50,9 +50,9 @@ experiment:
   exp_num_extra: '3D_5'
   exp_num_out: '3D_5'
 data:
-  data_path: fake/data/path/cm4_piControl_ocean_200yr_full.zarr
-  data_means_path: fake/data/path/cm4-piControl-ocean-200yr-dataset-stats/centering.nc
-  data_stds_path: fake/data/path/piControl-ocean-200yr-dataset-stats/scaling-full-field.nc
+  data_path: fake/path/to/data
+  data_means_path: fake/path/to/means
+  data_stds_path: fake/path/to/stds
   scaling_residuals_file: null
   depth_mode: all
   time_delta: 5
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,8 +11,6 @@ import xarray as xr
 import constants as c
 from config import TrainConfig
 
-PROJ_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), os.pardir)
-
 
 class DataSource(TypedDict):
     """In-memory `xarray.Dataset`s needed for tests."""
@@ -104,7 +102,9 @@ def data_source() -> DataSource:
 
 # TODO(alxmrs): Consider yielding multiple test configs.
 @pytest.fixture(scope="session")
-def train_config(data_source: DataSource) -> Generator[TrainConfig, Any, None]:
+def train_config(
+    data_source: DataSource, pytestconfig: pytest.Config
+) -> Generator[TrainConfig, Any, None]:
     with tempfile.TemporaryDirectory() as tmpdir:
 
         def _make_path(name_with_ext: str) -> str:
@@ -116,7 +116,7 @@ def train_config(data_source: DataSource) -> Generator[TrainConfig, Any, None]:
         data_source["stds"].to_netcdf(_make_path("stds.netcdf"))
 
         # Open default training script; modify it so it uses the temporary directory.
-        default_config = os.path.join(PROJ_ROOT, "configs", "train_cm4.test.yaml")
+        default_config = pytestconfig.rootpath / "configs" / "train_cm4.test.yaml"
         trainer = TrainConfig.from_yaml(default_config)
         data_config = dataclasses.replace(
             trainer.data,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -49,7 +49,7 @@ def device(request):
 def data_source() -> DataSource:
     """Returns in-memory `xarray.Dataset`s for tests."""
     summer_of_love = xr.cftime_range(
-        "1969-08-10", "1969-10-01", freq="D", calendar="noleap"
+        "1969-08-10", "1969-12-31", freq="5D", calendar="noleap"
     )
 
     coords = {
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,25 @@
+import dataclasses
+import os
+import tempfile
+from typing import Any, Generator, TypedDict
+
+import numpy as np
 import pytest
 import torch
+import xarray as xr
+
+import constants as c
+from config import TrainConfig
+
+PROJ_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), os.pardir)
+
+
+class DataSource(TypedDict):
+    """In-memory `xarray.Dataset`s needed for tests."""
+
+    data: xr.Dataset
+    means: xr.Dataset
+    stds: xr.Dataset
 
 
 def pytest_addoption(parser):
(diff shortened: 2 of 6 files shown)


suryadheeshjith:
Maybe it is easier to start with fake/mock data? This would probably mean we would have to directly interact with the Data classes rather than Trainer.

tests/conftest.py
```diff
@@ -1,5 +1,28 @@
...
 import torch
+import xarray as xr
+
+from config import TrainConfig
+
+DATA_ROOT = "https://nyu1.osn.mghpcc.org/m2lines-pubs/Samudra/"
+PROJ_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), os.pardir)
+
+# Tunable test data parameter.
+# Make sure to delete cached data if you change this number!
```
jder: Can we include this number in the cache path or something to prevent getting this wrong?

tests/conftest.py
```diff
@@ -25,3 +48,68 @@ def model2_path(request):
...
+
+        # When possible, get a small slice of the data for tests.
+        if "time" in data.dims and len(data.time) >= CACHE_SIZE__TIME:
+            data = data.isel(time=slice(CACHE_SIZE__TIME))
+
+        out[name] = data
+
+        # Write a local cache for future test runs.
+        if not cache_exists:
+            data.to_zarr(os.path.join(PROJ_ROOT, src))
```
jder: If we're thinking about CI, maybe worth putting this in the platform-supplied cache dir via https://pypi.org/project/platformdirs/ or something?

tests/conftest.py
```diff
@@ -25,3 +48,68 @@ def model2_path(request):
...
+    return out
+
+
+# TODO(alxmrs): Consider yielding multiple test configs.
+@pytest.fixture
+def train_config(test_data: DataSource) -> Generator[TrainConfig, Any, None]:
+    with tempfile.TemporaryDirectory() as tmpdir:
+
+        def _make_path(name_with_ext: str) -> str:
+            return os.path.join(tmpdir, name_with_ext)
```
jder, line 105: 🐑 If you made tmpdir a `pathlib.Path` the below can be eg `to_zarr(tmpdir / "data.zarr")`

tests/conftest.py
```diff
@@ -25,3 +48,68 @@ def model2_path(request):
...
+            cluster_data_dir=os.path.join(tmpdir, "cluster_data"),
+        )
+        test_data_trainer = dataclasses.replace(
+            trainer,
+            data=data_config,
+            experiment=experiment_config,
+        )
+
+        # After contextmanager closes, all test data will be automatically cleaned up.
+        yield test_data_trainer
```
jder, line 132: TIL fixtures can be generators

jder:
This generally makes sense to me as a starting point. Given the Trainers do some manipulation of the config etc before creating the dataloaders, I think an integration-style test like this is a good start. 

@suryadheeshjith what are you thinking of in terms of "easier"? One option might be to check in (via git-LFS) a small data sample rather than downloading them dynamically? Basically run your code once and check in the result, then have the config directly refer to those files (when run from the project dir)? Or is part of the goal to test the over-network-loading behavior?

suryadheeshjith:
I thought it would be easier to start with a unit test? Assuming we do change train_3D.py, these tests would break? Starting with a unit test would also allow us to use dummy data, rather than actual data from a small data sample / data over the network. We would also not require any config files.

```
def create_test_dataset():
    coords = {
        'lon': xr.DataArray(np.arange(0.5, 360, 1), dims=['x']),
        'lat': xr.DataArray(np.arange(-89.24, 90, 1), dims=['y']),
        'time': xr.DataArray(range(10), dims=['time']),
        'lev': xr.DataArray(
            [2.5, 10.0, 22.5, 40.0, 65.0, 105.0, 165.0, 250.0, 375.0,
             550.0, 775.0, 1050.0, 1400.0, 1850.0, 2400.0, 3100.0,
             4000.0, 5000.0, 6000.0],
            dims=['lev']
        ),
    }
    
    ds = xr.Dataset(
        {
            # 2D variables
            var: xr.DataArray(
                np.random.random([10, 180, 360]),
                dims=['time', 'y', 'x']
            )
            for var in ['hfds', 'tauuo', 'tauvo', 'zos']
        } | {
            # 3D variables
            var: xr.DataArray(
                np.random.random([10, 19, 180, 360]),
                dims=['time', 'lev', 'y', 'x']
            )
            for var in ['so', 'thetao', 'uo', 'vo']
        },
        coords=coords
    )
    
    return ds
```

suryadheeshjith:
I do think we should definitely have separate tests for all of the above!

tests/conftest.py
```diff
@@ -25,3 +45,108 @@ def model2_path(request):
...
+    coords = {
+        "lon": xr.DataArray(np.arange(0.5, 360, 1), dims=["lon"]),
+        "lat": xr.DataArray(np.arange(-89.24, 90, 1), dims=["lat"]),
+        "time": xr.DataArray(summer_of_love, dims=["time"]),
+        "lev": xr.DataArray(c.DEPTH_LEVELS, dims=["lev"]),
+    }
+
+    normal = np.random.normal(size=(len(coords["lat"]), len(coords["lon"])))
+
+    ds = xr.Dataset(
```
eric-czech: No strong opinion, but you could get rid of `_unpack_level_dimension` and use something like this instead:

```python
ds = xr.Dataset(
    # 2D variables
    {var: xr.DataArray(np.random.random([(time := 10), (y := 180), (x := 360)]), dims=["time", "y", "x"]) 
     for var in ["hfds", "tauuo", "tauvo", "zos"]}
    
    # 3D variables
    | {f"{var}_{lev}": xr.DataArray(np.random.random([time, y, x]), dims=["time", "y", "x"]) 
       for var in ["so", "thetao", "uo", "vo"] 
       for lev in range(19)}
    
    # Masks
    | {f"mask_{lev}": xr.DataArray(np.where(np.random.normal(size=(y, x)) > 0.5**lev, 1, 0), dims=["y", "x"]) 
       for lev in range(19)},
    
    coords=coords
)
```

Also, does the `lev` coord still exist when running this code as-is?  There shouldn't be any data variables with that dimension left after `_unpack_level_dimension`, so I'm wondering what happens to it.

alxmrs: That use of the walrus operator is pretty slick!

alxmrs: Yeah, baking in the lev now is the right way to go. My original thoughts on the two-step process was that I want to eventually get rid of the name-mangled data_vars soon. However, YAGNI, and I can change it later if I do.

That's a good insight; I just checked the debugger and the `lev` dimension does indeed go away.

tests/conftest.py
```diff
@@ -1,5 +1,25 @@
...
+
+import numpy as np
 import pytest
 import torch
+import xarray as xr
+
+import constants as c
+from config import TrainConfig
+
+PROJ_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), os.pardir)
```
eric-czech: You could use this pytest machinery instead: https://docs.pytest.org/en/stable/reference/customize.html#finding-the-rootdir

tests/test_datasets.py
```diff
@@ -0,0 +1,93 @@
...
+    return cfg, trainer.train_loader
+
+
+@pytest.fixture
+def val_loader_pair(trainer_pair):
+    cfg, trainer = trainer_pair
+    return cfg, trainer.val_loader
+
+
+def extract_sample_arrays(td: TrainData, steps: int) -> Tuple[np.ndarray, np.ndarray]:
```
eric-czech: Perhaps we're building up to this, but a DataLoader + collate_fn would normally handle this kind of batching logic.  I.e. it would probably be best to reimagine `TrainData` that way at some point.

alxmrs: I totally agree. On that note, I'm hoping we can use Xarray as late as possible, like right up to the collate_fn.

tests/conftest.py
alxmrs: TIL! Thank you, updated.

configs/train_cm4.test.yaml
suryadheeshjith, line 1: Nitpick but do you want to replace the unused values in the config like data paths etc. ('2024-11-12-cm4-piControl-ocean-200yr-dataset-stats/centering.nc') with dummy values? You can fill them in later when you have tests for them.

tests/test_datasets.py
```diff
@@ -0,0 +1,93 @@
...
+# TODO(alxmrs): How can we determine `n_samples` from the input config? Timeslice?
+def test_train__loads_correct_number_of_samples(train_loader_pair):
+    cfg, loader = train_loader_pair
+    n_samples = 6
+    assert (
+        len(list(loader)) == n_samples
+    ), f"Current config {cfg} only supports {n_samples} examples; got {len(loader)}."
+
+
+# TODO(alxmrs): What does the "2" dimension represent? Where is it in the config?
```
suryadheeshjith: 2 represents batch_size you pass in the config

tests/test_datasets.py
```diff
@@ -0,0 +1,93 @@
...
+def test_train__data_shape(train_loader_pair):
+    cfg, loader = train_loader_pair
+
+    exp = cfg.experiment
+
+    input_var_dim = len(INPT_VARS[exp.exp_num_in]) + len(EXTRA_VARS[exp.exp_num_extra])
```
suryadheeshjith: These would fail with a hist>0, I am assuming you will be fixing that in a future PR

configs/train_cm4.test.yaml
suryadheeshjith: Maybe even reducing the size of the network, I propose a config that might look like this?:

```
...
  data_path: /fake/data/path
  data_means_path: /fake/data/path
  data_stds_path: /fake/data/path
...
unet:
  ch_width: [2, 3]
  n_out: 4
  dilation: [1, 2]
  n_layers: [1, 1]
  pred_residuals: false
  last_kernel_size: 3
  pad: "circular"

  core_block:
    block_type: "conv_next_block"
    kernel_size: 3
    activation: "capped_gelu"
    upscale_factor: 2
    norm: "batch"

  corrector:
    non_negative_corrector_names:
      - "so"

  down_sampling_block: "avg_pool"
  up_sampling_block: "bilinear_upsample"
```

suryadheeshjith (commented):
LGTM overall! (this did not age well)

tests/test_datasets.py
alxmrs: Thanks, that's a good point. How is `hist` related to `input_var_dim`? I've been experimenting a bit and I haven't been able to figure out a formula (admittedly, by guessing).

alxmrs: Ah, nvm, I think I got it.

tests/conftest.py
```diff
@@ -25,3 +43,96 @@ def model2_path(request):
 @pytest.fixture(params=["cpu", pytest.param("cuda", marks=pytest.mark.cuda)])
 def device(request):
     return torch.device(request.param)
+
+
+@pytest.fixture(scope="session")
+def data_source() -> DataSource:
```
suryadheeshjith, line 49: Isnt DataSource type hinting incorrect? you are returning a dict?

suryadheeshjith: I am unsure why I excluded conftest.py from pre-commit checks

configs/train_cm4.test.yaml
```diff
@@ -0,0 +1,84 @@
...
+    project: 3D_ocean_emu_CM4
+    entity: testrobot
+  network: convnextunet
+  exp_num_in: '3D_5'
+  exp_num_extra: '3D_5'
+  exp_num_out: '3D_5'
+data:
+  data_path: fake/data/path/cm4_piControl_ocean_200yr_full.zarr
+  data_means_path: fake/data/path/cm4-piControl-ocean-200yr-dataset-stats/centering.nc
+  data_stds_path: fake/data/path/piControl-ocean-200yr-dataset-stats/scaling-full-field.nc
```
suryadheeshjith:
```suggestion
  data_path: fake/path/to/data
  data_means_path: fake/path/to/means
  data_stds_path: fake/path/to/stds
```

configs/train_cm4.test.yaml
```diff
@@ -0,0 +1,84 @@
+# A config file to test training.
```
suryadheeshjith, line 1: Since we want this PR to include only the bare minimum of testing, I’d like the config parameters set to values that allow all of us to visualize the data. There are multiple parameters like hist, steps, batch size, stride and the actual total number of timesteps in the data that changes the number of samples in the data so I want to be careful what we test first.

configs/train_cm4.test.yaml
```diff
@@ -0,0 +1,84 @@
...
+# Used in automated tests.
+debug: true
+disk_mode: true
+pin_mem: false
+save_freq: 5
+epochs: 120
+batch_size: 2 # 4
```
suryadheeshjith:
```suggestion
batch_size: 1
```

configs/train_cm4.test.yaml
```diff
@@ -0,0 +1,84 @@
...
+  start_time: '1969-08-15'
+  end_time: '1969-08-30'
+val:
+  start_time: '1969-09-01'
+  end_time: '1969-09-05'
+inference:
+  - start_time: '1969-09-06'
+    end_time: '1969-09-12'
+data_stride: [1]
+steps: [4]
```
suryadheeshjith:
```suggestion
steps: [1]
```

tests/conftest.py
```diff
@@ -25,3 +43,96 @@ def model2_path(request):
...
+    """Returns in-memory `xarray.Dataset`s for tests."""
+    out = {}
+
+    summer_of_love = xr.cftime_range(
+        "1969-08-10", "1969-10-01", freq="D", calendar="noleap"
```
suryadheeshjith:
```suggestion
        "1969-08-10", "1969-12-31", freq="5D", calendar="noleap"
```

suryadheeshjith: Lets stick to working with 5 day time-deltas

configs/train_cm4.test.yaml
suryadheeshjith: We are also missing tests for InferenceDataset which is important because it is a separate class used compared to what train and validation use. Will this be implemented in a later PR?

configs/train_cm4.test.yaml
```diff
@@ -0,0 +1,84 @@
...
+data_percent: 1.0
+train:
+  start_time: '1969-08-15'
+  end_time: '1969-08-30'
+val:
+  start_time: '1969-09-01'
+  end_time: '1969-09-05'
+inference:
+  - start_time: '1969-09-06'
+    end_time: '1969-09-12'
```
suryadheeshjith:
```suggestion
train:
  start_time: '1969-08-15'
  end_time: '1969-08-20'
val:
  start_time: '1969-09-04'
  end_time: '1969-09-09'
inference:
  - start_time: '1969-09-06'
    end_time: '1969-09-14'
```

suryadheeshjith: Now, our data for both train and validation will contain a single sample considering batch size = 1, steps = 1, stride = 1, hist = 0.

The sample for train will be: [Input : 1969-08-15, Target : 1969-08-20]
The sample for validation will be: [Input : 1969-09-04, Target : 1969-09-09]

tests/test_datasets.py
```diff
@@ -0,0 +1,101 @@
...
+    y_arrays = [td.get_label(s).numpy(force=True) for s in range(steps)]
+
+    return np.stack(x_arrays, axis=0), np.stack(y_arrays, axis=0)
+
+
+# TODO(alxmrs): How can we determine `n_samples` from the input config? Timeslice?
+#  Changing the "hist" parameter breaks this test.
+def test_train__loads_correct_number_of_samples(train_loader_pair):
+    cfg, loader = train_loader_pair
+    n_samples = 6
```
suryadheeshjith:
```suggestion
    n_samples = 1
```

tests/test_datasets.py
```diff
@@ -0,0 +1,101 @@
...
+        X, y = extract_sample_arrays(sample, cfg.steps[0])
+        assert X.shape == (cfg.steps[0], batch_size, input_var_dim, 180, 360)
+        assert y.shape == (cfg.steps[0], batch_size, output_var_dim, 180, 360)
+
+
+# TODO(alxmrs): How can we determine `n_samples` from the input config? Timeslice?
+#  Changing the "hist" parameter breaks this test.
+def test_val__loads_correct_number_of_samples(val_loader_pair):
+    cfg, loader = val_loader_pair
+    n_samples = 2
```
suryadheeshjith:
```suggestion
    n_samples = 1
```

suryadheeshjith (changes requested):
Review complete. I will need to write up a document for the working of the train and inference loaders so it is easy for everyone to understand what is happening (including me).

tests/conftest.py
```diff
@@ -25,3 +43,96 @@ def model2_path(request):
...
+
+        # Open default training script; modify it so it uses the temporary directory.
+        default_config = pytestconfig.rootpath / "configs" / "train_cm4.test.yaml"
+        trainer = TrainConfig.from_yaml(default_config)
+        data_config = dataclasses.replace(
+            trainer.data,
+            data_path=_make_path("data.zarr"),
+            data_means_path=_make_path("means.netcdf"),
+            data_stds_path=_make_path("stds.netcdf"),
+        )
```
jder, line 120: Definitely seems worth thinking through how to make this less fragile eg by passing root dirs into a build method or something (not for this PR).

tests/test_datasets.py
```diff
@@ -0,0 +1,101 @@
...
+from constants import EXTRA_VARS, INPT_VARS, OUT_VARS
+from datasets import TrainData
+from train_3D import Trainer
+
+# Note: Refactoring data loaders is planned for the near-term. Ideally,
+# these fixtures allow us to isolate data loader tests from their setup.
+
+
+# This micro-fixture is cached by pytest. Thus, we don't have to change
+# the factory methods that throw errors during double initialization.
```
jder, line 20: What factory methods that throw errors is this referring to?

tests/test_datasets.py
```diff
@@ -0,0 +1,101 @@
...
+@pytest.fixture(scope="session")
+def trainer_pair(train_config):
```
jder: Is it worth adding type hints for these? Why do you have these all return the train config instead of allowing the downstream test to request train_config if it wants it?

suryadheeshjith:
A broader question regarding testing, do we move the tests to live together with the code? And have conftest.py in the root directory? This would allow us to test across the source code without having to create a ton of files in the same directory or have a similar structure as in src/.

Similar to [ace](https://github.com/ai2cm/ace/tree/main)

jder (approved):
Nothing blocking from me, just some questions and 🐑 . (The 🐑 means a "sheepish suggestion" ie nit pick)

jder:
> A broader question regarding testing, do we move the tests to live together with the code? And have conftest.py in the root directory? This would allow us to test across the source code without having to create a ton of files in the same directory or have a similar structure as in src/.

I prefer this, too. Maybe we do this as a follow up after this PR lands?

configs/train_cm4.test.yaml
alxmrs: Thanks, these suggestions are really helpful. Let me see if I can add a quick inference dataset test here now.

alxmrs: Added inference tests. Thanks for calling this out.

tests/conftest.py
alxmrs: Yes, the type hinting is correct. It's a TypedDict.

alxmrs: Ok, I've included conftest to pre-commit and there was a type error. However, it was a problem with the type system, and I've now corrected it. Good catch.

tests/conftest.py
alxmrs: I tried that but decided not to go with it. When I apply the Pathlib object to the data_config values, which expects a string, it causes errors. Fixing these errors by wrapping the pathlib path with `str()` makes the code more complicated than keeping this function.

tests/test_datasets.py
alxmrs: Boy, there are a lot! These are all throughout the codebase. On first glance, they look like singletons, but instead of returning the singleton instance, they throw an error. 

e.g. see the first few results of `_instance`:
https://github.com/search?q=repo%3Asuryadheeshjith%2FOcean_Emulator%20_instance&type=code

tests/test_datasets.py
alxmrs: That's a good question. If I injected the config and the trainer in each test, I'd worry about them using mis-matched combinations of each item. This way, the config + trainer must be coupled. 

I'm not sure if this error will occur, but I'm worried about it. I could change it to this now and fix it later if it occurs. WDYT?

alxmrs: Just added types.

alxmrs:
> A broader question regarding testing,

Sounds good to me! We can do that in a follow-up PR.

suryadheeshjith (approved):
LGTM

PR #71, opened 2025-02-19
Making data loader tests more interpretable.

alxmrs:
This adds a simple test if the loaded data is non-zero. Further, it changes the randomly generated data to one related to lat-lng grid coordinates. The hope is that leads to making data more interpretable.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,6 +20,13 @@ class DataSource(TypedDict):
     stds: xr.Dataset
 
 
+class GridPoint(TypedDict):
+    lat: float
+    lng: float
+    days_since_start: int
+    data_var_index: int
+
+
 def pytest_addoption(parser):
     parser.addoption(
         "--model1", action="store", help="Path to the first model .pt file"
@@ -83,9 +90,9 @@ def data_source() -> DataSource:
     # A floating point digit-encoded grid.
     # ------------------------------------
     # Each number in this array is an interpretable float with the following scheme:
-    # AAAAGGGG.TTT[DD]
-    # - A := Latitude (originally a float with 2 decimal digits).
-    # - G := Longitude ""
+    # AAAAGGGG.TTTDD
+    # - A := Latitude, which ranges from 90.00 <--> -90.00
+    # - G := Longitude, which ranges from 000.0 <--> 360.0
     # - T := Time (the number of days since the start time).
     # - D := (optional) A int representing the index of the current data variable.
     interpretable_grid = rolled_out_grid + days_reshaped / 1000  # Float[30, 180, 360]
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -49,47 +49,53 @@ def device(request):
...
+        "1969-08-05", "1969-12-31", freq="5D", calendar="noleap"
     )
 
     coords = {
-        "lon": xr.DataArray(np.arange(0.5, 360, 1), dims=["lon"]),
-        "lat": xr.DataArray(np.arange(-89.24, 90, 1), dims=["lat"]),
-        "time": xr.DataArray(summer_of_love, dims=["time"]),
+        "lon": xr.DataArray(np.arange(0.5, 360, 1), dims=["lon"]),  # Float[360]
+        "lat": xr.DataArray(np.arange(-89.24, 90, 1), dims=["lat"]),  # Float[180]
+        "time": xr.DataArray(summer_of_love, dims=["time"]),  # CFTimeIndex[30]
```
jder, line 65: These comments are unenforced, right? Maybe would be clearer + more verifiable to use an assert instead?

tests/conftest.py
```diff
@@ -49,47 +49,53 @@ def device(request):
...
-        },
-        coords=coords,
+        size=(len(coords["lat"]), len(coords["lon"]))
+    )  # Float[180, 360]
+
+    # A grid of alternating lat/lng coordinate values across time.
+    # The zero axis (time) switches between a "lat" grid and a "lon" grid:
+    # Each "lat" col is a gradient 90° to -89° from top to bottom; all cols are equal.
+    # Each "lon" row is a gradient from 0.5° to 360° from left to right; rows are equal.
+    # This alternating grid of lat/lng values should help us validate data during tests.
```
jder: Alternative might be that each grid cell is `lat * 1000 + lon + time / 100000` or something.

tests/test_datasets.py
```diff
@@ -46,8 +46,19 @@ def inference_loader_pair(trainer_pair: TrainPair) -> LoaderPair:
...
+def td_loader_pair(request, train_loader_pair, val_loader_pair) -> LoaderPair:
+    if request.param == "train":
+        return train_loader_pair
+    else:
+        return val_loader_pair
+
+
+def extract_sample_arrays(td: TrainData) -> tuple[np.ndarray, np.ndarray]:
     """Extract underlying X, y pairs from TrainData object."""
+    steps = len(td.td_dict.keys())
```
jder: Maybe `td.steps` or `len(td)`?

jder (approved):
Couple of suggestions, but they can be follow ups too.

tests/conftest.py
alxmrs: I was hoping I'd be able to take these and turn them into jaxtyping types soon. I'd prefer that (a static type check) over a runtime assert.

Shall I remove these for now?

tests/test_datasets.py
alxmrs: Good point! I forgot about `td.steps`.

tests/conftest.py
alxmrs: I like that idea a lot!

tests/conftest.py
jder: Ah, sure, fine by me to leave and replace with static types later

tests/conftest.py
```diff
@@ -20,6 +20,13 @@ class DataSource(TypedDict):
     stds: xr.Dataset
 
 
+class GridPoint(TypedDict):
```
jder, line 23: Why TypedDict rather than a dataclass?

tests/conftest.py
```diff
@@ -49,51 +56,108 @@ def device(request):
...
+            np.where(normal > 0.5**lev, 1, 0), dims=["lat", "lon"]
+        )
+        for lev in range(len(c.DEPTH_I_LEVELS))
+    }
+    ds = xr.Dataset(vars_2d | vars_3d | masks, coords=coords)
 
     return {"data": ds, "means": ds.mean("time"), "stds": ds.std("time")}
 
 
+def parse_encoded_float(encoded: np.float64) -> GridPoint:
```
jder, line 123: FWIW if this we run into trouble with this reverse direction down the road, we could also implement only the structure-to-float direction and test equality on the floats themselves.

suryadheeshjith:
I did not have enough time to look at this merged PR but I really love the idea of being able to distinguish the data in the array/tensor. Some issues  that can be addressed in a later PR here #79.

tests/conftest.py
alxmrs: Working on round trip tests now :)

tests/conftest.py
alxmrs: No good reason, but now that I see this is a common pattern, I'm switching over!

PR #73, opened 2025-02-20
Created performance benchmarks.

alxmrs:
I've began adding standard performance benchmarking infrastructure using `pytest-benchmark`. This documents local use and sets up a standard Github Action (to host consistent performance changes in a gh pages website).


prof/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Developers may optionally run [performance profile traces](https://docs.python.org/3/library/profile.html) in unit tests
+via the `profile` pytest fixture. This will automatically save a timestamped copy of a cProfile trace to this directory.
+The intention is that we'll be able to compare project performance improvements over time as well as build a culture of
+testing performance.
+
+Here's how you can inspect the results of a `cProfile` profile dump:
+
+```bash
+uvx snakeviz prof/<your-profile-dump.prof>
```
jder: Maybe show how to run the profile in the first place, too?

jder: Oh, I see, it always runs. I was imagining we'd want to have these tests off by default and then run them explicitly, since we probably want to run them multiple times and compute averages etc. Can you say more about the workflow you're imagining here?

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: uv run pytest --benchmark-only --benchmark-json output.json
+
+      # gh-pages branch is automatically updated with extracted benchmark data.
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: Python Benchmark with pytest-benchmark
+          tool: 'pytest'
+          output-file-path: output.json
+          # Use personal access token instead of GITHUB_TOKEN due to https://github.community/t/github-action-not-triggering-gh-pages-upon-push/16096
```
jder: Comment maybe out of date? I can't see this link.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,48 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
jder, line 42: Given we're passing a sensitive token here, maybe we should at least fork the action to our own repo so we control the updates to that code?

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,48 @@
...
+
+# Do not run this workflow on pull request since this workflow has permission to modify contents.
+on:
+  push:
+    branches:
+      - master
+
+permissions:
+  contents: write  # Allow updating content on GitHub pages
+  deployments: write  # Allow to deploy to GitHub pages
```
jder, line 14: Is this actually publishing somewhere? If so can we add a link here or to the readme?

jder (commented):
Content & approach generally SGTM!

.github/workflows/benchmarks.yml
alxmrs: Just added a link to the README.

.github/workflows/benchmarks.yml
alxmrs: I'll do this in a later issue. But, this is a good idea.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,47 @@
+# Performance benchmarks with pytest
+# Based on this example: https://github.com/benchmark-action/github-action-benchmark/tree/master/examples/pytest
+name: Benchmark CPU
+
+# Do not run this workflow on pull request since this workflow has permission to modify contents.
+on:
+  push:
+    branches:
+      - main
```
jder, line 9: Should we add a `workflow_dispatch` too so that we can run when desired on PR branches?

README.md
```diff
@@ -1 +1,5 @@
 # Ocean Emulator
+
+[![Pre-commit](https://github.com/suryadheeshjith/Ocean_Emulator/actions/workflows/pre-commit.yml/badge.svg)](https://github.com/suryadheeshjith/Ocean_Emulator/actions/workflows/pre-commit.yml)
+[![Test CPU](https://github.com/suryadheeshjith/Ocean_Emulator/actions/workflows/test.yml/badge.svg)](https://github.com/suryadheeshjith/Ocean_Emulator/actions/workflows/test.yml)
+| [Perf Benchmarks](http://suryadheeshjith.github.io/Ocean_Emulator/dev/bench)
```
jder, line 5: Where does `dev/bench` come from? This is a default path from the benchmark action?

tests/test_datasets.py
```diff
@@ -194,3 +275,25 @@ def test_inference__data_is_not_zero(inference_loader_pair: LoaderPair):
...
+@pytest.mark.manual
+def test_profile__inference_loader__1gb(inference_loader_pair: LoaderPair, benchmark):
+    cfg, loader = inference_loader_pair
+
+    @benchmark
+    def bench():
+        for sample in loader:
+            dataset, n = sample
+            for X, y in dataset:
+                _, _ = X, y
```
jder, line 294: Should we name these _in_memory or something? These are testing in-memory datasets, yes?

jder (approved):
Looks great, just a few suggestions. Thanks!

.github/workflows/benchmarks.yml
alxmrs: I'm just googling workflow_dispatch. How does this work? And, you have recommendations for how I should configure one?

README.md
alxmrs: Yes, that is correct. See these docs: 

https://github.com/benchmark-action/github-action-benchmark#how-to-use:~:text=By%20default%2C%20this%20action%20assumes%20that%20gh%2Dpages%20is%20your%20GitHub%20Pages%20branch%20and%20that%20/dev/bench

tests/test_datasets.py
alxmrs: Thanks for clarifying. No, they are not in memory. While we do create test_data in memory, part of the test set up writes them to disk as zarr and netcdf. 

Related, however: I was thinking about adding the chunking scheme to the test name. From a quick investigation earlier, it looks like everything is in one big chunk, so I omitted writing it.

.github/workflows/benchmarks.yml
jder: I think all you need to add is a `workflow_dispatch:` as a peer to the `push:` key. That'll let us kick these off by hand on the actions page.

Issue #51, opened 2025-02-11
Model independence from stepper mechanics


suryadheeshjith:
[Base Model](https://github.com/suryadheeshjith/Ocean_Emulator/blob/master/src/models/base.py) has code that does the actual stepping which we want to separate. But, there are issues with doing this (especially with batchnorm), see more [here](https://github.com/suryadheeshjith/Ocean_Emulator/blob/master/notebooks/batchnorm_test.ipynb).

alxmrs:
The link to the notebook is broken, it can be found here: https://github.com/Open-Athena/Ocean_Emulator/blob/main/notebooks/batchnorm_test.py

PR #78, opened 2025-02-21
Allow specifying which backend we want to use for tests and in config

jder:
Code previously assumed if you had a GPU you wanted to have distributed = true and (AFAICT) have the right env vars set for that. This tries to let you pick if you want cpu, gpu, or distributed. (Or try to detect automatically.) 

This now lets us run tests on CPUs in CI. Next we'll set up GPU tests and run both.

### DIFF
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -25,7 +25,7 @@ inference:
 data_stride: [1]
 steps: [1]
 step_transition: []
-backend: cpu
+backend: auto
 
 experiment:
   name: train
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -25,14 +25,7 @@ inference:
 data_stride: [1]
 steps: [1]
 step_transition: []
-
-distributed:
-  enabled: true
-  dist_url: null
-  world_size: null
-  rank: null
-  gpu: null
-  dist_backend: null
+backend: cpu
 
 experiment:
   name: train
--- a/src/backend.py
+++ b/src/backend.py
@@ -1,0 +1,61 @@
+import logging
+
+import torch
+
+from config import DistributedConfig, EvalBackendConfig, TrainBackendConfig
+from utils.distributed import init_distributed_mode
+
+
+def init_train_backend(
+    backend: TrainBackendConfig,
+) -> tuple[torch.device, DistributedConfig | None]:
+    """Given backend config, get the device and (if any) distributed configuration."""
+    match backend:
+        case "cpu":
+            device = torch.device("cpu")
+            dist_cfg = None
+        case "cuda":
+            device = torch.device("cuda")
+            dist_cfg = None
+        case "nccl":
+            device = torch.device("cuda")
+            dist_cfg = init_distributed_mode()
+        case "auto" if torch.cuda.is_available():
+            logging.info("auto backend detected CUDA")
+            device = torch.device("cuda")
+            try:
+                dist_cfg = init_distributed_mode()
+                logging.info("succeeded in initializing distributed mode")
+            except RuntimeError as e:
+                logging.info(
+                    f"Failed to initialize distributed mode, running on single node.",
+                    exc_info=e,
+                )
+                dist_cfg = None
+        case "auto":
+            logging.info("auto backend: cuda not found, using CPU")
+            device = torch.device("cpu")
+            dist_cfg = None
+        case _:
+            raise ValueError(f"Invalid backend: {backend}")
+
+    return device, dist_cfg
+
+
+def init_eval_backend(backend: EvalBackendConfig) -> torch.device:
+    """Given evaluation backend config, get the device."""
+    match backend:
+        case "cpu":
+            device = torch.device("cpu")
+        case "cuda":
+            device = torch.device("cuda")
+        case "auto" if torch.cuda.is_available():
+            logging.info("auto backend detected CUDA")
+            device = torch.device("cuda")
+        case "auto":
+            logging.info("auto backend: cuda not found, using CPU")
+            device = torch.device("cpu")
+        case _:
+            raise ValueError(f"Invalid backend: {backend}")
+
+    return device
--- a/src/backend.py
+++ b/src/backend.py
@@ -40,6 +40,8 @@ def init_train_backend(
         case _:
             raise ValueError(f"Invalid backend: {backend}")
 
+    # We set this globally so we don't need to hand the device around.
+    # See https://github.com/suryadheeshjith/Ocean_Emulator/issues/87.
     set_device(device)
 
     return device, dist_cfg
--- a/src/backend.py
+++ b/src/backend.py
@@ -3,6 +3,7 @@ import logging
 import torch
 
 from config import DistributedConfig, EvalBackendConfig, TrainBackendConfig
+from utils.device import set_device
 from utils.distributed import init_distributed_mode
 
 
@@ -38,6 +39,8 @@ def init_train_backend(
             dist_cfg = None
         case _:
             raise ValueError(f"Invalid backend: {backend}")
+
+    set_device(device)
 
     return device, dist_cfg
 
--- a/src/stepper.py
+++ b/src/stepper.py
@@ -5,7 +5,7 @@ import torch
 
 from aggregator import InferenceEvaluatorAggregator
 from datasets import InferenceDataset, TrainData
-from src.models.base import BaseModel
+from models.base import BaseModel
 from utils.model import InfOutput, TrainOutput, ValOutput
 from utils.wandb import get_record_to_wandb
 from utils.writer import ZarrWriter
--- a/src/stepper.py
+++ b/src/stepper.py
@@ -6,6 +6,7 @@ import torch
 from aggregator import InferenceEvaluatorAggregator
 from datasets import InferenceDataset, TrainData
 from utils.device import using_gpu
+from src.models.base import BaseModel
 from utils.model import InfOutput, TrainOutput, ValOutput
 from utils.wandb import get_record_to_wandb
 from utils.writer import ZarrWriter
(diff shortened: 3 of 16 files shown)


configs/train_cm4.test.yaml
```diff
@@ -25,14 +25,7 @@ inference:
...
 step_transition: []
-
-distributed:
-  enabled: true
-  dist_url: null
-  world_size: null
-  rank: null
-  gpu: null
-  dist_backend: null
+backend: cpu
```
jder: Having this be `cpu` rather than `auto` is basically the fix to run tests (for now) on cpu. Previously the `distributed` blocks were unused (they were populated in code) except for the 'enabled' boolean. You can now control that with this backend option. (`nccl` vs `cuda` or `cpu`)

src/aggregator/inference/main.py
```diff
@@ -19,6 +19,7 @@ class InferenceEvaluatorAggregator:
 
     def __init__(
         self,
+        device: torch.device,
```
jder: Most of the changes in this PR are passing the device around. Not sure how much this is worth it vs setting the global device using `with torch.device(…)` -- do we want tensors to be allocated on the CPU by default and only moved to GPU manually with `to`?

src/backend.py
```diff
@@ -0,0 +1,59 @@
...
+        case "auto" if torch.cuda.is_available():
+            logging.info("auto backend detected CUDA")
+            device = torch.device("cuda")
+        case "auto":
+            logging.info("auto backend: cuda not found, using CPU")
+            device = torch.device("cpu")
+        case _:
+            raise ValueError(f"Invalid backend: {backend}")
+
+    return device
```
jder, line 66: This file now contains the logic to decide which device & distribution config to use based on the `backend` config option. These are run once and then the device is passed around rather than a global `get_device`.

src/datasets.py
```diff
@@ -76,9 +74,8 @@ def __init__(
         self.wet_surface = wet_surface.bool()
         self.size = len(self.rolling_indices)
 
-        if using_gpu():
-            self.wet = self.wet.pin_memory()
-            self.wet_surface = self.wet_surface.pin_memory()
+        self.wet = self.wet.pin_memory()
+        self.wet_surface = self.wet_surface.pin_memory()
```
jder: I think this is harmless?

src/stepper.py
```diff
@@ -26,12 +26,11 @@ def train_step(
     @staticmethod
     @torch.no_grad()
     def validate_step(
-        model: torch.nn.Module, batch: TrainData, loss_fn: Callable
+        model: BaseModel, batch: TrainData, loss_fn: Callable
     ) -> ValOutput:
         assert len(batch) == 1  # Assert we are using one step of input and output
         input = batch.get_input(0)
         label = batch.get_label(0)
-        model = model.module if using_gpu() else model
```
jder, line 34: This is actually a bug at the moment. The issue is that when using GPU we wrap the BaseModel into a `DistributedDataParallel` module. But that doesn't expose `forward_once`, so we unwrap here. We can of course pass the device along here and use the same conditional logic but I wonder if we could remove that by moving the logic currently in BaseModel.forward into Stepper and having UNet.forward be what's currently in forward_once. Is that better? Worse?

src/train_3D.py
```diff
@@ -541,7 +536,7 @@ def inference_one_epoch(self, epoch):
             )
 
             Stepper.inference(
-                model=self.model.module if using_gpu() else self.model,
+                model=self.model,
```
jder: This is also a bug for the same reason.

src/train_3D.py
```diff
@@ -736,14 +731,14 @@ def load_checkpoint(self, checkpoint_path, finetune=False):
             logging.info(f"Optimizer LR: {self.optimizer.param_groups[-1]['lr']}")
 
     def is_wandb_enabled(self):
-        if using_gpu():
-            return self.wandb_logger.enabled and is_main_process()
-        else:
-            return self.wandb_logger.enabled
+        return self.wandb_logger.enabled and is_main_process()
```
jder, line 736: is_main_process() already correctly handles the non-gpu case, I think?

src/datasets.py
jder: Ok, CI tells me it's not harmless, that's too bad.

src/aggregator/inference/main.py
alxmrs: > do we want tensors to be allocated on the CPU by default and only moved to GPU manually with to?

A great question! I think we'll have to decide based on a performance investigation. My intuition says "yes", though.

src/backend.py
```diff
@@ -0,0 +1,59 @@
...
+import torch
+
+from config import DistributedConfig, EvalBackendConfig, TrainBackendConfig
+from utils.distributed import init_distributed_mode
+
+
+def init_train_backend(
+    backend: TrainBackendConfig,
+) -> tuple[torch.device, DistributedConfig | None]:
+    match backend:
```
alxmrs, line 14: Nice use of pattern matching!

src/backend.py
alxmrs: This explanation is great! I think it would be nice to have as a docstring at the top of the file.

src/datasets.py
alxmrs: lool!

src/stepper.py
alxmrs: Great questions, I haven't looked at that part of the codebase yet, but hope to soon.

tests/conftest.py
```diff
@@ -40,9 +39,13 @@ def model2_path(request):
 
 
 # Run a test for both CPU and GPU, and allows selecting or skipping CUDA tests.
-@pytest.fixture(params=["cpu", pytest.param("cuda", marks=pytest.mark.cuda)])
-def device(request):
-    return torch.device(request.param)
+# TODO(jder): At the moment, we can't use this because Trainer can only be
+# inited once per process. Once that is we should accept it in `train_config` below.
+@pytest.fixture(
```
alxmrs, line 52: Thanks for this note. I like the change that you've made here.

I'm not sure if this is useful, but we could mark the scope for this fixture to be `session` and then it should only call this function once per testing session (per parameter). That's how I've gotten the trainer instantiation to work thus far. 

I think this comment already alludes to this, but: I like the idea of letting this fixture set the device in the train_config.

src/stepper.py
suryadheeshjith: > But that doesn't expose forward_once, so we unwrap here. We can of course pass the device along here and use the same conditional logic but I wonder if we could remove that by moving the logic currently in BaseModel.forward into Stepper and having UNet.forward be what's currently in forward_once.

This is a great catch. I am planning on doing exactly the same (#51) but what stopped me from doing this before were some issues with how batch normalization is implemented in pytorch alongside torch's DistributedDataParallel module. See more [here](https://github.com/pytorch/pytorch/issues/66504#issuecomment-1007806934). I also created a notebook that describes this issue [here](https://github.com/suryadheeshjith/Ocean_Emulator/blob/main/notebooks/batchnorm_test.ipynb).

src/backend.py
jder: Added it to the relevant functions.

src/stepper.py
jder: ok, for now I replaced this with a type test instead of passing the device around which seems more direct anyway.

tests/conftest.py
jder: Yeah, makes sense. Right now it is already session scoped but since it yields two configurations it runs into the same issue.

src/aggregator/inference/main.py
jder: OK, I left behavior unchanged for now.

jder:
ok, I think this is ready for a review. My main outstanding question: I think this probably breaks one or more notebooks since you must now pass a device to some classes you didn't have to before. @suryadheeshjith which notebooks should we be 
maintaining/validating as we make changes like this?

jder:
This is ready for another look. As discussed this no longer hands the device around -- I left a TODO with a link to the issue. We can now run tests on either CPU or GPU but not both together. 

@suryadheeshjith please let me know about this too

> @suryadheeshjith which notebooks should we be maintaining/validating as we make changes like this?

suryadheeshjith:
You may ignore notebooks for now. For future reference, the main notebook I use for evaluation is Samudra_Eval.ipynb. You can ignore it for now.

suryadheeshjith (approved):
LGTM

src/train_3D.py
```diff
@@ -31,21 +31,16 @@
 from constants import EXTRA_VARS, INPT_VARS, OUT_VARS, TensorMap, construct_metadata
 from datasets import InferenceDataset, InferenceDatasets, TrainDataset
 from models.unet import UNet
+from src.backend import init_train_backend
```
alxmrs: 🛑  I'm not a fan of including `src.` in the import path.

src/utils/device.py
```diff
@@ -1,11 +1,28 @@
...
+
 
 def using_gpu() -> bool:
     return get_device().type == "cuda"
 
 
+# TODO(jder): We'd like to remove this (and other global singletons)
+# by moving to some training framework + context objects once we have
+# benchmarking set up.
+def set_device(device: torch.device) -> None:
```
alxmrs, line 13: These LGTM, thanks!

alxmrs (approved):
I have one (small) blocker, but otherwise, this PR LGTM. After it's fixed, this looks ready to merge.

src/train_3D.py
alxmrs: Is there some way that you can make the import like `constants` and `datasets` above?

jder: This (and a few others I found) were added automatically by ruff. I think this is because `src` has an `__init__.py` in it and ideally we'd make it `src/samudra/__init__.py` (and maybe have tests live inside samudra as Surya said elsewhere). I didn't see a way to configure ruff otherwise (I tried a couple things like setting src = "src").

alxmrs: That makes sense! Thanks for letting me know. There may be a interesting `pyproject.toml` config we could add to make this work, but I agree that `src/samudra/` packaging is the way to go. For a future PR, ofc.

PR #83, opened 2025-02-28
Check in uv.lock

jder:
The uv docs recommend checking this file in so everyone can produce the same build. https://docs.astral.sh/uv/guides/projects/#uvlock

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,5484 @@
+version = 1
+revision = 1
+requires-python = ">=3.10"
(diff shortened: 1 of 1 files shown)


jder:
I should mention this failed the "large file" precommit check locally (it's ~500k) but I think this is OK.

suryadheeshjith:
I thought I removed the check

suryadheeshjith (approved):

PR #84, opened 2025-02-28
Work around typechecking bug in torch

jder:
Type checkers don't know about torch.utils because of https://github.com/pytorch/pytorch/issues/126401; seemed worth working around so we can move towards adding type hint checking in CI.


suryadheeshjith (approved):
LGTM

PR #88, opened 2025-03-03
Interpretable encoding scheme from dataset dimensions to data arrays.

### LINKED ISSUE #79: Dataloader tests issues
Some issues with the encoding scheme in data_source

1. Parsing/decoding uses divmod and does not work as expected with negatively valued latitudes. 
Example -86236804.98 : {'lat': -86.24, 'lng': 319.5, 'days_since_start': 19, 'data_var_index': 100}
2. Use of float(i + j + len(vars_2d)) in vars_3d while constructing the xarray dataset can have conflicting values with i+j being the same for different combinations of i and j. 
3. There are values like 18761085.040009998 in the data that get decoded to {'lat': 18.76, 'lng': 108.5, 'days_since_start': 40, 'data_var_index': 1}. There are probably some precision issues to be taken care of.

alxmrs:
Here, I've introduced dataclass that encodes and decodes a (mostly) interpretable `xarray.DataArray` of floats. These are manually tested with a fuzzer (hypothesis). These tests are turned off by default due to the choice to encode the numbers as significant digits in floats due to the tendency for rounding errors in `np.float64` values. For our purposes of making debuggable test data, these should be decent.

Fixes #79.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -53,7 +53,7 @@ class DataSourceDims:
         # So, to ensure that we can have negative latitudes, we parse the sign value
         # of the lat array.
         lat_sgn = np.sign(self.lat)
-        lat_sgn = np.where(lat_sgn == 0, 1, lat_sgn)
+        lat_sgn = np.where(lat_sgn == 0, 1, lat_sgn)  # change np.sign(0) to +1, not 0.
         self.lat = _trunc(self.lat.astype(np.float64), 2) % (lat_sgn * 100)
         if np.any(self.lat < -90.0) or np.any(self.lat > 90.0):
             raise ValueError("lat values must be between -90 and 90.")
@@ -68,6 +68,9 @@ class DataSourceDims:
             raise ValueError("days_since_start must be between 0 and 999.")
 
     def __eq__(self, other) -> bool:
+        # lat and lng values will often have floating point rounding errors.
+        # So, we declare arrays as equal within a generous tolerance.
+        # We can use exact equality with days_since_start since they are ints.
         return (
             np.allclose(self.lat, other.lat, rtol=0.1)
             and np.allclose(self.lng, other.lng, rtol=0.1)
@@ -159,7 +162,9 @@ class DataSourceDims:
             (DataSourceDims, int) - Parsed dims and data_var index.
         """
         encoded = da.to_numpy()
-        assert len(encoded.shape) == 3
+        assert (
+            len(encoded.shape) == 3
+        ), "DataArray must have (time, lat, lng) dimensions."
 
         scalar = encoded.flat[0]
         tim_dim = encoded[:, 0, 0]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,11 +33,12 @@ class DataSourceDims:
     """Dimension metadata to produce interpretable `xarray.DataArray`s.
 
     Each float in the encoded `xarray.DataArray` has the following scheme:
-    AAAAGGGG.TTTDD
+    (s)AAAAGGGG.TTTDDD
+    - s := Sign (+/-) of the Latitude. All other values are non-negative.
     - A := Latitude, which ranges from 90.00 <--> -90.00
     - G := Longitude, which ranges from 000.0 <--> 360.0
-    - T := Time (the number of days since the start time).
-    - D := (optional) A int representing the index of the current data variable.
+    - T := Time, the number of days since the start time.
+    - D := A int representing the index of the current data variable.
     """
 
     lat: NDArray[np.float64] = np.arange(-89.24, 90, 1, dtype=np.float64)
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -1,30 +1,183 @@
...
+
+    def __eq__(self, other) -> bool:
+        return (
+            np.array_equal(self.lat, other.lat)
+            and np.array_equal(self.lng, other.lng)
+            and np.array_equal(self.days_since_start, other.days_since_start)
+            and self.start_day == other.start_day
+        )
+
+    def parse_cftime_range(self, time_range: xr.CFTimeIndex) -> None:
```
jder: maybe `set_time_range`?

tests/conftest.py
```diff
@@ -1,30 +1,183 @@
...
+def _trunc(arr: NDArray[np.floating], decimals: int) -> NDArray[np.floating]:
+    factor = 10**decimals
+    return np.trunc(arr * factor) / factor
+
+
+@dataclasses.dataclass
+class DataSourceDims:
+    """Dimension metadata to produce interpretable `xarray.DataArray`s.
+
+    Each float in the `xarray.DataArray` has the following scheme:
```
jder:
```suggestion
    Each float in the encoded `xarray.DataArray` has the following scheme:
```

tests/conftest.py
```diff
@@ -1,30 +1,183 @@
...
+            and np.array_equal(self.lng, other.lng)
+            and np.array_equal(self.days_since_start, other.days_since_start)
+            and self.start_day == other.start_day
+        )
+
+    def parse_cftime_range(self, time_range: xr.CFTimeIndex) -> None:
+        self.start_day = time_range[0]
+        timedeltas = [date - self.start_day for date in time_range]
+        self.days_since_start = np.array(
+            [delta.total_seconds() / (24 * 3600) for delta in timedeltas],
```
jder: I see there's a `cftime.date2num` -- would that make this easier?

tests/conftest.py
```diff
@@ -1,30 +1,183 @@
...
+            dtype=np.int32,
+        )
+
+    def to_coords(self) -> dict[str, xr.DataArray]:
+        time = np.array(
+            [
+                self.start_day + cftime._cftime.timedelta(days=int(d))
+                for d in self.days_since_start
```
jder: Same here, would `cftime.num2date` be easier?

If not, it looks like this is a private alias for `datetime.timedelta`, so let's use that instead?

tests/test_datasets.py
```diff
@@ -1,8 +1,15 @@
...
+
+import cftime
 import numpy as np
 import pytest
-from conftest import parse_encoded_float
+from conftest import DataSourceDims
+from hypothesis import example, given, settings
+from hypothesis import strategies as st
+from hypothesis.extra.numpy import arrays
+from numpy._typing import NDArray
```
jder:
```suggestion
from numpy.typing import NDArray
```

tests/test_datasets.py
```diff
@@ -66,24 +73,98 @@ def extract_sample_arrays(td: TrainData) -> tuple[np.ndarray, np.ndarray]:
...
+        min_size=1,
+        max_size=1,
+    ).map(tuple)
+
+
+# Note(alxmrs): Full round trip corner cases are not possible today
+# (2025-03-03) due to floating point rounding woes. In the future,
+# there may be merit to an encoding scheme made of mangled complex
+# numbers. Not worth pursuing right now...
+@pytest.mark.manual
```
jder: Maybe `@pytest.mark.skip` until we want to pursue it?

tests/conftest.py
```diff
@@ -1,30 +1,183 @@
...
+
+    lat: NDArray[np.float64] = np.arange(-89.24, 90, 1, dtype=np.float64)
+    lng: NDArray[np.float64] = np.arange(0.5, 360, 1, dtype=np.float64)
+    days_since_start: NDArray[np.int32] = np.array([0, 5, 10])
+    start_day: cftime.datetime = cftime.DatetimeNoLeap.strptime(
+        "1969-08-05", "%Y-%m-%d", "noleap"
+    )
+
+    def __post_init__(self):
+        self.lat = _trunc(self.lat.astype(np.float64), 2) % 100
```
jder: 🔴 If the passed `lat` value is negative, this will make it positive (mod 100) and so potentially overlap with positive `lat` values. Maybe we want this to be % 200 or something so the ranges are distinct? 

In general manipulating the values in `__post_init__` makes it a little hard to believe the test since eg setting them all to zero would cause the test to pass. Maybe it would be better to assert the types and ranges here rather than coercing them?

tests/conftest.py
```diff
@@ -1,30 +1,183 @@
...
+        tim_sign = np.sign(tim_dim)
+        tim_sign = np.where(tim_sign == 0, 1, tim_sign)
+        lat_sign = np.sign(lat_dim)
+        lat_sign = np.where(lat_sign == 0, 1, lat_sign)
+        lng_sign = np.sign(lng_dim)
+        lng_sign = np.where(lng_sign == 0, 1, lng_sign)
+
+        days_since_start = ((tim_dim * 1000) % (tim_sign * 1000)).astype(np.int32)
+        lat = (lat_dim // (lat_sign * 10_000)) / 100
+        lng = lng_dim / 10 % (lng_sign * 1000)
```
jder: I don't understand the goal of this sign-related logic. It looks like these will always produce positive days/lat/lng values. Is that the goal? When do we expect the encoded floats to be negative? (Can we have them always be positive and remove this logic?)

jder (changes requested):
Just one blocking thing from me and a few suggestions/questions. If you would prefer to treat this as a hash instead and avoid having to resolve these, I'm ok with that too. (Let's just finish that conversation you started in slack about it.)

tests/conftest.py
alxmrs: TIL, thanks, this made it better.

tests/conftest.py
alxmrs: That idea is great, looks much cleaner.

tests/conftest.py
alxmrs: Thanks, this makes me reflect on the spec. I realized there is an implicit constraint -- only the lat values can be negative. I've now added more asserts and made this constraint explicit. I've also fixed the bug you found wrt to mod and sign.

tests/conftest.py
```diff
@@ -1,29 +1,203 @@
...
+        lat_sgn = np.sign(self.lat)
+        lat_sgn = np.where(lat_sgn == 0, 1, lat_sgn)  # change np.sign(0) to +1, not 0.
+        self.lat = _trunc(self.lat.astype(np.float64), 2) % (lat_sgn * 100)
+        if np.any(self.lat < -90.0) or np.any(self.lat > 90.0):
+            raise ValueError("lat values must be between -90 and 90.")
+
+        # All other values in the encoding must be non-negative.
+        self.lng = _trunc(self.lng.astype(np.float64), 1) % 1000
+        if np.any(self.lng < 0.0) or np.any(self.lng > 360.0):
+            raise ValueError("lng values must be between 0 and 360 degrees.")
```
jder, line 64: Should this test be on the self.lat/self.lng values *before* we do the trunc and mod?

jder (approved):
Awesome!

tests/conftest.py
alxmrs: I originally had that and my conclusion is -- _no_. The reason: every once in a while, we hit an error where [360.] as specified input gets interpreted as [360.09999999] and breaks the assert. I though the same thing at first, too.

jder: Hrm, I don't understand how that's possible given the constraints passed to hypothesis that has a max of 360.

alxmrs: I manually added the edge case example here, and that's what breaks this: https://github.com/suryadheeshjith/Ocean_Emulator/pull/88/files#diff-3b63cb613626e7d787b8e63ced54267d5d18966f5fbc3795812573b4e50a942dR116

And, I think that this is a regular type of thing to do / should be accommodated.

PR #90, opened 2025-03-04
A fix correctness fixes for the data source encoder.

alxmrs:
Had a discussion with @jder and we agreed that these changes would be more principled for generated test data. This avoid certain misconfiguration errors that could be confusing later on.


jder (approved):
Thanks!

PR #93, opened 2025-03-04
Trying to prevent a git error during benchmark publishing.

alxmrs:
See: https://docs.astral.sh/uv/concepts/projects/sync/#automatic-lock-and-sync


jder (approved):
As I said in slack, might want frozen instead, but fine to try.

PR #94, opened 2025-03-05
Benchmark training run

jder:
Adds a benchmark test for a mini training run for both GPU and CPU. Will put a few comments/questions below.

### DIFF
--- a/configs/eval_cm4.yaml
+++ b/configs/eval_cm4.yaml
@@ -10,7 +10,7 @@ record_every: 10
 backend: auto
 
 experiment:
-  name: eval
+  base_name: eval
   sub_name: local_cm4_samudra
   rand_seed: 15
   base_output_dir: .LOCAL
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -5,7 +5,7 @@ disk_mode: true
 pin_mem: true
 save_freq: 5
 epochs: 2
-batch_size: 2
+batch_size: 1
 learning_rate: 0.0001
 scheduler: false
 loss: mse
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -29,7 +29,7 @@ backend: auto
 
 experiment:
   base_name: train
-  sub_name: jder_test_fake_data
+  sub_name: test
   rand_seed: 15
   base_output_dir: .LOCAL
   wandb:
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -48,7 +48,7 @@ data:
   scaling_residuals_file: null
   depth_mode: all
   time_delta: 5
-  num_workers: 0
+  num_workers: 4
   hist: 0
 
 
--- a/configs/train_cm4.test.yaml
+++ b/configs/train_cm4.test.yaml
@@ -1,11 +1,11 @@
 # A config file to test training.
 # Used in automated tests.
-debug: true
+debug: false
 disk_mode: true
-pin_mem: false
+pin_mem: true
 save_freq: 5
-epochs: 120
-batch_size: 1
+epochs: 2
+batch_size: 2
 learning_rate: 0.0001
 scheduler: false
 loss: mse
@@ -28,16 +28,15 @@ step_transition: []
 backend: auto
 
 experiment:
-  name: train
-  sub_name: test_cm4_samudra
+  base_name: train
+  sub_name: jder_test_fake_data
   rand_seed: 15
   base_output_dir: .LOCAL
-  gantry: false
-  cluster_data_dir: will_be_replaced
   wandb:
     mode: disabled
-    project: 3D_ocean_emu_CM4
-    entity: testrobot
+    project: ocean-emulators
+    entity: m2lines
+    group: samudra-train-om4
   network: convnextunet
   exp_num_in: '3D_5'
   exp_num_extra: '3D_5'
@@ -49,7 +48,7 @@ data:
   scaling_residuals_file: null
   depth_mode: all
   time_delta: 5
-  num_workers: 4
+  num_workers: 0
   hist: 0
 
 
@@ -68,10 +67,6 @@ unet:
     activation: "capped_gelu"
     upscale_factor: 2
     norm: "batch"
-
-  corrector:
-    non_negative_corrector_names:
-      - "so"
 
   down_sampling_block: "avg_pool"
   up_sampling_block: "bilinear_upsample"
--- a/src/aggregator/main.py
+++ b/src/aggregator/main.py
@@ -40,6 +40,7 @@ class Aggregator:
             hist=hist,
             area_weights=area_weights,
             output_channels=output_channels,
+            record_step_20=(n_timesteps >= 20),
             log_global_mean_time_series=False,
             log_global_mean_norm_time_series=False,
         )
@@ -58,6 +59,7 @@ class Aggregator:
             hist=hist,
             area_weights=area_weights,
             output_channels=output_channels,
+            record_step_20=(n_timesteps >= 20),
             log_global_mean_time_series=True,
             log_global_mean_norm_time_series=True,
         )
--- a/src/aggregator/main.py
+++ b/src/aggregator/main.py
@@ -40,7 +40,7 @@ class Aggregator:
             hist=hist,
             area_weights=area_weights,
             output_channels=output_channels,
-            record_step_20=(n_timesteps >= 20),
+            record_step_20=(n_timesteps > 20),
             log_global_mean_time_series=False,
             log_global_mean_norm_time_series=False,
         )
@@ -59,7 +59,7 @@ class Aggregator:
             hist=hist,
             area_weights=area_weights,
             output_channels=output_channels,
-            record_step_20=(n_timesteps >= 20),
+            record_step_20=(n_timesteps > 20),
             log_global_mean_time_series=True,
             log_global_mean_norm_time_series=True,
         )
--- a/src/train_3D.py
+++ b/src/train_3D.py
@@ -58,6 +58,13 @@ class Trainer:
     model: UNet | nn.parallel.DistributedDataParallel
 
     def __init__(self, cfg: TrainConfig) -> None:
+        # Prep directory structure -- we do this first so it's set up before
+        # we try to initialize distributed training.
+        cfg.experiment.nets_dir.mkdir(parents=True, exist_ok=True)
+        cfg.experiment.output_dir.mkdir(parents=True, exist_ok=True)
+        cfg.save_yaml(str(cfg.experiment.output_dir / "config.yaml"))
+
+        # Backend
         self.device, self.distributed = init_train_backend(cfg.backend)
 
         # Adjust workers and memory pinning based on device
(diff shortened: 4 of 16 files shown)


configs/eval_cm4.yaml
```diff
@@ -10,7 +10,7 @@ record_every: 10
 backend: auto
 
 experiment:
-  name: eval
+  base_name: eval
```
jder, line 13: This was a bug fix. Previously we had

```python
def __post_init__(self):
  # ...
  self.name = f"{self.name}-{timestamp}"
```

But this means when we use `dataclasses.replace` or similar and this is called again, this adds the timestamp *again*. So, we now configure `base_name` and set `name` so it is idempotent.

configs/train_cm4.test.yaml
```diff
@@ -1,11 +1,11 @@
 # A config file to test training.
 # Used in automated tests.
-debug: true
+debug: false
 disk_mode: true
-pin_mem: false
+pin_mem: true
```
jder, line 5: These are to make this training run (slightly) more similar to the real thing.

configs/train_cm4.test.yaml
```diff
@@ -28,16 +28,15 @@ step_transition: []
...
   base_output_dir: .LOCAL
-  gantry: false
-  cluster_data_dir: will_be_replaced
   wandb:
     mode: disabled
-    project: 3D_ocean_emu_CM4
-    entity: testrobot
+    project: ocean-emulators
+    entity: m2lines
+    group: samudra-train-om4
```
jder, line 39: We should decide if we want this to log to wandb in CI. I think probably yes? Perhaps with a sub_name of the commit id or something? Thoughts?

configs/train_cm4.test.yaml
```diff
@@ -69,9 +68,5 @@ unet:
     upscale_factor: 2
     norm: "batch"
 
-  corrector:
-    non_negative_corrector_names:
-      - "so"
-
```
jder, line 75: This was causing me trouble, not sure it matters if we have it or not?

src/aggregator/main.py
```diff
@@ -40,6 +40,7 @@ def get_inline_inference_aggregator(
             hist=hist,
             area_weights=area_weights,
             output_channels=output_channels,
+            record_step_20=(n_timesteps > 20),
```
jder, line 43: Previously if we rolled out for fewer than 20 steps, this would crash later because no data was recorded by the aggregator created by this being True.

src/train_3D.py
```diff
@@ -92,13 +92,13 @@ def __init__(self, cfg: TrainConfig) -> None:
...
+        str_in = ", ".join([i for i in self.inputs])
+        str_ext = ", ".join([i for i in self.extra_in])
+        str_out = ", ".join([i for i in self.outputs])
 
-        logging.info(f"inputs: {self.str_in}")
-        logging.info(f"extra inputs: {self.str_ext}")
-        logging.info(f"outputs: {self.str_out}")
+        logging.info(f"inputs: {str_in}")
+        logging.info(f"extra inputs: {str_ext}")
+        logging.info(f"outputs: {str_out}")
```
jder, line 108: I was trying to debug and found this format kind of hard to parse. It was previously like `uo_0_uo_1_uo2...` and is now `uo_0, uo_1, uo_2...`. If this seems good, happy to make this change to eval too.

src/train_3D.py
```diff
@@ -253,6 +253,15 @@ def __init__(self, cfg: TrainConfig) -> None:
...
 
+        # Outputs
+        if not os.path.exists(cfg.experiment.nets_dir):
+            os.makedirs(cfg.experiment.nets_dir, exist_ok=True)
+
+        if not os.path.exists(cfg.experiment.output_dir):
+            os.makedirs(cfg.experiment.output_dir, exist_ok=True)
+
+        cfg.save_yaml(cfg.experiment.output_dir / "config.yaml")
+
```
jder: This was previously happening in main() instead of here, but we also want this when running a training run for test purposes. Any reason we shouldn't have it happen all the time (here)? Again if this makes sense, happy to make the analogous change for eval.

src/utils/device.py
```diff
@@ -11,6 +11,7 @@ def using_gpu() -> bool:
 # by moving to some training framework + context objects once we have
 # benchmarking set up.
 def set_device(device: torch.device) -> None:
+    global _CHOSEN_DEVICE
     _CHOSEN_DEVICE = device
```
jder, line 15: This was silently doing nothing. Great job, jder.

tests/conftest.py
```diff
@@ -261,7 +261,7 @@ def data_source() -> DataSource:
     }
     ds = xr.Dataset(vars_2d | vars_3d | masks, coords=coords)
 
-    return DataSource(data=ds, means=ds.mean("time"), stds=ds.std("time"))
+    return DataSource(data=ds, means=ds.mean(), stds=ds.std())
```
jder, line 264: These were the wrong shape when we went to use them at training time -- the means/stds appear to be across all axes.

configs/eval_cm4.yaml
alxmrs: Nice! Good catch. Sounds like debugging this config option will be better in the future because of this change.

configs/train_cm4.test.yaml
alxmrs: I'm curious how this will affect the CPU benchmarks. Is there some way we can set this flag based on the device fixture? Or, is that the current behavior?

jder: Yes, the train code already checks to see if the device is CPU and if so makes this False. 

https://github.com/suryadheeshjith/Ocean_Emulator/blob/5f0685bc93824c6020b68d0500e4950215fd61ec/src/train_3D.py#L64-L66

configs/train_cm4.test.yaml
```diff
@@ -1,11 +1,11 @@
...
-debug: true
+debug: false
 disk_mode: true
-pin_mem: false
+pin_mem: true
 save_freq: 5
-epochs: 120
-batch_size: 1
+epochs: 2
+batch_size: 2
```
alxmrs: IIUC, this will change the benchmarks. That's ok, but it may be hard to compare. 

Request: can you check if this will make a big difference by running a local benchmark run on your machine (say, comparing main and this branch)?

configs/train_cm4.test.yaml
alxmrs: Awesome, this is great.

configs/train_cm4.test.yaml
alxmrs: I have to admit, I haven't used wandb before. What kinds of things do we hope to measure by logging CI runs?

src/train_3D.py
alxmrs: I love tidy logs, thanks for this.

configs/train_cm4.test.yaml
jder: I was thinking of things like memory usage, time spent training per epoch. You can see something like what this will log here https://wandb.ai/m2lines/ocean-emulators/runs/2vm82o1g. At the moment this will create 5 separate runs (since the benchmark runs 5 times) which is maybe too much noise. Maybe we set it to be a single run?

src/train_3D.py
```diff
@@ -253,6 +253,15 @@ def __init__(self, cfg: TrainConfig) -> None:
             cfg.experiment.wandb.mode == "online", is_main_process()
         )
 
+        # Outputs
+        if not os.path.exists(cfg.experiment.nets_dir):
+            os.makedirs(cfg.experiment.nets_dir, exist_ok=True)
```
alxmrs: Are these already `pathlib.Path`s? If so, we can probably replace this with a one liner, see: https://stackoverflow.com/a/273227

src/utils/device.py
alxmrs: Lool! I was going to comment on that in the original PR, but I figured it was working. Great job, alxrms, too.

tests/conftest.py
alxmrs: Thanks, good catch.

tests/test_datasets.py
```diff
@@ -212,10 +212,22 @@ def test_val__data_shape(val_loader_pair: LoaderPair):
...
+        assert X.shape[1] == batch_size or (
+            i == num_samples - 1 and X.shape[1] < batch_size
+        )
+        assert X.shape[2:] == (input_var_dim, 180, 360)
+
+        assert y.shape[0] == cfg.steps[0]
+        assert y.shape[1] == batch_size or (
+            i == num_samples - 1 and y.shape[1] < batch_size
+        )
+        assert y.shape[2:] == (output_var_dim, 180, 360)
```
alxmrs, line 230: Thanks for doing this math!

tests/test_datasets.py
```diff
@@ -230,7 +242,7 @@ def test_inference__data_shape(inference_loader_pair: LoaderPair):
     cfg, loader = inference_loader_pair
 
     exp = cfg.experiment
-    batch_size = cfg.batch_size
+    batch_size = 1  # Inference always uses batch size 1
```
alxmrs, line 245: nice, and thanks for the comment.

tests/test_trainer.py
```diff
@@ -0,0 +1,16 @@
+import logging
+
+import pytest
+
+from config import TrainConfig
+from train_3D import Trainer
+
+
+@pytest.mark.manual
+def test_trainer__mini_benchmark(train_config: TrainConfig, caplog, benchmark):
```
alxmrs, line 10: TIL about the `caplog` fixture

configs/train_cm4.test.yaml
alxmrs: 🟡

alxmrs (approved):
LGTM! Good work getting training runs tested!

I just realized: I exclude CUDA marked tests from the benchmarking page -- can totally be fixed in a future PR.

configs/train_cm4.test.yaml
jder: You're just concerned about the batch size? I can put that back for now -- maybe we want separate configs for the training run as we try to make it more realistic.

configs/eval_cm4.yaml
jder: Yeah, in particular it means our tests were all set to have names with the timestamp twice since we use dataclasses.replace to override some config.

configs/train_cm4.test.yaml
alxmrs: Sorry, I get a 404 when I click the link. 
That sounds like a good use case -- and nice it's also a similar infra for debugging training runs. 

Maybe we could use wandb specific fixtures? The 5x runs do sound annoying in the tool, but I think they're useful in the benchmarking tool. 

(on a similar note: I've used memray in the past to measure memory --  https://bloomberg.github.io/memray/overview.html)

configs/train_cm4.test.yaml
alxmrs: Now that I think of it, the default test config I chose is pretty arbitrary. And, we're pretty early to setting the instrumentation up. If you think this is more realistic, then I'm happy to include it. There have been no major efforts to change performance merged yet. So, we're still able to measure improvements.

jder: I think it is very slightly more realistic :) -- will leave for now.

alxmrs: SGTM

jder: (I ended up reverting it because it breaks a test; can resolve when we go to make a more realistic run)

PR #97, opened 2025-03-05
Added some typing definitions (array and otherwise).

alxmrs:
A pre-requisite step in refactoring the data loader (for me, anyway) is just understanding what's going on. To that end, I've added beartype-enforced jaxtyping and standard typing definitions to parts of the codebase that the data loader touches. 

This does not aim for complete typing coverage of the codebase. Rather, these annotations have been added so I can feel confident in pulling array parsing routines up and down. Besides serving as useful documentation, these perform runtime type checks to ensure correctness.


src/datasets.py
```diff
@@ -1,15 +1,28 @@
...
-from typing import Dict, List, Tuple
 
 import numpy as np
 import torch
 import xarray as xr
 from einops import rearrange
+from jaxtyping import Float
 from torch.utils.data import Dataset
 
+from constants import (
```
alxmrs, line 10: Question for the reviewer: Should I introduce the following pattern to the codebase?
```python
import constants as c

def foo(x: c.ExtraVars) -> c.Extra:
    pass
```

src/constants.py
```diff
@@ -3,6 +3,22 @@
...
 import xarray as xr
+from jaxtyping import Bool, Float
+
+# Common Type Aliases
+# See "Existing jaxtyping annotations" section of
+#  https://docs.kidger.site/jaxtyping/api/array/#array
+Grid = Float[torch.Tensor, "180 360"]
+Input = Float[Grid, "input_vars"]  # equivalent to "input_vars lat lon"
+Extra = Float[Grid, "extra_vars"]
+TotalInput = Float[Grid, "total_vars"]
```
jder: Should this be `TotalInput = Float[Grid, "input_vars+extra_vars"]`?

jder:
Thanks, I started looking into this but I think I'm missing a piece of the high level here -- what's the goal of this? Is it just documentation? I was playing around with these annotations and couldn't get them to produce errors from mypy when shapes didn't match. I was also having some confusing errors from ruff and mypy when I tried to put these annotations elsewhere. [This](https://docs.kidger.site/jaxtyping/faq/#is-jaxtyping-compatible-with-static-type-checkers-like-mypypyrightpytype) kind of suggests that this is really for runtime type checking, not compile-time, but it doesn't look like we're using any of those.

alxmrs:
I'll give a longer response, but from your feedback, I did find a pretty good entrypoint for runtime checking via pytest: https://docs.kidger.site/jaxtyping/api/runtime-type-checking/#pytest-hook

alxmrs:
Runtime type-checking aside, you're right that the larger point of this is for documentation. It's documentation basically for me alone to help make cuts in refactoring in the codebase. With these annotations, I can better do things like replace `torch.Tensor`s with `xarray.DataArray`s or add another dimension to a collection of data and vectorize an operation (looking at you, `TrainData`), etc.

alxmrs:
Update: by adding a decorator (as documented), I am able to get runtime type checks in tests!

pyproject.toml
```diff
@@ -40,7 +41,7 @@ find = {}
 
 [tool.ruff]
 lint.select = [ "D", "E", "F", "I", "W" ]
-lint.ignore = [ "D1", "D200", "D205", "D212", "E203", "E402", "F541", "W293" ]
+lint.ignore = [ "D1", "D200", "D205", "D212", "E203", "E402", "F541", "F722", "W293" ]
```
jder, line 45:
```suggestion
# F722: https://docs.kidger.site/jaxtyping/faq/#flake8-or-ruff-are-throwing-an-error
lint.ignore = [ "D1", "D200", "D205", "D212", "E203", "E402", "F541", "F722", "W293" ]
```

src/constants.py
```diff
@@ -3,6 +3,23 @@
...
+Label = Float[Grid, "output_vars"]
+BatchInput = Float[Input, "batch"]  # equivalent to "batch input_vars lat lon"
+BatchExtra = Float[Extra, "batch"]
+BatchTotalInput = Float[TotalInput, "batch"]
+BatchLabel = Float[Label, "batch"]
```
jder: Like I said live, I kinda feel like it might just be noise to have batch versions of all these since many tensors flowing around have a batch dimension sometimes. Maybe adding `*batch` to Input/Extra/etc would be less noise and approximately as helpful? Up to you.

src/datasets.py
jder: I personally prefer importing the names. "constants." or "c." doesn't seem super helpful/meaningful to me.

src/datasets.py
```diff
@@ -211,35 +227,36 @@ def __getitem__(self, idx):
...
 class TrainData:
     def __init__(self, output_channels: int):
-        self.td_dict: Dict[int, Tuple[torch.Tensor, torch.Tensor]] = {}
+        self.td_dict: dict[int, Example] = {}
         self.output_channels = output_channels
         self.steps = 0
 
-    def insert(self, input: torch.Tensor, label: torch.Tensor):
-        self.td_dict[self.steps] = (input, label)
+    @jaxtyped(typechecker=typechecker)
```
jder: This seems pretty noisy, could we instead [automatically apply it to all functions in tests](https://docs.kidger.site/jaxtyping/api/runtime-type-checking/#pytest-hook) or similar?

src/constants.py
```diff
@@ -73,6 +94,7 @@
     "mask_18",
 ]
 
+InputVars = list[str]
 INPT_VARS = {
```
jder: Maybe makes sense to have `INPT_VARS: dict[str, InputVars] = `? 

Personally I might instead prefer to use `VarName = str` and then use `list[VarName]` in the various places below where you have `InputVars` or `ExtraVars` or `OutputVars` below. The "input"ness isn't really part of the *type* of these variables, it's part of the *value* of them. ie xr.Dataset is approximately a `dict[list[VarName], xr.DataArray]` so eg `data[input_vars + extra_vars]` makes perfect sense whereas if we think of these as separate types it's unclear what type `input_vars + extra_vars` is.

src/utils/data.py
```diff
@@ -29,18 +45,21 @@ def extract_wet_mask(data, outputs, hist):
     wet_inp = torch.from_numpy(wet_mask_np[depth_ind])
     wet_surface = torch.from_numpy(wet_surface_mask_np)
     wet_inp = torch.concat([wet_inp] * (hist + 1), dim=0)
-    return wet_inp, wet_surface
+    return wet_inp.bool(), wet_surface.bool()
```
jder, line 45: Why this change? I think Surya was saying that they're morally bools but in practice integers?

jder (changes requested):
Thank you for this, nothing makes me happier than adding types to python code. Just a few questions/comments/suggestions. I would really love to not have to `@jaxtyping` everything but the rest are all questions or subjective.

src/datasets.py
alxmrs: I tried to do this first, but I couldn't get it to work. Maybe we could pair on it and see if we can figure it out together?

src/utils/data.py
alxmrs: Argubaly, this is an error that was caught by the type checker. While these are returned as int arrays, they only ever have the values [0, 1]. Later on in the codebase, these two lines are called just like this to convert them to bools. Since they are only ever used as bools (we have type enforcement to guarantee this fact as well as anecdata from stepping through with my debugger), it seemed safe to change this. 

@suryadheeshjith: Am I missing anything by converting this to bools early rather than late?

src/constants.py
alxmrs: `xr.Dataset` is more like a `dict[VarName, xr.DataArray]`, though I don't meant to be nit-picky. I agree, in general, that we should choose a type definition with respect to how these values are used in the codebase. 

I also agree that `data[input_vars + extra_vars]` is indicative of the usage pattern. For this reason, I think that the type alias should be as it is now, or `list[str]`, rather than `list[InputVar]`. My mental model is: these are a selection of columns of data, or a subset of a `xr.Dataset`. (They could literally be `set[str]`, except I believe that we want them to be ordered.) Since they're used as a selection operation short-hand with respect to a tracked experiment, I think it makes sense to think of these things as a whole collection, albeit, a combinable one.

Either way, I like your first idea to annotate the config dictionaries as well.

src/constants.py
alxmrs: Good suggestion! It is much simpler/elegant now.

src/constants.py
alxmrs: Good shout; see comment above.

src/datasets.py
alxmrs: SGTM.

src/datasets.py
alxmrs: Actually, I figured it out!

src/datasets.py
```diff
@@ -1,15 +1,31 @@
...
+    InputVars,
+    Label,
+    OutputVars,
+    TotalInput,
+)
 from utils.data import Normalize
 from utils.device import get_device, using_gpu
 
+TDInput: TypeAlias = TotalInput | Float[TotalInput, "*batch"]
+TDLabel: TypeAlias = Label | Float[Label, "*batch"]
```
jder:
```suggestion
TDInput: TypeAlias = Float[TotalInput, "*batch"]
TDLabel: TypeAlias = Float[Label, "*batch"]
```

I think the * means "0 or more" so this is equivalent?

src/datasets.py
```diff
@@ -211,35 +227,31 @@ def __getitem__(self, idx):
...
     def __init__(self, output_channels: int):
-        self.td_dict: Dict[int, Tuple[torch.Tensor, torch.Tensor]] = {}
+        self.td_dict: dict[int, Example] = {}
         self.output_channels = output_channels
         self.steps = 0
 
-    def insert(self, input: torch.Tensor, label: torch.Tensor):
-        self.td_dict[self.steps] = (input, label)
+    def insert(self, input_: TDInput, label: TDLabel):
+        self.td_dict[self.steps] = (input_, label)
```
jder, line 232: Do we still need this underscore?

jder (approved):
Looks great, thanks!

src/datasets.py
alxmrs: This is unrelated to adding typing. But `input` is a default function in python (requests user input). Thus, it shouldn't generally be used for a variable name.

src/datasets.py
alxmrs: Oh! I didn't know that. I think I misunderstood your suggestion, in that case. Then, let me see if I can make all the rules simpler!

src/datasets.py
jder: Oh, of course. That bites me all the time. Thanks for the fix :)

alxmrs:
With my latest change (which looks reasonable), I get this weird internal jaxtyping error:

```
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/Users/alxmrs/git/Ocean_Emulator/.venv/lib/python3.10/site-packages/_pytest/main.py", line 279, in wrap_session
INTERNALERROR>     config._do_configure()
INTERNALERROR>   File "/Users/alxmrs/git/Ocean_Emulator/.venv/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1118, in _do_configure
INTERNALERROR>     self.hook.pytest_configure.call_historic(kwargs=dict(config=self))
INTERNALERROR>   File "/Users/alxmrs/git/Ocean_Emulator/.venv/lib/python3.10/site-packages/pluggy/_hooks.py", line 535, in call_historic
INTERNALERROR>     res = self._hookexec(self.name, self._hookimpls.copy(), kwargs, False)
INTERNALERROR>   File "/Users/alxmrs/git/Ocean_Emulator/.venv/lib/python3.10/site-packages/pluggy/_manager.py", line 120, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>   File "/Users/alxmrs/git/Ocean_Emulator/.venv/lib/python3.10/site-packages/pluggy/_callers.py", line 139, in _multicall
INTERNALERROR>     raise exception.with_traceback(exception.__traceback__)
INTERNALERROR>   File "/Users/alxmrs/git/Ocean_Emulator/.venv/lib/python3.10/site-packages/pluggy/_callers.py", line 103, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/Users/alxmrs/git/Ocean_Emulator/.venv/lib/python3.10/site-packages/jaxtyping/_pytest_plugin.py", line 53, in pytest_configure
INTERNALERROR>     raise RuntimeError(message.format(", ".join(already_imported_packages)))
INTERNALERROR> RuntimeError: jaxtyping cannot check these packages because they are already imported: datasets
```

alxmrs:
Filed an issue into jaxtyping: https://github.com/patrick-kidger/jaxtyping/issues/306

PR #98, opened 2025-03-06
Ensuring uniqueness property of interpretable data.

alxmrs:
Also, added a minor equality fix which tolerates noisier data. Luckily, the uniqueness property is still preserved.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -74,8 +74,8 @@ class DataSourceDims:
         # So, we declare arrays as equal within a generous tolerance.
         # We can use exact equality with days_since_start since they are ints.
         return (
-            np.allclose(self.lat, other.lat, rtol=1)
-            and np.allclose(self.lng, other.lng, rtol=0.1)
+            np.allclose(self.lat, other.lat, atol=0.1)
+            and np.allclose(self.lng, other.lng, atol=0.01)
             and np.array_equal(self.days_since_start, other.days_since_start)
             and self.start_day == other.start_day
         )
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -74,7 +74,7 @@ def __eq__(self, other) -> bool:
         # So, we declare arrays as equal within a generous tolerance.
         # We can use exact equality with days_since_start since they are ints.
         return (
-            np.allclose(self.lat, other.lat, rtol=0.1)
+            np.allclose(self.lat, other.lat, rtol=1)
```
jder: `rtol` of 1 means that for a lat of 90, we would accept anything in the range +/- `abs(90) * 1` -- ie [0, 180]. Maybe this should be `atol=0.1` or something?

alxmrs: I thought so too. When configured this way, hypothesis finds an error case: `0.06` != `0.05` for `lat`. Am I misinterpreting this error, and thus, the solution to it? 

Maybe this is your point: what's the difference between `atol` and `rotl`?

jder: The np docs describe them as:

> The relative difference (rtol * abs(b)) and the absolute difference atol are added together to compare against the absolute difference between a and b.

So atol means "absolute difference permitted" and rtol means "difference permitted relative to the size of the value". Since we know the scales + size of errors expected, we could use either but atol seems more direct + interpretable to me. With an atol of 0.1, `0.06` and `0.05` should be within tolerance.

alxmrs: TIL! Thank you, that was the real bug!

jder (approved):
Thanks!

Issue #87, opened 2025-03-03
Can't run CUDA and non-CUDA tests together


jder:
We can't create multiple Trainers within the same Python process due to re-initialization of singletons. Create a context object and pass them around instead, or remove them by using a training framework (for device in particular, once we have benchmarks set up).

PR #99, opened 2025-03-06
fix benchmarks when run together

jder:
Sorry, I clearly [did not test](https://github.com/suryadheeshjith/Ocean_Emulator/actions/runs/13708924698/job/38341018074) the benchmarks well enough when run together with the new trainer benchmark. 

Today we can't create more than one trainer in a process. It's on my plate to fix (https://github.com/suryadheeshjith/Ocean_Emulator/issues/87) but in the meantime let's unbreak the benchmark by re-using the existing trainer.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,6 +12,7 @@ from typing_extensions import Self
 
 import constants as c
 from config import TrainBackendConfig, TrainConfig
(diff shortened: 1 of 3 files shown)


alxmrs (approved):
LGTM!

PR #100, opened 2025-03-07
fix training when run by main script

jder:
Turns out the handle_logging requires the directory to be there first, so create them first (but also try again to recreate them in the constructor)

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -196,6 +196,11 @@ class TrainConfig:
         with open(save_path, "w") as f:
             yaml.dump(config_dict, f, default_flow_style=False)
 
+    def prepare_output_dirs(self) -> None:
+        self.experiment.nets_dir.mkdir(parents=True, exist_ok=True)
+        self.experiment.output_dir.mkdir(parents=True, exist_ok=True)
+        self.save_yaml(str(self.experiment.output_dir / "config.yaml"))
+
 
 # See backend.py for how these are turned into concrete devices
 EvalBackendConfig = Literal["cpu", "cuda", "auto"]
@@ -264,3 +269,6 @@ class EvalConfig:
 
         with open(save_path, "w") as f:
             yaml.dump(config_dict, f, default_flow_style=False)
+
+    def prepare_output_dirs(self) -> None:
+        self.experiment.output_dir.mkdir(parents=True, exist_ok=True)
--- a/src/config.py
+++ b/src/config.py
@@ -199,7 +199,6 @@ class TrainConfig:
     def prepare_output_dirs(self) -> None:
         self.experiment.nets_dir.mkdir(parents=True, exist_ok=True)
         self.experiment.output_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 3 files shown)


src/config.py
```diff
@@ -196,6 +196,10 @@ def save_yaml(self, save_path: str):
         with open(save_path, "w") as f:
             yaml.dump(config_dict, f, default_flow_style=False)
 
+    def prepare_output_dirs(self) -> None:
```
alxmrs, line 199: I have a crazy idea that I think would be really simple. What if we move this to the  `ExperimentConfig` dataclass `__post_init__` method? When reading how this function (and the one below) is used in practice, it is called right before read/write (or around object instantiation). Further, both of these top level classes are modified the internal `experiment` data class -- often, the same values of the dataclass. 

It seems like to me that there actually is a pre-condition that is coupled to `ExperimentConfig` (or experiments) that the generated directories must exist. Thus, I think it this bit of logic should belong to that config dataclass (and the side-effect should be documented in that dataclass).

jder: I don't think that works because those dataclasses are often loaded from a not-really-real config and then modified as in your test config setup. Both the initial load and later modifications call `__post_init__`, so it would create the directories before we had a chance to modify it.

alxmrs: I just checked all the instances of `output_dir` and `nets_dir` in my IDE. It looks like after they are set in the experiment `__post_init__`, they are never mutated, only referenced. I believe that means this is safe (especially since creating the directory is idempotent to some degree)?

I think I must be misunderstanding something: Are you saying `ExperimentConfig` is initialized multiple times via the test config set up? Can you explain this more?

jder: Yes, I'm saying they are initialized multiple times. We have [this code](https://github.com/suryadheeshjith/Ocean_Emulator/blob/97f2bb5015b51f43ffffd5435c924eb8b1438cd4/tests/conftest.py#L284-L294) right now:

```python
        trainer = TrainConfig.from_yaml(default_config)
        data_config = dataclasses.replace(
            trainer.data,
            data_path=_make_path("data.zarr"),
            data_means_path=_make_path("means.netcdf"),
            data_stds_path=_make_path("stds.netcdf"),
        )
        experiment_config = dataclasses.replace(
            trainer.experiment,
            cluster_data_dir=os.path.join(tmpdir, "cluster_data"),
        )
```

In the first line, it creates a TrainConfig and the nested ExperimentConfig. This calls the ExperimentConfig's `__post_init__`. We then call `dataclasses.replace`, passing in that ExperimentConfig instance, which creates a new ExperimentConfig with a different cluster_data_dir and calls `__post_init__` on that. 

Agreed creating the directories is idempotent if the directory doesn't change, but given this pattern it seems very possible the directories do in fact change when we update/override the config before finally passing it to Trainer.

alxmrs: Cool, thanks for the discussion on this. I'm convinced my suggestion is only neat but not, in fact, the right way to go.

alxmrs (approved):
LGTM -- happy 100th PR!

PR #107, opened 2025-03-11
Simple utility script to clone Samudra data locally.

alxmrs:
Clones OM4 in about 600s using ~100MB chunks. Saves means and stds as NetCDF. Is packaged as a `uv` script (it self-describes dependencies).

This change is in service of setting up remote GPU runs and chunk tuning.

### DIFF
--- a/scripts/clone_data.py
+++ b/scripts/clone_data.py
@@ -34,7 +34,7 @@ def main(dest_root: pathlib.Path, time_slice: slice) -> None:
                 DATA_ROOT + name, engine="zarr", chunks={"time": 700}
             )
             data = data.isel(time=time_slice)
-            data.chunk({"time": 1})
+            data = data.chunk({"time": 1})
         else:
             data = xr.open_dataset(DATA_ROOT + name, engine="zarr", chunks={})
 
--- a/scripts/clone_data.py
+++ b/scripts/clone_data.py
@@ -2,8 +2,10 @@
 # /// script
 # requires-python = ">=3.10"
 # dependencies = [
-#     "xarray[io]",
-#     "dask",
+#   "xarray[io]",
+#   "dask",
+#   "requests",
+#   "aiohttp",
 # ]
 # ///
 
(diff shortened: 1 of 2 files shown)


suryadheeshjith (approved):
LGTM! I am curious as to why exactly we are saving means and stds as netcdf files

scripts/clone_data.py
```diff
@@ -0,0 +1,70 @@
...
+        ("OM4_stds", "netcdf"),
+    ]:
+        dest = dest_root / name
+
+        if name == "OM4":
+            data = xr.open_dataset(
+                DATA_ROOT + name, engine="zarr", chunks={"time": 700}
+            )
+            data = data.isel(time=time_slice)
+            data.chunk({"time": 1})
```
suryadheeshjith:
```suggestion
            data = data.chunk({"time": 1})
```

chunk() is not an in-place operation

suryadheeshjith: Based on my tests with it, it seems to slow down saving when we rechunk with time=1

jder:
@alxmrs what version of uv are you using? It seems like we keep getting this trivial change back and forth in our uv.lock and I think it must be mismatching of versions. I'm running uv 0.6.5.

jder:
@suryadheeshjith all the other scripts in this directory start with a `.` -- is that historical or is there some reason this one should too?

jder (approved):
LGTM after Surya's feedback.

suryadheeshjith:
No reason, just to indicate it is a bash script

alxmrs:
I just committed surya's suggestion; going to test it out now (on my new dev machine! 😀 ).

alxmrs:
>  I am curious as to why exactly we are saving means and stds as netcdf files

I though that the train pipeline expects that these are nc files -- am I wrong about this?

scripts/clone_data.py
alxmrs: That would make sense to me! Though, changing the chunks at rest is a significant data engineering change. I'd rather save stuff that like for the future.

suryadheeshjith:
~~It need not be but it can!~~ I am mistaken, it does need to be netcdf files. I am getting confused working with a million repositories. sorry.

alxmrs:
>  I'm running uv 0.6.5.

Yeah, I've been having some weird uv differences. I just did a fresh install of `uv` on my remote VM, and it is version: `uv 0.6.6`

alxmrs:
Just ran this script on a fresh VM. The specs have changed a bunch, so it's not quite a fair comparison, but this still downloads all the data in about 10m or 600s.

Issue #106, opened 2025-03-11
Tune `open_zarr()` chunks parameter, providing good defaults.


alxmrs:
I'm working on a simple PR that will make the `chunks={}` argument configurable. From meeting with @jder yesterday, it seems like `chunks=dict(time=700)` should lead to some improvement; how much is a matter of measuring properly. To that end, in addition to parameterizing this, I'm messing a bit with the trainer's logging system.

In the future, this could be a hparam that we perform sweeps on (which also is dependent on the hw the trainer runs on).

jder:
Starting with some context: profiling the om4 training run on a `a2-highgpu-1g` with 1 A100 shows that each training step spends ~2/3 of its time waiting for data loaders (using 4 data loaders). https://openathena.slack.com/archives/C08CYM42DT3/p1741385964346899

As a first cut, we tried comparing:
* [Baseline](https://github.com/suryadheeshjith/Ocean_Emulator/tree/u/jder/experiments/profile-2025-03-11) with a single in-thread data loader, training on a single `a2-highgpu-1g` instance on the om4 data, one epoch of 16 steps. 
* Same as baseline, but in the open_zarr calls, passing `chunks=dict(time="100mb")` to try and get it to load more data at once.
* Same as baseline but pointing at underling zarr files rechunked to have chunks with t=700. 

Result was
* Baseline: train time 131.89s
* t100mb on read: train time 129.23s
* t=700 rechunked: train time 130.30
* Baseline again: train time 128.08

alxmrs:
Is train time the right metric? I think we want to be measuring the length of the delay before the first data allocation on to the GPU (or something similar). 

BTW, can I fork off of your branch?

jder:
We can certainly measure that but wouldn’t you expect overall train time to go down too? (Which seems more directly like the thing that matters). Feel free to fork off the branch.

alxmrs:
I might expect that. How is train time measured? To me, it seems to not include any data loading / preparation stuff. e.g., it's not recording any of the GPU idle time.

jder:
This is the train time -- it's wall-clock time for the epoch https://github.com/suryadheeshjith/Ocean_Emulator/blob/3eb5b3bb7d123cfc0adcd72a5264fa0d68b25c5e/src/train_3D.py#L416-L418

alxmrs:
ah. This seems to confirm my suspicions. I don't think it's capturing the idling GPU.

suryadheeshjith:
[...]

PR #108, opened 2025-03-12
Upgrade to python 3.11, zarr 3, and various other updates

jder:
I originally did this for performance reasons (which [didn't pan out yet](https://github.com/suryadheeshjith/Ocean_Emulator/issues/106#issuecomment-2717976095)) but seems like it's healthy to do this anyway.


.github/workflows/benchmarks.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           # It is easiest if this cuda version (cu121) matches the cuda version below.
           # (We could relax this since installing torch below pulls in the right version of cuda
           # libraries, but then you have a mismatch between the cuda toolkit and those libraries, which seems possibly bad.)
-          image: pytorch-2-2-cu121-v20250205-ubuntu-2204-py310-conda
+          image: pytorch-2-2-cu121-v20250205-ubuntu-2204-conda
```
jder, line 33: This is actually still python 3.10 (we install 3.11 below anyway) but seemed weird to explicitly request the wrong version.

.github/workflows/benchmarks.yml
```diff
@@ -64,14 +64,14 @@ jobs:
...
+          python-version: "3.11"
 
       # once we've solved https://github.com/suryadheeshjith/Ocean_Emulator/issues/87 we can make a single pytest run
       - name: Run CPU benchmarks
-        run: uv run --frozen pytest -m "not cuda" --benchmark-only --benchmark-json output-cpu.json
+        run: uv run --locked pytest -m "not cuda" --benchmark-only --benchmark-json output-cpu.json
 
       - name: Run GPU benchmarks
-        run: uv run --frozen pytest -m "cuda" --benchmark-only --benchmark-json output-gpu.json
+        run: uv run --locked pytest -m "cuda" --benchmark-only --benchmark-json output-gpu.json
```
jder, line 74: Pretty sure we want `--locked` which errors out if the lock file is out of date rather than `--frozen` doesn't even check if the lock file is out of date. (Neither will edit the lock file.)

pyproject.toml
```diff
@@ -5,23 +5,38 @@ requires = [ "setuptools", "wheel" ]
...
+  "jaxtyping>=0.2.38",
+  "matplotlib>=3.10.1",
+  "numpy>=1.26.4,<2",     # There is some incompatibility between a dependency and numpy 2+
+  "pyyaml>=6.0.2",
+  "torch~=2.2.1",
+  "torchinfo>=1.8",
+  "wandb>=0.19.8",
+  "xarray[io]>=2025.1.2",
+  "zarr>=3.0.5",
+]
```
jder, line 39: Open to feedback about whether I should be more or less specific with these requirements, this is the default `uv` uses except for numpy (as commented) and torch (which I think we probably want to test specifically on upgrade)

src/datasets.py
```diff
@@ -365,7 +365,7 @@ def __getitem__(self, idx: int):
 
     def _get_x_index(
         self, idx: int, step: int, prev_rolling_idx: int | None
-    ) -> Integer[xr.Variable, "window_dim time"]:
+    ) -> xr.Variable:
```
jder, line 368: While theoretically jaxtyping is supposed to support anything with `.shape` and `.dtype`, in practice it seems to have a name collision with [`NamedArray._check_shape`](https://github.com/pydata/xarray/blob/852ad1135b9b8c46d8e301c9badbe898bd6d262d/xarray/namedarray/core.py#L535) which is a parent class of `xr.Variable`. I'll file a bug with jaxtyping.

tests/conftest.py
```diff
@@ -41,9 +41,15 @@ class DataSourceDims:
...
-    days_since_start: NDArray[np.int32] = np.array([0, 5, 10])
+    lat: NDArray[np.float64] = dataclasses.field(
+        default_factory=lambda: np.arange(-89.24, 90, 1, dtype=np.float64)
+    )
+    lng: NDArray[np.float64] = dataclasses.field(
+        default_factory=lambda: np.arange(0.5, 360, 1, dtype=np.float64)
+    )
+    days_since_start: NDArray[np.int32] = dataclasses.field(
+        default_factory=lambda: np.array([0, 5, 10])
+    )
```
jder, line 52: Python 3.11 rejects this use of mutable default values, so wrapped in factory functions.

.github/workflows/benchmarks.yml
alxmrs: Lol, that SGTM

.github/workflows/benchmarks.yml
alxmrs: 🐑  IIUC we can omit `--locked` or `--frozen` because of the `git stash` step right after this.

.python-version
```diff
@@ -1 +1 @@
-3.10
\ No newline at end of file
+3.11
```
alxmrs, line 1: Wohoo! 🚀

pyproject.toml
```diff
@@ -5,23 +5,38 @@ requires = [ "setuptools", "wheel" ]
 
 [project]
 name = "ocean-emulators"
+version = "2025.1.0"
```
alxmrs: We prefer calver over semver? (I prefer the latter). 

Minor warning: the guy who convinced a lot of pydata projects to use calver has since shown some hesitations with it: https://jacobtomlinson.dev/posts/2023/sometimes-i-regret-using-calver/

I could see an argument that we're more like a distribution than a library. What are your thoughts on this?

pyproject.toml
```diff
@@ -5,23 +5,38 @@ requires = [ "setuptools", "wheel" ]
...
+dependencies = [
+  "cftime>=1.6.4.post1",
+  "dacite>=1.9.2",
+  "dask>=2025.2",
+  "einops>=0.8.1",
+  "jaxtyping>=0.2.38",
+  "matplotlib>=3.10.1",
+  "numpy>=1.26.4,<2",     # There is some incompatibility between a dependency and numpy 2+
+  "pyyaml>=6.0.2",
+  "torch~=2.2.1",
```
alxmrs, line 34: `~fancy~`!

pyproject.toml
alxmrs: Looks great so far!

pyproject.toml
```diff
@@ -54,9 +79,7 @@ markers = [
   "manual: marks tests as manually run (deselect with '-m \"not manual\"').",
   "cuda: marks tests that need a CUDA-enabled device to run (deselect with '-m \"not cuda\"').",
 ]
-pythonpath = [
-  "src",
-]
+pythonpath = [ "src" ]
```
alxmrs, line 82: While we're here: do you also want to create an `oceans_emulator` package under `src`?

alxmrs: Happy to do this in a future PR, too.

src/datasets.py
alxmrs: Nice! Cool bug!

tests/conftest.py
alxmrs: That's definitely better! Thank you.

alxmrs (approved):
Much better, thanks for doing this! 

I'd love to have a conversation about semver vs calver (I prefer semver!). Though, the setting is different for ML experimentation. It seems pretty reversible, however, since we're not distributing packages right now.

pyproject.toml
suryadheeshjith: I prefer semantic versioning as well! We could also use a combination like ``v1.16.20250310``

uv.lock
suryadheeshjith, line 1: I see this being constantly changed as well!

suryadheeshjith (approved):
LGTM

src/datasets.py
jder: For posterity https://github.com/patrick-kidger/jaxtyping/issues/308

.github/workflows/benchmarks.yml
jder: I think it's probably good practice to use locked everywhere in ci -- if the lockfile is not consistent with pyproject.toml we should error (as opposed to updating the lockfile and then throwing it away). I'll add this to the other places, too, actually.

jder:
@alxmrs there are a couple warnings in tests now. Any idea what these are or if we should resolve before merging?

```
tests/test_datasets.py: 14 warnings
  /Users/jder/oa/Ocean_Emulator/tests/conftest.py:103: CFWarning: this date/calendar/year zero convention is not supported by CF
    cftime.num2date(num, units, calendar=self.start_day.calendar)

tests/test_datasets.py::test_train__loads_correct_number_of_samples[cpu]
  /Users/jder/oa/Ocean_Emulator/.venv/lib/python3.11/site-packages/zarr/api/asynchronous.py:203: UserWarning: Consolidated metadata is currently not part in the Zarr format 3 specification. It may not be supported by other zarr implementations and may change in the future.
    warnings.warn(
```

pyproject.toml
jder: Changed to 1.0 (and removed version from `src/__init__.py`)

pyproject.toml
jder: Yep, I can do this as a follow-up.

suryadheeshjith:
Will look into it

alxmrs:
>  Any idea what these are or if we should resolve before merging?

The Zarr / consolidated metadata looks significant. I think that should be addressed here or in a follow-up PR.

suryadheeshjith:
I am not sure if the consolidated metadata warning is a huge issue or not, but the fix seems to be to set consolidated=False while [creating](https://docs.xarray.dev/en/stable/generated/xarray.open_zarr.html) the zarr and reading it in train_3d.py. I am not sure why this is happening since consolidated=None by [default](https://docs.xarray.dev/en/stable/generated/xarray.open_zarr.html).

jder:
ok, I'm going to go ahead and merge since this is conflicty and then can follow up with resolving the warnings.

PR #110, opened 2025-03-12
Move everything into src/ocean_emulators, get rid of ocean_emulators_main

### LINKED ISSUE #75: Remove ocean_emulators_main module
Remove this module since we do not depend on it anymore. If we do require some functionality (like [xarray-schema](https://github.com/m2lines/ocean_emulators/blob/main/ocean_emulators/schema.py)), we can port the code over to this repository.

jder:
Follow-up to https://github.com/suryadheeshjith/Ocean_Emulator/pull/108 which moves everything into a named module and fixes #75.


jder:
This is more-or-less ready to review. I'm working on ignoring and/or fixing the now-working mypy errors so will ping again once I've done that. There are about 30 more to go. :)

.pre-commit-config.yaml
```diff
@@ -6,26 +6,26 @@ repos:
...
+# https://whynothugo.nl/journal/2023/01/12/notes-on-pre-commit/#using-the-system-plugin-language
+- repo: local
+  hooks:
+  - id: mypy
+    name: mypy
+    language: system
+    require_serial: true
+    entry: uv run --dev mypy
+    types_or: [python, pyi]
+    args: ["--config", "pyproject.toml"]
```
jder, line 28: Now that we have new import paths, we can get mypy working. This does that.

pyproject.toml
```diff
@@ -85,5 +87,8 @@ testpaths = [ "tests" ]
 addopts = "--jaxtyping-packages=data,datasets,beartype.beartype"
 
 [tool.mypy]
-ignore_missing_imports = true
 check_untyped_defs = true
+
+[[tool.mypy.overrides]]
+module = [ "cftime.*" ]
+follow_untyped_imports = true
```
jder, line 94: cftime doesn't have type information but only exposes a few functions, so we tell mypy to just do its best.

src/ocean_emulators/aggregator/inference/main.py
```diff
@@ -4,8 +4,8 @@
 import wandb
 import xarray as xr
 
-from utils.data import Normalize, get_norm_unnorm_dicts
-from utils.model import InfOutput, SingleTimeseriesOutput
+from ocean_emulators.utils.data import Normalize, get_norm_unnorm_dicts
+from ocean_emulators.utils.model import InfOutput, SingleTimeseriesOutput
```
jder, line 8: I used `ocean_emulators...` for import paths rather than relative imports, because my IDE (and I assume most IDEs) use that by default, but open to feedback. I didn't go through and fix existing relative paths, maybe we can do that over time.

src/ocean_emulators/aggregator/train.py
```diff
@@ -44,7 +44,8 @@ def get_logs(self, label: str = "train") -> Dict[str, torch.Tensor]:
             **var_loss_dict,
             **channel_loss_dict,
         }
+        meaned_logs = {}
         for key in sorted(logs.keys()):
-            logs[key] = float(all_reduce_mean(logs[key].detach()).cpu().numpy())
+            meaned_logs[key] = float(all_reduce_mean(logs[key].detach()).cpu().numpy())
 
-        return logs
+        return meaned_logs
```
jder, line 50: I'm not actually sure the right semantics here -- doing this causes errors downstream which makes it unclear to me how this was supposed to work. @suryadheeshjith do you know what the expected types here are?

suryadheeshjith: ~~Why are you creating a new dict meaned_logs?~~ seems fine. It looks right, the returned dictionary seems correctly typed as Dict[str, float]

jder: The type of `logs` is `Dict[str, Tensor]` but then here we are storing floats, not tensors. So created a new dict to return which is of type `Dict[str, float]` and updated the return type. Not sure if that's the right fix, though.

suryadheeshjith: What error are you facing right now?

jder: https://github.com/suryadheeshjith/Ocean_Emulator/actions/runs/13956359284/job/39068325637?pr=110 is all of them, the main one I'm not sure about is:

```
src/ocean_emulators/aggregator/validate/main.py:78: error: Signature of "get_logs" incompatible with supertype "TrainAggregator"  [override]
src/ocean_emulators/aggregator/validate/main.py:78: note:      Superclass:
src/ocean_emulators/aggregator/validate/main.py:78: note:          def get_logs(self, label: str = ...) -> dict[str, float]
src/ocean_emulators/aggregator/validate/main.py:78: note:      Subclass:
src/ocean_emulators/aggregator/validate/main.py:78: note:          def get_logs(self, label: str) -> Any
```

jder: ie not sure what the type of get_logs should be

suryadheeshjith: The error is coming from validate/main.py, try type hinting get_logs() there

jder: Thanks, yeah, I guess I'm trying to ask: do you know what the type of `get_logs` should be? If`Dict[str, float]` makes sense then I can carry that throughout.

suryadheeshjith: Oh yeah, it should be Dict[str, float].

suryadheeshjith: Rechecking the code, the get_logs() function of the ValidateAggregator can return a dict that has values that need not be float. If it is causing problems, we can ignore types for now and find a better solution (maybe decouple the train and validation aggregators) in a later PR.

suryadheeshjith: I believe the get_logs() return type in the ValidationAggregator class should be something like Dict[str, float | wandb.Image]]

.github/workflows/pre-commit.yml
```diff
@@ -15,6 +15,9 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
```
jder, line 20: This is now needed for pre-commit to run.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "mypy-type-checker.args": [
+        "\"--config=pyproject.toml\""
+    ]
+}
```
jder: Not sure what we want our policy to be about checking in workspace settings files. If you're using the mypy extension in vscode, you need this to make it work correctly.

src/ocean_emulators/aggregator/inference/main.py
```diff
@@ -238,11 +239,7 @@ def __init__(
             metadata=metadata,
         )
         self._aggregators = aggregators
-        self._summary_aggregators = {
-            name: aggregators[name]
-            for name in ["time_mean"]  # ["time_mean", "annual"]
-        }
-        self._time_dependent_aggregator_names: list[str] = []  # ["annual"]
```
jder, line 245: `_time_dependent_aggregator_names` appeared to be unused and to be complex from a typing POV below, so I removed it. @suryadheeshjith that seem ok?

src/ocean_emulators/aggregator/validate/main.py
```diff
@@ -34,8 +36,15 @@ def __init__(
...
+    # TODO(jder): we could remove this by moving from inheritance
+    # to composition with the TrainAggregator functionality.
+    def record_batch(self, batch):
+        raise NotImplementedError(
+            "Call record_validation_batch instead of record_batch"
+        )
+
     @torch.no_grad()
-    def record_batch(self, batch: ValOutput):
+    def record_validation_batch(self, batch: ValOutput):
```
jder, line 47: You can't override record_batch taking ValOutput since the super class says it must take TrainOutput. So I broke out a separate method you should call instead.

src/ocean_emulators/aggregator/validate/main.py
```diff
@@ -104,7 +116,7 @@ def _get_loss_scaled_mse_components(
             rmse_key = f"{label}/mean/weighted_rmse/{var}"
             if rmse_key in validation_metrics:
                 scaled_squared_errors[var] = (
-                    validation_metrics[rmse_key] / self._loss_scaling[var].item()
+                    validation_metrics[rmse_key] / self._loss_scaling[var]
```
jder, line 119: I think this was an actual bug -- this dict appears to have floats in it, not tensors.

src/ocean_emulators/aggregator/validate/map.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, metadata: Dict[str, Dict[str, str]] = {}, hist: int = 0):
     @torch.no_grad()
     def record_batch(
         self,
-        loss: float,
+        loss,
```
jder, line 41: This was actually a tensor, I think.

src/ocean_emulators/aggregator/validate/reduced.py
```diff
@@ -100,13 +101,14 @@ def _get_variable_metrics(self, gen_data):
     @torch.no_grad()
     def record_batch(
         self,
+        *,
+        loss: torch.Tensor = torch.tensor(np.nan),
```
jder, line 105: To match the other signatures

src/ocean_emulators/aggregator/validate/sub_aggregator.py
```diff
@@ -0,0 +1,23 @@
...
+        self,
+        *,
+        loss: torch.Tensor,
+        target_data,
+        gen_data,
+        input_data,
+        target_data_norm,
+        gen_data_norm,
+        input_data_norm,
+    ): ...
```
jder, line 23: Created this abstract class so we can reason over the dict of these things in `ValidateAggregator`

.vscode/settings.json
suryadheeshjith: We could include a sample settings.json file on CONTRIBUTING.md?

src/ocean_emulators/aggregator/inference/main.py
suryadheeshjith: Yes!

src/ocean_emulators/aggregator/validate/main.py
suryadheeshjith: Cool!

src/ocean_emulators/models/base.py
```diff
@@ -40,7 +40,9 @@ def forward(
             if step == 0:
                 input_tensor = train_data.get_initial_input()
             else:
-                input_tensor = train_data.merge_prognostic_and_boundary(
+                # TODO(jder): this function seems to be unused, resolve
+                # as part of https://github.com/suryadheeshjith/Ocean_Emulator/issues/51
+                input_tensor = train_data.merge_prognostic_and_boundary(  # type: ignore[attr-defined]
```
jder, line 45: This method doesn't exist on train_data, so silencing this warning for now, unless I misunderstood something? @suryadheeshjith

src/ocean_emulators/aggregator/validate/main.py
suryadheeshjith: Ah yeah, I did not test this function and it has never been used

suryadheeshjith: Super useful later on though

src/ocean_emulators/utils/wandb.py
```diff
@@ -210,26 +218,23 @@ def log_inference_metrics(
...
-                predictions,
-                targets,
-                targets_unnormalized,
-                model_pred_unnormalized,
-            ]
+        if (
+            predictions is not None
+            and targets is not None
+            and targets_unnormalized is not None
+            and model_pred_unnormalized is not None
```
jder, line 225: This is sad but the type checker didn't understand the previous form meant these things were non-none

src/ocean_emulators/utils/wandb.py
```diff
@@ -210,26 +218,23 @@ def log_inference_metrics(
...
         ):
             for i, var in enumerate(outputs):
                 fig = plt.figure(figsize=(10, 5))
                 plt.plot(
                     range(targets.shape[0]),
-                    targets_unnormalized[:, :, :, i].mean(axis=(1, 2)),
+                    targets_unnormalized[:, :, :, i].mean(dim=(1, 2)),
```
jder, line 231: TIL that pytorch silently supports both of these but unfortunately the typing declarations only show dim, not axis. AFAICT they are equivalent but this is undocumented.

jder:
ok, I think I fixed and/or silenced all the rest of the new mypy errors, so this is ready for a final review. 🙇

suryadheeshjith:
We definitely need that regression test before merging in this PR. I hope that's okay!

.vscode/settings.json
jder: Done, thanks!

jder:
> We definitely need that regression test before merging in this PR. I hope that's okay!

Seems reasonable to me. Let's discuss what it should look like in slack?

src/ocean_emulators/models/base.py
suryadheeshjith: This [PR](https://github.com/suryadheeshjith/Ocean_Emulator/commit/9bf9391bb918a926b0d2c530fde14fae0d78cfe2) removed this function, very confused how this slipped past all the train eval tests.

jder: AFAICT it's because this `forward` function is unused -- we always use `inference` which calls `forward_once` or the stepper, which does the same thing.

suryadheeshjith: Ah, we need a test to include forward() then

.github/workflows/pre-commit.yml
alxmrs: Cool, good to know.

.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "ocean_emulators_main"]
```
alxmrs, line 1: Hell yeah.

.pre-commit-config.yaml
alxmrs: Great work!

CONTRIBUTING.md
```diff
@@ -154,6 +154,18 @@ git push --force-with-lease
 
 10. Celebrating submitting your patch to Ocean Emulator — well done!
 
+## VS Code Integration
```
alxmrs, line 157: Thanks for adding this!

src/ocean_emulators/aggregator/inference/main.py
alxmrs: 100% agree -- I think these package focused imports are more stable than relative paths. And, I'm happy to change these over time.

src/ocean_emulators/aggregator/validate/reduced.py
alxmrs: Wow, I'm thrilled that we have mypy checks now!!

src/ocean_emulators/models/base.py
alxmrs: Sorry, I think I caused this issue!

alxmrs (approved):
LGTM! These look like a heroic number of type fixes. Great work.

src/ocean_emulators/models/base.py
jder: Hopefully fixed in #129

jder:
Discussed offline and decided to merge!

alxmrs:
Wohoo!

PR #111, opened 2025-03-12
uv script to prototype Zarr opening + update to Zarr cloning script.

alxmrs:
# Experiments

So far, it looks like we _should_ be using Zarr v2 + time_chunks=700, as this is the fastest way to open this particular dataset!

## Opening Remote OM4 data
```
# Zarr v2 (I think this uses consolidated=True by default!)
uv run scripts/open_zarr_tuning.py                        
3.11.11 (main, Jan 14 2025, 23:36:41) [Clang 19.1.6 ]
zarr-version=2.18.4,xarray-version=2025.1.2
ELAPSED: 7.5452s. Config: Namespace(target=None, n_iters=8, zarr_concurrency=None, time_chunks=None)

# Zarr v2 + time=10
uv run scripts/open_zarr_tuning.py --time_chunks=10
3.11.11 (main, Jan 14 2025, 23:36:41) [Clang 19.1.6 ]
zarr-version=2.18.4,xarray-version=2025.1.2
ELAPSED: 7.1839s. Config: Namespace(target=None, n_iters=8, zarr_concurrency=None, time_chunks=10)

# Zarr v2 + time=100
uv run scripts/open_zarr_tuning.py --time_chunks=100
3.11.11 (main, Jan 14 2025, 23:36:41) [Clang 19.1.6 ]
zarr-version=2.18.4,xarray-version=2025.1.2
ELAPSED: 7.1952s. Config: Namespace(target=None, n_iters=8, zarr_concurrency=None, time_chunks=100)

# Zarr v2 + time=700
uv run scripts/open_zarr_tuning.py --time_chunks=700
3.11.11 (main, Jan 14 2025, 23:36:41) [Clang 19.1.6 ]
zarr-version=2.18.4,xarray-version=2025.1.2
ELAPSED: 7.0575s. Config: Namespace(target=None, n_iters=8, zarr_concurrency=None, time_chunks=700)



# Zarr v3 (defaults)
uv run scripts/open_zarr_tuning.py 
3.11.11 (main, Jan 14 2025, 23:36:41) [Clang 19.1.6 ]
zarr-version=3.0.5,xarray-version=2025.1.2
ELAPSED: 11.2039s. Config: Namespace(target=None, n_iters=8, zarr_concurrency=10, time_chunks=None)

# Zarr v3 + manually setting `consolidated=True`
 uv run scripts/open_zarr_tuning.py 
3.11.11 (main, Jan 14 2025, 23:36:41) [Clang 19.1.6 ]
[...]

PR #117, opened 2025-03-14
Fix: Opening with consolidated metadata.

alxmrs:
I expect that this will fix the performance regression from switching between Zarr v2 --> Zarr v3.

### DIFF
--- a/src/train_3D.py
+++ b/src/train_3D.py
@@ -161,6 +161,7 @@ class Trainer:
             self.data = xr.open_zarr(
                 os.path.join(self.data_dir, self.data_path),
                 chunks={},
+                consolidated=True,
             )
         self.data_mean = xr.open_dataset(
             os.path.join(self.data_dir, self.data_means_path),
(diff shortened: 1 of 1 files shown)


jder:
The [xarray docs](https://docs.xarray.dev/en/stable/generated/xarray.open_zarr.html) say the default (None) tries to read consolidated metadata and then falls back to non-consolidated. That seems like what we want, no? Or is that the bug? FWIW it seems like the public Zarr store does not have consolidated metadata: https://nyu1.osn.mghpcc.org/m2lines-pubs/Samudra/OM4/zarr.json

alxmrs:
That's a good shout. I think the docs may not be totally up-to-date. In #111, I've found that after setting `consolidated=True`, the dataset loaded significantly faster; see: https://github.com/suryadheeshjith/Ocean_Emulator/pull/111#issuecomment-2722846780.

jder (approved):
Seems like the metadata is under .zmetadata in the zarr 2 format, so this seems fine assuming the xarray docs are wrong for the zarr 3 library.

PR #118, opened 2025-03-18
Rework round-tripping/interpretable data with bit fields

jder:
When I went to make another change, this test started failing again, I think because it generated new test cases. Try to resolve the instability by avoiding complexity of rounding and numeric error by packing data as smaller bitfields into the float64.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dependencies = [
 dev = [
   "beartype>=0.20",
   "hypothesis[numpy]>=6.127.2",
+  "ipykernel>=6.29.5",
   "pytest>=8.3.4",
   "pytest-benchmark[histogram]>=5.1",
   "pytest-profiling>=1.8.1",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -28,17 +28,24 @@ def _trunc(arr: NDArray[np.floating], decimals: int) -> NDArray[np.floating]:
     return np.trunc(arr * factor) / factor
 
 
+HEADER_BIT_OFFSET = 56
+LAT_BIT_OFFSET = 40
+LNG_BIT_OFFSET = 24
+DAYS_BIT_OFFSET = 8
+VAR_BIT_OFFSET = 0
+
+
 @dataclasses.dataclass
 class DataSourceDims:
     """Dimension metadata to produce interpretable `xarray.DataArray`s.
 
-    Each float in the encoded `xarray.DataArray` has the following scheme:
-    (s)AAAAGGGG.TTTDDD
-    - s := Sign (+/-) of the Latitude. All other values are non-negative.
-    - A := Latitude, which ranges from 90.00 <--> -90.00 (only decimals=2).
-    - G := Longitude, which ranges from 000.0 <--> 360.0 (only decimals=1).
-    - T := Time, the number of days since the start time.
-    - D := A int representing the index of the current data variable.
+    Each float in the encoded `xarray.DataArray` is interpreted as a uint64 broken
+    into the following fields, MSB first:
+      * 8 bits fixed as 0100 0000 (a non-NaN exponent)
+      * lat encoded as a float16
+      * lng encoded as a float16
+      * days_since_start encoded as a uint16
+      * data_var_index encoded as a uint8
     """
 
     lat: NDArray[np.float64] = dataclasses.field(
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -42,6 +42,7 @@ dependencies = [
 dev = [
   "beartype>=0.20",
   "hypothesis[numpy]>=6.127.2",
+  "ipykernel>=6.29.5",
```
jder, line 45: Drive by, I think most people will want this in dev

tests/conftest.py
```diff
@@ -23,65 +23,118 @@ class DataSource:
...
-    - T := Time, the number of days since the start time.
-    - D := A int representing the index of the current data variable.
+    Each float in the encoded `xarray.DataArray` is interpreted as a uint64 broken
+    into the following fields, MSB first:
+      * 8 bits fixed as 0100 0000
+        * which overlaps with float64 exponent to make it non-NaN and non-subnormal
+      * lat encoded as a float16
+      * lng encoded as a float16
+      * days_since_start encoded as a uint16
+      * data_var_index encoded as a uint8
```
jder, line 99: Might not be safe to assume we can fit the data var in a uint8; we could give this more space but not without some trickiness (eg simulating a uint12 or shrinking the float16s somehow)

tests/test_datasets.py
```diff
@@ -123,6 +123,14 @@ def vector_of(max_vec_size: int, min_vec_size=1):
...
     calendar="noleap",
 )
+@example(
+    data_var_index=0,
+    lat=np.array([2.0]),
+    lng=np.array([1.375]),
+    days_since_start=np.array([0], dtype=np.int32),
+    start_day=datetime.datetime(2000, 1, 1, 0, 0),
+    calendar="noleap",
+)
```
jder, line 133: this was the previously-failing example

tests/conftest.py
jder: (Or using a float128)

alxmrs: Before, I only supported up to 999 data vars, so I think this is already an improvement!

tests/conftest.py
```diff
@@ -23,65 +23,118 @@ class DataSource:
...
-    - s := Sign (+/-) of the Latitude. All other values are non-negative.
-    - A := Latitude, which ranges from 90.00 <--> -90.00 (only decimals=2).
-    - G := Longitude, which ranges from 000.0 <--> 360.0 (only decimals=1).
-    - T := Time, the number of days since the start time.
-    - D := A int representing the index of the current data variable.
+    Each float in the encoded `xarray.DataArray` is interpreted as a uint64 broken
+    into the following fields, MSB first:
+      * 8 bits fixed as 0100 0000
+        * which overlaps with float64 exponent to make it non-NaN and non-subnormal
+      * lat encoded as a float16
```
alxmrs, line 96: Is this a single lat value, or a range of lat values? I'm a bit confused.

alxmrs: Oh! Nvm, I see how this fits in my bad.

tests/conftest.py
```diff
@@ -23,65 +23,118 @@ class DataSource:
...
-        self.lng = _trunc(self.lng.astype(np.float64), 1)
         if np.any(self.lng < 0.0) or np.any(self.lng > 360.0):
-            raise ValueError("lng values must be between 0 and 360 degrees.")
-        self.lng %= 1000
+            raise ValueError("lng values are expected to be between 0 and 360 degrees.")
 
-        self.days_since_start = self.days_since_start.astype(np.int32)
-        if np.any(self.days_since_start < 0) or np.any(self.days_since_start > 999):
-            raise ValueError("days_since_start must be between 0 and 999.")
+        self._days_since_start_field.ensure_value_fits(self.days_since_start)
```
alxmrs, line 143: Nice

tests/conftest.py
```diff
@@ -23,65 +23,118 @@ class DataSource:
...
+        # lat and lng values round-trip via float16s, so declare them equal if they
+        # would encode to the same float16.
         # We can use exact equality with days_since_start since they are ints.
         return (
-            np.allclose(self.lat, other.lat, atol=0.1)
-            and np.allclose(self.lng, other.lng, atol=0.01)
+            np.array_equal(self.lat.astype(np.float16), other.lat.astype(np.float16))
+            and np.array_equal(
+                self.lng.astype(np.float16), other.lng.astype(np.float16)
+            )
```
alxmrs, line 153: This is sweet! It should definitely remove a large class of floating point errors in the roundtrip.

tests/conftest.py
```diff
@@ -112,40 +165,42 @@ def to_coords(self) -> dict[str, xr.DataArray]:
...
             axis=0,
         )
-        latlng_grid_3sf = np.around(latlng_grid, decimals=2)
 
-        template_grid = (
-            latlng_grid_3sf[0, :, :] * 1_000_000.0 + latlng_grid_3sf[1, :, :] * 10.0
-        )
+        template_grid = self._lat_field.encode(
+            latlng_grid[0, :, :]
+        ) + self._lng_field.encode(latlng_grid[1, :, :])
```
alxmrs, line 207: It's really nice to see an "encode" and "decode" than have to do the order-of-magnitude math, thanks this is a great improvement.

tests/conftest.py
```diff
@@ -112,40 +165,42 @@ def to_coords(self) -> dict[str, xr.DataArray]:
...
-        data_index_digits = float(data_var_index) / 1_000_000
+        interpretable_grid = (
+            rolled_out_grid
+            + self._days_since_start_field.encode(days_reshaped)
+            + self._data_var_index_field.encode(np.array(data_var_index))
+            + self._header_field.encode(self._header_value)
+        )
         return xr.DataArray(
-            interpretable_grid + data_index_digits,
+            interpretable_grid.view(np.float64),
```
alxmrs, line 219: 🐑 or more of a question: I thought we were packing bits into a uint64 -- how does this work with float64?

tests/test_datasets.py
```diff
@@ -72,7 +72,7 @@ def vector_of(max_vec_size: int, min_vec_size=1):
 
 
 @given(
-    data_var_index=st.integers(min_value=0, max_value=999),
+    data_var_index=st.integers(min_value=0, max_value=255),
```
alxmrs, line 75: Ah, I just realized now that uint8 means the range of values is much lower. I thought we used about ~200 data vars in all, so it's up to you if you like a tight bound or not.

tests/test_datasets.py
alxmrs: Nice!

alxmrs (approved):
Looks good! Thanks for writing this, happy to see this area of code be really stable. I've mainly just added a few questions to understand what's going on.

tests/test_datasets.py
jder: Yeah, I suggest we leave it for now and if we end up wanting more data vars in test we can expand it.

tests/conftest.py
jder: For posterity, it now only supports up to 255 as discussed below.

tests/conftest.py
jder: So we pack everything into a uint64 and then reinterpret those same bytes as a float64. These are the same size and due to the header position + value we picked above, we end up with a reasonable value when interpreted as a float. I added a comment in DataSourceDims which illustrates this a bit more.

PR #123, opened 2025-03-18
Wrote an Xarray-based OM4 Dataloader.

### LINKED ISSUE #119: Move data loader to use xarray throughout

alxmrs:
Fixes #119. This new loader is behind a feature flag. After I manually profile the new loader, I will make a follow-up PR to make this one the default.

### DIFF
--- a/src/ocean_emulators/constants.py
+++ b/src/ocean_emulators/constants.py
@@ -26,9 +26,6 @@ Boundary = Float[Grid, "*batch boundary_vars"]
 Input: TypeAlias = Float[Grid, "*batch total_vars"]
 Label: TypeAlias = Prognostic
 
-GridMask = Bool[torch.Tensor, "180 360"]
-PrognosticMask = Bool[GridMask, "prognostic_vars"]
-
 Example = tuple[Input, Label] | tuple[xr.Dataset, xr.Dataset]
 
 GridMask = Bool[Array, "180 360"]
--- a/src/ocean_emulators/constants.py
+++ b/src/ocean_emulators/constants.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict
+from typing import Dict, TypeAlias, TypeVar
 
 import torch
 import xarray as xr
@@ -10,7 +10,11 @@ from ocean_emulators.utils.multiton import Multiton
 # Common Type Aliases
 # See "Existing jaxtyping annotations" section of
 #  https://docs.kidger.site/jaxtyping/api/array/#array
-Grid = Float[torch.Tensor, "180 360"]
+
+# Our Arrays will be either `torch.Tensor`s or `xarray.DataArray`s.
+Array = TypeVar("Array", torch.Tensor, xr.DataArray)
+
+Grid = Float[Array, "180 360"]
 Prognostic = Float[
     Grid, "*batch prognostic_vars"
 ]  # equivalent to "*batch prognostic_vars lat lon"
@@ -19,9 +23,15 @@ Boundary = Float[Grid, "*batch boundary_vars"]
 #   In practice you should usually only use symbolic axes in annotations
 #   for return types, referring only to axes annotated for arguments.
 # So, we'll leave this default and use symbolic axes locally.
-Input = Float[Grid, "*batch total_vars"]
+Input: TypeAlias = Float[Grid, "*batch total_vars"]
+Label: TypeAlias = Prognostic
 
 GridMask = Bool[torch.Tensor, "180 360"]
+PrognosticMask = Bool[GridMask, "prognostic_vars"]
+
+Example = tuple[Input, Label] | tuple[xr.Dataset, xr.Dataset]
+
+GridMask = Bool[Array, "180 360"]
 PrognosticMask = Bool[GridMask, "prognostic_vars"]
 
 # Experiment prognostic and boundary variables
@@ -128,6 +138,12 @@ def construct_metadata(data: xr.Dataset) -> Dict[str, Dict[str, str]]:
                 "units": "Unknown",
             }
     return metadata
+
+
+LOADER_FLAGS = [
+    "Eager-TrainData-OM4-Loader",
+    "Lazy-Xarray-OM4-Loader",
+]
 
 
 # TODO(#95): See if this can be removed and replaced.
--- a/src/ocean_emulators/constants.py
+++ b/src/ocean_emulators/constants.py
@@ -138,8 +138,8 @@ def construct_metadata(data: xr.Dataset) -> Dict[str, Dict[str, str]]:
 
 
 LOADER_FLAGS = [
-    "Eager-TrainData-OM4-Loader",
-    "Lazy-Xarray-OM4-Loader",
+    "om4-eager",
+    "om4-lazy",
 ]
 
 
--- a/src/ocean_emulators/constants.py
+++ b/src/ocean_emulators/constants.py
@@ -1,3 +1,4 @@
+import enum
 import logging
 from typing import Dict, TypeAlias, TypeVar
 
@@ -137,10 +138,9 @@ def construct_metadata(data: xr.Dataset) -> Dict[str, Dict[str, str]]:
     return metadata
 
 
-LOADER_FLAGS = [
-    "om4-eager",
-    "om4-lazy",
-]
+class LoaderVersion(enum.Enum):
+    OM4_EAGER = "om4-eager"
+    OM4_LAZY = "om4-lazy"
 
 
 # TODO(#95): See if this can be removed and replaced.
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -9,12 +9,12 @@ from jaxtyping import Float, Integer
 from torch.utils.data import Dataset
 
 from ocean_emulators.constants import (
-    LOADER_FLAGS,
     Boundary,
     BoundaryVarNames,
     Example,
     GridMask,
     Input,
+    LoaderVersion,
     Prognostic,
     PrognosticMask,
     PrognosticVarNames,
@@ -26,7 +26,7 @@ from ocean_emulators.utils.device import get_device, using_gpu
 class OM4Dataset(Dataset):
     """A `torch.Dataset` for Zarr-backed OM4 data."""
 
-    FLAG = LOADER_FLAGS[1]
+    FLAG = LoaderVersion.OM4_LAZY
 
     def __init__(
         self,
@@ -411,7 +411,7 @@ class TrainDataset(Dataset):
             step=3->[[9, 10, 11], [12, 13, 14]]
     """
 
-    FLAG = LOADER_FLAGS[0]
+    FLAG = LoaderVersion.OM4_EAGER
 
     def __init__(
         self,
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -5,22 +5,152 @@ import numpy as np
 import torch
 import xarray as xr
 from einops import rearrange
-from jaxtyping import Float
+from jaxtyping import Float, Integer
 from torch.utils.data import Dataset
 
 from ocean_emulators.constants import (
+    LOADER_FLAGS,
     Boundary,
     BoundaryVarNames,
+    Example,
     GridMask,
     Input,
     Prognostic,
     PrognosticMask,
     PrognosticVarNames,
 )
-from ocean_emulators.utils.data import Normalize
+from ocean_emulators.utils.data import Normalize, mask, unflatten_masks
 from ocean_emulators.utils.device import get_device, using_gpu
 
-Example = tuple[Input, Prognostic]
+
+class OM4Dataset(Dataset):
+    """A `torch.Dataset` for Zarr-backed OM4 data."""
+
+    FLAG = LOADER_FLAGS[1]
+
+    def __init__(
+        self,
+        data: xr.Dataset,
+        input_vars: PrognosticVarNames,
+        extra_vars: BoundaryVarNames,
+        output_vars: PrognosticVarNames,  # TODO(alxmrs): rm
+        hist: int,
+        steps: int,
+        stride: int = 1,
+        is_inference: bool = False,
+    ) -> None:
+        # Ensure that there is a `wetmask` DataArray with a supported `lev` dimension.
+        data = unflatten_masks(data)
+
+        input_ = data[input_vars]
+        extra_ = data[extra_vars]
+        output_ = data[output_vars]
+
+        # Normalize data. E.g. mean=zero, std=1., NaN --> 0.0
+        norm = Normalize.get_instance()
+        norm_input = norm.normalize_inputs(input_)
+        norm_extra = norm.normalize_boundary(extra_)
+        norm_output = norm.normalize_outputs(output_)
+
+        # Set non-ocean areas to zero.
+        self.input = mask(norm_input, data.wetmask)
+        self.extra = mask(norm_extra, data.wetmask)
+        self.output = mask(norm_output, data.wetmask)
+
+        self.hist: int = hist
+        self.steps: int = steps
+        self.stride: int = stride
+        self.is_inference: bool = is_inference
+
+        self._size: int = (
+            data.time.size
+            - self.steps * (self.hist + 1) * self.stride
+            - self.hist * self.stride
+        )
+
+        # TODO(alxmrs): When we want to support inference later on, we will need to
+        #  calculate different steps.
+        total_steps: int = 2 * self.hist + 2
+        # Calculate the number of windows
+        num_windows = data.time.size - (total_steps - 1) * self.stride
+        # Create base indices
+        indices = np.arange(num_windows)
+        indices_da = xr.DataArray(indices, dims=["window"])
+        # Create window dimension
+        window_dim = xr.DataArray(np.arange(total_steps), dims=["time"])
+        # Construct rolling indices
+        self.rolling_indices: Integer[xr.DataArray, "window time"] = (
+            indices_da + stride * window_dim
+        )
+
+    def __len__(self) -> int:
+        return self._size
+
+    # TODO(#124): Vectorize `step`
+    def window_from(
+        self, idx: int | slice, step: int
+    ) -> Integer[xr.Variable, "window time"]:
+        """Coalesce index values to a window of the input data."""
+        # First, parse int inputs as a slice.
+        if isinstance(idx, int):
+            if idx < 0 or idx >= len(self):
+                raise IndexError(
+                    f"index {idx!r} out of bounds. Must be between 0 and {len(self)}."
+                )
+
+            if not self.is_inference:
+                idx = idx + step * (self.hist + 1) * self.stride
+            idx = slice(idx, idx + 1, 1)
+
+        # Validate and normalize all slices.
+        if self.is_inference:
+            if idx.start < 0 or idx.stop < 0 or idx.step < 0:
+                raise IndexError(
+                    f"index {idx!r} invalid: negative values not supported."
+                )
+            if idx.start > self._size or idx.stop > self._size:
+                raise IndexError(
+                    f"index {idx!r} out of bounds. "
+                    f"All bounds must be less than or equal to {len(self)}."
+                )
+
+        if idx.start is None:
+            idx = slice(0, idx.stop, idx.step)
+        if idx.stop is None:
+            idx = slice(idx.start, len(self), idx.step)
+
+        window_index = self.rolling_indices.isel(window=idx)
+        window = xr.Variable(["window", "time"], window_index)
+
+        return window
+
+    def __getitem__(self, idx: int) -> Example:
+        inputs = []
+        labels = []
+
+        # TODO(#124): Vectorize this operation! Or, is it fine because it's lazy?
+        for step in range(self.steps):
+            window = self.window_from(idx, step)
+
+            # This point in time splits the training data and the label data!
+            time_split = self.hist + 1
+
+            prognostic = self.input.isel(time=window).isel(time=slice(None, time_split))
+            boundary = self.extra.isel(time=window).isel(time=self.hist)
+            input_ = xr.merge(
+                [prognostic, boundary], compat="no_conflicts"
+            )  # Combines variables.
+
+            label = self.output.isel(time=window).isel(time=slice(time_split, None))
+
+            inputs.append(input_)
+            labels.append(label)
+
+        # Combines data along a new dimension "step"
+        input_ = xr.concat(inputs, dim="step")
+        label = xr.concat(labels, dim="step")
+
+        return input_, label
 
 
 class InferenceDataset(Dataset):
@@ -280,6 +410,8 @@ class TrainDataset(Dataset):
             step=3->[[9, 10, 11], [12, 13, 14]]
     """
 
+    FLAG = LOADER_FLAGS[0]
+
     def __init__(
         self,
         data: xr.Dataset,
@@ -377,9 +509,8 @@ class TrainDataset(Dataset):
 
         start = idx + step * (self.hist + 1) * self.stride
         end = start + 1
-        idx_slice = slice(
-            start, end
-        )  # Create a slice for similar indexing as in InferenceDataset
+        # Create a slice for similar indexing as in InferenceDataset
+        idx_slice = slice(start, end)
         rolling_idx = self.rolling_indices.isel(window_dim=idx_slice)
         # Convert to tests, tests are outdated since changing time definition
         # if prev_rolling_idx is not None:
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -31,9 +31,8 @@ class OM4Dataset(Dataset):
     def __init__(
         self,
         data: xr.Dataset,
-        input_vars: PrognosticVarNames,
-        extra_vars: BoundaryVarNames,
-        output_vars: PrognosticVarNames,  # TODO(alxmrs): rm
+        depth_vars: PrognosticVarNames,
+        surface_vars: BoundaryVarNames,
         hist: int,
         steps: int,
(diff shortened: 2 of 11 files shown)


src/config.py
```diff
@@ -34,6 +34,7 @@ class DataConfig:
     time_delta: int = 5
     num_workers: int = 4
     hist: int = 1
+    loader_version = "1.0"
```
alxmrs: I am very welcoming of advice for how we should name and/or version the data loaders!

src/constants.py
```diff
@@ -8,17 +8,23 @@
 # Common Type Aliases
 # See "Existing jaxtyping annotations" section of
 #  https://docs.kidger.site/jaxtyping/api/array/#array
-Grid = Float[torch.Tensor, "180 360"]
-Input = Float[Grid, "*batch input_vars"]  # equivalent to "*batch input_vars lat lon"
-Extra = Float[Grid, "*batch extra_vars"]
+
+# Our Arrays will be either `torch.Tensor`s or `xarray.DataArray`s.
+Arr = TypeVar("Arr", torch.Tensor, xr.DataArray)
```
alxmrs: Arr? Array? Any name suggestions?

src/utils/data.py
```diff
@@ -31,20 +33,92 @@ def extract_wet_mask(
...
+    for name, da in data_.items():
+        # Parse the level index info from the variable name.
+        tokens = str(name).split("_")
+        if len(tokens) == 4:  # OM4 data format (e.g., {variable}_lev_{level}_{decimal})
+            # TODO(alxmrs): Is the OM4 data wrong? Is preprocessing done somewhere?
+            # In this format, "level" is a member of DEPTH_LEVELS, _not_ DEPTH_I_LEVELS.
+            # Thus, we need to convert it to the corresponding DEPTH_I_LEVELS index.
+            _, _, level, _ = tokens
+            closest_level = min(DEPTH_LEVELS, key=lambda x: abs(x - float(level)))
+            level = str(DEPTH_LEVELS.index(closest_level))
```
alxmrs: I don't know if this case needs to be supported. Further, I'm not sure how the OM4 data that's public now maps to how it seems to be treated throughout the codebase.

src/utils/train.py
```diff
@@ -27,6 +30,36 @@ def collate_train_data(data: Sequence[TrainData]) -> TrainData:
...
 
+def collate_om4(examples: Sequence[Example]) -> Example:
+    """Combine several deferred Examples into single a `torch.Tensor` example pair."""
+    inputs: list[xr.DataArray] = []
+    labels: list[xr.DataArray] = []
+
+    for input_, label in examples:
+        inputs.append(
+            input_.to_array()
+            .compute()
```
alxmrs: This is when the data actually gets loaded!

tests/test_datasets.py
```diff
@@ -285,6 +287,60 @@ def test_inference__data_is_not_zero(inference_loader_pair: LoaderPair):
             ), "Label data should not be a zeros matrix!"
 
 
+def test_om4__is_equal_to_v1_data_loader(train_loader_pair: LoaderPair):
```
alxmrs, line 304: A flaw (or benefit) of this test: it compares the new data loader to the original data loader pipeline.

tests/test_utils_data.py
```diff
@@ -0,0 +1,22 @@
...
+def test_mask_roundtrip(data_source):
+    data = data_source.data
+
+    unflattened = unflatten_masks(data.copy())
+    flattened = flatten_masks(unflattened.copy())
+
+    assert flattened == data, "Assume a safe roundtrip"
+
+
+def test_mask__zeros_data(data_source):
```
alxmrs, line 15: Admittedly, I could have written more extensive tests of the new mask code that I wrote. I've kicked this can down the road because 1) I value more integration-style tests more to prove correctness (see my comparison to the v1 data loader above) and 2) I expect the shape of the data to change soon (in #120).

suryadheeshjith:
I will need to review this PR!

src/constants.py
alxmrs: I avoided "Array" because that is a think from jaxtyping (admittedly, which depends on jax -- so I didn't use it).

src/datasets.py
```diff
@@ -4,24 +4,142 @@
...
+        window_index = self.rolling_indices.isel(window=idx)
+        window = xr.Variable(["window", "time"], window_index)
+
+        return window
+
+    def __getitem__(self, idx: int) -> Example:
+        inputs = []
+        labels = []
+
+        # TODO(#124): Vectorize this operation! Or, is it fine because it's lazy?
```
alxmrs: I think I might implement this in this PR (i.e. tomorrow).

src/constants.py
jder: I think `Array` is good. jaxtyping doesn't include Array unless jax is installed, which it isn't. Alternatively we could track at the type level whether we have torch tensors or xarray data arrays (ie not have this Array alias). Might help prevent bugs where we expect one or the other, and be more clear about when things get converted, but might also be too much noise -- happy either way.

src/datasets.py
```diff
@@ -4,24 +4,142 @@
...
+        hist: int,
+        steps: int,
+        stride: int = 1,
+        is_inference: bool = False,
+    ) -> None:
+        # Ensure that there is a `wetmask` DataArray with a supported `lev` dimension.
+        data = unflatten_masks(data)
+
+        # Apply oceans mask to the data up-front.
+        masked = mask(data)
```
suryadheeshjith: Masking then normalizing is not the same as normalizing and then masking. The order needs to be consistent with what we have right now.

suryadheeshjith: I am surprised tests are passing for this case because the behavior of the models trained in each way are very different.

src/datasets.py
suryadheeshjith: Would be great to vectorize this in a future PR!

src/datasets.py
alxmrs: I thought this was the case as well (and I tried it both ways), but I went with this one because 1) it was simpler and 2) it passed tests.

alxmrs: Let me add more tests to confirm or deny this property.

tests/test_datasets.py
```diff
@@ -285,6 +287,63 @@ def test_inference__data_is_not_zero(inference_loader_pair: LoaderPair):
...
+    # we use sorting as a simple way to compare the two loaders (without having to
+    # monkeypatch the train loader fixture).
+    original_samples = sorted(
+        [extract_sample_arrays(sample) for sample in loader], key=key
+    )
+    om4_samples = sorted([as_numpy(sample) for sample in om4_loader], key=key)
+
+    atol = 0.01
+    for (x_orig, y_orig), (x_new, y_new) in zip(original_samples, om4_samples):
+        x_not_close = np.isclose(x_orig, x_new, atol=atol) == False  # noqa: E712
```
suryadheeshjith: I am starting to get a little worried on the use of atol in the test_datasets file :/

src/datasets.py
alxmrs: Working on this now!

src/datasets.py
alxmrs: Ok, you were right on two counts: the low tolerance in tests was covering the errors exposed by masking before normalizing. Once I switched these operations, I passed the test with a high-error tolerance. Thanks for calling these things out.

src/datasets.py
```diff
@@ -4,24 +4,144 @@
...
+    def __init__(
+        self,
+        data: xr.Dataset,
+        input_vars: InputVars,
+        extra_vars: ExtraVars,
+        output_vars: OutputVars,
+        hist: int,
+        steps: int,
+        stride: int = 1,
+        is_inference: bool = False,
```
suryadheeshjith: is this being used during inference as well? Will you be testing this in a later PR?

suryadheeshjith: I would remove this if we are not using it for inference

uv.lock
suryadheeshjith, line 1: This is still updating?

src/datasets.py
alxmrs: The plan was to use it for inference, and I have not tested that! My original plan was to verify this in a later PR, but I may as well do it now.

uv.lock
alxmrs: Looks like I had a merge conflict (I also change it a bit for debugging). I hope changes here are resolved now.

src/ocean_emulators/datasets.py
```diff
@@ -5,24 +5,149 @@
...
+        # Construct rolling indices
+        self.rolling_indices: Integer[xr.DataArray, "window time"] = (
+            indices_da + stride * window_dim
+        )
+
+    def __len__(self) -> int:
+        return self._size
+
+    # TODO(#124): Vectorize `step`
+    # TODO(alxmrs): Why doesn't jaxtyping work here?
```
jder: FYI jaxtyping 0.3.0 fixed [the bug I reported](https://github.com/patrick-kidger/jaxtyping/issues/308) about xarray compatibility, so you might want to try bumping the version.

src/ocean_emulators/constants.py
```diff
@@ -8,17 +8,23 @@
...
+Grid = Float[Arr, "180 360"]
+Prognostic = Float[Grid, "*batch input_vars"]  # = "*batch input_vars lat lon"
+Boundary = Float[Grid, "*batch extra_vars"]
 # A note from jaxtyping (why we can't do "input_vars+extra_vars"):
 #   In practice you should usually only use symbolic axes in annotations
 #   for return types, referring only to axes annotated for arguments.
 # So, we'll leave this default and use symbolic axes locally.
-TotalInput = Float[Grid, "*batch total_vars"]
-Label = Float[Grid, "*batch output_vars"]
+Input: TypeAlias = Float[Grid, "*batch total_vars"]
```
jder, line 27: 🐑 Seems kind of confusing that `input_vars` is the axis for Prognostic and `total_vars` is the axis for Input. Why did this change from TotalInput?

src/ocean_emulators/datasets.py
```diff
@@ -393,7 +517,7 @@ def _get_x_index(
         x_index = xr.Variable(["window_dim", "time"], rolling_idx)
         return x_index
 
-    def _get_input(self, x_index) -> Input:
+    def _get_input(self, x_index) -> Prognostic:
```
jder: 🐑 Seems like a bit of a smell that "get input" produces something not called Input. Maybe a question for @suryadheeshjith what would be more natural here (eg should this be get_prognostic? or should this type Input?)

src/config.py
jder: My vote would be like feature flags, so maybe `loader = "xarray-e2e"` or something?

src/ocean_emulators/utils/data.py
```diff
@@ -31,20 +33,103 @@ def extract_wet_mask(
...
     wet_surface = torch.from_numpy(wet_surface_mask_np)
     wet_inp = torch.concat([wet_inp] * (hist + 1), dim=0)
     return wet_inp.bool(), wet_surface.bool()
 
 
+def _parse_lev_from_output_var(outputs: OutputVars) -> list[int]:
+    """Parse the `lev` dimension from the output var names. Default: 0 for surface."""
+    depth_inds = []
+    for var_depth_i in outputs:
+        # Examples: "so_18", "zos"
```
jder, line 49: <3

src/ocean_emulators/utils/data.py
```diff
@@ -31,20 +33,103 @@ def extract_wet_mask(
...
+        var_split = var_depth_i.split("_")
+        if len(var_split) == 1:
+            depth_inds.append(0)
+        else:
+            depth_inds.append(int(var_split[-1]))
+
+    return depth_inds
+
+
+def flatten_masks(data: xr.Dataset) -> xr.Dataset:
```
jder, line 59: 🐑 Maybe this should be in test code?

src/ocean_emulators/utils/data.py
```diff
@@ -31,20 +33,103 @@ def extract_wet_mask(
...
+
+def unflatten_masks(data: xr.Dataset) -> xr.Dataset:
+    """Adds a "wetmask" `xarray.DataArray` with dimensions (lev, y, x)."""
+    if "wetmask" not in data.variables:
+        assert MASK_VARS[0] in data.variables, "Wet mask must have masks as data vars!"
+
+        wetmask = data[MASK_VARS].to_array(dim="lev", name="wetmask")
+
+        data["wetmask"] = wetmask.assign_coords(lev=data.lev)
+        data = data.drop_vars(MASK_VARS)
```
jder: Unless I'm missing something we're both doing some mutation to the underlying data and also returning a new dataset. Can we avoid doing both?

src/ocean_emulators/utils/data.py
```diff
@@ -31,20 +33,103 @@ def extract_wet_mask(
...
+
+
+def mask(data: xr.Dataset, wetmask: xr.DataArray | None = None) -> xr.Dataset:
+    """Apply a wetmask (areas of the ocean) to all variables in the dataset."""
+    # I revised on this via Project Pythia's tutorial:
+    #  https://foundations.projectpythia.org/core/xarray/computation-masking.html#masking-data
+    data_ = data.copy()
+
+    if wetmask is None:
+        wetmask = data_.wetmask
```
jder: 🐑 maybe we can remove this defaulting if we are always getting a wetmask in actual code?

src/ocean_emulators/utils/data.py
```diff
@@ -31,20 +33,103 @@ def extract_wet_mask(
...
+        # Parse the level index info from the variable name.
+        is_surface = False
+        tokens = str(name).split("_")
+        if len(tokens) >= 4:  # OM4 data format (e.g., {variable}_lev_{level}_{decimal})
+            # TODO(alxmrs): Is the OM4 data wrong? Is preprocessing done somewhere?
+            # In this format, "level" is a member of DEPTH_LEVELS, _not_ DEPTH_I_LEVELS.
+            # Thus, we need to convert it to the corresponding DEPTH_I_LEVELS index.
+            _, _, level, *_ = tokens
+            closest_level = min(DEPTH_LEVELS, key=lambda x: abs(x - float(level)))
+            level = DEPTH_I_LEVELS[DEPTH_LEVELS.index(closest_level)]
```
jder, line 113: Presumably some OM4 data is not in this format since there's no way to set is_surface = True here? Can you add a comment or example covering that case?

src/ocean_emulators/utils/train.py
```diff
@@ -27,6 +30,43 @@ def collate_train_data(data: Sequence[TrainData]) -> TrainData:
...
+            .einops.rearrange(
+                "step window (time variable) lat lon",
+            )
+        )
+
+    input_batch: xr.DataArray = xr.concat(inputs, dim="step")
+    labels_batch: xr.DataArray = xr.concat(labels, dim="step")
+
+    input_tensor = torch.from_numpy(input_batch.to_numpy())
+    labels_tensor = torch.from_numpy(labels_batch.to_numpy())
```
jder: At the moment we do:

1. compute
2. rearrange
3. concat
4. torch.from_numpy


might be worth trying instead:

1. concat
2. rearrange
3. compute
4. torch.from_numpy

To compute as late as possible? Perhaps in a follow-up?

src/ocean_emulators/datasets.py
```diff
@@ -5,24 +5,149 @@
...
+        self.stride: int = stride
+        self.is_inference: bool = is_inference
+
+        self._size: int = (
+            data.time.size
+            - self.steps * (self.hist + 1) * self.stride
+            - self.hist * self.stride
+        )
+
+        # This class will be used only for training
```
jder: Is this comment out of date?

src/ocean_emulators/datasets.py
```diff
@@ -5,24 +5,149 @@
...
+        for step in range(self.steps):
+            window = self.window_from(idx, step)
+
+            # This point in time splits the training data and the label data!
+            time_split = self.hist + 1
+
+            prognostic = self.input.isel(time=window).isel(time=slice(None, time_split))
+            boundary = self.extra.isel(time=window).isel(time=self.hist)
+            input_ = xr.merge(
+                [prognostic, boundary], compat="override"
```
jder: Why "override" rather than something like "no_conflicts"?

jder (commented):
Thanks! This is some very exciting work, looking forward to seeing benchmark results. 

Mostly sheep and questions from me. I suggest we try to tie this off (eg postpone things like supporting inference if possible) and get it merged. There might also be some de-duplication we want to do with #134 in supporting OM4?

src/ocean_emulators/datasets.py
suryadheeshjith: About to change a bunch of the naming in #134

src/ocean_emulators/datasets.py
alxmrs: Wow, what a fast fix!!

src/constants.py
alxmrs: I found it to be a bit noisy. 

In that case, Array is a good name!

src/ocean_emulators/constants.py
alxmrs: Good shout, I'll rewrite these.

src/ocean_emulators/datasets.py
alxmrs: I'll keep it for now, and assume that it will get fixed soon.

src/ocean_emulators/utils/data.py
alxmrs: I actually plan to use this in the "application layer" really soon. I'm in a bit of a catch 22 in this PR and #128.

src/ocean_emulators/utils/data.py
alxmrs: SGTM

src/ocean_emulators/utils/data.py
alxmrs: Sure, and if I find a use for this, I can always add it back.

src/ocean_emulators/utils/train.py
alxmrs: The rearrange step requires that data actually exist. Otherwise, I agree with you that computing last would be better.

I'll see if I can concat before rearranging, however. That's a good shout and I think it will be more performant.

alxmrs: Ok, thanks for the suggestion, I think it's even lazier.

src/ocean_emulators/datasets.py
alxmrs: Boy, is it! In fact, it shows me that it should actually be a TODO.

src/ocean_emulators/datasets.py
alxmrs: I am having data dimension errors, so this could be the source of it (though, I have only started investigating, focusing on what I can catch via testing first). I'll give it a shot.

alxmrs: Reading the docs now, yeah, this is worthy of careful configuration + tests.

src/ocean_emulators/utils/data.py
alxmrs: :)

src/ocean_emulators/datasets.py
alxmrs: I'll follow up with this in future tests before we turn this feature flag on and focus on merging this now.

alxmrs:
(I hope neither of you see this so late on a Friday, but this PR is now ready for another round of review.)

src/ocean_emulators/constants.py
```diff
@@ -220,6 +226,12 @@ def construct_metadata(data: xr.Dataset) -> Dict[str, Dict[str, str]]:
     return metadata
 
 
+LOADER_FLAGS = [
+    "Eager-TrainData-OM4-Loader",
+    "Lazy-Xarray-OM4-Loader",
+]
```
jder: This seems pretty verbose for a human to have to type into a config file. (and eg get all the casing right). How about "eager" and "lazy"?

src/ocean_emulators/datasets.py
```diff
@@ -5,24 +5,154 @@
...
+        self.is_inference: bool = is_inference
+
+        self._size: int = (
+            data.time.size
+            - self.steps * (self.hist + 1) * self.stride
+            - self.hist * self.stride
+        )
+
+        # TODO(alxmrs): When we want to support inference later on, we will need to
+        #  calculate different steps.
```
jder, line 68: How about an `assert !is_inference` for now?

src/ocean_emulators/datasets.py
```diff
@@ -282,6 +412,8 @@ class TrainDataset(Dataset):
             step=3->[[9, 10, 11], [12, 13, 14]]
     """
 
+    FLAG = LOADER_FLAGS[0]
```
jder: This seems easy to mess up to me eg if someone adds a new loader flag to the beginning of the list. Maybe we have `FLAG = "eager"` here and then LOADER_FLAGS can be `= [TrainDataset.FLAG, OM4Dataset.FLAG]`?

tests/test_datasets.py
```diff
@@ -293,6 +295,61 @@ def test_inference__data_is_not_zero(inference_loader_pair: LoaderPair):
...
+    # monkeypatch the train loader fixture).
+    original_samples = sorted(
+        [extract_sample_arrays(sample) for sample in loader], key=key
+    )
+    om4_samples = sorted(
+        [extract_sample_arrays(sample) for sample in om4_loader], key=key
+    )
+
+    for (x_orig, y_orig), (x_new, y_new) in zip(original_samples, om4_samples):
+        x_not_close = np.isclose(x_orig, x_new) == False  # noqa: E712
```
jder, line 341: why not `not np.isclose(x_orig, x_new)`?

tests/test_datasets.py
```diff
@@ -293,6 +295,61 @@ def test_inference__data_is_not_zero(inference_loader_pair: LoaderPair):
...
+        y_not_close = np.isclose(y_orig, y_new) == False  # noqa: E712
+
+        x_not_close_index = list(zip(*np.where(x_not_close)))
+        y_not_close_index = list(zip(*np.where(y_not_close)))
+
+        assert np.allclose(x_orig, x_new), (
```
jder: Should this be `assert not np.any(x_not_close)`? Kinda weird we're doing the comparisons twice. (Or maybe we should have `x_close` rather than `x_not_close` and then this is np.all?)

tests/test_datasets.py
```diff
@@ -293,6 +295,61 @@ def test_inference__data_is_not_zero(inference_loader_pair: LoaderPair):
...
+
+    # Why are we sorting here? Well, the default data loader uses a random sampler. So
+    # we use sorting as a simple way to compare the two loaders (without having to
+    # monkeypatch the train loader fixture).
+    original_samples = sorted(
+        [extract_sample_arrays(sample) for sample in loader], key=key
+    )
+    om4_samples = sorted(
+        [extract_sample_arrays(sample) for sample in om4_loader], key=key
+    )
```
jder, line 338: 🐑 Seems a bit surprising that key(x) is unique across the dataset to me; would a key of `stack(x[0].flat, x[1].flat)` or something work?

Alternatively, it might end up being cleaner to use torch.set_seed to to get the same shuffled output from each loader (which would require building the old dataloader yourself, I think.)

src/ocean_emulators/utils/train.py
```diff
@@ -27,6 +30,30 @@ def collate_train_data(data: Sequence[TrainData]) -> TrainData:
...
+    """Combine several deferred Examples into single a `torch.Tensor` example pair."""
+    inputs_batch = xr.concat([x.to_array() for x, _ in examples], dim="step")
+    labels_batch = xr.concat([y.to_array() for _, y in examples], dim="step")
+
+    inputs = inputs_batch.compute().einops.rearrange(
+        "step window (time variable) lat lon"
+    )
+    labels = labels_batch.compute().einops.rearrange(
+        "step window (time variable) lat lon"
+    )
```
jder: I see you included `xarray-einstats` -- it looks currently unused but I'm assuming your plan is to try using that before compute() in a follow up?

jder (approved):
Thanks, looks good to me! I think we should change the flag stuff as described but nothing here is blocking a merging this, since it's disabled by default.

src/ocean_emulators/datasets.py
alxmrs: One reason I've factored it this way is that when I import datasets into constants, I get a circular import. Maybe I can add an enum to constants? Any idea for how to work with that constraint?

tests/test_datasets.py
alxmrs: Great question: when I did that, it gave an `a.any()` / `a.all()` style numpy error.

src/ocean_emulators/utils/train.py
alxmrs: I cannot use it before `compute()`, it throws an error when I do this. Let me try one more time to confirm.

I need to import the package in order to use the `einops` accessor.

src/ocean_emulators/constants.py
alxmrs: SGTM. Though, I'll add an om4 prefix, too.

src/ocean_emulators/datasets.py
alxmrs: Good shout.

src/ocean_emulators/utils/train.py
alxmrs: Ok, I actually read the full error message (below), and it looks like there _is_ a way to make this work (maybe). Let me give that a shot (thanks for your persistence here).

```
 ValueError: apply_ufunc encountered a chunked array on an argument, but handling for chunked arrays has not been enabled. Either set the ``dask`` argument or load your data into memory first with ``.load()`` or ``.compute()``

```

alxmrs: It seems like I've got it to work! I added a link to docs to help us tune this down the road.

PR #128, opened 2025-03-20
Adding remote data to tests + caching test data

alxmrs:
This PR addressed part of #146. It adds ~30 days worth of OM4 to the test data sources. To make this more performant, we add a local data caching mechanism.


alxmrs:
I think I need either #123 or #134 to be merged before I can make progress on this PR since the OM4 data does not naturally contain masking layers, but the default code path expects that it does. 
CC: @suryadheeshjith @jder

tests/conftest.py
```diff
@@ -343,36 +346,123 @@ def backend(request) -> TrainBackendConfig:
...
+            # for optimal download time. In local experiments, this time range (which
+            # matches the simulated data) is about 30 items.
+
+            data = retry_with_backoff(
+                (
+                    lambda: xr.open_zarr(REMOTE_DATA + "OM4", chunks=dict(time=50))
+                    .sel(time=slice("1975-08-05", "1976-03-31"))
+                    .compute()
+                )
+            )
```
alxmrs, line 446: This pattern was suggestion by Claude. Not to throw them under the bus; I kind of like it. 

I'd prefer not to add a retry library just for test utils.

configs/train_cm4.test.yaml
suryadheeshjith, line 1: Needs a file rename. We should probably separate out config files that use dummy data and simulation data like OM4/CM4

tests/conftest.py
```diff
@@ -1,24 +1,31 @@
...
+from aiohttp import ServerDisconnectedError
 from numpy.typing import ArrayLike, NDArray
 from typing_extensions import Self
 
 import ocean_emulators.constants as c
 from ocean_emulators.config import TrainBackendConfig, TrainConfig
 from ocean_emulators.utils.multiton import MultitonScope
 
+REMOTE_DATA = "https://nyu1.osn.mghpcc.org/m2lines-pubs/Samudra/"
+DEFAULT_CONFIG = "train_cm4.test.yaml"
```
suryadheeshjith: Same as above, rename to train_dummy or maybe train_default

configs/train_cm4_2step.test.yaml
suryadheeshjith, line 1: same as above

tests/conftest.py
```diff
@@ -343,36 +346,123 @@ def backend(request) -> TrainBackendConfig:
...
+
+@pytest.fixture(scope="session", params=["simulated", "remote-om4"])
+def data_source(request, pytestconfig) -> DataSource:
+    """Returns remote and in-memory `xarray.Dataset`s for tests."""
+    # Use cached data if available.
+    if cached_data := maybe_read_cache(cache_dir(pytestconfig), request.param):
+        return cached_data
+
+    match request.param:
+        case "simulated":
```
suryadheeshjith: rename "simulated" to "dummy" or something else

tests/test_datasets.py
```diff
@@ -64,6 +65,11 @@ def extract_sample_arrays(td: TrainData) -> tuple[np.ndarray, np.ndarray]:
     return np.stack(x_arrays, axis=0), np.stack(y_arrays, axis=0)
 
 
+def calc_num_samples(cfg: TrainConfig, time_slice: slice) -> int:
```
suryadheeshjith, line 68: I would like tests for this across configurations of the data!

suryadheeshjith: Just a sanity check 😁

alxmrs: Great idea!

configs/train_cm4.test.yaml
alxmrs: Good point, these are actually om4!

tests/conftest.py
alxmrs: Good shout, I like train_default.

tests/conftest.py
alxmrs: I'm torn between "mock" and "fake", but I prefer these to "dummy".

suryadheeshjith (approved):
LGTM

alxmrs:
Thanks! Merging!

PR #129, opened 2025-03-20
Fix multi-step training

### LINKED ISSUE #127: Train test should include multi-step training
Need to test multi-step training, replace step = [1] by step = [2] [here](https://github.com/suryadheeshjith/Ocean_Emulator/blob/main/configs/train_cm4.test.yaml#L26).

jder:
Make sure our mini training test runs multi-step training, fix bug this was hiding. 

Our validation tests were expecting the same number of steps but AFAICT validation always uses 1 step, so corrected this.

This lets us run tests for the current train config (the default), or for a specific one, or all configs. 

Fixes #127

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -225,14 +225,14 @@ def test_val__data_shape(val_loader_pair: LoaderPair):
     for i, sample in enumerate(loader):
         X, y = extract_sample_arrays(sample)
 
+        assert X.shape[0] == 1  # validation always has 1 step
         # Last validation batch may have fewer samples
-        assert X.shape[0] == cfg.steps[0]
         assert X.shape[1] == batch_size or (
             i == num_samples - 1 and X.shape[1] < batch_size
         )
         assert X.shape[2:] == (input_var_dim, 180, 360)
 
-        assert y.shape[0] == cfg.steps[0]
+        assert y.shape[0] == 1
         assert y.shape[1] == batch_size or (
             i == num_samples - 1 and y.shape[1] < batch_size
         )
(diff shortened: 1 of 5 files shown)


tests/test_datasets.py
```diff
@@ -175,7 +175,7 @@ def test_test_util__data_source_roundtrip(
 #  Changing the "hist" parameter breaks this test.
 def test_train__loads_correct_number_of_samples(train_loader_pair: LoaderPair):
     cfg, loader = train_loader_pair
-    n_samples = 13
+    n_samples = 12
```
jder: At some point it would be nice to compute this (or split the test config so we don't need to share between this test and other tests) but for now just updated the value.

jder:
FWIW the mini training run doesn't currently run at test time -- only as a benchmark. This is not ideal but a bit annoying to solve with our current test setup. But at least we'll catch it immediately post-merge.

jder:
For now I kicked off a benchmark on this branch so we can see it succeed https://github.com/suryadheeshjith/Ocean_Emulator/actions/runs/13976774712

suryadheeshjith (approved):

src/datasets.py
```diff
@@ -241,6 +241,11 @@ def get_input(self, step: int) -> TotalInput:
     def get_label(self, step: int) -> Label:
         return self.td_dict[step][1]
 
+    def merge_prognostic_and_boundary(self, prognostic: torch.Tensor, step: int):
```
alxmrs: Thanks!

tests/test_datasets.py
alxmrs: agreed! I'm not totally sure how it derives from the data IIRC.

alxmrs (approved):
LGTM

PR #132, opened 2025-03-21
Introduce Multitons

### LINKED ISSUE #87: Can't run CUDA and non-CUDA tests together

jder:
Please see docs & tests -- we basically can create scopes in which our singletons exist, and have those scopes get turned on when we use methods on the Trainer/Eval classes. 

I made GPU runners run CPU tests to prove this works, we can revert that before merging.

Fixes #87

### DIFF
--- a/src/ocean_emulators/eval.py
+++ b/src/ocean_emulators/eval.py
@@ -30,12 +30,10 @@ from ocean_emulators.utils.device import using_gpu
 from ocean_emulators.utils.distributed import is_main_process, set_seed
 from ocean_emulators.utils.logging import handle_logging, handle_warnings
 from ocean_emulators.utils.model import get_model_summary
-from ocean_emulators.utils.multiton import scoped
 from ocean_emulators.utils.wandb import WandBLogger
 
 
 class Eval:
-    @scoped
     def __init__(self, cfg: EvalConfig) -> None:
         cfg.prepare_output_dirs()
 
@@ -209,7 +207,6 @@ class Eval:
             long_rollout=True,
         )
 
-    @scoped
     def run(self) -> None:
         start_time = time.time()
         with self.tensor_map, self.normalize, self.wandb_logger:
--- a/src/ocean_emulators/eval.py
+++ b/src/ocean_emulators/eval.py
@@ -30,10 +30,12 @@ from ocean_emulators.utils.device import using_gpu
 from ocean_emulators.utils.distributed import is_main_process, set_seed
 from ocean_emulators.utils.logging import handle_logging, handle_warnings
 from ocean_emulators.utils.model import get_model_summary
+from ocean_emulators.utils.multiton import scoped
 from ocean_emulators.utils.wandb import WandBLogger
 
 
 class Eval:
+    @scoped
     def __init__(self, cfg: EvalConfig) -> None:
         cfg.prepare_output_dirs()
 
@@ -207,9 +209,11 @@ class Eval:
             long_rollout=True,
         )
 
+    @scoped
     def run(self) -> None:
         start_time = time.time()
-        inf_stats = self.standalone_inference()
+        with self.tensor_map, self.normalize, self.wandb_logger:
+            inf_stats = self.standalone_inference()
         time_elapsed = time.time() - start_time
 
         log_stats = {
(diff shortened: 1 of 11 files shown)


jder:
https://github.com/suryadheeshjith/Ocean_Emulator/actions/runs/13983043958/job/39151878490?pr=132 shows that we can run GPU + CPU tests in the same process. The failing tests are because we're extracting pieces of the Trainer and using them independently of the trainer -- shouldn't be too hard to fix tomorrow. (But is a clear annoyance of this approach.)

jder:
OK, I have something slightly more magical which I think solves this reasonably well, so this is ready for review.

jder:
ok, here's the passing run showing both GPU + CPU tests running together https://github.com/suryadheeshjith/Ocean_Emulator/actions/runs/13983281185/job/39152534928?pr=132. Will now revert that change.

src/ocean_emulators/eval.py
suryadheeshjith, line 1: Do we need to use this outside tests?

src/ocean_emulators/utils/multiton.py
suryadheeshjith, line 1: Brilliant! 😎

src/ocean_emulators/utils/multiton.py
```diff
@@ -0,0 +1,108 @@
...
+        # gets the instance created above
+        instance = SomeMultiton.instance()
+
+    """
+
+    def __init__(self):
+        self.scope = {}
+
+    def __enter__(self):
+        # TODO(jder): Could make this thread/async-safe if needed.
```
suryadheeshjith, line 35: Agreed, how can we test this? I do see the assertion in __exit__ but where else could things go wrong?

suryadheeshjith (approved):

src/ocean_emulators/eval.py
jder: Sorry, this was an earlier plan that I decided against, reverted.

src/ocean_emulators/utils/multiton.py
jder: If someone is making scopes active from multiple threads at once, the `__exit__` check should catch the problem. It might cause confusing results before that check runs, though. We could add a lock here to try and stop people from doing this ahead of time but I don't think we're using threads at all at the moment so didn't seem necessary. Up to you, though.

jder:
@suryadheeshjith I removed some more of the magic and the code in eval.py, LMK how you feel about the multithreading thing and I can merge.

src/ocean_emulators/utils/multiton.py
suryadheeshjith: Lets keep it as is then.

tests/conftest.py
```diff
@@ -355,28 +356,44 @@ def _make_path(name_with_ext: str) -> str:
...
         )
 
+        # Create a fresh scope to use in any tests using this config
+        # (see set_scope below)
+        scope = MultitonScope()
+        setattr(test_data_config, "_multiton_scope", scope)
+
         # After contextmanager closes, all test data will be automatically cleaned up.
-        yield test_data_trainer
+        yield test_data_config
```
alxmrs, line 372: First, the Multiton logic is brilliant. I haven't seen this pattern before, but it's quite suave. 

Second, and 🐑: Is it possible to use a normal ctx manager here before the yield? That should work on the pytest side, is there a reason why it hasn't been used? I also see it was used below. (Does it have to do with pytest scope?)

jder: I don't think we can, because I think pytest could:

1. create a train_config at session level
2. create another train_config at session level
3. use first train config for a given test
4. use second train config for a given test

I think if we used a normal context manager the yield in 1 will extend through to the end of 4, and the yield in 2 will extend through the end of 4. Meaning that 3 will see the wrong scope (ie for the second config instead of the first). 

Unless I'm missing a guarantee pytest gives us about the ordering of these fixtures?

alxmrs: Thanks, this makes it really concrete to me how Multitons work with scope in Pytest. 

> Unless I'm missing a guarantee pytest gives us about the ordering of these fixtures?

Maybe? But I don't think it's worth investigating of you have it working now.

PR #135, opened 2025-03-21
Restore benchmarks to old behavior

jder:
This was a silly bug but after looking at this we probably want to leave them separate so the benchmark web format doesn't change.


alxmrs (approved):
It was worth a shot! LGTM

PR #137, opened 2025-03-21
Update ruff to avoid fighting with VS Code version

jder:
These keep flipping back and forth for me since the VS Code extension version is more recent.


alxmrs (approved):
Approved when tests pass.

suryadheeshjith:
I was battling with this a bit earlier today, glad you found the fix!

PR #139, opened 2025-03-21
Resolve CFTime warning

### LINKED ISSUE #112: Resolve warnings from zarr & cftime upgrades
```
tests/test_datasets.py: 14 warnings
  /Users/jder/oa/Ocean_Emulator/tests/conftest.py:103: CFWarning: this date/calendar/year zero convention is not supported by CF
    cftime.num2date(num, units, calendar=self.start_day.calendar)

tests/test_datasets.py::test_train__loads_correct_number_of_samples[cpu]
  /Users/jder/oa/Ocean_Emulator/.venv/lib/python3.11/site-packages/zarr/api/asynchronous.py:203: UserWarning: Consolidated metadata is currently not part in the Zarr format 3 specification. It may not be supported by other zarr implementations and may change in the future.
    warnings.warn(
```

_Originally posted by @jder in https://github.com/suryadheeshjith/Ocean_Emulator/issues/108#issuecomment-2718901726_

jder:
The underlying [warning ](https://github.com/Unidata/cftime/blob/eab0212df73decd5a420adaae7938b942d131c95/src/cftime/_cftime.pyx#L1151) was due to us passing a date with year < 0 for a calendar that doesn't support that. 

This was for 2 reasons: first, we had no minimum in the possible generated dates and second because the fromordinal function we were using takes the number of days since year -4713 (yes, negative four thousand years BCE) but the python datetime toordinal produces the number of days since year 1. So, we have a minimum now and we are using a conversion function which is a little less silly. 

Fixes #112

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -88,7 +88,11 @@ def vector_of(max_vec_size: int, min_vec_size=1):
         shape=vector_of(50),
         elements=st.integers(min_value=0, max_value=999),
     ),
-    start_day=st.datetimes(),
+    start_day=st.datetimes(
+        min_value=datetime.datetime(
+            1900, 1, 1
+        ),  # to quiet cftime warning about year < 0
+    ),
     calendar=st.sampled_from(["noleap", "standard"]),
 )
 @example(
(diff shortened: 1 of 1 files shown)


tests/test_datasets.py
```diff
@@ -88,47 +88,49 @@ def vector_of(max_vec_size: int, min_vec_size=1):
         shape=vector_of(50),
         elements=st.integers(min_value=0, max_value=999),
     ),
-    start_day=st.datetimes(),
+    start_day=st.dates(
+        min_value=datetime.date(1900, 1, 1),  # to quiet cftime warning about year < 0
```
alxmrs, line 92: Lol, amazing!

alxmrs (approved):
Thanks for doing this, I can only imagine how finicky it might have been to uncover this from docs, etc.

Issue #116, opened 2025-03-14
Merge in om4 compatibility & public refactors


jder:
After March 20: Merge in the om4-checks branch changes and the [public Samudra changes](https://openathena.slack.com/archives/C08CYM42DT3/p1741783501273679) into this codebase, with a release plan to the public codebase moving forward.

jder:
Let's do this now so we can profile more easily.

suryadheeshjith:
I can work on this if you haven't started working on it already

jder:
Great! I haven't yet started so it's all yours.

jder:
You might want to wait until #110 lands (hopefully tomorrow) before making any major changes to avoid all kinds of conflicts

PR #134, opened 2025-03-21
[Draft] Public repo refactors

suryadheeshjith:
* Ability to train on OM4 or CM4
* Bunch of renaming and removal of unused experiment configs in constants
* Added metadata
* The scripts for training on gantry, empire ai etc. may not work and I am planning to fix that in a later PR

Closes Issues #116 , #74 , #80 , #101 , #69

Issue #101, opened 2025-03-07
Error Logging


suryadheeshjith:
Errors are not logged correctly to error.log.

jder:
Might also be nice to mark weights & biases as failed when we end training due to an exception

Issue #74, opened 2025-02-20
Number of rollout steps != Number of model steps.


suryadheeshjith:
In get_time_slice() [here](https://github.com/suryadheeshjith/Ocean_Emulator/blob/master/src/utils/data.py), comment mentions that num_steps does not include initial prognostic but it does. Some additional refactoring - 

1. Rename to num_rollout_steps or some equivalent in the function and the rest of the code. 
2. This value should only be passed to inference aggregator in train_3D.py. Like eval.py, 

In Train, we are passing num_steps
```
Stepper.inference(
                model=self.model.module if using_gpu() else self.model,
                dataset=inference_dataset,
                inf_aggregator=inf_aggregator,
                epoch=epoch,
                num_model_steps_forward=num_steps,
            )
```

In eval.py, We pass a parameter from config
```
Stepper.inference(
            model=self.model,
            dataset=self.inference_dataset,
            inf_aggregator=inf_aggregator,
            epoch=0,
            output_dir=self.output_dir,
            model_path=self.model_path,
            num_model_steps_forward=self.num_model_steps_forward,
            record_every=self.record_every,
            save_zarr=self.save_zarr,
        )
```

3. Since this parameter is being used only in inference, maybe we can think of some other way to pass this.

PR #142, opened 2025-03-24
March refactors - functionality

suryadheeshjith:
* Ability to train on OM4 or CM4
* Added metadata
* Reduced checkpointing
* Removed unnecessary zos check in wandblogger
* Logger fix
* Logging in eval fix
* Train/val/inference time range extraction fix

Following Draft PR https://github.com/suryadheeshjith/Ocean_Emulator/pull/134

Closes Issues #116 , #74 , #101 , #69 , #59


src/ocean_emulators/constants.py
```diff
@@ -122,11 +166,17 @@ def construct_metadata(data: xr.Dataset) -> Dict[str, Dict[str, str]]:
...
         except AttributeError:
-            logging.info(f"{var} has no long_name or units attribute")
-            metadata[str(var)] = {
-                "long_name": "Unknown",
-                "units": "Unknown",
-            }
+            if var in DEFAULT_METADATA.keys():
+                metadata[str(var)] = DEFAULT_METADATA[str(var)]
+            elif str(var).split("_")[0] in DEFAULT_METADATA.keys():
+                metadata[str(var)] = DEFAULT_METADATA[str(var).split("_")[0]]
```
alxmrs: Consider using the walrus operator here:
```suggestion
            elif (key := str(var).split("_")[0]) in DEFAULT_METADATA.keys():
                metadata[str(var)] = DEFAULT_METADATA[key]
```

src/ocean_emulators/train.py
```diff
@@ -609,11 +621,10 @@ def init_data_loaders(self, cur_step: int) -> None:
...
                     data=self.data.sel(
-                        time=get_time_slice(
-                            self.train_times,
-                            time_delta=self.time_delta,
-                            hist=self.hist,
-                        )[0]
+                        time=slice(
+                            self.train_times.start_time,
+                            self.train_times.end_time,
+                        )
```
alxmrs: I like this version a lot! A great improvement.

alxmrs: 🐑  Consider storing a slice in `train_times` if this is how it's going to be used anyway: 

```

self.train_times = slice([start_time], [end_time])
# ...
self.data.sel(time=self.train_times)
```
(This applies to all similar usages, e.g. val_times).

src/ocean_emulators/train.py
```diff
@@ -679,58 +689,54 @@ def init_data_loaders(self, cur_step: int) -> None:
             collate_fn=collate_train_data,
         )
 
-    def save_checkpoint(self, epoch, v_loss, inf_loss):
-        save_best_val = False
-        # Check for best validation loss
-        if v_loss < self.best_val_loss:
+    def save_all_checkpoints(self, epoch, v_loss, inf_loss):
```
alxmrs: Do you mind adding types to these arguments?

src/ocean_emulators/utils/data.py
```diff
@@ -53,24 +55,17 @@ def spherical_area_weights(data: xr.Dataset) -> Grid:
     return weights
 
 
-def get_time_slice(
-    time_config: config.TimeConfig, time_delta: int = 5, hist: int = 1
-) -> tuple[slice, int]:
+def get_inference_steps(time_config, time_delta=5, hist=1):
```
alxmrs: Do you mind adding types to these arguments?

src/ocean_emulators/utils/data.py
```diff
@@ -122,6 +117,83 @@ def get_norm_unnorm_dicts(
     return data_dict, data_unnorm_dict
 
 
+def compute_anomalies(data: xr.Dataset, var: str) -> xr.Dataset:
+    """
+    Compute the anomalies of a data variable.
+    """
+    climatology = data[var].groupby("time.dayofyear").mean("time").compute()
```
alxmrs: In the future, we might consider calculating this up-front for all variables.

Also, consider "pulling the loop down" for this function (assuming we aim to apply it to multiple variables): https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

src/ocean_emulators/utils/data.py
```diff
@@ -122,6 +117,83 @@ def get_norm_unnorm_dicts(
...
+        data[var] - climatology.sel(dayofyear=day_of_year)
+    ).compute()
+    return data
+
+
+def rename_vars(data: xr.Dataset) -> xr.Dataset:
+    """
+    Rename variables if required.
+    """
+    for var in data.variables:
```
alxmrs, line 238: Let's make a copy of the input data to make this function immutable (if we are returning a dataset).

src/ocean_emulators/utils/data.py
```diff
@@ -122,6 +117,83 @@ def get_norm_unnorm_dicts(
...
+    climatology = data[var].groupby("time.dayofyear").mean("time").compute()
+    # Remove the seasonal cycle (climatology) from the detrended data
+    day_of_year = data[var]["time"].dt.dayofyear
+    data[var + "_anomalies"] = (
+        data[var] - climatology.sel(dayofyear=day_of_year)
+    ).compute()
+    return data
+
+
+def rename_vars(data: xr.Dataset) -> xr.Dataset:
```
alxmrs, line 233: I would love a few unit tests for this function.

src/ocean_emulators/utils/data.py
```diff
@@ -122,6 +117,83 @@ def get_norm_unnorm_dicts(
...
+            lev_in_depth_idx = DEPTH_LEVELS.index(lev_in_depth)
+            data = data.rename({var_str: var + "_" + str(lev_in_depth_idx)})
+    return data
+
+
+def validate_data(
+    data: xr.Dataset,
+    data_mean: xr.Dataset,
+    data_std: xr.Dataset,
+) -> Tuple[xr.Dataset, xr.Dataset, xr.Dataset]:
```
alxmrs: Let's use `tuple` (the built-in).

src/ocean_emulators/utils/data.py
```diff
@@ -122,6 +117,83 @@ def get_norm_unnorm_dicts(
...
+    """
+    Validate the data such that we have the correct format for training.
+    """
+    # Check if Mask variables exist
```
alxmrs, line 263: Consider making copies of the inputs or at least taking care that the transformations are immutable.

src/ocean_emulators/constants.py
```diff
@@ -112,6 +112,50 @@
...
+        "long_name": "Surface ocean heat flux from "
+        "SW+LW+latent+sensible+masstransfer+frazil+seaice_melt_heat",
+        "units": "W/m^2",
+    },
+    "hfds_anomalies": {
+        "long_name": "hfds anomalies",
+        "units": "W/m^2",
+    },
+}
+
```
jder, line 165: What's stopping us from writing this metadata into the underlying OM4 zarr file rather than having this in code? (Just trying to understand what constraints we have since I know Alex is about to want to make changes to the data storage eg chunking etc)

src/ocean_emulators/eval.py
```diff
@@ -273,7 +276,13 @@ def main():
     handle_warnings()
 
     Evaluator = Eval(cfg)
-    Evaluator.run()
+
+    try:
+        Evaluator.run()
+    except Exception as e:
+        # Log the exception with traceback
+        logging.exception(e)
```
jder:
```suggestion
        logging.exception("Evaluation failed with an exception")
```

src/ocean_emulators/utils/train.py
```diff
@@ -33,3 +34,23 @@ def collate_inference_data(
...
+    def latest_checkpoint_path_with_epoch(self, epoch: int) -> Path:
+        return self.checkpoint_dir / f"ckpt_{epoch}.pt"
+
+    @property
+    def best_inference_checkpoint_path(self) -> Path:
+        return self.checkpoint_dir / "best_inference_ckpt.pt"
+
+    @property
+    def best_validation_checkpoint_path(self) -> Path:
+        return self.checkpoint_dir / "best_validation_ckpt.pt"
```
jder, line 87: 🐑 Personally I might name these "latest.pt", "epoch_{epoch}.pt", "best_inference.pt" and "best_validation.pt".

src/ocean_emulators/train.py
```diff
@@ -679,58 +674,54 @@ def init_data_loaders(self, cur_step: int) -> None:
             collate_fn=collate_train_data,
         )
 
-    def save_checkpoint(self, epoch, v_loss, inf_loss):
-        save_best_val = False
-        # Check for best validation loss
-        if v_loss < self.best_val_loss:
+    def save_all_checkpoints(self, epoch: int, v_loss: float, inf_loss: float):
+        save_best_checkpoint = False
```
jder: 🐑 How about `is_best_val_loss`?

src/ocean_emulators/eval.py
jder: (Looks like this should also be changed for train.py)

jder (approved):
Awesome! Excited to get this in so we can get everything in a single place. Thank you for doing this! Just a few questions and some optional 🐑  from me.

src/ocean_emulators/utils/data.py
suryadheeshjith: Neat

src/ocean_emulators/utils/data.py
suryadheeshjith: Ill add them!

src/ocean_emulators/constants.py
suryadheeshjith: Nothing is, its just that we now have too many sources of data and it would be good to have some temporary quick fixes.

src/ocean_emulators/utils/train.py
suryadheeshjith: Keeping it as is since we might be saving data stores that are torch tensors as well

src/ocean_emulators/train.py
suryadheeshjith: A function?

jder: No, sorry, I was just suggesting a rename of the variable.

suryadheeshjith: Much better!

suryadheeshjith:
@alxmrs Made some variable changes to the OM4 Data loader

alxmrs:
Thanks, that's fine by me. I don't mind either way; I chose the other names because they were short.

src/ocean_emulators/config.py
```diff
@@ -26,6 +26,9 @@ class TimeConfig:
     start_time: str
     end_time: str
 
+    def __post_init__(self):
```
alxmrs: Nice

🐑  instead of a __post_init__, you could also make this a getter-only property.

src/ocean_emulators/constants.py
```diff
@@ -25,9 +25,8 @@
 #   for return types, referring only to axes annotated for arguments.
 # So, we'll leave this default and use symbolic axes locally.
 Input: TypeAlias = Float[Grid, "*batch total_vars"]
-Label: TypeAlias = Prognostic
 
-Example = tuple[Input, Label] | tuple[xr.Dataset, xr.Dataset]
+Example = tuple[Input, Prognostic] | tuple[xr.Dataset, xr.Dataset]
```
alxmrs, line 29: It doesn't make too much of a difference, but you don't like the `Label` type alias?

alxmrs: 🐑

suryadheeshjith: I just made it consistent with the rest of the codebase, we can switch to having a duplicate type alias in a later PR if needed.

src/ocean_emulators/train.py
```diff
@@ -617,23 +621,12 @@ def init_data_loaders(self, cur_step: int) -> None:
...
-            hist=self.hist,
-        )[0]
-
         match self.loader_version:
             case TrainDataset.FLAG:
                 train_data: ConcatDataset = ConcatDataset(
                     [
                         TrainDataset(
-                            self.data.sel(time=train_slice),
+                            data=self.data.sel(time=self.train_times.time_slice),
```
alxmrs, line 629: This is slick!

alxmrs (approved):
Great work! This all looks great.

src/ocean_emulators/constants.py
alxmrs: SGTM

suryadheeshjith:
@jder While adding a test, I had the issue - 

tests/test_utils_data.py:4: error: Skipping analyzing "scipy.stats": module is installed, but missing library stubs or py.typed marker  [import-untyped]
tests/test_utils_data.py:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 2 source files)

I fixed pyproject.toml but is this the way to do it?

jder:
@suryadheeshjith I think that works; we can probably get better errors with https://github.com/scipy/scipy-stubs but can also be a follow-up.

jder:
Woo! Excited to get this landed, should let us get going on regression tests + some benchmarking stuff on OM4 data. 🎉

Issue #140, opened 2025-03-22
Hypothesis tests for interpretable test data generator are flaky (they often exceed the deadline).


alxmrs:
CC: @jder.

jder:
How does this show up? I either haven't seen it or haven't known what I was seeing.

alxmrs:
Check out the last few actions run on this PR: https://github.com/suryadheeshjith/Ocean_Emulator/pull/123

alxmrs:
Fixed in #143.

PR #143, opened 2025-03-24
Omitting filtering by guaranteeing inputs are unique.

alxmrs:
Fixing #140 by making omitting slow step (`itertools.product`), instead guaranteeing that all inputs are unique.

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -75,13 +75,13 @@ def vector_of(max_vec_size: int, min_vec_size=1):
 @given(
     data_var_index=st.integers(min_value=0, max_value=255),
     lat=arrays(
-        dtype=np.float64,
+        dtype=np.float16,
         shape=vector_of(50),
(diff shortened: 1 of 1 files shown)


jder:
The problem is that you're asking for unique float64s but in fact we only guarantee that unique float16s are mapped to unique values. You can either generate unique float16s directly or else compare the DataSourceDims for uniqueness (since they internally map to float16). If you want to do the latter we probably need to override `__hash__` too. But FWIW I think it doesn't really test the thing @suryadheeshjith was asking for which was uniqueness *for the synthetic data* unless we also verify the synthetic data is unique as float16s. (Which I am pretty sure it is today)

alxmrs:
Thanks, that was a good insight. I now got the tests to pass with a much tighter deadline.

jder (approved):
Makes sense to me, thanks!

alxmrs:
Thanks for the review!

Issue #126, opened 2025-03-19
Migrate Training code away from TrainData


alxmrs:

jder:
FWIW I think we should probably keep TrainData but change its internals to be more efficient.

alxmrs:
I agree, I don't think we want this, actually. In similar projects, this is called a "Batch". Closing this issue.

PR #148, opened 2025-03-25
Fixes to the lazy data loader given more realistic data.

alxmrs:
Extracted some pre-conditions from the `validate_data` function into smaller parts and applying some of them to all input from the current data loader.

In addition, I've caught a minor issue in this loader's collate_fn as well as my new data loader test.

This fixes were found during #128.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -39,7 +39,7 @@ class OM4Dataset(Dataset):
         is_inference: bool = False,
     ) -> None:
         # Ensure that a `wetmask` DataArray exists along a `lev` dimension.
-        data_ = data.copy().pipe(unflatten_masks)
+        data_ = data.pipe(unflatten_masks)
 
         prognostic = data_[prognostic_var_names]
         boundary = data_[boundary_var_names]
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -39,7 +39,7 @@ class OM4Dataset(Dataset):
         is_inference: bool = False,
     ) -> None:
         # Ensure that a `wetmask` DataArray exists along a `lev` dimension.
-        data_ = data.pipe(unflatten_masks)
+        data_ = unflatten_masks(data)
 
         prognostic = data_[prognostic_var_names]
         boundary = data_[boundary_var_names]
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -19,7 +19,13 @@ from ocean_emulators.constants import (
     PrognosticMask,
     PrognosticVarNames,
 )
-from ocean_emulators.utils.data import Normalize, mask, unflatten_masks
+from ocean_emulators.utils.data import (
+    Normalize,
+    mask,
+    rename_coords,
+    rename_vars,
+    unflatten_masks,
+)
 from ocean_emulators.utils.device import get_device, using_gpu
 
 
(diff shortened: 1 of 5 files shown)


src/ocean_emulators/utils/data.py
```diff
@@ -256,38 +265,12 @@ def validate_data(
...
-        data_copy = data_copy.drop_vars("wetmask")
+    data_copy = data.copy().pipe(flatten_masks).pipe(rename_vars).pipe(rename_coords)
 
     # Check if data variables are in the right format
     # This check is to ensure we convert data to the correct format
-    data_copy = rename_vars(data_copy)
-    data_mean_copy = rename_vars(data_mean_copy)
-    data_std_copy = rename_vars(data_std_copy)
-
-    # OM4 data has coordinates we don't need
```
suryadheeshjith, line 284: Why is this removed?

src/ocean_emulators/utils/data.py
```diff
@@ -248,6 +243,20 @@ def rename_vars(data: xr.Dataset) -> xr.Dataset:
     return data_copy
 
 
+def rename_coords(data: xr.Dataset) -> xr.Dataset:
```
jder: 🐑 How about `ensure_lat_lon` or `with_lat_lon_coords` or similar? (Could also do something similar for "rename_vars")

src/ocean_emulators/datasets.py
```diff
@@ -38,20 +44,23 @@ def __init__(
         stride: int = 1,
         is_inference: bool = False,
     ) -> None:
-        # Ensure that there is a `wetmask` DataArray with a supported `lev` dimension.
-        data = unflatten_masks(data)
+        # Ensure that the input data is in the correct format. This means:
+        # - Rename variables to match the `PrognosticVarNames` and `BoundaryVarNames`.
+        # - Extract a `wetmask` DataArray that is alone a `lev` dimension.
+        # - Drop unnecessary coordinates and make sure `x`/`y` are `lon`/`lat`.
+        data_ = data.pipe(rename_vars).pipe(unflatten_masks).pipe(rename_coords)
```
jder: Do we need to do rename_vars/rename_coords since it's already done in validate_data?

jder: I also don't understand why we're flattening in validate_data and then unflattening here. Perhaps the step in validate_data could ensure we have both flat and unflat masks if we need both depending on the type of dataset we're using?

src/ocean_emulators/utils/data.py
jder: I think this just moved into rename_coords, now called in the first line of this function

jder (approved):
Just a couple questions about the sequencing/duplication between validate_data and the stuff we're doing in OM4Dataset, but generally makes sense to me!

src/ocean_emulators/utils/data.py
suryadheeshjith: Got it!

src/ocean_emulators/utils/data.py
alxmrs: Interesting, I'll think about alternative names 🤔

src/ocean_emulators/datasets.py
alxmrs: I honestly think that these type of data preconditions should live in the data loaders and not in the training pipeline altogether; e.g. I'd eventually like to migrate `validate_data` into `TrainDataset`. I think this would be preferable for a few reasons: 
- Different data loaders will expect different preconditions from the data. The flatten vs unflatten is a perfect example of this (TrainDataset expects "flat" variables; my data loader expects the opposite
- I think it may make it easier to swap out different input datasets and massage them into a common example interface for training (`TrainData` now, something else later #126).

alxmrs: Decided in chat: use `validate_data` now, consider pre-conditions later on.

src/ocean_emulators/utils/data.py
```diff
@@ -106,12 +106,7 @@ def mask(data: xr.Dataset, wetmask: xr.DataArray) -> xr.Dataset:
         # If the name has four tokens, then it definitely is at some depth level (i.e.,
         # not at the surface).
         if len(tokens) >= 4:  # OM4 data format (e.g., {variable}_lev_{level}_{decimal})
-            # TODO(alxmrs): Is the OM4 data wrong? Is preprocessing done somewhere?
-            # In this format, "level" is a member of DEPTH_LEVELS, _not_ DEPTH_I_LEVELS.
-            # Thus, we need to convert it to the corresponding DEPTH_I_LEVELS index.
-            _, _, level, *_ = tokens
-            closest_level = min(DEPTH_LEVELS, key=lambda x: abs(x - float(level)))
-            level = DEPTH_I_LEVELS[DEPTH_LEVELS.index(closest_level)]
+            raise ValueError("please call `rename_vars` before masking!")
```
```suggestion
            raise ValueError("please call `vars_as_level_index` before masking!")
```

alxmrs: Good catch!

jder (approved):

PR #149, opened 2025-03-26
Fix parsing of loader_version config parameter.

alxmrs:


suryadheeshjith (approved):

jder:
What was the problem here? Can you add some context to the PR for posterity?

alxmrs:
dacite doesn't know how to parse an enum (as far as I configured it). Before, we were parsing it as a str, but I made a mistake of choosing a bad default str: this PR changes the default from `LoaderVersion.OM4_EAGER` to `om4-eager`.

PR #151, opened 2025-03-28
Covering more configurations (hist>0, multiple loader_versions); fixed the lazy data loader to pass all cases.

alxmrs:
This PR is a combination of #152 and #154. It adds tests for several history values as well as all the loaders covered by a feature flag. In addition, it adds fixes in the new data loader for the resulting broken configurations.


suryadheeshjith (approved):
LGTM

PR #152, opened 2025-03-28
Lazy data loader now supports history>0.

alxmrs:
These fixes were found during #151; I've extracted the fixes into a smaller, reviewable PR. These changes mimic the original data loader implementation, but do all the work in Xarray, and thus are still lazily evaluated. This fix (and a couple other ones floating around) are necessary to perform an e2e performance test.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -7,6 +7,7 @@ import xarray as xr
 from einops import rearrange
 from jaxtyping import Float, Integer
 from torch.utils.data import Dataset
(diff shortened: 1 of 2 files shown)


suryadheeshjith:
Will you be adding tests or are they in #154 ? Can you merge them here?

alxmrs:
> Will you be adding tests or are they in https://github.com/suryadheeshjith/Ocean_Emulator/pull/154 ? Can you merge them here?

Funnily enough, at this point #151 is just these fixes + the associated tests. How about I get that ready for review?

alxmrs:
Fixed in #151.

PR #153, opened 2025-03-28
Eval checkpoint load fix

suryadheeshjith:
Needed to remove the "module." component in the keys of the checkpoint


src/ocean_emulators/eval.py
```diff
@@ -187,6 +185,15 @@ def __init__(self, cfg: EvalConfig) -> None:
         self.model_path = cfg.ckpt_path
         self.init_inference_store()
 
+    def load_checkpoint(self, ckpt_path):
```
alxmrs: 🐑  would you mind adding a type to `ckpt_path` (is it a `pathlib.Path` or `str`)?

src/ocean_emulators/eval.py
```diff
@@ -187,6 +185,15 @@ def __init__(self, cfg: EvalConfig) -> None:
...
+        checkpoint = torch.load(ckpt_path, map_location=torch.device("cuda"))
```
alxmrs: Do we want to get the device from the designated backend?

src/ocean_emulators/eval.py
```diff
@@ -187,6 +185,15 @@ def __init__(self, cfg: EvalConfig) -> None:
...
+        model_state_dict = checkpoint["model"]
+        new_state_dict = OrderedDict()
+        for k, v in model_state_dict.items():
+            name = k[7:]  # remove `module.`
```
alxmrs: Consider using `removeprefix()` instead: https://docs.python.org/3/library/stdtypes.html#str.removeprefix

src/ocean_emulators/eval.py
alxmrs: Ok, it looks like this was just moved -- feel free to disreguard!

alxmrs (approved):
A few minor recommendations, but overall this fix LGTM.

jder:
@suryadheeshjith Is it easy to write a test for this functionality?

suryadheeshjith:
Yepp, we can add a test

PR #154, opened 2025-03-28
Cover loader versions and hist>0 in test fixtures.

alxmrs:
These increase combinatorially increase the number of tests covered by also checking for hist values from 0-2 as well as checking different loader versions. These tests were originally found in #151.


alxmrs:
I think #152 needs to be merged before this one.

suryadheeshjith (approved):
LGTM

alxmrs:
Fixed in #151.

PR #156, opened 2025-03-31
Fixed anomalies statistics for anomalies use

suryadheeshjith:
Need to compute statistics when you compute anomalies before training

### DIFF
--- a/src/ocean_emulators/utils/data.py
+++ b/src/ocean_emulators/utils/data.py
@@ -205,11 +205,18 @@ def get_anomalies_vars(var_names: BoundaryVarNames) -> tuple[str, ...]:
     return tuple([var for var in var_names if var.endswith("_anomalies")])
 
 
-def compute_anomalies(data: xr.Dataset, anomalies_vars: tuple[str, ...]) -> xr.Dataset:
+def compute_anomalies(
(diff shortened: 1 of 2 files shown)


src/ocean_emulators/utils/data.py
```diff
@@ -205,11 +205,18 @@ def get_anomalies_vars(var_names: BoundaryVarNames) -> tuple[str, ...]:
     return tuple([var for var in var_names if var.endswith("_anomalies")])
 
 
-def compute_anomalies(data: xr.Dataset, anomalies_vars: tuple[str, ...]) -> xr.Dataset:
+def compute_anomalies(
+    data: xr.Dataset,
+    data_mean: xr.Dataset,
+    data_std: xr.Dataset,
```
alxmrs, line 211: Should we move `DataSource` from conftest to here (in a future PR)? It seems like these three datasets go together quite often.

alxmrs (approved):
LGTM!

src/ocean_emulators/utils/data.py
suryadheeshjith: Lets keep this for now but true, they do move around together most times.

PR #158, opened 2025-03-31
Compact script should compact means and stds, too.

alxmrs:


suryadheeshjith (approved):
LGTM

jder:
Aren't the means/stds trivially small?

alxmrs:
> Aren't the means/stds trivially small?

Yes, but I'd like the data to be a consistent shape to make it easier to apply preconditions.

PR #159, opened 2025-04-01
Added movie to notebook

suryadheeshjith:
Also linked ReviewNB to this repository for easy notebook diff


suryadheeshjith:
Merging this since it is a notebook I use quite often and hope we come to a consensus on how to store notebooks and this one in particular which will be executed after every training run.

PR #162, opened 2025-04-01
Vectorizing steps when creating examples in the lazy loader.

### LINKED ISSUE #124: Vectorize creating examples (avoid iterating through each `step`).

alxmrs:
Fixes #124.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -123,7 +123,7 @@ class OM4Dataset(Dataset):
 
     def __getitem__(self, idx: int) -> Example:
         windows = [self.window_from(idx, step) for step in range(self.steps)]
-        window = xr.concat(windows, dim="window")
+        window = xr.concat(windows, dim="step")
 
         # This point in time splits the training data and the label data!
         time_split = self.hist + 1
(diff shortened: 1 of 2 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -89,7 +89,7 @@ def __len__(self) -> int:
     # TODO(#124): Vectorize `step`
```
```suggestion
```

src/ocean_emulators/datasets.py
```diff
@@ -119,61 +119,47 @@ def window_from(
...
 
     def __getitem__(self, idx: int) -> Example:
-        inputs = []
-        labels = []
-
-        # TODO(#124): Vectorize this operation! Or, is it fine because it's lazy?
-        for step in range(self.steps):
-            window = self.window_from(idx, step)
+        windows = [self.window_from(idx, step) for step in range(self.steps)]
+        window = xr.concat(windows, dim="window")
```
jder: I think this should be `dim="step"` -- we still want/need to have each step of the rollout separately. I think `window` is currently a unary dimension so this is effectively the same thing but I think keeping the dimension named "window" is somewhat confusing (and would be wrong if it's every not unary). We might just drop the window dimension if it's always unary but hard for me to tell if that's true in all cases.

jder (changes requested):

src/ocean_emulators/datasets.py
alxmrs: Now that we're trying to vectorize the operation, `window` is no longer a unary dimension. From my reading, the function of the loop was to translate several `step`s into a several unary `window`s. In other words, these concepts are related. 

I do agree that this is a bit confusing. I'll try to see if I can replace it with "step", but I think that the `isel()` method depends on the values being named `window`. Let me check...

alxmrs: Ok, I didn't think it would work, and it does! Great suggestion. How it works with isel and einsum breaks my brain a little bit. It reads a lot clearer now, and the change is even smaller.

jder (approved):
Thanks!

PR #163, opened 2025-04-01
Switched land values to nan for plotting and metric calculation

suryadheeshjith:
- RMSE during train/eval will increase because of this change as we are considering land as nans and not 0s anymore
- Removed some redundant normalization functions
- Normalization functions weirdly changed the data type from float32 to float64 so I converted the final result back to float32
- Fixed plotting to handle Nan values, now land looks a lot better in the maps
- Added tests for aggregator metric weighted_mean and normalization functions

### DIFF
--- a/src/ocean_emulators/aggregator/metrics.py
+++ b/src/ocean_emulators/aggregator/metrics.py
@@ -3,32 +3,6 @@ from typing import Optional
 import torch
 
 
-def weighted_mean_with_nan_as_zero(
-    tensor: torch.Tensor,
-    weights: Optional[torch.Tensor] = None,
-    dim: tuple[int, ...] = (-2, -1),
-    keepdim: bool = False,
-) -> torch.Tensor:
-    """Computes the weighted mean across the specified list of dimensions.
-
-    Args:
-        tensor: torch.Tensor
-        weights: Weights to apply to the mean.
-        dim: Dimensions to compute the mean over.
-        keepdim: Whether the output tensor has `dim` retained or not.
-
-    Returns:
-        a tensor of the weighted mean averaged over the specified dimensions `dim`.
-    """
-    if weights is None:
-        return tensor.mean(dim=dim, keepdim=keepdim)
-
-    weights = weights.to(tensor.device)
-    return (tensor * weights).sum(dim=dim, keepdim=keepdim) / weights.expand(
-        tensor.shape
-    ).sum(dim=dim, keepdim=keepdim)
-
-
 def weighted_mean(
     tensor: torch.Tensor,
     weights: Optional[torch.Tensor] = None,
@@ -54,6 +28,34 @@ def weighted_mean(
         dim=dim, keepdim=keepdim
     )
     return (tensor * weights).nansum(dim=dim, keepdim=keepdim) / denom
+
+
+def weighted_mean_with_nan_as_zero(
+    tensor: torch.Tensor,
+    weights: Optional[torch.Tensor] = None,
+    dim: tuple[int, ...] = (-2, -1),
+    keepdim: bool = False,
+) -> torch.Tensor:
+    """This function used purely for testing.
+
+    Similar to weighted_mean, but treats assumes input with Nans as zeros.
+
+    Args:
+        tensor: torch.Tensor
+        weights: Weights to apply to the mean.
+        dim: Dimensions to compute the mean over.
+        keepdim: Whether the output tensor has `dim` retained or not.
+
+    Returns:
+        a tensor of the weighted mean averaged over the specified dimensions `dim`.
+    """
+    if weights is None:
+        return tensor.mean(dim=dim, keepdim=keepdim)
+
+    weights = weights.to(tensor.device)
+    return (tensor * weights).sum(dim=dim, keepdim=keepdim) / weights.expand(
+        tensor.shape
+    ).sum(dim=dim, keepdim=keepdim)
 
 
 def area_weighted_mean(
--- a/src/ocean_emulators/aggregator/metrics.py
+++ b/src/ocean_emulators/aggregator/metrics.py
@@ -11,6 +11,9 @@ def weighted_mean(
 ) -> torch.Tensor:
     """Computes the weighted mean across the specified list of dimensions.
 
+    This function treats NaNs as land cells and does not include them in the
+    computation.
+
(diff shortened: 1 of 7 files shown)


src/ocean_emulators/aggregator/metrics.py
```diff
@@ -11,6 +11,35 @@ def weighted_mean(
 ) -> torch.Tensor:
     """Computes the weighted mean across the specified list of dimensions.
 
+    Args:
+        tensor: torch.Tensor
```
jder, line 18: How about documenting here how we treat NaNs in the tensor?

src/ocean_emulators/aggregator/metrics.py
```diff
@@ -11,6 +11,35 @@ def weighted_mean(
...
+
+def weighted_mean_with_nan_as_zero(
+    tensor: torch.Tensor,
+    weights: Optional[torch.Tensor] = None,
+    dim: tuple[int, ...] = (-2, -1),
+    keepdim: bool = False,
+) -> torch.Tensor:
+    """This function used purely for testing.
+
+    Similar to weighted_mean, but treats assumes input with Nans as zeros.
```
jder: Any reason to keep this around? Seems like we don't use it. If we do want to keep it around, I'm not sure what it means to "treats assumes input with Nans as zeros". How about "assumes there are no NaNs in tensor" or just "returns NaN if there are any NaNs in tensor"? (And maybe change the function name to match eg `weighted_mean_no_nans`?)

src/ocean_emulators/utils/data.py
```diff
@@ -370,9 +364,12 @@ def normalize_tensor_prognostic(
         norm = (data - tensor_mean) / tensor_std
         if fill_nan:
             norm = norm.nan_to_num(nan=fill_value)
+        norm = norm.to(data.dtype)
         return norm
 
-    def unnormalize_tensor_prognostic(self, data: torch.Tensor) -> torch.Tensor:
+    def unnormalize_tensor_prognostic(
+        self, data: torch.Tensor, apply_nan=False
```
jder: How about `mask_value: float = 0` and then people can pass `mask_value = float("nan")` if they want? Otherwise it's a little unclear to me from the name what "apply"ing a NaN is. Or could add docs clarifying?

tests/test_aggregator_metrics.py
```diff
@@ -0,0 +1,52 @@
...
+    assert torch.allclose(result_with_nan, expected_with_nan)
+
+    # Test case 3: Without weights
+    result_no_weights = weighted_mean(tensor)
+    expected_no_weights = tensor.nanmean(dim=(-2, -1))
+    assert torch.allclose(result_no_weights, expected_no_weights)
+
+    # Test case 4: With keepdim=True
+    result_keepdim = weighted_mean(tensor, weights, keepdim=True)
+    assert result_keepdim.shape == (1, 1)
```
jder, line 24: Thanks for the tests! It's a little funny to me when the test code and the "expected" code are basically the same implementation. What about doing something like this where we know the answer for test & the below tests?

```
    # each value * weight should == 1
    tensor = torch.tensor([[1.0, 2.0], [4.0, 8.0]])
    weights = torch.tensor([[1.0, 0.5], [0.25, 0.125]])
    result = weighted_mean(tensor, weights)
    assert torch.allclose(result, 4.0)
```

tests/test_aggregator_metrics.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_area_weighted_mean():
+    data = torch.tensor([[1.0, 2.0], [3.0, 4.0]])
+    area_weights = torch.tensor([[0.1, 0.2], [0.3, 0.4]])
+    result = area_weighted_mean(data, area_weights)
+    expected = weighted_mean(data, area_weights)
+    assert torch.allclose(result, expected)
```
jder: I'm not sure what this is testing. `area_weighted_mean` just calls `weighted_mean` internally.

tests/test_utils_data.py
```diff
@@ -123,11 +126,61 @@ def test_compute_anomalies():
...
+            "var_1": (["lat", "lon"], [[1.0]]),
+            "var_2": (["lat", "lon"], [[2.0]]),
+        },
+        coords={"lat": [0], "lon": [0]},
+    )
+
+    # Create test wet mask
+    wet_mask = torch.tensor([[1.0, 0.0], [0.0, 1.0]])
+    # Initialize Normalize instance
+    with MultitonScope():
```
jder, line 161: This makes sense to me! Might have a comment saying this is creating a non-global Normalize instance so it doesn't affect other tests.

jder (approved):
Just some suggestions for minor cleanups from me, generally looks good!

tests/test_utils_data.py
alxmrs: I'm going to borrow this pattern :)

src/ocean_emulators/aggregator/metrics.py
alxmrs: 🐑  Further on this: instead of a new method, could the Nan support be handled as a flag? The two routines seems really similar.

alxmrs (approved):
LGTM too :)

src/ocean_emulators/aggregator/metrics.py
suryadheeshjith: So this function is what weighted_mean previously was and is actually an incorrect function when we pass in Nans to the data (which we didn't before since the land cells were 0s and not Nan). Maybe we could move this into the test file instead?

tests/test_aggregator_metrics.py
suryadheeshjith: I can remove this

tests/test_utils_data.py
suryadheeshjith: Okay, I was hoping this was how we wanted to test singletons!

src/ocean_emulators/aggregator/metrics.py
suryadheeshjith: Actually, I will just remove this function!

src/ocean_emulators/utils/data.py
suryadheeshjith: Renamed to fill_value and added some documentation

PR #164, opened 2025-04-01
Setting scope where it's needed to omit unnecessary test combinations.

alxmrs:
This ensures that multiton scope is only applies in the datasets and trainer tests, where `TrainConfig` is already in use.


suryadheeshjith (approved):
This LGTM but maybe @jder should have a look as well

alxmrs:
Closing in favor of https://github.com/suryadheeshjith/Ocean_Emulator/pull/166.

PR #165, opened 2025-04-02
Adding Ocean Heat Content Corrector

suryadheeshjith:
- Refactoring current corrector module
- Added tests for Relu corrector
- Added Ocean Heat Corrector module
- Added tests for ocean heat content and ocean heat corrector
- Ocean heat corrector requires static data - sea surface fraction and geothermal heat flux. I am just reading them from separate zarrs and combining it with the data. The static data comes from the CM4 dataset and it is reasonable to use this data for OM4.
- The ocean heat content is not entirely numerically correct as we are taking normalized area weights, but this is fine as we are computing a ratio and we are also using these weights for the surface and sea floor heat fluxes.
- [Big Update] Since we need to correct the fields at every time step and is determined by the surface fluxes, we now pass the boundary conditions at each input time step in contrast to the latest input time step.

NOTE: Constant multiplying like RHO and CP is not redundant as we are adding the heat content quantity with heat flux and they need to be in the same units.

- [x] Add SSF to hfgeou
- [x] Add static data to OM4 data
- [x] Add notebook to save static data to OM4
- [x] ~~Configure option to use last or all boundary conditions~~ This seems to add quite a bit of bloat to the codebase. The assumption here was to be able to continue testing previous models but it seems like it would just be easier to roll back to previous experiment branches and run the code there.
- [x] Detach input from graph
- [x] ~~Configure option to set optimizer~~ In favor of #222
- [x] ~~Register buffers for non trainable params~~
- [x] ~~Add tests to ensure OHC correction does not cause overflows~~ Easier to just view the logs on wandb than have explicit tests for data.
- [x] Wandb logging of OHC correction / OHC
- [x] Jaxtyping in corrector

For additional details of implementation of Ocean Heat Corrector - follow this [doc](https://www.overleaf.com/read/fnqqnqphfyxq#278184).


suryadheeshjith:
I currently don't like the way I am passing the static data to the model, maybe a better way would be to pass it through the Stepper.

src/ocean_emulators/config.py
```diff
@@ -37,6 +37,7 @@ class DataConfig:
     data_means_path: str = "CM4_5daily_v0.4.0_means"
     data_stds_path: str = "CM4_5daily_v0.4.0_stds"
     scaling_residuals_file: Optional[str] = None
+    static_data_paths: Optional[Dict[str, str]] = None
```
jder: Why not list the possible paths here?

src/ocean_emulators/constants.py
```diff
@@ -100,6 +125,10 @@
     "mask_18",
 ]
 
+RHO_0 = 1035.0
+CP_SW = 3992.0
```
jder: Can you give a link to a reference for what these are?

src/ocean_emulators/constants.py
```diff
@@ -100,6 +125,10 @@
...
+N_SECONDS = 5 * 24 * 60 * 60  # 5 day average
```
jder: How about SECONDS_PER_DAY and then multiply it by `time_delta` to avoid a mis-match there? Or else delete `time_delta` so there's no chance of confusion.

src/ocean_emulators/models/corrector.py
```diff
@@ -27,55 +99,233 @@ def __init__(self, corrector_config, hist):
...
+        dHC_expected = (
+            self.area_weighted_func(
+                surface_heat_flux * self.sea_surface_fraction_tensor
             )
-            fts_reshaped = rearrange(fts_reshaped, "n hist c h w -> (n hist) c h w")
+            * N_SECONDS
+        )  # [J]
+
+        # Apply geothermal heat flux
+        dHC_expected += self.area_weighted_func(self.hfgeou_tensor) * N_SECONDS
```
jder: Maybe this is a naïve question, but am I right that we’re considering geothermal heat flux to be independent of ocean state? If so I think we can/should compute this product in `__init__`?

src/ocean_emulators/models/corrector.py
```diff
@@ -1,22 +1,94 @@
...
         super().__init__()
-        self.corrector_config = corrector_config
-        self.non_negative_corrector_names: Optional[List[str]] = (
-            corrector_config.non_negative_corrector_names
+        self.hist = hist
+        self.tensor_map = tensor_map
+        self.normalize = normalize
+        self.num_prognostic_channels = len(self.tensor_map.prognostic_var_names)
+
+    def _rearrange_fts(self, fts: Tensor) -> Tensor:
```
jder: You might consider using jaxtyping for these functions + classes to help document/check the shape manipulation happening here. Or maybe we can come up with names more semantically meaningful than "rearrange"? eg flatten_to_planes or something? 

What does "fts" stand for?

src/ocean_emulators/models/corrector.py
```diff
@@ -27,55 +99,233 @@ def __init__(self, corrector_config, hist):
...
+) -> Tuple[Tensor, Tensor]:
+    """Compute the global heat content of the ocean.
+
+    Args:
+        T: Temperature tensor of shape (batch_size, depth, height, width)
+        dz: Depth tensor of shape (depth,)
+        area_weighted_func: Area weighted function
+
+    Returns:
+        Global heat content tensor of shape (batch_size,)
```
jder: Looks like this comment is out of date? The non-global version looks unused in the corrector, maybe we should not be returning it here?

src/ocean_emulators/models/corrector.py
```diff
@@ -27,55 +99,233 @@ def __init__(self, corrector_config, hist):
...
+        normalize: Normalize,
+        hfgeou_tensor: torch.Tensor,
+        sea_surface_fraction_tensor: torch.Tensor,
+    ):
+        super().__init__(hist, tensor_map, normalize)
+        # Area weights are not on the correct scale.
+        self.area_weights = area_weights
+        self.area_weighted_func = partial(
+            area_weighted_sum, area_weights=self.area_weights
+        )
```
jder, line 178: Might be slightly more clear to write this as a member function? ie `def area_weighted(self, data): return area_weighted_sum(...)`

tests/test_corrector.py
```diff
@@ -0,0 +1,220 @@
...
+    T = T * wet_mask
+    dz = tensor_map.dz
+    area_weights = torch.ones(wet_mask.shape).to(get_device())
+    area_weighted_func = partial(area_weighted_sum, area_weights=area_weights)
+
+    global_HC_t, HC_t = compute_ocean_heat_content(T, dz, area_weighted_func)
+    assert global_HC_t == 86766120
+    assert HC_t[0, 0, 0, 0] == 4131720
+    assert HC_t[0, 1, 0, 0] == 8263440
+    assert HC_t[0, 2, 0, 0] == 16526880
```
jder: Were these numbers derived somehow? Can you leave a comment explaining how they are known to be correct? (And below)

jder (commented):
OK, did a first pass and had some suggestions. I didn't look too closely at the actual corrector but I can do so next. 

Would an alternative to the static_data_paths be to include the new variables in the OM4 zarr? Any reason we can't consider doing that?

suryadheeshjith:
Nothing stopping us from adding the static data to the store itself, again, it's that we have multiple sources of the same data in many place

src/ocean_emulators/constants.py
suryadheeshjith: I will add a comment

src/ocean_emulators/models/corrector.py
suryadheeshjith: Yepp, this can be precomputed. I will do that!

src/ocean_emulators/models/corrector.py
suryadheeshjith: fts = features. I renamed some of the functions to make things more semantically clear

src/ocean_emulators/models/corrector.py
suryadheeshjith: I have used this format throughout the codebase, we can change this in a later PR

suryadheeshjith:
To be honest, we would need this new static data available everywhere so might as well just augment it offline. I can remove the augment function

src/ocean_emulators/config.py
suryadheeshjith: Because if the data is not present, which is not in the data for the tests, we would provide paths that do not exist. I could possibly add a comment?

suryadheeshjith: The other option is to instead pass these paths that may or may not exist and then check if the ocean heat corrector flag is set in the config instead

src/ocean_emulators/models/corrector.py
suryadheeshjith: Do we still want JAX typing here? We are just flattening/unflattening the history dimension

tests/test_corrector.py
suryadheeshjith: Done

src/ocean_emulators/config.py
jder: I don't understand why we'd provide paths which don't exist. What I meant was to replace `static_data_paths: Optional[Dict[str, str]] = None` with `sea_surface_fraction_path: Path | None = None` and `hfgeou_path: Path | None = None`

jder:
> To be honest, we would need this new static data available everywhere so might as well just augment it offline. I can remove the augment function

Were you still planning to do this? I see `augment_static_data` is still here.

src/ocean_emulators/utils/data.py
```diff
@@ -138,18 +139,18 @@ def spherical_area_weights(data: xr.Dataset) -> Grid:
...
 
     Args:
         time_config: Time configuration
-        time_delta: Time delta in days
         hist: Number of rollout steps
 
     Returns:
         num_steps: Number of rollout steps
     """
+    time_delta = 5
```
jder, line 153: Thanks for removing this configuration, but I think this is still the same possible bug. Can we make `TIME_DELTA = 5` in constants and use it both here and in the computation of `SECONDS_PER_5DAY` (maybe `SECONDS_PER_TIMESTEP`?)

suryadheeshjith:
Yepp I will remove it! I will ping you when this needs to be reviewed again

src/ocean_emulators/models/corrector.py
```diff
@@ -27,55 +95,233 @@ def __init__(self, corrector_config, hist):
...
+        fts_input_boundary = fts_input_boundary.detach()
+
+        fts = self._flatten_hist(fts)
+        fts = self._unnormalize_fts_prognostic(fts)
+
+        fts_input, fts_boundary = self._flatten_input(fts_input_boundary)
+        fts_input, fts_boundary = self._unnormalize_fts_input(fts_input, fts_boundary)
+
+        # The input and output mapping of the variables are the same
+        T_input = fts_input[:, self.thetao_idx]  # (batch, depth, lat, lon)
```
jder, line 207: The first index here is not `batch`, is it? We've just flattened it so it's batch * history? I'm actually a little confused how we expect this to interact with history. ie I was expecting we'd somewhere want T_input to be computed from just the most recent input timestep but I don't see that anywhere.

src/ocean_emulators/models/corrector.py
jder: Up to you; I would find it helpful to have types in order to understand the code. I think of it basically like the comments about the shape you have in a few places, but enforced.

jder (commented):
Sorry, I was mid-review when you said you'd ping again later, leaving a couple comments I had already written down.

src/ocean_emulators/models/corrector.py
suryadheeshjith: I need to update the comment. fts_input_boundary is the current input to the model, does that answer your question?

jder: No, I think I'm still confused. Maybe we can chat live for a minute whenever you're around? (Or happy to wait for an updated comment if you like)

src/ocean_emulators/models/corrector.py
```diff
@@ -27,55 +95,233 @@ def __init__(self, corrector_config, hist):
...
+        self.hfds_idx = self.tensor_map.INPT_BOUNDARY_IDX["hfds"]
+
+        self.thetao_idx = self.thetao_idx.to(get_device())
+        self.hfds_idx = self.hfds_idx.to(get_device())
+        self.dz = self.dz.to(get_device())
+        self.hfgeou_tensor = hfgeou_tensor.to(get_device())
+        self.sea_surface_fraction_tensor = sea_surface_fraction_tensor.to(get_device())
+
+        self.dHC_geothermal = (
+            self.area_weighted_func(self.hfgeou_tensor) * SECONDS_PER_5DAY
```
jrayshi: self.hfgeou_tensor should also multiply by the sea_surface_fraction_tensor.
=> self.area_weighted_func(self.hfgeou_tensor * self.sea_surface_fraction_tensor) * SECONDS_PER_5DAY

suryadheeshjith: That seems unintuitive to me, wouldn't we have to use a similar tensor but for the sea floor?

src/ocean_emulators/datasets.py
```diff
@@ -139,9 +139,9 @@ def __getitem__(self, idx: int) -> Example:
...
         boundary = (
             self.boundary.isel(time=window)
-            .isel(time=self.hist)
-            .to_array("var", "boundary")
-            .transpose("step", "var", "lat", "lon")
+            .isel(time=slice(None, time_split))
+            .to_array()
+            .einops.rearrange("step (time variable)=var lat lon", dask="allowed")
             .drop_vars("var", errors="ignore")
         )
```
alxmrs, line 146: If boundary has the same einsum as prognostic, then my preference is to merge these xr.Datasets first and apply the transformations once.

src/ocean_emulators/aggregator/metrics.py
```diff
@@ -3,6 +3,39 @@
...
+        a tensor of the weighted sum averaged over the specified dimensions `dim`.
+    """
+    if weights is None:
+        return tensor.nansum(dim=dim, keepdim=keepdim)
+
+    weights = weights.to(tensor.device)
+    return (tensor * weights).nansum(dim=dim, keepdim=keepdim)
+
+
+def area_weighted_sum(
```
alxmrs, line 30: What is the value of having this function (alias)?

src/ocean_emulators/models/corrector.py
jder: We talked and resolved this: the code is computing the correction across the batch and across the (hist + 1) timesteps independently and applying them to the relevant (hist + 1) timesteps in the output.

jder:
I was thinking more about this and potential precision issues and wondering if it might be worth adding a wandb metric which tracks the difference between the total heat and the target total heat both before and after the corrector. (Or a test which assets those are well-behaved on realistic-sized grids.)

src/ocean_emulators/models/corrector.py
suryadheeshjith: Based on our conversations, I will be adding the sea-surface fraction in the computation of geothermal heatflux. My understanding is that the heat flux emanates from not only the sea floor but also the land structure below the sea-surface.

suryadheeshjith:
That's a good idea!

src/ocean_emulators/datasets.py
suryadheeshjith: They are arranged as [Prog1, Prog2, Bound1, Bound2] though, not [Prog1, Bound1, Prog2, Bound2]. I am okay with figuring out a better way to merge these operations in a later PR

suryadheeshjith:
This PR has been replaced by #224

Issue #105, opened 2025-03-11
Data loader performance optimization


jder:
Creating an umbrella issue for tracking data loader performance. We can create sub/linked issues for particular threads if needed.

alxmrs:
A few more interesting ideas: 
- Shared memory LRU Cache of Zarr data stored in Redis
- Pre-shuffle the data (then we'll have nice linear reads)
- Looking at the data loader, it looks like we're loading almost twice as much data as we need to since the output of the previous step is the input of the next step. 
  - Train data could be slices into a commonly loaded dataset. The final data structure we produce in the batch could produce the time step and ocean state into input/output pairs at train time.

jder:
Here are some experiments @alxmrs ran: https://docs.google.com/document/d/19Mvct9YidGzzPoCqYjT8P9OyiSJtq2xZ4qMn9HnuSIo/edit?tab=t.0

jder:
I ran some experiments today as well, raw code/config is [here](https://github.com/suryadheeshjith/Ocean_Emulator/compare/u/jder/experiments/profile-2025-04-02?expand=1) and some rough analysis [here](https://docs.google.com/spreadsheets/d/1rgkgEzFar4fFcbuSXNCKz1QuoZ9fQp_UIXWSVKNEveE/edit?gid=0#gid=0). 

They were all using the full samudra model size, batch size 2, 4 steps of rollout and history = 0.

Small tests (10 train steps only, 0 workers meaning in-thread data loading)
* baseline (eager loader) average "data" time: 3.21s
* lazy loader, unvectorized: 2.96s
* lazy loader, vectorized: 2.18s

Big tests (1325 train steps, 0 workers)
* baseline: 4.49s
* lazy, unvectorized: 3.60s
* lazy, vectorized: 2.84s

Big tests 4 workers (averages don't include first train step)
* baseline: 0.20s (spikes to 1.2s every ~4 steps)
* lazy, unvectorized: 0s
* lazy, vectorized: 0s

My takeaways:
* It is somewhat surprising that the total number of train steps dramatically impacts how much time *each step* takes to load data. I think this is likely because with only 10 train steps, there is a lot of overlap between data used for adjacent steps (due to rollout length) and so we probably are hitting the file cache. 
[...]

PR #168, opened 2025-04-02
Turn off dask by default

jder:
According to [results here](https://github.com/suryadheeshjith/Ocean_Emulator/issues/105#issuecomment-2773940880), turning off dask is 20x faster for the eager data loader and 10x faster for the lazy data loader. This disables dask by default but you can toggle it back on.


alxmrs (approved):
LGTM, LFGGGG

suryadheeshjith (approved):
Great catch!

jder:
I suggest we wait to merge this until we fix our benchmarks so we can see the effect there, so let's wait until after we investigate those.

alxmrs:
Sounds good to me!

PR #169, opened 2025-04-02
Fix bug in detection of broken test config, fix test_trainer__mini_2step

jder:
Our check to see if the "only_configs" were real was (a) not running and (b) broken. Fixed the check and also this test which wasn't running and was broken in config and also exercised a subtle bug in the new lazy loader.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -300,15 +300,16 @@ def check_skip_configs(request, config_name):
     """
     only_configs = request.node.get_closest_marker("only_configs")
     if only_configs:
+        for desired_config in only_configs.args:
+            if desired_config not in ALL_CONFIGS:
+                raise ValueError(
+                    f"Test config {desired_config} is not"
+                    f" in the list of all configs: {ALL_CONFIGS}"
+                )
         if config_name not in only_configs.args:
             pytest.skip(
                 f"Skipping test for {config_name} because"
                 f" it is not in {only_configs.args}"
-            )
-        if config_name not in ALL_CONFIGS:
-            raise ValueError(
-                f"Test config {config_name} is not"
-                f" in the list of all configs: {ALL_CONFIGS}"
             )
     else:
         all_configs = request.node.get_closest_marker("all_configs")
--- a/tests/test_trainer.py
+++ b/tests/test_trainer.py
@@ -15,7 +15,7 @@ def test_trainer__mini_benchmark(trainer_pair: TrainPair, caplog, benchmark):
         trainer.run()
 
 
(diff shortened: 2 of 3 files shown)


tests/test_trainer.py
```diff
@@ -15,7 +15,7 @@ def run():
         trainer.run()
 
 
-@pytest.mark.only_configs("train_cm4_2step.test.yaml")
+@pytest.mark.only_configs("train_default_2step.test.yaml")
```
alxmrs, line 18: Ah, good catch. This may have been my mad during a rename.

tests/conftest.py
```diff
@@ -300,16 +300,17 @@ def check_skip_configs(request, config_name):
     """
     only_configs = request.node.get_closest_marker("only_configs")
     if only_configs:
+        for desired_config in only_configs.args:
+            if desired_config not in ALL_CONFIGS:
+                raise ValueError(
+                    f"Test config {desired_config} is not"
+                    f" in the list of all configs: {ALL_CONFIGS}"
```
alxmrs, line 307: Good idea to raise an error -- it's nice to prevent this kind of string matching error.

alxmrs (approved):
LGTM

suryadheeshjith (approved):
LGTM

jder:
@alxmrs @suryadheeshjith I have added the fix for the mutation/gradient bug here since it was preventing merging. Will merge when green unless you folks want to take a look.

suryadheeshjith:
What issue was this causing?

jder:
@suryadheeshjith it was causing this exception:

```
E       RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.FloatTensor [1, 24, 180, 360]], which is output 0 of AsStridedBackward0, is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).

.venv/lib/python3.11/site-packages/torch/autograd/__init__.py:266: RuntimeError
```

The problem is that we were mutating the input variable, and in the lazy loader the inputs share data. I suspect this is a feature of the lazy loader which is why we're seeing reduced memory usage there but it could also be some kind of bug.

suryadheeshjith:
Ok, good catch. I will check if this affects the previous dataloader

suryadheeshjith:
Okay, looks like the previous/eager dataloader does not have an issue since we are only updating the input values in the TrainData object and when we re-read similar data, the data is once again being read using xarray, phew! Good fix regardless! Thanks

PR #170, opened 2025-04-03
Torch arrays should be floats, not doubles.

alxmrs:
This should fix the broken benchmark run.


jder (approved):
This is just in tests that this is failing? Should we be casting the data somewhere test-specific? But looks innocent enough so fine with this too.

alxmrs:
No, I think this might also matter for performance, too. Now, all the training examples take half the bytes!

alxmrs:
This also matches the old loader behavior. I think I should add a test to check for this.

PR #171, opened 2025-04-03
Adding EMA weights

### LINKED ISSUE #81: EMA of weights
* Need to push EMA weights functionality to main
* Solves issue #65

suryadheeshjith:
- Adding EMA of weights
- Saving EMA weights for validation and inference checkpoints 
Closes #81

### DIFF
--- a/configs/train_om4.yaml
+++ b/configs/train_om4.yaml
@@ -8,9 +8,6 @@ batch_size: 4
 learning_rate: 0.0002
 scheduler: true
 loss: mse
-validate_using_ema: true
-ema_decay: 0.999
-faster_decay_at_start: true
 finetune: false
 resume_ckpt_path: null
 inference_epochs: [-1]
--- a/configs/train_om4.yaml
+++ b/configs/train_om4.yaml
@@ -9,6 +9,8 @@ learning_rate: 0.0002
 scheduler: true
 loss: mse
 validate_using_ema: true
+ema_decay: 0.999
+faster_decay_at_start: true
 finetune: false
 resume_ckpt_path: null
 inference_epochs: [-1]
--- a/configs/train_om4.yaml
+++ b/configs/train_om4.yaml
@@ -8,6 +8,7 @@ batch_size: 4
 learning_rate: 0.0002
 scheduler: true
 loss: mse
+validate_using_ema: true
 finetune: false
 resume_ckpt_path: null
 inference_epochs: [-1]
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -130,6 +130,7 @@ class TrainConfig:
     finetune: bool = False
     resume_ckpt_path: Optional[str] = None
     debug: bool = False
+    validate_using_ema: bool = True
     backend: TrainBackendConfig = "auto"
 
     # Data parameters at root level
@@ -183,6 +184,7 @@ class TrainConfig:
             "loss": self.loss,
             "finetune": self.finetune,
             "resume_ckpt_path": self.resume_ckpt_path,
+            "validate_using_ema": self.validate_using_ema,
             "backend": self.backend,
             "data_percent": self.data_percent,
             "data_stride": self.data_stride,
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -131,6 +131,8 @@ class TrainConfig:
     resume_ckpt_path: Optional[str] = None
     debug: bool = False
     validate_using_ema: bool = True
+    ema_decay: float = 0.999
+    faster_decay_at_start: bool = True
     backend: TrainBackendConfig = "auto"
 
     # Data parameters at root level
@@ -185,6 +187,8 @@ class TrainConfig:
             "finetune": self.finetune,
             "resume_ckpt_path": self.resume_ckpt_path,
             "validate_using_ema": self.validate_using_ema,
+            "ema_decay": self.ema_decay,
+            "faster_decay_at_start": self.faster_decay_at_start,
             "backend": self.backend,
             "data_percent": self.data_percent,
             "data_stride": self.data_stride,
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -130,7 +130,7 @@ class TrainConfig:
     finetune: bool = False
     resume_ckpt_path: Optional[str] = None
     debug: bool = False
-    validate_using_ema: bool = True
+    test_using_ema: bool = True
     ema_decay: float = 0.999
     faster_decay_at_start: bool = True
     backend: TrainBackendConfig = "auto"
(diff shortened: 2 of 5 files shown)


configs/train_om4.yaml
```diff
@@ -8,6 +8,9 @@ batch_size: 4
 learning_rate: 0.0002
 scheduler: true
 loss: mse
+validate_using_ema: true
+ema_decay: 0.999
+faster_decay_at_start: true
```
jder: I would suggest omitting values in here which are the defaults.

src/ocean_emulators/train.py
```diff
@@ -729,31 +739,41 @@ def init_data_loaders(self, cur_step: int) -> None:
...
-                f"{self.ckpt_paths.best_inference_checkpoint_path}"
-            )
-            self.best_inf_loss = inf_loss
-            self.save_checkpoint(epoch, self.ckpt_paths.best_inference_checkpoint_path)
-        if is_best_val_loss:
-            self.save_checkpoint(epoch, self.ckpt_paths.best_validation_checkpoint_path)
+        if self.validate_using_ema:
+            best_checkpoint_context = self._ema_context
+        else:
+            best_checkpoint_context = contextlib.nullcontext  # type: ignore
```
alxmrs: TIL! This is neat.

src/ocean_emulators/config.py
```diff
@@ -130,6 +130,9 @@ class TrainConfig:
     finetune: bool = False
     resume_ckpt_path: Optional[str] = None
     debug: bool = False
+    validate_using_ema: bool = True
```
jder: It's validate *and* inference, right? How about "test_using_ema" or "eval_using_ema"?

src/ocean_emulators/train.py
```diff
@@ -313,8 +315,16 @@ def __init__(self, cfg: TrainConfig) -> None:
                 device_ids=[self.distributed.gpu],
             )
 
+        # EMA
+        self._ema = EMATracker(
+            self.model,
+            decay=cfg.ema_decay,
+            faster_decay_at_start=cfg.faster_decay_at_start,
+        )
+
```
jder, line 324: Maybe I'm just missing something obvious, but I don't ever see us calling the `__call__` to update the ema weights?

src/ocean_emulators/train.py
```diff
@@ -529,7 +539,6 @@ def train_one_epoch(self, epoch):
 
         return train_aggregator.get_logs()
 
-    @torch.no_grad()
```
jder, line 532: Why did this move?

src/ocean_emulators/train.py
```diff
@@ -729,31 +739,41 @@ def init_data_loaders(self, cur_step: int) -> None:
...
-            self.best_inf_loss = inf_loss
-            self.save_checkpoint(epoch, self.ckpt_paths.best_inference_checkpoint_path)
-        if is_best_val_loss:
-            self.save_checkpoint(epoch, self.ckpt_paths.best_validation_checkpoint_path)
+        if self.validate_using_ema:
+            best_checkpoint_context = self._ema_context
+        else:
+            best_checkpoint_context = contextlib.nullcontext  # type: ignore
+
+        with best_checkpoint_context():
```
jder: How about `with self._validation_context()`? I think that's equivalent to the best_checkpoint_context we're building here?

src/ocean_emulators/train.py
```diff
@@ -794,10 +821,38 @@ def load_checkpoint(self, checkpoint_path, finetune=False):
...
+        """
+        A context where the stepper uses the EMA model.
+        """
+        self._ema.store(parameters=self.model.parameters())
+        self._ema.copy_to(model=self.model)
+        try:
+            yield
+        finally:
+            self._ema.restore(parameters=self.model.parameters())
+
```
jder, line 852: It seems easy to forget to use `_validation_context` or `_ema_context` when accessing the model. What do you think about renaming `self.model` to `self._model` and then always accessing it via a context manager `@contextmanager def model(for_eval: bool) -> Samudra | DistributedDataParallel`? Calling code would then look like: `with self.model(for_eval=True) as model: …`. Could also call `model.train(True)` or `model.eval()` inside that.

jder (changes requested):
A bunch of optional ideas from me, except for the bit about _ema never being `__call__`ed which I think we should try to resolve before merging.

src/ocean_emulators/train.py
```diff
@@ -794,10 +821,38 @@ def load_checkpoint(self, checkpoint_path, finetune=False):
...
+    def _ema_context(self):
+        """
+        A context where the stepper uses the EMA model.
+        """
+        self._ema.store(parameters=self.model.parameters())
+        self._ema.copy_to(model=self.model)
+        try:
+            yield
+        finally:
+            self._ema.restore(parameters=self.model.parameters())
```
alxmrs, line 851: Now that I see how this is called: I think you may want to add this context manager logic directly to the ema class itself -- e.g. `copy_to` and `restore` seem like enter and exit methods. 

Of course, I could be reading this wrong.

src/ocean_emulators/utils/ema.py
```diff
@@ -0,0 +1,189 @@
...
+            faster_decay_at_start: Whether to use the number of updates to determine
+                the decay rate. If True, the decay rate will be min(decay, (1 +
+                num_updates) / (10 + num_updates)). If False, the decay rate
+                will be decay.
+        """
+        super().__init__()
+        if decay < 0.0 or decay > 1.0:
+            raise ValueError("Decay must be between 0 and 1")
+
+        self._module_name_to_ema_name = {}
```
alxmrs, line 67: I'm having a hard time parsing it, but something is rubs me the wrong way if we need to consume model in both the init and in major methods. Could model be consumed just in the init or just in the primary methods?

src/ocean_emulators/train.py
suryadheeshjith: I missed the call, I am not sure how I did not push that. I am adding some logging with wandb as well

src/ocean_emulators/train.py
suryadheeshjith: Just being more specific!

src/ocean_emulators/train.py
suryadheeshjith: Good point!

src/ocean_emulators/train.py
suryadheeshjith: It seems like the code would get a lot more verbose if we implement it like this. I would like to avoid having to use a context manager everytime I access the model

src/ocean_emulators/utils/ema.py
suryadheeshjith: We need it at init to set up ema model weights and then the new weights each time to update the ema weights in an ema fashion.

suryadheeshjith: i.e. EMA_params = decay * EMA_params + (1 - decay) * model_params where model_params is the new params and EMA_params is the initially saved and continuously updating params

src/ocean_emulators/train.py
suryadheeshjith: I thought of that but it seemed more elegant to put both context managers _validation_context and _ema_context together.

src/ocean_emulators/utils/ema.py
alxmrs: Thanks, that makes sense to me. I appreciate the explanation. :)

jder (approved):
LGTM, thanks!

PR #172, opened 2025-04-03
Datasets test for TrainData + Test fix

suryadheeshjith:
- Just a test for traindata ensuring underlying data is not manipulated by TrainData. Added a test just in case (even with #169  )
- Removed lat/lon coords in tests data statistics

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -415,7 +415,7 @@ def traindataset_input():
     wet = torch.ones(2, 2, 2)
     wet_surface = torch.ones(2, 2)
 
-    # Initialize Normalize instance
+    # Initialize and yield within the MultitonScope
     with MultitonScope():
         _ = Normalize.init_instance(
             data_mean=data_mean,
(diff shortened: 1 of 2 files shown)


suryadheeshjith:
It is very possible I could use a config, maybe I just wasted my time creating a fixture :)

suryadheeshjith:
Open to removing the test if it seems redundant

jder (approved):
Seems OK to me, though if you want to create a new fixture here I suggest yielding within the multiton scope so that the test will have the right Normalize instance present while the test runs. (Or use one of the existing fixtures)

suryadheeshjith:
Since I am creating the traindataset object within the same multiton, wouldn't I be using the correct normalize instance?

jder:
Yes, but for example if your test code calls collate_train_data and that uses Normalize.get_instance(), it will not get the right one.

suryadheeshjith:
Sounds good!

suryadheeshjith:
I hope this is what you meant @jder

jder (approved):
Yep, exactly.

jder:
Oh, I wanted to ask: why did you remove the lat/lon thing?

suryadheeshjith:
We should have a single mean / variance value for every channel, it made no sense to have lat / lon coordinates for it. I also had a bug in this test when I tried keeping it.

PR #176, opened 2025-04-04
Disabling concurrent pr runs

suryadheeshjith:
- Prior commits with action runs in the same PR are cancelled in favor of new ones.
- Ensure we still enable concurrency when merged to main.


suryadheeshjith:
Successfully cancelled!

jder (approved):
Looks good to me! Maybe worth adding a comment that the goal is to cancel anything running already for the same pull request?

suryadheeshjith:
Would be nice to test if concurrent action runs still work when we have two PRs merged to main, I will merge this once some other PR is ready to be merged

PR #177, opened 2025-04-04
Automatically omit extra benchmarks.

alxmrs:
Adding a hook into all generated test combinations. For benchmarks, I add skip marks to all tests that don't fit the happy path test conditions. 

I think this infrastructure could be useful to omit testing other unwanted fixture combinations.


jder (approved):
Thanks! @suryadheeshjith Alex and I discussed and this is a short-term fix to reduce benchmarking, we're hoping to refactor test config a little and so not need it long term.

jder:
I'm assuming we're not going to merge this in favor of #179

PR #178, opened 2025-04-04
Torch-first data loader without dask

jder:
I did some profiling and continue to see a huge amount of time spent in xarray indexing. I built a dataloader which goes as quickly as possible into (CPU) torch and does not use dask. On my laptop this is 3x faster than the eager data loader, and on my test GPU machine it's about 5x faster in the data loader benchmarks. Also seems to eliminate any waiting for data on my GPU machine with 4 workers when running with batch size == 2. 

I manually verified this produces the same data as the eager loader in the debugger -- hoping to add a test for this once @alxmrs is done refactoring the test configuration story a little. 

- [x] add tests for matching old behavior, shape, etc before merging
- [x] should we be masking with NaN?

### DIFF
--- a/configs/train_om4.yaml
+++ b/configs/train_om4.yaml
@@ -1,5 +1,5 @@
 
-debug: true
+debug: false
 disk_mode: true
 pin_mem: true
 save_freq: 5
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -41,7 +41,6 @@ class DataConfig:
     num_workers: int = 4
     hist: int = 1
     loader_version: str = str(LoaderVersion.OM4_EAGER.value)
-    use_dask: bool = True
 
 
 @dataclass
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -594,3 +594,172 @@ class TrainDataset(Dataset):
         label = torch.from_numpy(label).float()
         label = torch.where(self.wet, label, 0.0)
         return label
+
+
+class TorchTrainDataset(Dataset):
+    """
+    This class is used for training and validation.
+
+    It creates rolling indices to keep track of histories/past states. But different
+    from InferenceDataset, as it creates rolling indices based on stride. By default,
+    the sliding window / stride is 1.
+
+    We make use of TrainData class to store a single sample.
+
+    For example,
+    Hist=0 ; TD: step=0->[0, 1]; step=1->[1, 2]; step=2->[2, 3]; step=3->[3, 4]
+    Hist=1 ; TD: step=0->[[0, 1], [2, 3]]; step=1->[[2, 3], [4, 5]];
+            step=2->[[4, 5], [6, 7]]; step=3->[[6, 7], [8, 9]]
+    Hist=2 ; TD: step=0->[[0, 1, 2], [3, 4, 5]];
+            step=1->[[3, 4, 5], [6, 7, 8]];
+            step=2->[[6, 7, 8], [9, 10, 11]];
+            step=3->[[9, 10, 11], [12, 13, 14]]
+    """
+
+    FLAG = LoaderVersion.OM4_TORCH
+
+    def __init__(
+        self,
+        data: xr.Dataset,
+        prognostic_var_names: PrognosticVarNames,
+        boundary_var_names: BoundaryVarNames,
+        wet: PrognosticMask,
+        wet_surface: GridMask,
+        hist: int,
+        steps: int,
+        stride: int = 1,
+    ):
+        super().__init__()
+        self.device = get_device()
+
+        self.hist: int = hist
+        self.steps: int = steps
+        self.stride: int = stride
+
+        self.num_prognostic_channels: int = (hist + 1) * len(prognostic_var_names)
+        self._prognostic_vars: xr.Dataset = data[prognostic_var_names]
+        self._boundary_vars: xr.Dataset = data[boundary_var_names]
+
+        # This class will be used only for training
+        total_steps: int = 2 * self.hist + 2
+
+        # Calculate the number of windows
+        num_windows = data.time.size - (total_steps - 1) * self.stride
+
+        # Create base indices
+        indices = np.arange(num_windows)
+        indices_da = xr.DataArray(indices, dims=["window_dim"])
+
+        # Create window dimension
+        window_dim = xr.DataArray(np.arange(total_steps), dims=["time"])
+
+        # Construct rolling indices
+        self.rolling_indices: Float[xr.DataArray, "window_dim time"] = (
+            indices_da + stride * window_dim
+        )
+
+        self.wet = wet.bool()
+        self.wet_surface = wet_surface.bool()
+
+        self.size: int = (
+            data.time.size
+            - self.steps * (self.hist + 1) * self.stride
+            - self.hist * self.stride
+        )
+
+        if using_gpu():
+            self.wet = self.wet.pin_memory()
+            self.wet_surface = self.wet_surface.pin_memory()
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __getitem__(self, idx: int):
+        TD = TrainData(self.num_prognostic_channels)
+        prev_rolling_idx = None
+        for step in range(self.steps):
+            x_index = self._get_x_index(idx, step, prev_rolling_idx)
+
+            prognostic_all = torch.from_numpy(
+                self._prognostic_vars.isel(time=x_index)
+                .to_array()
+                .transpose(
+                    "variable", "time", "lat", "lon"
+                )  # this should be a no-op, for documentation
+                .to_numpy()
+            )
+            boundary = torch.from_numpy(
+                self._boundary_vars.isel(time=x_index)
+                .isel(time=self.hist, drop=True)
+                .to_array()
+                .transpose("variable", "lat", "lon")
+                .to_numpy()
+            )
+
+            input_, label = self._get_input_and_label(prognostic_all, boundary)
+
+            TD.insert(
+                input_=input_,
+                label=label,
+            )
+
+        return TD
+
+    def _get_input_and_label(
+        self,
+        # time includes (self.hist + 1) past steps and the (label) future steps
+        prognostic_all: Float[torch.Tensor, "variable time lat lon"],
+        boundary: Float[torch.Tensor, "variable lat lon"],
+    ) -> tuple[Input, Prognostic]:
+        normalize = Normalize.get_instance()
+
+        # grab past steps and prep for model
+        input_ = self._prep_prognostic_steps(prognostic_all[:, : self.hist + 1, :, :])
+
+        # add in boundary to final input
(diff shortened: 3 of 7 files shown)


configs/train_om4.yaml
```diff
@@ -1,5 +1,5 @@
 
-debug: true
```
jder, line 2: I use this config as "most like real life" so felt like we should remove this.

src/ocean_emulators/config.py
```diff
@@ -41,7 +41,6 @@ class DataConfig:
     num_workers: int = 4
     hist: int = 1
     loader_version: str = str(LoaderVersion.OM4_EAGER.value)
-    use_dask: bool = True
```
jder, line 44: We now infer this based on which loader you use.

tests/test_datasets.py
```diff
@@ -455,8 +455,9 @@ def test_profile__loader__1gb(td_loader_pair: LoaderPair, benchmark):
 
     @benchmark
     def bench():
-        for sample in loader:
-            _ = sample
+        indices = np.random.randint(0, len(loader), size=len(loader))
+        for idx in indices:
+            _ = loader.dataset[int(idx)]
```
jder: Felt like random indexes was more representative

alxmrs:
This is excellent. I really am happy to get this landed. Now that I've written an Xarray-based loader, I have a better understanding of what's going on. I think two additional improvements we could adapt from the lazy loader are: 
- Making the data up front once (not on the `__getitem__` loop)
- Vectorizing steps

jder:
Agreed about the vectorizing, it should not be hard. I'm not sure we can do the masking up front, though, without loading all data into ram.

src/ocean_emulators/datasets.py
```diff
@@ -594,3 +594,172 @@ def _get_label(self, x_index) -> Prognostic:
...
+
+        self.hist: int = hist
+        self.steps: int = steps
+        self.stride: int = stride
+
+        self.num_prognostic_channels: int = (hist + 1) * len(prognostic_var_names)
+        self._prognostic_vars: xr.Dataset = data[prognostic_var_names]
+        self._boundary_vars: xr.Dataset = data[boundary_var_names]
+
+        # This class will be used only for training
```
alxmrs:
```suggestion
        # This class will be used only for training and validation
```

src/ocean_emulators/datasets.py
```diff
@@ -594,3 +594,172 @@ def _get_label(self, x_index) -> Prognostic:
...
+        return total_input, label
+
+    def _prep_prognostic_steps(
+        self, prognostic_steps: Float[torch.Tensor, "variable time lat lon"]
+    ) -> Input:
+        normalize = Normalize.get_instance()
+        prognostic_steps = rearrange(
+            prognostic_steps,
+            "variable time lat lon -> time variable lat lon",
+        )
```
alxmrs, line 734: 🐑  Is a named `transpose` available? it may be slightly simpler.

src/ocean_emulators/datasets.py
```diff
@@ -594,3 +594,172 @@ def _get_label(self, x_index) -> Prognostic:
...
+        ).float()
+
+        # flatten time and variable dimensions into a set of channels for model
+        prognostic_steps = rearrange(
+            prognostic_steps,
+            "time variable lat lon -> (time variable) lat lon",
+        )
+
+        # post-normalize, mask out values where there is no ocean
+        prognostic_steps = torch.where(self.wet, prognostic_steps, 0.0)
```
alxmrs, line 748: Future PR: we should consider moving this to the `__init__`.

tests/test_datasets.py
alxmrs: SGTM

src/ocean_emulators/datasets.py
```diff
@@ -594,3 +594,172 @@ def _get_label(self, x_index) -> Prognostic:
...
+
+        return prognostic_steps
+
+    def _get_x_index(
+        self, idx: int, step: int, prev_rolling_idx: int | None
```
alxmrs: 🐑  I know you're just copying over the previous version, but maybe we can get rid of `prev_rolling_idx`?

src/ocean_emulators/datasets.py
alxmrs: Is this needed for normalize?

src/ocean_emulators/datasets.py
```diff
@@ -594,3 +594,172 @@ def _get_label(self, x_index) -> Prognostic:
...
+            input_, label = self._get_input_and_label(prognostic_all, boundary)
+
+            TD.insert(
+                input_=input_,
+                label=label,
+            )
+
+        return TD
+
+    def _get_input_and_label(
```
alxmrs, line 707: I like how you combine all prognostic access into a single function.

src/ocean_emulators/datasets.py
alxmrs: Could this be done in the caller function?

src/ocean_emulators/datasets.py
```diff
@@ -594,3 +594,172 @@ def _get_label(self, x_index) -> Prognostic:
...
+        self,
+        # time includes (self.hist + 1) past steps and the (label) future steps
+        prognostic_all: Float[torch.Tensor, "variable time lat lon"],
+        boundary: Float[torch.Tensor, "variable lat lon"],
+    ) -> tuple[Input, Prognostic]:
+        normalize = Normalize.get_instance()
```
suryadheeshjith, line 713: Does this have to be here? Are you getting errors when you get the instance in init?

src/ocean_emulators/datasets.py
suryadheeshjith: I think the entire loader here is created so dask is not being used and this would mean normalize would have to occur during loading since we cannot normalize the entire dataset at once

alxmrs: Same with the normalization

suryadheeshjith (approved):
LGTM overall

src/ocean_emulators/train.py
```diff
@@ -157,7 +158,8 @@ def __init__(self, cfg: TrainConfig) -> None:
         self.data_stds_path = cfg.data.data_stds_path
         self.scaling_residuals_file = cfg.data.scaling_residuals_file
 
-        if cfg.data.use_dask:
+        self.use_dask = cfg.data.loader_version != LoaderVersion.OM4_TORCH.value
```
alxmrs, line 163: 🐑  maybe this could be made overridable by the config

src/ocean_emulators/datasets.py
suryadheeshjith: Oh I thought you were referencing the normalization code

src/ocean_emulators/utils/data.py
```diff
@@ -391,3 +391,22 @@ def unnormalize_tensor_prognostic(
...
+        if data.ndim == 3:
+            tensor_mean = tensor_mean.reshape([-1, 1, 1])
+            tensor_std = tensor_std.reshape([-1, 1, 1])
+        else:
+            raise ValueError(f"Invalid data shape: {data.shape}")
+
+        norm = (data - tensor_mean) / tensor_std
+        if fill_nan:
+            norm = norm.nan_to_num(nan=fill_value)
+        norm = norm.to(data.dtype)
```
alxmrs, line 411: may not be necessary since we cast to `float` right after in the calling function.

src/ocean_emulators/datasets.py
alxmrs: Sorry Surya and Jesse, I missed both your comments. Ah, I now understand that this saves us from loading all the data into memory. 

In that case, I wonder if we could push it to the data engineering. Short of that, I think this is fine.

alxmrs (approved):
Approved, will review again after I merge my testing PR.

src/ocean_emulators/datasets.py
jder: I removed the fetch of the instance in init; generally I try to avoid having state when I don't need it but it would work either way.

src/ocean_emulators/datasets.py
jder: I'm not sure what you mean by "a named transpose" -- this is a torch vector at this point so it doesn't know what axes are what. And yes, it's required for the next step as the comment says. 

It could be done by the caller but then it would be duplicated for input + label. Why would you prefer the caller?

src/ocean_emulators/datasets.py
jder: Yeah, agreed that doing this in the data someday would be nice.

src/ocean_emulators/utils/data.py
jder: This matches all the other normalize functions in this file so would rather leave it. (The caller's float will be a no-op if so, right?)

src/ocean_emulators/train.py
jder: I think using dask with the torch loader will be slower and not using dask with the other loaders will load all data into memory, so I'm not sure why we would want to configure it separately.

src/ocean_emulators/datasets.py
suryadheeshjith: Alright

jder:
@alxmrs can I get you to take a quick look at the [new tests](https://github.com/suryadheeshjith/Ocean_Emulator/pull/178/commits/22514eeaea0e0e51f96d015c681d33422f9af7ce) before I merge?

src/ocean_emulators/datasets.py
alxmrs: Happy to have it here, thank for pointing out things that I missed.

tests/test_datasets.py
```diff
@@ -347,10 +371,17 @@ def test_inference__data_is_not_zero(inference_loader_pair):
             )
 
 
-def test_om4__is_equal_to_v1_data_loader(train_config):
+ORIGINAL_LOADER_VERSION = LoaderVersion.OM4_EAGER
```
alxmrs, line 374: This LGTM, but can we move it to the top of the file?

tests/test_datasets.py
```diff
@@ -347,10 +371,17 @@ def test_inference__data_is_not_zero(inference_loader_pair):
...
+
+
+@pytest.mark.parametrize(
+    "loader_version", [v for v in LoaderVersion if v != ORIGINAL_LOADER_VERSION]
+)
```
alxmrs, line 379: Nice, good idea.

alxmrs (approved):
LGTM

tests/test_datasets.py
jder: Why? It's only meaningful in the context of this test. If you don't want it floating here, maybe I make a class for it + the test to live in?

alxmrs: I don't have a good reason except to say that I'm used to seeing constants at the top of a file. It's more of a 🐑 ish suggestion; I'm happy to leave it here.

PR #179, opened 2025-04-04
Simplify pytest fixtures to reduce excessive combinations.

alxmrs:
There are now only 56 tests (excluding cuda); there are some combinatorial things, but they are used way more sparingly. I also removed the `only_configs`/`all_configs` marking logic in favor of something more standard to pytest (`indirect` parametrized tests). Further, I got rid of the manual weights tests because they went unused for so long.

Last but not least: I'm running all the non-benchmark tests on 2 workers with xdist.


tests/conftest.py
```diff
@@ -288,7 +288,7 @@ def loader_version(request: pytest.FixtureRequest) -> str:
     return request.param
 
 
-@pytest.fixture(scope="session", params=[0, 1, 2], ids=lambda x: f"hist{x}")
+@pytest.fixture(scope="session", params=[0, 1], ids=lambda x: f"hist{x}")
```
alxmrs: @suryadheeshjith: Is there ever a scenario where `hist=2`?

tests/test_datasets.py
```diff
@@ -313,64 +353,43 @@ def test_inference__data_is_not_zero(inference_loader_pair: LoaderPair):
...
-def test_om4__is_equal_to_v1_data_loader(train_loader_pair: LoaderPair):
-    cfg, loader = train_loader_pair
-
-    data_path = os.path.join(cfg.experiment.data_dir, cfg.data.data_path)
-    ds = xr.open_dataset(data_path, chunks={})
+def test_om4__is_equal_to_v1_data_loader(train_config):
+    with (
+        make_loader(train_config) as loader,
+        make_loader(train_config, version=LoaderVersion.OM4_LAZY) as om4_loader,
+    ):
```
alxmrs, line 354: This should make it easier to test new loaders in the future.

tests/conftest.py
suryadheeshjith: Testing hist=1 should hopefully be equivalent to testing hist=2

suryadheeshjith: But yeah, my training runs with hist=2 hasn't seen gains over hist=1

tests/test_datasets.py
```diff
@@ -12,50 +14,113 @@
...
+@contextlib.contextmanager
+def make_loader(
+    cfg,
+    time_slice: slice | None = None,
+    drop_last: bool = True,
+    version: LoaderVersion = LoaderVersion.OM4_EAGER,
+) -> Generator[DataLoader, None, None]:
+    if time_slice is None:
+        time_slice = cfg.train.time_slice
+    if isinstance(version, str):
```
jder: version is typed as LoaderVersion so hopefully this is never true? Can we make the `loader_version` fixture return a LoaderVersion?

tests/test_datasets.py
```diff
@@ -193,80 +258,46 @@ def test_test_util__data_source_roundtrip(
...
+def test_loader__data_shape(train_config, history, loader_version):
+    train_config.data.hist = history
 
-    num_samples = len(loader)
-    for i, sample in enumerate(loader):
-        X, y = extract_sample_arrays(sample)
+    with make_loader(train_config, version=loader_version) as loader:
+        exp = train_config.experiment
+        batch_size = train_config.batch_size
+        hist = history + 1
```
jder: how about `num_input_timesteps = history + 1`?

tests/test_datasets.py
```diff
@@ -12,50 +14,113 @@
...
+                collate_fn = collate_om4
+            case _:
+                raise ValueError(f"Unknown loader version: {version}")
+
+        loader = DataLoader(
+            data,
+            batch_size=cfg.batch_size,
+            drop_last=drop_last,
+            collate_fn=collate_fn,
+        )
```
jder, line 114: This is a lot of duplication with Trainer and makes me wonder how often we're going to miss bugs when we change that code but not this code. WDYT?

tests/test_datasets.py
```diff
@@ -193,80 +258,46 @@ def test_test_util__data_source_roundtrip(
...
+                180,
+                360,
+            )
+            assert y.shape == (
+                train_config.steps[0],
+                batch_size,
+                output_var_dim,
+                180,
+                360,
+            )
```
jder, line 294: Am I correct in reading that we're no longer testing the val loader?

tests/test_datasets.py
```diff
@@ -278,25 +309,31 @@ def test_inference__data_shape(inference_loader_pair: LoaderPair):
     )
     output_var_dim = len(PROGNOSTIC_VARS[exp.prognostic_vars_key]) * hist
 
-    for sample in loader:
+    samples = list(loader)
+    assert len(samples) == 1, (
+        f"Current config {cfg} only supports 1 examples for inference; "
```
jder: I don't understand this error message -- what does cfg have to do with this limitation?

tests/test_datasets.py
```diff
@@ -450,17 +466,25 @@ def test_train_dataset(traindataset_input):
...
+            for sample in loader:
+                _ = sample
 
 
 @pytest.mark.manual
-def test_profile__inference_loader__1gb(inference_loader_pair: LoaderPair, benchmark):
+@pytest.mark.parametrize(
+    "data_source,config_name", [("mock", DEFAULT_CONFIG)], indirect=True
+)
+def test_profile__inference_loader__1gb(inference_loader_pair, benchmark):
```
jder, line 484: Am I reading this correctly that we're only benchmarking the mock data now? If we're going to pick one, why not pick the real data? (Same for the two tests in test_trainer)

tests/test_datasets.py
```diff
@@ -193,80 +258,46 @@ def test_test_util__data_source_roundtrip(
...
+            f"got {len(samples)}."
+        )
 
-def test_inference__loads_correct_number_of_samples(inference_loader_pair: LoaderPair):
-    cfg, loader = inference_loader_pair
-    n_samples = 1
-    assert len(list(loader)) == n_samples, (
-        f"Current config {cfg} only supports {n_samples} examples; got {len(loader)}."
-    )
+        for sample in samples:
```
jder: WDYT about doing this for eg 1 or 2 samples instead of all of them? Not sure it buys us much to do it exhaustively and it takes a long time.

tests/test_datasets.py
```diff
@@ -12,50 +14,113 @@
...
-@pytest.fixture
-def inference_loader_pair(trainer_pair: TrainPair) -> LoaderPair:
-    cfg, trainer = trainer_pair
-    return cfg, trainer.inference_loader
+    try:
+        TensorMap.get_instance()
+    except ValueError:
+        TensorMap.init_instance(
+            cfg.experiment.prognostic_vars_key, cfg.experiment.boundary_vars_key
+        )
```
jder, line 68: Could move this down into the MultitonScope and unconditionally init rather than this (possibly brittle) check?

jder (changes requested):
This is definitely much cleaner, thank you! I wish the default behavior was still to run tests once but we can just be careful when adding them.

A few suggestions/questions I'd like to resolve, but main thing I'd like to understand is the plan for `make_loader`. If you really want to merge this as-is to reduce cost I think that's ok, but let's not forget to circle back and de-duplicate with Trainer.

tests/test_datasets.py
alxmrs: Good shout.

tests/test_datasets.py
alxmrs: Yes, thank you. I think I went blind to that block of code.

tests/test_datasets.py
alxmrs: I totally agree. Per our morning meeting today: let's move on the plan to merge this and refactor the code after to have a function that makes data loaders.

tests/test_datasets.py
alxmrs: I like that idea.

tests/test_datasets.py
alxmrs: That's correct. I made this choice because it uses the same underlying data loader, the difference is just how it's configured.

tests/test_datasets.py
alxmrs: The number of samples is determined by the time slice we use for inference. Let me make that clearer.

tests/test_datasets.py
alxmrs: That's correct. And, I'm choosing the mock data for consistency with the original data loader tests. The mock data is useful bc it demonstrates how effective loading 1GB of data is. If we want to test the new data, I think we should make a new test (and call it `_om4` or similar). WDYT? Should I add that here?

jder (approved):
Discussed in slack, LGTM once we have the 2-step test include om4 data.

jder:
@alxmrs FYI this was failing GPU tests due to the change I made over the weekend. Will put up a fix. @suryadheeshjith we might want to configure the repo so that statuses are green before you can merge to main?

PR #185, opened 2025-04-08
Added a Claude-generated, ruff checked log parsing script.

alxmrs:
I've been using this script for parsing profile output to coalesce it into tabular data for analysis. I thought I would pull it out of the profile branch and into main for future use.


scripts/parse_training_logs.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python3
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "pandas>2.0",
+#     "pandas-stubs",
```
jder, line 6: I would personally prefer dependencies go in one place, our pyproject.toml, rather than here also. It looks like pandas is included here but not there? And pandas-stubs has different version requirements across the two.

scripts/parse_training_logs.py
```diff
@@ -0,0 +1,147 @@
...
+# ]
+# ///
+"""
+Training Log Parser CLI
+
+This script parses training log files into a CSV format using pandas and argparse.
+
+Generated by Claude 3.5 Haiku in a quick conversation:
+    https://claude.ai/share/b1edac41-72c5-436e-8540-201f56f3c83c
```
jder, line 15: I'm not sure why this is here. If you're saying that you disclaim responsibility for this code then I don't think we should check it in.

scripts/parse_training_logs.py
```diff
@@ -0,0 +1,147 @@
...
+        "Learning Rate",
+        "Current Loss",
+        "Average Loss",
+        "Time",
+        "Avg Time",
+        "Data Time",
+        "Avg Data Time",
+        "Max CPU Mem",
+        "Max GPU Mem",
+    ]
```
jder, line 88: This seems very brittle to me; we could extract these from the log messages themselves mostly?

jder (commented):
So my vote is that we don't check this in if it can be easily re-generated with an LLM. It doesn't look very durable to me and ideally we'd be reading this from wandb or some other structured format.

alxmrs:
It makes sense to prefer structured output over directly parsing logs; we can omit checking in this script.

scripts/parse_training_logs.py
alxmrs: My intention is to give Claude credit. Maybe that's a silly thing to do :)

PR #187, opened 2025-04-09
Use stricter notion of equality for data loaders

jder:
I was trying to figure out why the lazy loader's losses were quite different from eager and torch loaders. I was not successful but given this test currently passes it seems like it was worth checking it in. Thoughts?

PR #188, opened 2025-04-09
Testing the new loaders!

suryadheeshjith:
Checking data loaders and solving PR #187

### DIFF
--- a/tests/test_loaders.py
+++ b/tests/test_loaders.py
@@ -4,10 +4,15 @@ import torch
 import xarray as xr
 
 from ocean_emulators.constants import DEPTH_LEVELS, TensorMap
-from ocean_emulators.datasets import OM4Dataset, TorchTrainDataset, TrainDataset
+from ocean_emulators.datasets import (
+    OM4Dataset,
+    TorchTrainDataset,
+    TrainData,
+    TrainDataset,
+)
 from ocean_emulators.utils.data import Normalize, extract_wet_mask, validate_data
 from ocean_emulators.utils.multiton import MultitonScope
-from ocean_emulators.utils.train import collate_om4
+from ocean_emulators.utils.train import collate_om4, collate_train_data
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


alxmrs:
One idea about the difference in values is that it may come from different approaches to masking.

suryadheeshjith:
Not sure what you mean, I pass the data through the validate function to ensure the data is the same. But, one crucial aspect of the testing was that the data needed to be scaled!

tests/test_loaders.py
```diff
@@ -0,0 +1,148 @@
...
+
+        data = xr.Dataset(
+            {
+                **{
+                    f"thetao_{lev}": (
+                        ["time", "lat", "lon"],
+                        np.random.randn(10, lats, lons) * 10 + 4,
+                    )
+                    for lev in range(levels)
+                },
```
suryadheeshjith: For example, here, scaled by 10

alxmrs:
Very interesting! I think I'm following. Does this mean that the problem is likely in normalize?

suryadheeshjith:
I will add the following tests later! This is ready for review

suryadheeshjith:
I am not sure if the problem is from Normalize(), it would be good to try removing the bias / scale and verify whether the tests still pass

jder:
@suryadheeshjith we're planning to delete the lazy loader rather than try to fix it. Do you think we should get to the bottom of why this test is failing as well or are you OK with just deleting it?

suryadheeshjith:
I am okay with anything. If we aren't seeing any performance gains even with the bug, maybe it's worth not pursuing this further.

jder:
Pasting from [slack](https://openathena.slack.com/archives/C08CYM42DT3/p1744819529637999?thread_ts=1744218302.106789&cid=C08CYM42DT3):

> ok, I believe I have tracked this down to some extent. Turns out the original test (test_new_loaders__are_equal_to_v1_data_loader) also fails with the lazy loader if the mock data includes a wetmask that varies over time. Similarly, the new test (test_loaders) passes if I remove the time dimension from the wetmask. Not sure what the actual bug in the lazy loader is (or if we should even expect wetmasks to vary over time?) but I think this explains the discrepancy between the two tests and we can safely delete one of them.

jder:
Closing in favor of #187 as discussed in slack above.

Issue #102, opened 2025-03-07
Use logging instances, not global loggers.


alxmrs:
Right now, the codebase uses this logging pattern:
```python
import logging

# ... in a routine
logging.info('stmt')
```

The best practice for using the Python standard logging library looks more like this:

```python
import logging

logger = logging.getLogger(__name__)

# ... in a routine
logger.info('stmt')
```

This is a minor difference, but it provides a practical value: namely, when reading the logs, we can easily interpret what file each trace came from.

PR #189, opened 2025-04-09
Using local logger in train.py

alxmrs:
Starts to address #102 (incrementally).

### DIFF
--- a/src/ocean_emulators/utils/logging.py
+++ b/src/ocean_emulators/utils/logging.py
@@ -14,13 +14,12 @@ def handle_logging(cfg):
     # Set up logging
     logger = logging.getLogger()  # Use the root logger or specify a name if needed
     logger.setLevel(logging.DEBUG if cfg.debug else logging.INFO)
+    fmt = logging.Formatter("%(asctime)s - %(levelname)s - %(module)s - %(message)s")
 
     # STDOUT handler
     stdout_handler = logging.StreamHandler(sys.stdout)
     stdout_handler.setLevel(logging.DEBUG if cfg.debug else logging.INFO)
-    stdout_handler.setFormatter(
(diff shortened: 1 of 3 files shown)


src/ocean_emulators/utils/logging.py
```diff
@@ -1,46 +1,59 @@
...
 import torch
 
 
-def handle_logging(cfg):
+def handle_logging(cfg) -> QueueListener:
     # Set up logging
     logger = logging.getLogger()  # Use the root logger or specify a name if needed
     logger.setLevel(logging.DEBUG if cfg.debug else logging.INFO)
+    fmt = logging.Formatter(
+        "%(threadName)s - %(asctime)s - %(levelname)s - %(message)s"
```
alxmrs: Log format now has a prefix to state what thread things are being run in.

jder (changes requested):
Using the module-specific logger LGTM!

I don't think we should do the queue logger thing without good performance evidence because it's a real pain to have your log output not be up to date when eg debugging. I would be surprised if logging to stdout and a file is taking much time given the log volumes we have.

alxmrs:
> I don't think we should do the queue logger thing without good performance evidence because it's a real pain to have your log output not be up to date when eg debugging. 

That's a really going point I didn't consider. As a compromise: what if I just use the bg thread for the file loggers? See the latest commit.

jder (approved):
Thanks!

jder:
Does the existing format already include the module name?

alxmrs:
Let me make sure it's in there before I merge.

suryadheeshjith:
Can you change eval.py as well?

alxmrs:
Sure, happy to do that in this PR.

PR #191, opened 2025-04-09
Using a perf counter to measure time for performance.

alxmrs:


jder (approved):
Thanks! Will be much happier next time there's a leap second or DST change.

PR #192, opened 2025-04-09
Add aggregate data benchmarks to W&B Logs.

alxmrs:
I also added fine-grained debug logging statements to the core of each data loader.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -135,7 +135,7 @@ class InferenceDataset(Dataset):
         data = torch.cat((prognostic, boundary), dim=1)
         return data
 
-    @elapsed(log_level=logging.DEBUG)
+    @elapsed(level=logging.DEBUG)
     def __getitem__(self, idx):
         x_index = self._get_x_index(idx)
         data_in = self._get_prognostic(x_index)
@@ -394,7 +394,7 @@ class TrainDataset(Dataset):
     def __len__(self) -> int:
         return self.size
 
-    @elapsed(log_level=logging.DEBUG)
+    @elapsed(level=logging.DEBUG)
     def __getitem__(self, idx: int):
         TD = TrainData(self.num_prognostic_channels)
         prev_rolling_idx = None
@@ -602,7 +602,7 @@ class TorchTrainDataset(Dataset):
     def __len__(self) -> int:
         return self.size
 
-    @elapsed(log_level=logging.DEBUG)
+    @elapsed(level=logging.DEBUG)
     def __getitem__(self, idx: int):
         TD = TrainData(self.num_prognostic_channels)
         x_indexes = [self._get_x_index(idx, step) for step in range(self.steps)]
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -163,6 +163,10 @@ class OM4Dataset(Dataset):
             .rename({"var": "variable"})
         )
 
+        logger.debug(
+            "[OM4Dataset] get_item took %.4f seconds", time.perf_counter() - get_start
+        )
+
         return input_, label
 
 
@@ -288,7 +292,7 @@ class InferenceDataset(Dataset):
         label = self._get_label(x_index)
 
         logger.debug(
-            "[InferenceDataset] get_item took %.5f seconds",
+            "[InferenceDataset] get_item took %.4f seconds",
             time.perf_counter() - get_start,
         )
         return (data_in, label)
@@ -565,7 +569,7 @@ class TrainDataset(Dataset):
             )
 
         logger.debug(
-            "[TrainDataset] get_item took %.5f seconds", time.perf_counter() - get_start
+            "[TrainDataset] get_item took %.4f seconds", time.perf_counter() - get_start
         )
         return TD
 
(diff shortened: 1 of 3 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -507,6 +539,9 @@ def __getitem__(self, idx: int):
                 label=label,
             )
 
+        logger.debug(
+            "[TrainDataset] get_item took %.5f seconds", time.perf_counter() - get_start
```
alxmrs: How many sigfigs do we want for seconds when debugging? I think 5 is too many; maybe 4 is better?

src/ocean_emulators/utils/logging.py
```diff
@@ -153,6 +153,8 @@ def log_every(self, iterable, print_freq, header=None):
         end = time.time()
         iter_time = SmoothedValue(fmt="{value:.3f}({avg:.3f})", window_size=print_freq)
         data_time = SmoothedValue(fmt="{value:.3f}({avg:.3f})", window_size=print_freq)
+        self.meters["iter_time"] = iter_time
+        self.meters["data_time"] = data_time
```
alxmrs, line 155: This was the simplest cut that I could think of to plumb this data to wandb; I do think this area in general needs refactoring / simplification.

alxmrs:
Hey @suryadheeshjith, may I have your early feedback with this PR? Specifically, would you mind sharing the modifications we did in the profiling branch together? Do the changes here match what we ended up with?

src/ocean_emulators/datasets.py
suryadheeshjith: 4 is good

src/ocean_emulators/datasets.py
```diff
@@ -160,6 +169,10 @@ def __getitem__(self, idx: int) -> Example:
             .rename({"var": "variable"})
         )
 
+        logger.debug(
```
suryadheeshjith: Since this is a pretty critical part of loading, I hope this line of code doesn't cause any overheads

src/ocean_emulators/datasets.py
suryadheeshjith: Same comment as above.

src/ocean_emulators/utils/logging.py
```diff
@@ -165,44 +167,26 @@ def log_every(self, iterable, print_freq, header=None):
...
-                            meters=str(self),
-                            time=str(iter_time),
-                            data=str(data_time),
-                            cpu_memory=resource.getrusage(
-                                resource.RUSAGE_SELF
-                            ).ru_maxrss
-                            / KB,
-                        )
-                    )
+                    named_metrics["gpu_memory"] = torch.cuda.max_memory_allocated() / MB
```
suryadheeshjith, line 186: Nice

suryadheeshjith (approved):
LGTM, just hoping the logging doesn't cause issues in full training runs.

src/ocean_emulators/datasets.py
alxmrs: I've been reading up on this just to be sure, and from what I can tell, this shouldn't cause performance issues. It helps that we're not using Python string interpolation (i.e. `f` strings), which means that string formatting will only happen when the line is called. Second, it helps that `time.perf_counter()` is a very inexpensive operation. 

It is possible that these lines slow down computation depending on the log storage medium (e.g. if we wrote to a file, I think it would slow down the core loop), however, this only would incur that penalty when debug log statements are turned on. In debug mode, I don't think we care too much about this, right?

suryadheeshjith: Yes, I meant in non-debug modes

src/ocean_emulators/utils/logging.py
```diff
@@ -165,44 +167,26 @@ def log_every(self, iterable, print_freq, header=None):
         log_msg = self.delimiter.join(log_msg_list)
         KB = 1024.0
         MB = 1024.0 * 1024.0
+        # dummy data for iter_time; makes sure the first value is not None
+        iter_time.update(time.perf_counter() - end)
```
jder: This seems suspicious to me, why do we need dummy data? I don't see it being used before the update below.

src/ocean_emulators/datasets.py
```diff
@@ -82,6 +86,10 @@ def __init__(
         self.rolling_indices: Integer[xr.DataArray, "step time"] = (
             indices_da + stride * window_dim
         )
+        logger.info(
+            "[OM4Dataset] initialized in %.2f seconds",
+            time.perf_counter() - loader_start,
+        )
```
jder: :sheep: `"%s initialized in …", self.__class__`?

jder: 🐑 Alternatively, given the number of times this is copy-pasted, maybe you'd prefer:

```python

@log_timing(level=DEBUG)
def __init__(…)

```

jder (approved):
So I'm fine with these changes modulo the one question about dummy data. But I don't think this will give us visibility into how long data loading itself is taking in real training runs since those runs are with debug = False and the data loading `__getitem__` time is not logged to wandb. Is that not worth doing right now or a planned follow up?

alxmrs:
@jder, I liked your suggestion so much that I implemented it and now need another review :)

jder (approved):
New decorator LGTM. 🐑 maybe `log_elapsed` would be more clear? 

Still not certain about the other comment re: dummy data.

alxmrs:
Thanks for the additional review! I've just patched the dummy value issue.

PR #193, opened 2025-04-10
Inline inference refactors

### LINKED ISSUE #182: More efficient Inline Inference
* Record + Wandb logging needs to be done more efficiently since GPU utilization is very low during this stage of training
* Record logging needs to be done in batches rather than single timestep
* Preferably log to wandb once at the end

suryadheeshjith:
Components — target extraction, aggregator recording, wandb logging, and writing — are now handled in batches rather than singular timesteps. This allowed me to remove the record_every parameter as well. Additionally, some other minor optimizations were made.

Here is a comparison of the time taken in minutes, before and after optimization:

| Dataset (Period) | Old Inference (With Write) | Old Inference (Without Write)  | Optimized Inference (With Write) | Optimized Inference (Without Write) |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| OM4 (~8 years)  | [14:08](https://wandb.ai/m2lines/ocean-emulators/runs/pjzsf71m) | [13:14](https://wandb.ai/m2lines/ocean-emulators/runs/025m3u3s) | [03:52](https://wandb.ai/m2lines/ocean-emulators/runs/rubapmx5)  |  [02:54](https://wandb.ai/m2lines/ocean-emulators/runs/i9kkjb85) |
| CM4 (20 years)  | [24:52](https://wandb.ai/m2lines/ocean-emulators/runs/zcbmfvme) | [22:45](https://wandb.ai/m2lines/ocean-emulators/runs/v8zazvdi) | [05:27 ](https://wandb.ai/m2lines/ocean-emulators/runs/1ziu9lrt) | [03:17](https://wandb.ai/m2lines/ocean-emulators/runs/mxk43iwq)  |

I made a comparison between writing the actual zarr and skipping write because during train we don't actually save the data produced by the inference rollout but in standalone inference, we do.
 
Training should now be cut down from ~4 days to 2.5 days and Inference should now go from 1152 SYPD (~100x) to 4800 SYPD (400x). 

Add tests
- [x] Ensure correct data fed to model (Use time indexed values for data)
- [x] ~~Check if batched output metrics are different from single output metrics~~ This test is not needed since Alex caught my bug
- [x] Fix time returned so we save appropriate time in zarr

NOTE: The above times with the optimization actually had a minor bug in choosing the next prognostic but there is no change in performance because I just had to change an index. See this [comment](https://github.com/suryadheeshjith/Ocean_Emulator/pull/193#discussion_r2037893244). A fixed run with OM4 with write is logged [here](https://wandb.ai/m2lines/ocean-emulators/runs/b3qznqb0) and time taken was: 03:46

Closes #182

### DIFF
--- a/src/ocean_emulators/aggregator/inference/main.py
+++ b/src/ocean_emulators/aggregator/inference/main.py
@@ -104,7 +104,6 @@ class InferenceEvaluatorAggregator:
         if len(data.target) == 0:
             raise ValueError("No target values in data")
         total_len = len(data.time)
-        assert data.prediction.shape == data.target.shape
         assert data.prediction.shape[0] == total_len // (self.hist + 1)
         target_norm_dict, target_unnorm_dict = get_norm_unnorm_dicts(
             data.target,
--- a/src/ocean_emulators/aggregator/inference/main.py
+++ b/src/ocean_emulators/aggregator/inference/main.py
@@ -103,14 +103,19 @@ class InferenceEvaluatorAggregator:
             raise ValueError("No prediction values in data")
         if len(data.target) == 0:
             raise ValueError("No target values in data")
+        total_len = len(data.time)
+        assert data.prediction.shape == data.target.shape
+        assert data.prediction.shape[0] == total_len // (self.hist + 1)
         target_norm_dict, target_unnorm_dict = get_norm_unnorm_dicts(
             data.target,
+            long_rollout=True,
             input_type="target",
             num_prognostic_channels=self.num_prognostic_channels,
             hist=self.hist,
         )
         gen_norm_dict, gen_unnorm_dict = get_norm_unnorm_dicts(
             data.prediction,
+            long_rollout=True,
             input_type="input",
             num_prognostic_channels=self.num_prognostic_channels,
             hist=self.hist,
@@ -151,6 +156,7 @@ class InferenceEvaluatorAggregator:
 
         data_norm_dict, data_unnorm_dict = get_norm_unnorm_dicts(
             initial_prognostic,
+            long_rollout=True,
             input_type="input",
             num_prognostic_channels=self.num_prognostic_channels,
             hist=self.hist,
@@ -257,8 +263,12 @@ class InferenceAggregator:
         if len(data.data) == 0:
             raise ValueError("data is empty")
 
+        total_len = len(data.time)
+        assert data.data.shape[0] == total_len // (self.hist + 1)
+
         _, data_unnorm_dict = get_norm_unnorm_dicts(
             data.data,
+            long_rollout=True,
             input_type="gen",
             num_prognostic_channels=self.num_prognostic_channels,
             hist=self.hist,
@@ -288,6 +298,7 @@ class InferenceAggregator:
             )
         _, data_unnorm_dict = get_norm_unnorm_dicts(
             initial_prognostic,
+            long_rollout=True,
             input_type="input",
             num_prognostic_channels=self.num_prognostic_channels,
             hist=self.hist,
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -279,9 +279,9 @@ class InferenceDataset(Dataset):
     def _get_x_index(self, idx):
         if isinstance(idx, slice):
             if (
-                (idx.start and idx.start < 0)
-                or (idx.stop and idx.stop < 0)
-                or (idx.step and idx.step < 0)
+                (idx.start is not None and idx.start < 0)
+                or (idx.stop is not None and idx.stop < 0)
+                or (idx.step is not None and idx.step < 0)
             ):
                 raise IndexError("Sorry, negative indexing is not supported!")
             if idx.step is None:
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -195,6 +195,7 @@ class InferenceDataset(Dataset):
         self.num_prognostic_channels = (hist + 1) * len(prognostic_var_names)
         self._prognostic_vars = data[prognostic_var_names]
         self._boundary_vars = data[boundary_var_names]
+        self._times = data.time
 
         time_indices = np.arange(data.time.size)
         indices = xr.DataArray(
@@ -236,19 +237,23 @@ class InferenceDataset(Dataset):
 
     @property
     def initial_prognostic(self):
-        data = self.__getitem__(0)[0]
-        return data[:, : self.num_prognostic_channels]
+        x_index = self._get_x_index(0)
+        data_in = self._get_prognostic(x_index)
+        return data_in
 
-    def inference_target(self, step: int):
-        return self.__getitem__(step)[1]
+    def inference_target(self, step: int | slice):
+        x_index = self._get_x_index(step)
+        label = self._get_label(x_index)
+        return label
 
     def get_initial_input(self):
         data = self.__getitem__(0)[0]
         return data
 
     # TODO: This is a placeholder for now since time returned is incorrect
-    def get_input_time(self, step: int):
-        return self._prognostic_vars.time[step]
+    def get_input_time(self, step: int | slice):
+        x_index = self._get_x_index(step)
+        return self._times[x_index]
 
     def merge_prognostic_and_boundary(self, prognostic: torch.Tensor, step: int):
         x_index = self._get_x_index(step)
@@ -266,16 +271,20 @@ class InferenceDataset(Dataset):
 
     def _get_x_index(self, idx):
         if isinstance(idx, slice):
-            if idx.start < 0 or idx.stop < 0 or idx.step < 0:
+            if (
+                (idx.start and idx.start < 0)
+                or (idx.stop and idx.stop < 0)
+                or (idx.step and idx.step < 0)
+            ):
                 raise IndexError("Sorry, negative indexing is not supported!")
-            elif idx.start >= self.size or idx.stop >= self.size:
-                raise IndexError(f"Index {idx} out of range with size {self.size}")
             elif idx.start is None and idx.stop is None:
                 idx = slice(0, self.size, idx.step)
             elif idx.start is None:
                 idx = slice(0, idx.stop, idx.step)
             elif idx.stop is None:
                 idx = slice(idx.start, self.size, idx.step)
+            elif idx.step is None:
+                idx = slice(idx.start, idx.stop, 1)
         elif isinstance(idx, int):
             if idx < 0:
                 raise IndexError("Sorry, negative indexing is not supported!")
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -250,10 +250,17 @@ class InferenceDataset(Dataset):
         data = self.__getitem__(0)[0]
         return data
 
-    # TODO: This is a placeholder for now since time returned is incorrect
-    def get_input_time(self, step: int | slice):
-        x_index = self._get_x_index(step)
-        return self._times[x_index]
+    def get_target_time(self, start_step: int, num_steps: int):
+        x_index = self._get_x_index(start_step)
+        batch_index = x_index.values[0]
+        steps_predicted = len(batch_index) // 2
+        start_target_index = batch_index[steps_predicted]
+
+        return self._times.isel(
+            time=slice(
+                start_target_index, start_target_index + num_steps * steps_predicted
+            )
+        )
 
     def merge_prognostic_and_boundary(self, prognostic: torch.Tensor, step: int):
         x_index = self._get_x_index(step)
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -532,16 +532,6 @@ class TrainDataset(Dataset):
         # Create a slice for similar indexing as in InferenceDataset
         idx_slice = slice(start, end)
         rolling_idx = self.rolling_indices.isel(window_dim=idx_slice)
-        # Convert to tests, tests are outdated since changing time definition
-        # if prev_rolling_idx is not None:
-        #     assert (
-        #         prev_rolling_idx.isel(time=slice(self.hist + 1, None))
-        #         - rolling_idx.isel(time=slice(0, self.hist + 1))
-        #     ).sum() == 0  # Prev output = Cur Input
-        #     assert (
-        #         rolling_idx.diff("time") == self.stride
-        #     ).all()  # Stride is maintained
-        #     assert rolling_idx.isel(time=-1) < self.size  # Last index check
 
         x_index = xr.Variable(["window_dim", "time"], rolling_idx)
         return x_index
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -350,7 +350,6 @@ class InferenceDataset(Dataset):
             "window_dim time variable lat lon -> window_dim (time variable) lat lon",
         )
         label = torch.from_numpy(label).float()
-        # label = label * self.wet
         label = torch.where(self.wet, label, 0.0)
         return label
 
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -277,14 +277,14 @@ class InferenceDataset(Dataset):
                 or (idx.step and idx.step < 0)
             ):
                 raise IndexError("Sorry, negative indexing is not supported!")
-            elif idx.start is None and idx.stop is None:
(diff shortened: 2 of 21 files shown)


suryadheeshjith (commented):
Oddly there seems to be some differences in the old and optimized rollouts. I will need to add tests

suryadheeshjith:
I've added a bunch of parameter sweep tests since they almost take no time to complete. Happy to remove some of them though.

src/ocean_emulators/datasets.py
```diff
@@ -266,16 +271,20 @@ def __getitem__(self, idx):
...
-            elif idx.start >= self.size or idx.stop >= self.size:
-                raise IndexError(f"Index {idx} out of range with size {self.size}")
             elif idx.start is None and idx.stop is None:
                 idx = slice(0, self.size, idx.step)
             elif idx.start is None:
                 idx = slice(0, idx.stop, idx.step)
             elif idx.stop is None:
                 idx = slice(idx.start, self.size, idx.step)
+            elif idx.step is None:
+                idx = slice(idx.start, idx.stop, 1)
```
alxmrs: Great catching this missing case!

🐑  you might be able to reduce the number of conditionals by using `if` instead of `elif`. I have an example of this in the `window_from` function of the `OM4Dataset` class.

src/ocean_emulators/datasets.py
```diff
@@ -523,16 +532,6 @@ def _get_x_index(
...
-        # Convert to tests, tests are outdated since changing time definition
-        # if prev_rolling_idx is not None:
-        #     assert (
-        #         prev_rolling_idx.isel(time=slice(self.hist + 1, None))
-        #         - rolling_idx.isel(time=slice(0, self.hist + 1))
-        #     ).sum() == 0  # Prev output = Cur Input
-        #     assert (
-        #         rolling_idx.diff("time") == self.stride
-        #     ).all()  # Stride is maintained
-        #     assert rolling_idx.isel(time=-1) < self.size  # Last index check
```
alxmrs, line 535: Thanks for deleting this.

src/ocean_emulators/datasets.py
suryadheeshjith: Good catch, fixed!

src/ocean_emulators/models/base.py
```diff
@@ -82,8 +83,12 @@ def inference(
         steps_completed=0,
         num_steps=None,
         epoch=None,
-    ) -> list[torch.Tensor]:
-        outputs: list[torch.Tensor] = []
+    ) -> InfOutput:
```
alxmrs, line 86: Great work replacing the list with a tensor.

src/ocean_emulators/datasets.py
```diff
@@ -236,19 +237,23 @@ def __len__(self):
...
+    def inference_target(self, step: int | slice):
+        x_index = self._get_x_index(step)
+        label = self._get_label(x_index)
+        return label
 
     def get_initial_input(self):
         data = self.__getitem__(0)[0]
         return data
 
     # TODO: This is a placeholder for now since time returned is incorrect
```
alxmrs: Is this comment still up-to-date?

src/ocean_emulators/models/base.py
```diff
@@ -99,7 +104,7 @@ def inference(
                 )
             else:
                 input_tensor = dataset.merge_prognostic_and_boundary(
-                    prognostic=outputs[-1],
+                    prognostic=pred_tensor[-1].unsqueeze(0),
```
alxmrs: I need some help understanding how this works: Does this pass in the last step of the `pred_tensor` array of zeros in to the merge?

src/ocean_emulators/models/base.py
```diff
@@ -99,7 +104,7 @@ def inference(
...
                     step=steps_completed + step,
```
alxmrs, line 106: `steps_completed` --> is this how we do a sort of gander accumulation or step rollout for inference?

src/ocean_emulators/models/base.py
suryadheeshjith: AH! You caught a nasty bug! This explains the difference in rollout behavior. Thank you 🙏

I meant to put pred_tensor[step-1]

src/ocean_emulators/models/base.py
suryadheeshjith: Yepp, we are performing windowed rollouts so we can rollout to infinite steps

src/ocean_emulators/models/base.py
```diff
@@ -99,7 +104,7 @@ def inference(
                 )
             else:
```
alxmrs, line 103: So, bear with me, I'm still trying to understand how this loops works. I'm also trying to imagine how this could be vectorized (though, I do think it is good the way it is now). Some notes:
- `get_initial_input()` called with step=0 seems equivalent to getting the initial boundary and initial prognostic and merging them, like we do in the `elif` below. (I need help to verify this, but that is may takeaway reading the code above). I think it's possible to combine the `if` and `elif` into one block, given some refactoring
- I still am not sure how `pred_tensor[-1]` functions (see question below), but I am curious if the `else` block could be combined with the above blocks like so: could we somehow treat `initial_prognostic` as values in the `num_steps=0` dimension of `pred_tensor`?
- if the above points are true and these conditionals could be made into one block, then is it possible to generalize them wrt the `step`? i.e. I think if we took another look at the loop, it could be vectorized into an array operation.

alxmrs: This change is already quite an improvement, so we can definitely save these notes for a future time.

alxmrs: Given the fix you just made on 107, I'm happy to say the vectorization logic is way more possible now :)

src/ocean_emulators/datasets.py
suryadheeshjith: I will remove it once I fix the logic of saving times to zarr

src/ocean_emulators/models/base.py
```diff
@@ -99,7 +104,7 @@ def inference(
...
                 input_tensor = dataset.merge_prognostic_and_boundary(
-                    prognostic=outputs[-1],
+                    prognostic=pred_tensor[step - 1].unsqueeze(0),
```
alxmrs, line 105: Why is it `step -1` and not `step`?

src/ocean_emulators/stepper.py
```diff
@@ -104,44 +103,19 @@ def inference(
...
-                    if writer:
-                        writer.write()
-                    all_logs = []
-
-            if len(all_logs) > 0:
-                logging.info(f"Inference [epoch {epoch}]: wandb logging...")
-                record_logs(all_logs)
-                if writer:
-                    writer.write()
+            initial_prognostic = IO.prediction[-1].unsqueeze(0).clone()
```
alxmrs, line 114: Given the change above, should this also be `step` or `step - 1`?

src/ocean_emulators/stepper.py
```diff
@@ -104,44 +103,19 @@ def inference(
...
+            if writer:
+                writer.record_batch(IO)
+                writer.write()
```
alxmrs: Feels good to replace so much logic with just three lines :)

src/ocean_emulators/train.py
```diff
@@ -597,7 +597,7 @@ def inference_one_epoch(self, epoch):
                     dataset=inference_dataset,
                     inf_aggregator=inf_aggregator,
                     epoch=epoch,
-                    num_model_steps_forward=num_steps,
+                    num_model_steps_forward=num_steps // 2,
```
alxmrs: I'm new to the Stepper area of the code -- what does it mean to cut the number of steps in half?

src/ocean_emulators/models/base.py
suryadheeshjith: Because we want to use the previous step's prognostic.

src/ocean_emulators/stepper.py
suryadheeshjith: Here, we would use the last one since our pred_tensor or IO.prediction is a tensor corresponding to a windowed rollout

src/ocean_emulators/models/base.py
alxmrs: ah, that makes a lot of sense, thank you.

src/ocean_emulators/train.py
suryadheeshjith: Previously, during training, we were rolling out everything at once and then performing aggregator recording and wandb logging after for each timestep. Now we produce a rollout halfway, log things and then rollout the rest. With this implementation, I was getting OOM errors when I rollout out all steps at once.

src/ocean_emulators/datasets.py
```diff
@@ -266,11 +271,15 @@ def __getitem__(self, idx):
...
+                or (idx.stop and idx.stop < 0)
+                or (idx.step and idx.step < 0)
+            ):
                 raise IndexError("Sorry, negative indexing is not supported!")
-            elif idx.start >= self.size or idx.stop >= self.size:
-                raise IndexError(f"Index {idx} out of range with size {self.size}")
-            elif idx.start is None and idx.stop is None:
+            if idx.step is None:
+                idx = slice(idx.start, idx.stop, 1)
+            if idx.start is None and idx.stop is None:
```
alxmrs, line 289: I think we could go even further:
```suggestion
            if idx.start is None:
                idx = slice(0, idx.stop, idx.step)
            if idx.stop is None:
                idx = slice(idx.start, self.size, idx.step)
```

alxmrs: (probably not a good idea to accept this suggestion; I couldn't modify the code below)

src/ocean_emulators/utils/data.py
```diff
@@ -190,7 +190,8 @@ def get_norm_unnorm_dicts(
         data = data[:, :num_prognostic_channels]
 
     # Separate history from channels
-    data_reshaped = rearrange(data, "n (hi c) h w -> n hi c h w", hi=hist + 1)
+    data_reshaped = rearrange(data, "n (hi c) h w -> (n hi) c h w", hi=hist + 1)
+    data_reshaped = data_reshaped.unsqueeze(0)  # batch dim for minimal code-change
```
alxmrs: nice

src/ocean_emulators/utils/model.py
```diff
@@ -38,7 +39,7 @@ def __init__(
         self,
         prediction: torch.Tensor,
         target: torch.Tensor,
-        time: torch.Tensor,
+        time: xr.DataArray,
```
alxmrs, line 43: Thanks, nice.

tests/test_stepper.py
```diff
@@ -0,0 +1,227 @@
+import numpy as np
+import pytest
```
alxmrs, line 2: Awesome, great to have stepper tests.

src/ocean_emulators/models/base.py
suryadheeshjith: I am not sure what you mean by vectorized since we need to sequentially produce new states. 

For example this is a representation of the model rollout:
x1' = Model(x0)
x2' = Model(x1')
x3' = Model(x2')

Can you elaborate on how we can vectorized this aspect? Or maybe I missed your point

suryadheeshjith: But yes, I can clean up the conditionals and remove that one if condition

src/ocean_emulators/datasets.py
suryadheeshjith: I think it is good the way it is right now, There won't be any cases missed (Unless I am mistaken?)

src/ocean_emulators/models/base.py
alxmrs: I didn't understand that this was more like a diffeq, where the next prediction depends on the previous. Given this, there's no sense trying to eliminate the loop.

suryadheeshjith:
This PR is ready for review

suryadheeshjith:
The om4-torch loader version doesn’t handle long windowed rollouts well — it seems to freeze as soon as it tries to load the target data into memory. Here are two runs that fail with longer rollouts: [Run 1](https://wandb.ai/m2lines/ocean-emulators/runs/8khd08jb), [Run 2](https://wandb.ai/m2lines/ocean-emulators/runs/zuwizagk). And here’s one that works using shorter windowed rollouts: [Run 3](https://wandb.ai/m2lines/ocean-emulators/runs/2f5zuole). While the shorter rollout setup takes more time per epoch, it’s still faster overall compared to using just the om4-torch loader without inline batched inference. Training on larger data like CM4 might warrant the switch to the om4-torch loader though,

src/ocean_emulators/train.py
jder: Probably safer if we did something like num_model_steps_forward=(min(num_steps, MAX_ROLLOUT_STEPS_AT_ONCE))?

src/ocean_emulators/stepper.py
```diff
@@ -97,51 +96,26 @@ def inference(
...
 
+            logs = inf_aggregator.record_batch(IO)
+            record_logs(logs)
```
jder, line 123: Do I understand this change correctly that we could have gotten these same performance improvements (but not code cleanup) by setting record_every to a large number? 

Can we add a logging.info here that we're about to do the recording?

src/ocean_emulators/utils/data.py
```diff
@@ -190,7 +191,12 @@ def get_norm_unnorm_dicts(
         data = data[:, :num_prognostic_channels]
 
     # Separate history from channels
-    data_reshaped = rearrange(data, "n (hi c) h w -> n hi c h w", hi=hist + 1)
+    if long_rollout:
+        data_reshaped = rearrange(data, "n (hi c) h w -> (n hi) c h w", hi=hist + 1)
+        data_reshaped = data_reshaped.unsqueeze(0)  # add artificial batch dim
+    else:
+        data_reshaped = rearrange(data, "n (hi c) h w -> n hi c h w", hi=hist + 1)
+
```
jder, line 199: I don't understand this change, can you leave a comment explaining why long_rollout wants to have `(n hi)` in a single dimension? Or is there a more meaningful name we can give to `long_rollout` that explains this behavior?

src/ocean_emulators/aggregator/inference/main.py
```diff
@@ -103,14 +103,19 @@ def record_batch(self, data: InfOutput):
             raise ValueError("No prediction values in data")
         if len(data.target) == 0:
             raise ValueError("No target values in data")
+        total_len = len(data.time)
+        assert data.prediction.shape == data.target.shape
+        assert data.prediction.shape[0] == total_len // (self.hist + 1)
```
jder, line 107: Maybe some of these asserts should be in `InfOutput.__init__`?

src/ocean_emulators/utils/data.py
jder: (Generally I'm a little confused about what that InfOutput and ValOutput are expected to have, perhaps some jaxtyping or commenting those would make this all clearer?)

jder (approved):
Makes sense to me! Just a few minor suggestions & a couple questions for my own understanding.

src/ocean_emulators/stepper.py
suryadheeshjith: No, the problem was even if you set record_every to a large number, you are still aggregator recording a single timestep which is suboptimal. The functionality of logging multiple steps at a time already existed within the aggregators though. I just underestimated how bad recording a single timestep was in terms of performance.

suryadheeshjith: I will add the logging info line

src/ocean_emulators/utils/data.py
suryadheeshjith: I can try adding Jaxtyping here. The main reason is that during validation, we are enforcing a single timestep output and that means the batch is separate from the predictions (each prediction in the batch are unrelated to other batch predictions). But for inference / long rollout, we have the predictions stacked in its batch dimension, and thus our full set of steps would correspond to combining it with hist

suryadheeshjith: This does limit future use cases where we want to rollout more steps for validation, but I think this is an ok fix for now

src/ocean_emulators/aggregator/inference/main.py
suryadheeshjith: I moved the simpler assertions to ValOutput and InfOutput

src/ocean_emulators/utils/data.py
suryadheeshjith: I jumped the gun without taking care of this, will add a fix now!

PR #198, opened 2025-04-11
Inline Inference - Clarity Refactor

suryadheeshjith:
- Added a comment to provide clarity on what we are doing during inference and validation.


suryadheeshjith:
@jder Can you help me with the jax typing? Where exactly do you suggest I add them?

jder:
I was thinking of putting them on eg the return type of this function and possibly the argument types?

suryadheeshjith:
Would probably need a few iterations to get this right. I believe there are a bunch of places where we can replace the jaxtyping, especially in the datasets file. It would be nice to explicitly mention places that have the time dimension.

jder (approved):
Thanks, this is super helpful!

PR #199, opened 2025-04-11
Cleaning up benchmarks

### LINKED ISSUE #186: Clean up benchmark page, remove redundant benchmark plots
After expanding (#151) and contracting (#179) the number of test combinations, the benchmark page has been rendered a bit confusing. It has too many plots, many of which are for tests that we no longer profile. Thus, we should mess with the benchmark data source and delete a lot of the redundant benchmark records.

alxmrs:
Fixes #186 by removing all benchmark records with history!=0 or using a non-`mock` data source (i.e. remote data).


jder (approved):
Thanks! Why did you pick preserving history=0 vs history=1? I think the latter is what we typically use for "real" runs?

alxmrs:
That's totally true. However, I picked hist=0 to preserve the longevity of data. With this change, all the benchmark plots are roughly equivalent, except the generated names have changed. This means that in the future, we could gather all the data together and create one super chart.

PR #201, opened 2025-04-12
Better colors for wandb map plots

suryadheeshjith:
Better coloring of land cells in the plots


alxmrs (approved):
LGTM

src/ocean_emulators/aggregator/plotting.py
```diff
@@ -1,5 +1,5 @@
 import gc
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Tuple
```
alxmrs: 🐑 /nit: I prefer if we use built-in collections for type annotations.

PR #202, opened 2025-04-12
Minor stepper refactors and norm metrics fix 

suryadheeshjith:
- Fix the calculation of metrics for normalized data. We were still considering 0s at land.
- Renamed output types and moved them to output.py

### DIFF
--- a/src/ocean_emulators/aggregator/inference/main.py
+++ b/src/ocean_emulators/aggregator/inference/main.py
@@ -112,7 +112,7 @@ class InferenceEvaluatorAggregator:
             data.target,
             wet=self.wet,
             long_rollout=True,
-            input_type="target",
+            input_type="prognostic",
             num_prognostic_channels=self.num_prognostic_channels,
             hist=self.hist,
         )
@@ -120,7 +120,7 @@ class InferenceEvaluatorAggregator:
             data.prediction,
             wet=self.wet,
             long_rollout=True,
-            input_type="input",
+            input_type="prognostic",
             num_prognostic_channels=self.num_prognostic_channels,
             hist=self.hist,
         )
--- a/src/ocean_emulators/aggregator/inference/main.py
+++ b/src/ocean_emulators/aggregator/inference/main.py
@@ -5,12 +5,12 @@ import wandb
 import xarray as xr
 
 from ocean_emulators.utils.data import Normalize, get_norm_unnorm_dicts
-from ocean_emulators.utils.model import InfOutput, SingleTimeseriesOutput
+from ocean_emulators.utils.output import ModelInferenceOutput
 from ocean_emulators.utils.wandb import Metrics, MetricsDict
 
 from ..validate.reduced import MeanAggregator as OneStepMeanAggregator
-from .reduced import MeanAggregator, SingleTargetMeanAggregator
-from .time_mean import TimeMeanAggregator, TimeMeanEvaluatorAggregator
+from .reduced import MeanAggregator
+from .time_mean import TimeMeanEvaluatorAggregator
 
 
 class InferenceEvaluatorAggregator:
@@ -24,6 +24,7 @@ class InferenceEvaluatorAggregator:
         metadata: Dict[str, Dict[str, str]],
         hist: int,
         area_weights: torch.Tensor,
+        wet: torch.Tensor,
         num_prognostic_channels: int,
         record_step_20: bool = True,
         log_global_mean_time_series: bool = True,
@@ -37,6 +38,7 @@ class InferenceEvaluatorAggregator:
                 used in generating logged image captions.
             hist: Number of timesteps of history.
             area_weights: Area weights for the data.
+            wet: Wet mask for the data.
             num_prognostic_channels: Number of prognostic channels in the data.
             record_step_20: Whether to record the mean of the 20th steps.
             log_global_mean_time_series: Whether to log global mean time series metrics.
@@ -92,13 +94,14 @@ class InferenceEvaluatorAggregator:
         self._normalize = Normalize.get_instance()
         self.num_prognostic_channels = num_prognostic_channels
         self.hist = hist
+        self.wet = wet
 
     @property
     def log_time_series(self) -> bool:
         return self._log_time_series
 
     @torch.no_grad()
-    def record_batch(self, data: InfOutput):
+    def record_batch(self, data: ModelInferenceOutput):
         if len(data.prediction) == 0:
             raise ValueError("No prediction values in data")
         if len(data.target) == 0:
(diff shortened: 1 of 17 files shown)


src/ocean_emulators/aggregator/inference/main.py
```diff
@@ -217,147 +223,6 @@ def _get_inference_logs_slice(self, step_slice: slice):
         return to_inference_logs(logs)
 
 
-class InferenceAggregator:
```
suryadheeshjith, line 220: Removed this class since we are not using it anywhere. Might add it in the future if we do need it!

src/ocean_emulators/aggregator/train.py
```diff
@@ -15,7 +15,7 @@ def __init__(self):
         self._loss_per_channel = torch.tensor(torch.nan)
 
     @torch.no_grad()
-    def record_batch(self, batch: TrainOutput):
+    def record_batch(self, batch: TrainStepOutput):
```
jder: :sheep: these names make me think that this should be record_step taking `step: TrainStepOutput`? Or the type should be `TrainBatchOutput`?

src/ocean_emulators/utils/data.py
```diff
@@ -198,7 +199,9 @@ def get_norm_unnorm_dicts(
         data_reshaped = rearrange(data, "n (hi c) h w -> n hi c h w", hi=hist + 1)
 
     # Get normalized dict
-    data_dict = convert_tensor_out_to_dict(data_reshaped)
+    data_normalized = data_reshaped.clone()
```
jder, line 211: Why do we need the clone? I don't see any mutations to `data_normalized` or `data_reshaped`.

src/ocean_emulators/utils/data.py
```diff
@@ -198,7 +199,9 @@ def get_norm_unnorm_dicts(
...
+    data_normalized = torch.where(wet == 0, float("nan"), data_normalized)
```
jder, line 212: I'm a bit confused why this method needs to do this. It's converting from normalized-data to normalized-data-with-nans-for-land? Is this because this method is only intended to be used by aggregators that want to compute stats across wet cells? Would maybe be nice to try and distinguish that at a type level someday but maybe for now this method could be `get_norm_unnorm_analysis_dicts` or something? Not sure "analysis" is the right name.

src/ocean_emulators/eval.py
```diff
@@ -122,7 +125,9 @@ def __init__(self, cfg: EvalConfig) -> None:
         self.wet, self.wet_surface = extract_wet_mask(
             self.data, self.prognostic_var_names, cfg.data.hist
         )
-        wet_without_hist, _ = extract_wet_mask(self.data, self.prognostic_var_names, 0)
+        self.wet_without_hist, _ = extract_wet_mask(
+            self.data, self.prognostic_var_names, 0
+        )
```
jder, line 131: Why did you want to assign this to self here? It feels worse to me because `self.wet_without_history` is sometimes on self.device and sometimes not. If it's important that Normalize is handed the non-device wet mask then I think we should have this be a local `wet_without_history_cpu` or something. If it's not important, maybe do the `to(self.device)` here?

src/ocean_emulators/train.py
```diff
@@ -200,7 +200,9 @@ def __init__(self, cfg: TrainConfig) -> None:
         self.wet, self.wet_surface = extract_wet_mask(
             self.data, self.prognostic_var_names, cfg.data.hist
         )
-        wet_without_hist, _ = extract_wet_mask(self.data, self.prognostic_var_names, 0)
+        self.wet_without_hist, _ = extract_wet_mask(
+            self.data, self.prognostic_var_names, 0
+        )
```
jder, line 205: Same comment/question here about overwriting self.wet_without_hist

tests/test_utils_data.py
```diff
@@ -166,7 +170,7 @@ def normalize_input():
             boundary_var_names=["var_2"],
             wet_mask=wet_mask,
         )
-    return normalize, wet_mask
+        yield normalize, wet_mask
```
jder, line 173: Nice catch

tests/test_utils_data.py
```diff
@@ -186,3 +190,104 @@ def test_unnormalize_prognostic_tensor(normalize_input, fill_value):
...
+        tensor_map = TensorMap.init_instance("thetao_surface", "hfds")
+
+        wet_mask_ = np.array([[1, 0, 1], [0, 1, 0], [1, 0, 1]])
+        wet_full = np.tile(wet_mask_, (total_time_steps, levels, 1, 1))
+
+        # Even thetao, odd hfds for every time step
+        # Ex, timestep 0: thetao = 0, hfds = 1
+        # Ex, timestep 1: thetao = 2, hfds = 3
+        # Ex, timestep 2: thetao = 4, hfds = 5
+        # ...
```
jder, line 212: We don't actually care about this structure, right? I assume this is just copy-pasted? Maybe once we've gotten rid of multitons it will be easier to share this kind of setup between tests.

tests/test_utils_data.py
```diff
@@ -186,3 +190,104 @@ def test_unnormalize_prognostic_tensor(normalize_input, fill_value):
...
+                "time": np.arange(total_time_steps),
+                "lev": DEPTH_LEVELS,
+                "lat": np.arange(lats),
+                "lon": np.arange(lons),
+            },
+        )
+        data_mean = data.mean() * 0.0
+        data_std = data.std() * 0.0 + 1.0
+        data, data_mean, data_std = validate_data(data, data_mean, data_std)
+        wet, wet_surface = extract_wet_mask(data, tensor_map.prognostic_var_names, hist)
```
jder: wet/wet_surface are unused.
```suggestion
```

jder (approved):
Just some questions/suggestions from me but generally makes sense!

tests/test_utils_data.py
suryadheeshjith: Yes

src/ocean_emulators/aggregator/train.py
suryadheeshjith: Fixed

src/ocean_emulators/utils/data.py
suryadheeshjith: We are using data_reshaped for both the normalized (renamed to data_normalized) and unnormalized dicts. I treated it as a "just in case" clone

src/ocean_emulators/utils/data.py
suryadheeshjith: I will rename the function since we use them only for aggregators

Issue #208, opened 2025-04-14
Resolve performance issue with torch loader & new inference code


jder:
When we have the new torch loader on, dask is off and so when we load the labels for long rollouts, it does so sequentially which takes a long time. https://openathena.slack.com/archives/C08CYM42DT3/p1744383970262879?thread_ts=1744304306.945759&cid=C08CYM42DT3

jder:
@alxmrs I was thinking of stealing this from you since I know you're pretty busy this week. Do you have thoughts about how to fix this, in particular the comment in #211 says that it should make this trivial but I'm not sure what you had in mind.

alxmrs:
I would love for you to seal this from me! I was thinking for the inference code portion, we would turn dask on when calling `from_config`.

jder:
Did some digging into this. I think the issue is that the torch loader appears to be using about 2x the (CPU) memory as the eager loader, which was causing the runs discussed in that thread to OOM. (FYI @suryadheeshjith). I would guess this is easily resolvable but not "trivial" anymore :)

alxmrs:
My mistake! I don't have too much context into this issue, but I was hoping it could be solved by turning on/off dask. At least that is an easier thing to do.

Issue #190, opened 2025-04-09
Refactor train and eval files - prevent rewriting setup


suryadheeshjith:

jder:
@alxmrs was planning to take this so assigning to him

alxmrs:
Thanks! That's right.

alxmrs:
This should be closed with #211 (and the subsequent DataContainer feature). Closing.

PR #175, opened 2025-04-04
[Ugly draft] Lazy loader can work with compact input data.

alxmrs:
Fixes #120.

PR #211, opened 2025-04-15
Introducing `DataSource` to train and eval scripts.

### LINKED ISSUE #218: Data Class to combine various sources of data and their statistics
- This would be useful when we want to train on multiple datasets - ex. OM4, CM4, etc
- Each data sample must have their associated data, mean and standard deviation

alxmrs:
This makes progress towards #190. Specifically, this PR argues that the three Xarray datasets (data, means, stds) are actually part of the same concern, one that is fundamentally related to normalization. I anticipate that this affordance will make it easier to modify core aspects to the data loader in the future. In particular, the addition of the `DataSource.filter()` method should provide a single entry point to add data compaction support -- see #215 (initially prototyped in #175). In addition, this change should make #208 trivial to implement.

Fixes #218.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -1,3 +1,4 @@
+import dataclasses
 import logging
 from typing import Any
 
@@ -20,7 +21,7 @@ from ocean_emulators.constants import (
     PrognosticMask,
     PrognosticVarNames,
 )
-from ocean_emulators.utils.data import Normalize
+from ocean_emulators.utils.data import Normalize, DataSource, to_tensor, normalize, normalize_tensor
 from ocean_emulators.utils.device import get_device, using_gpu
 
 
@@ -40,7 +41,7 @@ class InferenceDataset(Dataset):
 
     def __init__(
         self,
-        data,
+        src: DataSource,
         prognostic_var_names,
         boundary_var_names,
         wet,
@@ -52,10 +53,14 @@ class InferenceDataset(Dataset):
         self.device = get_device()
 
         self.hist = hist
+        self.src = src
+        data = src.data
 
         self.num_prognostic_channels = (hist + 1) * len(prognostic_var_names)
-        self._prognostic_vars = data[prognostic_var_names]
-        self._boundary_vars = data[boundary_var_names]
+        self._prog_src = src.filter(prognostic_var_names)
+        self._bound_src = src.filter(boundary_var_names)
+        self._prognostic_vars = self._prog_src.data
+        self._boundary_vars = self._bound_src.data
         self._times = data.time
 
         time_indices = np.arange(data.time.size)
@@ -85,8 +90,8 @@ class InferenceDataset(Dataset):
                 )
             )
 
-        self.wet = wet.bool()
-        self.wet_surface = wet_surface.bool()
+        self.wet: torch.Tensor = wet.bool()
+        self.wet_surface: torch.Tensor = wet_surface.bool()
         self.size = len(self.rolling_indices)
 
         if using_gpu():
@@ -169,9 +174,8 @@ class InferenceDataset(Dataset):
         data_in_ds: xr.Dataset = self._prognostic_vars.isel(time=x_index).isel(
             time=slice(None, self.hist + 1)
         )
-        data_in_ds = Normalize.get_instance().normalize_prognostic(
-            data_in_ds
-        )  # TODO: Weird error when I get_instance in init
+        src_in = dataclasses.replace(self._prog_src, data=data_in_ds)
+        data_in_ds = normalize(src_in)
         data_in_np: np.ndarray = (
             data_in_ds.to_array()
             .transpose("window_dim", "time", "variable", "lat", "lon")
@@ -195,9 +199,8 @@ class InferenceDataset(Dataset):
         data_in_boundary_ds: xr.Dataset = self._boundary_vars.isel(time=x_index).isel(
             time=self.hist
         )
-        data_in_boundary_ds = Normalize.get_instance().normalize_boundary(
-            data_in_boundary_ds
-        )
+        src_in_boundary = dataclasses.replace(self._bound_src, data=data_in_boundary_ds)
+        data_in_boundary_ds = normalize(src_in_boundary)
         data_in_boundary_np: np.ndarray = (
             data_in_boundary_ds.to_array()
             .transpose("window_dim", "variable", "lat", "lon")
@@ -211,7 +214,8 @@ class InferenceDataset(Dataset):
         label_ds: xr.Dataset = self._prognostic_vars.isel(time=x_index).isel(
             time=slice(self.hist + 1, None)
         )
-        label_ds = Normalize.get_instance().normalize_prognostic(label_ds)
+        src_label = dataclasses.replace(self._prog_src, data=label_ds)
+        label_ds = normalize(src_label)
         label_np: np.ndarray = (
             label_ds.to_array()
             .transpose("window_dim", "time", "variable", "lat", "lon")
@@ -305,7 +309,7 @@ class TrainDataset(Dataset):
 
     def __init__(
         self,
-        data: xr.Dataset,
+        src: DataSource,
         prognostic_var_names: PrognosticVarNames,
         boundary_var_names: BoundaryVarNames,
         wet: PrognosticMask,
@@ -320,10 +324,11 @@ class TrainDataset(Dataset):
         self.hist: int = hist
         self.steps: int = steps
         self.stride: int = stride
+        data = src.data
+        self._prog_src = src.filter(prognostic_var_names)
+        self._bound_src = src.filter(boundary_var_names)
 
         self.num_prognostic_channels: int = (hist + 1) * len(prognostic_var_names)
-        self._prognostic_vars: xr.Dataset = data[prognostic_var_names]
-        self._boundary_vars: xr.Dataset = data[boundary_var_names]
 
         # This class will be used only for training
         total_steps: int = 2 * self.hist + 2
@@ -359,10 +364,8 @@ class TrainDataset(Dataset):
         # Normalize
         logging.info("Normalizing inputs")
         self.normalize = Normalize.get_instance()
-        self._prognostic_vars = self.normalize.normalize_prognostic(
-            self._prognostic_vars
-        )
-        self._boundary_vars = self.normalize.normalize_boundary(self._boundary_vars)
+        self._prognostic_vars = normalize(self._prog_src)
+        self._boundary_vars = normalize(self._bound_src)
 
     def __len__(self) -> int:
         return self.size
@@ -490,7 +493,7 @@ class TorchTrainDataset(Dataset):
 
     def __init__(
         self,
-        data: xr.Dataset,
+        src: DataSource,
         prognostic_var_names: PrognosticVarNames,
         boundary_var_names: BoundaryVarNames,
         wet: PrognosticMask,
@@ -507,8 +510,16 @@ class TorchTrainDataset(Dataset):
         self.stride: int = stride
 
         self.num_prognostic_channels: int = (hist + 1) * len(prognostic_var_names)
-        self._prognostic_vars: xr.Dataset = data[prognostic_var_names]
-        self._boundary_vars: xr.Dataset = data[boundary_var_names]
+        self.src = src
+        data = src.data
+        self._prog_src = src.filter(prognostic_var_names)
+        self._bound_src = src.filter(boundary_var_names)
+        self._prognostic_vars: xr.Dataset = self._prog_src.data
+        self._boundary_vars: xr.Dataset = self._bound_src.data
+
+        # cache tensors for faster access during training.
+        to_tensor(self._prog_src, device=self.device)
+        to_tensor(self._bound_src, device=self.device)
 
         # This class will be used only for training and validation
         total_steps: int = 2 * self.hist + 2
@@ -581,15 +592,21 @@ class TorchTrainDataset(Dataset):
         prognostic_all: Float[torch.Tensor, "step variable time lat lon"],
         boundary: Float[torch.Tensor, "step variable lat lon"],
     ) -> tuple[Input, Prognostic]:
-        normalize = Normalize.get_instance()
-
         # grab past steps and prep for model
         input_ = self._prep_prognostic_steps(
             prognostic_all[:, :, : self.hist + 1, :, :]
         )
 
         # add in boundary to final input
-        boundary = normalize.normalize_tensor_boundary(boundary).float()
+        _, boundary_means, boundary_stds = to_tensor(
+            self._bound_src, device=self.device
+        )
+        boundary = normalize_tensor(
+            boundary,
+            boundary_means.reshape([-1, 1, 1]),
+            boundary_stds.reshape([-1, 1, 1]),
+        ).float()
+
         boundary = torch.where(self.wet_surface, boundary, 0.0)
         total_input = torch.cat((input_, boundary), dim=1)  # dim=1 -> variables
 
@@ -600,7 +617,6 @@ class TorchTrainDataset(Dataset):
     def _prep_prognostic_steps(
         self, prognostic_steps: Float[torch.Tensor, "step variable time lat lon"]
     ) -> Input:
-        normalize = Normalize.get_instance()
         # Time needs to be before step for normalization to work (even though it would
         # make more sense for `step` to the be first dimension). Don't worry: we fix
         # this later.
@@ -609,9 +625,13 @@ class TorchTrainDataset(Dataset):
             "step variable time lat lon -> time step variable lat lon",
         )
 
+        _, prog_means, prog_stds = to_tensor(self._prog_src, device=self.device)
+
         # normalize expects variables in third dimension
-        prognostic_steps = normalize.normalize_tensor_prognostic(
-            prognostic_steps
+        prognostic_steps = normalize_tensor(
+            prognostic_steps,
+            prog_means.reshape([1, -1, 1, 1]),
+            prog_stds.reshape([1, -1, 1, 1]),
         ).float()
 
         # flatten time and variable dimensions into a set of channels for model
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -52,13 +52,11 @@ class InferenceDataset(Dataset):
         self.device = get_device()
 
         self.hist = hist
-        data = src.data
 
         self.num_prognostic_channels = (hist + 1) * len(prognostic_var_names)
+        data = src.data
         self._prognostic_src = src.filter(prognostic_var_names)
         self._boundary_src = src.filter(boundary_var_names)
-        self._prognostic_vars = self._prognostic_src.data
-        self._boundary_vars = self._boundary_src.data
         self._times = data.time
 
         time_indices = np.arange(data.time.size)
@@ -169,10 +167,10 @@ class InferenceDataset(Dataset):
         return x_index
 
     def _get_prognostic(self, x_index):
-        data_in_ds: xr.Dataset = self._prognostic_vars.isel(time=x_index).isel(
-            time=slice(None, self.hist + 1)
+        data_in_src = self._prognostic_src.map_data(
+            lambda ds: ds.isel(time=x_index).isel(time=slice(None, self.hist + 1))
         )
-        data_in_ds = self._prognostic_src.normalize(data_in_ds)
+        data_in_ds = data_in_src.normalize()
         data_in_np: np.ndarray = (
             data_in_ds.to_array()
             .transpose("window_dim", "time", "variable", "lat", "lon")
@@ -193,10 +191,10 @@ class InferenceDataset(Dataset):
         With hist > 0, the boundary condition considered is always the last step of
         the input.
         """
-        data_in_boundary_ds: xr.Dataset = self._boundary_vars.isel(time=x_index).isel(
-            time=self.hist
+        data_in_boundary_src = self._boundary_src.map_data(
+            lambda ds: ds.isel(time=x_index).isel(time=self.hist)
         )
-        data_in_boundary_ds = self._boundary_src.normalize(data_in_boundary_ds)
+        data_in_boundary_ds = data_in_boundary_src.normalize()
         data_in_boundary_np: np.ndarray = (
             data_in_boundary_ds.to_array()
             .transpose("window_dim", "variable", "lat", "lon")
@@ -207,10 +205,10 @@ class InferenceDataset(Dataset):
         return data_in_boundary
 
     def _get_label(self, x_index):
-        label_ds: xr.Dataset = self._prognostic_vars.isel(time=x_index).isel(
-            time=slice(self.hist + 1, None)
+        label_src = self._prognostic_src.map_data(
+            lambda ds: ds.isel(time=x_index).isel(time=slice(self.hist + 1, None))
         )
-        label_ds = self._prognostic_src.normalize(label_ds)
+        label_ds = label_src.normalize()
         label_np: np.ndarray = (
             label_ds.to_array()
             .transpose("window_dim", "time", "variable", "lat", "lon")
@@ -225,7 +223,9 @@ class InferenceDataset(Dataset):
         return label
 
     def get_coords_dict(self):
-        return {co: self._prognostic_vars[co] for co in self._prognostic_vars.coords}
+        return {
+            co: self._prognostic_src.data[co] for co in self._prognostic_src.data.coords
+        }
 
 
 class InferenceDatasets(Dataset):
@@ -507,8 +507,6 @@ class TorchTrainDataset(Dataset):
         data = src.data
         self._prognostic_src = src.filter(prognostic_var_names)
         self._boundary_src = src.filter(boundary_var_names)
-        self._prognostic_vars: xr.Dataset = self._prognostic_src.data
-        self._boundary_vars: xr.Dataset = self._boundary_src.data
 
         # This class will be used only for training and validation
         total_steps: int = 2 * self.hist + 2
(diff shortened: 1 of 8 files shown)


src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +30,106 @@
...
+            + _shape_hash(self.stds)
+        )
+
+    def filter(
+        self,
+        var_names: PrognosticVarNames | BoundaryVarNames,
+        *,
+        name: str | None = None,
+    ) -> Self:
+        """Filter the data source to only include the specified variables."""
```
alxmrs, line 46: In a future PR, this will make it really easy to support compact data. Something like: https://github.com/suryadheeshjith/Ocean_Emulator/pull/175/files#diff-3e496072412d1e191c8f724220b196643bbb36e110f3452931a25815161a139bR381

src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +30,106 @@
...
+            self,
+            name=name or self.name,
+            data=self.data.copy(),
+            means=self.means.copy(),
+            stds=self.stds.copy(),
+        )
+
+    @classmethod
+    def from_config(
+        cls, cfg: TrainConfig | EvalConfig, *, use_dask: bool | None = None
```
alxmrs: This argument here will make it really easy to turn on/off dask for inference/training in the future.

src/ocean_emulators/utils/data.py
```diff
@@ -289,52 +389,101 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
...
+
+
+@cache
+def to_tensor(
+    src: DataSource, device: torch.device
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Convert data source to tensors (with caching)."""
+    data_np = src.data.to_array().to_numpy().reshape(-1)
+    means_np = src.means.to_array().to_numpy().reshape(-1)
+    stds_np = src.stds.to_array().to_numpy().reshape(-1)
```
alxmrs: In the future, I think I'll be able to avoid reshaping for the sake of caching.

tests/test_datasets.py
```diff
@@ -52,37 +57,23 @@ def make_loader(
...
 
     with MultitonScope():
         TensorMap.init_instance(
             cfg.experiment.prognostic_vars_key, cfg.experiment.boundary_vars_key
         )
-        ds_, means_, stds_ = validate_data(ds, ds_means, ds_stds)
-        wet, wet_surface = extract_wet_mask(ds_, prognostic, cfg.data.hist)
-        Normalize.init_instance(means_, stds_, prognostic, boundary, wet)
+        val = validate_data(raw)
+        wet, wet_surface = extract_wet_mask(val.data, prognostic, cfg.data.hist)
```
alxmrs: In the near future, I also plan to make `extract_wet_mask` take in a `DataSource`. I didn't want to make this PR too big.

src/ocean_emulators/utils/data.py
alxmrs: Demoed in #215.

src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +30,106 @@
 from ocean_emulators.utils.multiton import Multiton
 
 
+@dataclasses.dataclass
+class DataSource:
+    """Data source for the model."""
+
+    name: str
```
jder, line 35: As far as I can see, this name is only ever used in test code. Do we need it? It kinda muddies the semantics of this class for me. For example, what does it mean to filter the datasource but not change its name? Should the dask-ness of a DataSource affect caching of the to_tensor form? etc

tests/test_datasets.py
alxmrs: Maybe this doesn't need to be refactored this way 🤔.

src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +30,106 @@
...
+
+        def _shape_hash(ds: xr.Dataset) -> int:
+            return hash(
+                "_".join(
+                    [
+                        f"{k}-{v}:{d}"
+                        for k, da in ds.items()
+                        for v, d in da.sizes.items()
+                    ]
+                )
```
```suggestion
            return hash(
                tuple(
                    (k, v, d)
                    for k, da in ds.items()
                    for v, d in da.sizes.items()
                )
            )
```

src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +30,106 @@
...
+        name: str | None = None,
+    ) -> Self:
+        """Filter the data source to only include the specified variables."""
+        data = self.data[var_names]
+        means = self.means[var_names]
+        stds = self.stds[var_names]
+
+        return dataclasses.replace(
+            self, name=name or self.name, data=data, means=means, stds=stds
+        )
```
jder, line 53: Maybe more clear to do Datasource(…) rather than dataclasses.replace since we're replacing everything?

src/ocean_emulators/utils/data.py
jder: I suggest we move the logic about loader version outside this function and make use_dask non-optional. The semantics here are pretty subtle and I'd rather the caller know there's a connection between the loader and the daskness. (Since we don't create the loader here.)

src/ocean_emulators/utils/data.py
```diff
@@ -249,52 +349,101 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
...
-    data_std: xr.Dataset,
-) -> tuple[xr.Dataset, xr.Dataset, xr.Dataset]:
+def validate_data(src: DataSource) -> DataSource:
     """
     Validate the data such that we have the correct format for training.
     """
-    data_copy = (
-        data.copy()
-        .pipe(flatten_masks)
+    src_ = src.copy(name=src.name + "_validated")
```
jder: Why not `src = `? We never want to refer to `src` again.

src/ocean_emulators/datasets.py
```diff
@@ -507,8 +514,16 @@ def __init__(
...
+        self.src = src
+        data = src.data
+        self._prog_src = src.filter(prognostic_var_names)
+        self._bound_src = src.filter(boundary_var_names)
+        self._prognostic_vars: xr.Dataset = self._prog_src.data
+        self._boundary_vars: xr.Dataset = self._bound_src.data
+
+        # cache tensors for faster access during training.
+        to_tensor(self._prog_src, device=torch.device("cpu"))
+        to_tensor(self._bound_src, device=torch.device("cpu"))
```
jder: Won't this load all of these into memory? (We should really add a test for this somehow.)

src/ocean_emulators/utils/data.py
```diff
@@ -249,52 +349,101 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
     return data_copy
 
 
-def validate_data(
-    data: xr.Dataset,
-    data_mean: xr.Dataset,
-    data_std: xr.Dataset,
-) -> tuple[xr.Dataset, xr.Dataset, xr.Dataset]:
+def validate_data(src: DataSource) -> DataSource:
```
jder, line 371: Nice!

src/ocean_emulators/datasets.py
```diff
@@ -52,10 +58,14 @@ def __init__(
...
+        self.src = src
+        data = src.data
 
         self.num_prognostic_channels = (hist + 1) * len(prognostic_var_names)
-        self._prognostic_vars = data[prognostic_var_names]
-        self._boundary_vars = data[boundary_var_names]
+        self._prog_src = src.filter(prognostic_var_names)
+        self._bound_src = src.filter(boundary_var_names)
+        self._prognostic_vars = self._prog_src.data
+        self._boundary_vars = self._bound_src.data
```
jder: I think it would probably be more clear to name this (and elsewhere) `_prognostic_src` and `_boundary_src` rather than `_prog_src` and `_bound_src` -- I don't think anywhere else in the codebase abbreviates those words that way. I also think we can probably get rid of `_prognostic_vars` and use `_prognostic_src.data` instead? The main places we use `_prognostic_vars` already assumes these are equal.

src/ocean_emulators/datasets.py
```diff
@@ -169,9 +179,8 @@ def _get_prognostic(self, x_index):
         data_in_ds: xr.Dataset = self._prognostic_vars.isel(time=x_index).isel(
             time=slice(None, self.hist + 1)
         )
-        data_in_ds = Normalize.get_instance().normalize_prognostic(
-            data_in_ds
-        )  # TODO: Weird error when I get_instance in init
+        src_in = dataclasses.replace(self._prog_src, data=data_in_ds)
+        data_in_ds = normalize(src_in)
```
jder: I love that this doesn't require the singleton. WDYT about adding a DataSource.isel_time(…) or something so we can avoid pulling the data out and then pushing it back in? Alternatively you could get rid of the existing filter/select methods and add a map_data method which can do any of these things.

src/ocean_emulators/datasets.py
```diff
@@ -52,10 +58,14 @@ def __init__(
         self.device = get_device()
 
         self.hist = hist
+        self.src = src
```
jder: We don't seem to use self.src, I suggest we delete it. (Same for TorchTrainDataset)

src/ocean_emulators/datasets.py
```diff
@@ -581,15 +596,21 @@ def _get_input_and_label(
...
-        boundary = normalize.normalize_tensor_boundary(boundary).float()
+        _, boundary_means, boundary_stds = to_tensor(
+            self._bound_src, device=boundary.device
+        )
+        boundary = normalize_tensor(
+            boundary,
+            boundary_means.reshape([-1, 1, 1]),
+            boundary_stds.reshape([-1, 1, 1]),
+        ).float()
+
```
jder, line 589: Can we return this (and below) to the smaller amount of code it was before with something like `self._boundary_src.normalize_tensor(boundary, variable_axis=0)`? The dataset can cache the tensor form of the stds/mean internally if we need to for performance, but I don't think we need to, this is all on CPU anyway.

src/ocean_emulators/eval.py
```diff
@@ -99,28 +98,9 @@ def __init__(self, cfg: EvalConfig) -> None:
...
-            os.path.join(self.data_dir, self.data_stds_path),
-            engine="netcdf4" if self.data_stds_path.endswith(".nc") else "zarr",
-            chunks={},
-        )
-
-        self.data, self.data_mean, self.data_std = validate_data(
-            data, data_mean, data_std
-        )
+        raw = DataSource.from_config(cfg)
+        val = self.src = validate_data(raw)
```
jder: I don't understand the name `val`. Maybe just use `self.src`?

src/ocean_emulators/eval.py
```diff
@@ -99,28 +98,9 @@ def __init__(self, cfg: EvalConfig) -> None:
...
+        self.data, self.data_mean, self.data_std = val.data, val.means, val.stds
```
jder: Same comment here -- we could probably use `self.src.data` in the rest of this class.

src/ocean_emulators/train.py
```diff
@@ -156,43 +157,11 @@ def __init__(self, cfg: TrainConfig) -> None:
...
-            engine="netcdf4" if self.data_stds_path.endswith(".nc") else "zarr",
-            chunks=chunks,
-        )
-
-        self.data, self.data_mean, self.data_std = validate_data(
-            data, data_mean, data_std
-        )
+        raw = DataSource.from_config(cfg)
+        val = self.src = validate_data(raw)
+        self.data, self.data_mean, self.data_std = val.data, val.means, val.stds
```
jder, line 166: Yay! So much de-duplication. Same nits here as in eval about `val` and self.data*

tests/test_datasets.py
```diff
@@ -51,37 +56,23 @@ def make_loader(
...
     prognostic = PROGNOSTIC_VARS[cfg.experiment.prognostic_vars_key]
     boundary = BOUNDARY_VARS[cfg.experiment.boundary_vars_key]
 
     with MultitonScope():
         TensorMap.init_instance(
             cfg.experiment.prognostic_vars_key, cfg.experiment.boundary_vars_key
         )
-        ds_, means_, stds_ = validate_data(ds, ds_means, ds_stds)
-        wet, wet_surface = extract_wet_mask(ds_, prognostic, cfg.data.hist)
-        Normalize.init_instance(means_, stds_, prognostic, boundary, wet)
```
jder, line 77: I think we probably still need to init Normalize since other train-related code under test might expect it to be initialized? (But I guess if they pass that's not true?)

tests/test_utils_data.py
```diff
@@ -148,13 +153,14 @@ def normalize_input():
         coords={"lat": [0], "lon": [0]},
     )
 
+    test = DataSource("test", data_mean, data_mean, data_std)
```
jder, line 158: Passing data_mean in for data seems worth a comment at least?

src/ocean_emulators/utils/data.py
```diff
@@ -249,52 +349,101 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
...
+) -> torch.Tensor:
+    unnorm = data * stds + means
+    unnorm = unnorm.to(data.dtype)
+    return unnorm
+
+
+def mask_tensor(
+    data: torch.Tensor, wet_mask: torch.Tensor, fill_value=float("nan")
+) -> torch.Tensor:
+    return torch.where(wet_mask.to(data.device) == 0, fill_value, data)
```
jder: mask_tensor and unnormalize_tensor appear to be unused

src/ocean_emulators/utils/data.py
```diff
@@ -249,52 +349,101 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
...
+    stds_np = src.stds.to_array().to_numpy().reshape(-1)
+
+    data_tensor = torch.from_numpy(data_np).to(device)
+    means_tensor = torch.from_numpy(means_np).to(device)
+    stds_tensor = torch.from_numpy(stds_np).to(device)
+
+    return data_tensor, means_tensor, stds_tensor
+
+
+def normalize(src: DataSource, fill_nan=True, fill_value=0.0) -> xr.Dataset:
```
jder: Should this be a member function on DataSource?

src/ocean_emulators/utils/data.py
```diff
@@ -249,52 +349,101 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
...
-        )
 
-    return data_copy, data_mean_copy, data_std_copy
+    out = compute_anomalies(src_, anomalies_vars) if anomalies_vars else src_
+
+    return out
+
+
+@cache
+def to_tensor(
```
jder: I think this function is pretty dangerous, especially since its name does not warn you about the fact it *permanently* caches the passed data on the given device. Looking at where it is used, I think we can remove this function altogether if you are happy with my suggestion elsewhere about caching means/stds as tensors on DataSource.

src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +30,106 @@
...
+class DataSource:
+    """Data source for the model."""
+
+    name: str
+    data: xr.Dataset
+    means: xr.Dataset
+    stds: xr.Dataset
+
+    def __hash__(self) -> int:
+        """Hashing good enough for dictionary keys."""
```
jder: Why do we need to store a DataSource as a dictionary key?

jder (changes requested):
I think this is a nice win and enables lots more stuff down the road, thanks for taking this on.

Most of my comments are small and are naming-related or unused variables/code, but the `to_tensor` stuff I think is a foot gun and worth resolving.

I have some fuzzier/hand-wavier questions about the semantics of DataSource in addition to the piece about the `name` member. I think it might be worth figuring out how to enforce some more invariants via this type. For example, is there ever a need to have a DataSource which has not had `validate_data` called on it? If no, I suggest we move that logic into the __init__ so code can rely on the fact that being handed a DataSource means it already has been validated. (Or else split into RawDataSource and DataSource.) Similarly it seems sad that given a DataSource you have no idea if it's prognostic, boundary, everything, etc. We could instead make this clear with a type parameter or subclass. This stuff doesn't need to happen right now but might be worth a quick live chat?

src/ocean_emulators/datasets.py
alxmrs: I like the idea of `map_data`! I'll come back to this comment and see how it fits in to the review.

src/ocean_emulators/utils/data.py
alxmrs: I liked keeping it around for the sake of knowing how the data was modified via the debugger. We'll get a record of what config option was opened and how it was processed in the name right now.

I was on the fence about adding a name with it filtered. WDYT if we go the other way and add names to filtered data?

> Should the dask-ness of a DataSource affect caching of the to_tensor form?

IMO, yes it should, and I believe that it does today, based on how `from_config` sets the name of the data source.

src/ocean_emulators/utils/data.py
alxmrs: This is needed for the `@cache` decorator, which uses `DataSource` as a dict key.

src/ocean_emulators/utils/data.py
alxmrs: Good point :) though, I figured that a replace would be robust to change if we later on added new fields to the data class. WDYT here? I am very persuaded by YAGNI these days.

src/ocean_emulators/utils/data.py
alxmrs: That makes sense to me. I think the name mangling is a force of habit at this point (probably one that I should break).

src/ocean_emulators/utils/data.py
alxmrs: I am really back and forth on it; I think there are good reasons to add it there. Let's talk it over?

src/ocean_emulators/utils/data.py
alxmrs: Yes, they are right now. I thought I would use these in a near-term PR, but I'm not so sure anymore. I'll delete these.

tests/test_datasets.py
alxmrs: We do need it in the `Trainer`, but no longer in a data loader code! :)

tests/test_utils_data.py
alxmrs: Great point.

src/ocean_emulators/datasets.py
alxmrs: With `map` and `map_data`, it's looking pretty slick :)

src/ocean_emulators/utils/data.py
alxmrs: Per our discussion, we're including this in the data type.

src/ocean_emulators/datasets.py
```diff
@@ -600,24 +600,20 @@ def _get_input_and_label(
...
-            prognostic_steps
+        prognostic_steps = self._prognostic_src.normalize_tensor(
+            prognostic_steps, variable_axis=2
         ).float()
 
         # flatten time and variable dimensions into a set of channels for model
         prognostic_steps = rearrange(
             prognostic_steps,
-            "time step variable lat lon -> step (time variable) lat lon",
+            "step time variable lat lon -> step (time variable) lat lon",
```
alxmrs: I think the reason I added that long comment above and couldn't figure out how to get time working in the right place was that I kept forgetting to also change this line. Better late than never!

src/ocean_emulators/datasets.py
```diff
@@ -320,10 +319,11 @@ def __init__(
         self.hist: int = hist
         self.steps: int = steps
         self.stride: int = stride
+        data = src.data
+        self._prognostic_src = src.filter(prognostic_var_names, prefix="prognostic")
+        self._boundary_src = src.filter(boundary_var_names, prefix="boundary")
```
jder: 🐑 I don't think we need these member variables?

src/ocean_emulators/datasets.py
```diff
@@ -550,17 +548,18 @@ def __getitem__(self, idx: int):
         x_index = xr.concat(x_indexes, dim="step")
 
         prognostic_all = torch.from_numpy(
-            self._prognostic_vars.isel(time=x_index)
-            .to_array()
+            self._prognostic_src.map_data(lambda ds: ds.isel(time=x_index))
+            .data.to_array()
```
jder: Sorry, my suggestion was maybe a bad one here, but if you're going to grab `.data` I don't think there's a good reason to map_data first, right? (Same with boundary just below)

src/ocean_emulators/datasets.py
```diff
@@ -600,24 +600,20 @@ def _get_input_and_label(
...
             prognostic_steps,
-            "step variable time lat lon -> time step variable lat lon",
+            "step variable time lat lon -> step time variable lat lon",
         )
 
         # normalize expects variables in third dimension
-        prognostic_steps = normalize.normalize_tensor_prognostic(
-            prognostic_steps
+        prognostic_steps = self._prognostic_src.normalize_tensor(
+            prognostic_steps, variable_axis=2
```
jder: I think now that we can pass the axis here, there's no need to rearrange just above -- we can start with normalizing along axis 1, right?

src/ocean_emulators/utils/data.py
jder: I would prefer using the constructor because it would force us to review and consider if filtering should change whatever new data members we have. But either way is fine with me.

src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +28,126 @@
...
+        """Map the function over just data in DataSource."""
+        if suffix is None:
+            suffix = func.__qualname__
+        data = func(self.data.copy())
+        return dataclasses.replace(self, name=f"{self.name}_{suffix}", data=data)
+
+    def slice(self, time: TimeConfig) -> Self:
+        """Slice the data source to only include the specified time slice."""
+        data = self.data.sel(time=time.time_slice)
+        return dataclasses.replace(self, name=f"{time=}[{self.name}]", data=data)
```
jder, line 83: Sorry, what does the `=` do in `{time=}`?

src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +28,126 @@
...
+        """Normalize input data treated as torch Tensors."""
+        device = data.device
+        reshape_vars = [1] * data.ndim
+        reshape_vars[variable_axis] = -1
+
+        # TODO(alxmrs): Do we have to reshape twice?
+        means_np = self.means.to_array().to_numpy().reshape(-1)
+        stds_np = self.stds.to_array().to_numpy().reshape(-1)
+        means = torch.from_numpy(means_np).to(device).reshape(reshape_vars)
+        stds = torch.from_numpy(stds_np).to(device).reshape(reshape_vars)
```
jder: Might consider dropping the `to(device)` so that if someone passes a non-CPU tensor here it will explode and we'll have to go solve the caching thing rather than silently being slow.

tests/test_datasets.py
```diff
@@ -44,44 +49,31 @@ def inference_loader_pair(trainer_pair: TrainPair) -> tuple[TrainConfig, DataLoa
...
 def make_loader(
     cfg,
-    time_slice: slice | None = None,
+    time_slice: TimeConfig | None = None,
     drop_last: bool = True,
     version: LoaderVersion = LoaderVersion.OM4_EAGER,
 ) -> Generator[DataLoader, None, None]:
     if time_slice is None:
-        time_slice = cfg.train.time_slice
+        time_slice = cfg.train
```
jder: maybe `time_config` since it's not a slice?

src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +28,126 @@
...
+            suffix = func.__qualname__
+        data = func(self.data.copy())
+        return dataclasses.replace(self, name=f"{self.name}_{suffix}", data=data)
+
+    def slice(self, time: TimeConfig) -> Self:
+        """Slice the data source to only include the specified time slice."""
+        data = self.data.sel(time=time.time_slice)
+        return dataclasses.replace(self, name=f"{time=}[{self.name}]", data=data)
+
+    def normalize(self, fill_nan=True, fill_value=0.0) -> xr.Dataset:
```
jder, line 85: Maybe `normalized_data` or `data_normalized` to distinguish more clearly from `normalize_tensor`? (ie it's not just tensor vs not, it's also that the data source's own data is returned as opposed to the dat passed in.)

src/ocean_emulators/utils/data.py
```diff
@@ -27,6 +28,126 @@
...
+
+        root = cfg.experiment.data_dir
+
+        if "*" in cfg.data.data_path:
+            kwargs: dict[str, Any] = dict(
+                engine="netcdf4", chunks={"time": 1, "lat": 180, "lon": 360}
+            )
+        else:
+            kwargs = dict(chunks=chunks, consolidated=True)
+        data = xr.open_dataset(root / cfg.data.data_path, **kwargs)
```
jder: FWIW I find this harder to read than the other version was (and less type safe since it can't validate argument types)

jder (approved):
Thanks for all the improvements! Just a few small nits from me now, feel free to take or leave what you like.

Looks like you decided against moving validate_data into the DataSource initializer. I still think we should do this but I'm ok with deferring for now, but can you please remove the name `val` from all the places you call `validate_data`? It is commonly used in the codebase to mean "validation" as opposed to "train" or "inference" data.

src/ocean_emulators/datasets.py
alxmrs: We do use them below.

src/ocean_emulators/datasets.py
alxmrs: Yes, fair point!

src/ocean_emulators/datasets.py
jder: I meant that they can be locals rather than member variables.

alxmrs: These fields are used in the methods `_get_prognostic` and `_get_boundary`.

jder: Not on this (`TrainDataset`) class.

alxmrs: Ah, thank you. I was confused reading the diff, I thought this was the `InferenceDataset` class!

src/ocean_emulators/utils/data.py
alxmrs: https://stackoverflow.com/a/59661963

This is short-hand for `time={time}` in fstrings.

src/ocean_emulators/utils/data.py
alxmrs: That's a good point; this is a slightly different API surface. WDYT of this new name?

alxmrs: (sorry, specifically: `norm_of` below)

src/ocean_emulators/utils/data.py
jder: Thanks! TIL.

PR #212, opened 2025-04-15
Vectorizing the Torch Loader

### LINKED ISSUE #205: Vectorize torch loader

alxmrs:
Fixes #205.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -741,7 +741,7 @@ class TorchTrainDataset(Dataset):
             "step variable time lat lon -> time step variable lat lon",
         )
 
-        # normalize expects variables in second dimension
+        # normalize expects variables in third dimension
         prognostic_steps = normalize.normalize_tensor_prognostic(
             prognostic_steps
         ).float()
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -726,7 +726,7 @@ class TorchTrainDataset(Dataset):
         # add in boundary to final input
         boundary = normalize.normalize_tensor_boundary(boundary).float()
         boundary = torch.where(self.wet_surface, boundary, 0.0)
-        total_input = torch.cat((input_, boundary), dim=1)
+        total_input = torch.cat((input_, boundary), dim=1)  # dim=1 -> variables
 
         # grab future steps, repeat as we do for input
         label = self._prep_prognostic_steps(prognostic_all[:, :, self.hist + 1 :, :, :])
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -736,6 +736,9 @@ class TorchTrainDataset(Dataset):
         self, prognostic_steps: Float[torch.Tensor, "step variable time lat lon"]
     ) -> Input:
         normalize = Normalize.get_instance()
+        # Time needs to be before step for normalization to work (even though it would
+        # make more sense for `step` to the be first dimension). Don't worry: we fix
+        # this later.
         prognostic_steps = rearrange(
             prognostic_steps,
             "step variable time lat lon -> time step variable lat lon",
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -681,8 +681,8 @@ class TorchTrainDataset(Dataset):
 
     def __getitem__(self, idx: int):
         TD = TrainData(self.num_prognostic_channels)
-        windows = [self._get_x_index(idx, step) for step in range(self.steps)]
-        x_index = xr.concat(windows, dim="step")
+        x_indexes = [self._get_x_index(idx, step) for step in range(self.steps)]
+        x_index = xr.concat(x_indexes, dim="step")
 
         prognostic_all = torch.from_numpy(
             self._prognostic_vars.isel(time=x_index)
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -681,30 +681,32 @@ class TorchTrainDataset(Dataset):
 
     def __getitem__(self, idx: int):
         TD = TrainData(self.num_prognostic_channels)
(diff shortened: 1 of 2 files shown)


suryadheeshjith:
For testing, do you want to merge #188 here?

src/ocean_emulators/datasets.py
```diff
@@ -653,13 +653,13 @@ def __init__(
...
         indices = np.arange(num_windows)
-        indices_da = xr.DataArray(indices, dims=["window_dim"])
+        indices_da = xr.DataArray(indices, dims=["step"])
 
         # Create window dimension
         window_dim = xr.DataArray(np.arange(total_steps), dims=["time"])
 
         # Construct rolling indices
-        self.rolling_indices: Float[xr.DataArray, "window_dim time"] = (
+        self.rolling_indices: Float[xr.DataArray, "step time"] = (
```
suryadheeshjith: Do we want to save the types in constants.py or use any existing type alias from there?

src/ocean_emulators/datasets.py
```diff
@@ -681,61 +681,64 @@ def __len__(self) -> int:
...
-                .to_numpy()
-            )
-            boundary = torch.from_numpy(
-                self._boundary_vars.isel(time=x_index)
-                .isel(time=self.hist, drop=True)
-                .to_array()
-                .transpose("variable", "lat", "lon")
-                .to_numpy()
-            )
+        x_indexes = [self._get_x_index(idx, step) for step in range(self.steps)]
```
suryadheeshjith, line 684: 🚀

alxmrs:
> For testing, do you want to merge https://github.com/suryadheeshjith/Ocean_Emulator/pull/188 here?

That's a good question. I'm happy to wait to merge this PR until 188 is landed. I'll add my review there to try to get that in faster.

alxmrs:
Actually, what do you think about us deleting the lazy loader before merging in these new tests? See #213.

suryadheeshjith:
Cool, I will wait for that to land first!

src/ocean_emulators/datasets.py
```diff
@@ -653,13 +653,13 @@ def __init__(
 
         # Create base indices
         indices = np.arange(num_windows)
-        indices_da = xr.DataArray(indices, dims=["window_dim"])
+        indices_da = xr.DataArray(indices, dims=["step"])
```
jder: I think this is more confusing with the name `step` here. The index in this list has logically nothing to do with the use of "step" in `__getitem__` or elsewhere in the code, which is an index into the range `[0, self.steps)` within a single timestep's rollout. 

I think the fact that `_get_x_index` below takes `step: int` and then indexes this array with `step=window_index` (and `step=step` would be a bad bug) is a pretty good indication that these are not the same thing and should not be named the same way. 

Perhaps instead of `window_dim` you'd prefer `window_index` since it goes from 0 to num_windows?

src/ocean_emulators/datasets.py
```diff
@@ -681,61 +681,64 @@ def __len__(self) -> int:
...
+        self, prognostic_steps: Float[torch.Tensor, "step variable time lat lon"]
     ) -> Input:
         normalize = Normalize.get_instance()
         prognostic_steps = rearrange(
             prognostic_steps,
-            "variable time lat lon -> time variable lat lon",
+            "step variable time lat lon -> time step variable lat lon",
         )
 
         # normalize expects variables in second dimension
```
```suggestion
        # normalize expects variables in third dimension
```

src/ocean_emulators/datasets.py
```diff
@@ -681,61 +681,64 @@ def __len__(self) -> int:
...
 
     def _prep_prognostic_steps(
-        self, prognostic_steps: Float[torch.Tensor, "variable time lat lon"]
+        self, prognostic_steps: Float[torch.Tensor, "step variable time lat lon"]
     ) -> Input:
         normalize = Normalize.get_instance()
         prognostic_steps = rearrange(
             prognostic_steps,
-            "variable time lat lon -> time variable lat lon",
+            "step variable time lat lon -> time step variable lat lon",
```
jder, line 744: Not sure it matters, but I think `step time variable lat lon` makes more sense to me? `step` should logically always be first, right?

jder (approved):
Thank you! This is great, just a couple understandability suggestions from me.

src/ocean_emulators/datasets.py
alxmrs: I am not sure if this is something that will be used more than once in the long term.

src/ocean_emulators/datasets.py
alxmrs: This is a fair point. Indeed, I almost tripped myself up and introduced the kind of bug you predict around `step`. Instead of `window_dim`, I think I prefer just `window`. I'm happy to change it.

src/ocean_emulators/datasets.py
alxmrs: Yeah, that makes plenty of sense.

src/ocean_emulators/datasets.py
alxmrs: Just to be clear: we still want to keep `step` in the `__getitem__` call?

src/ocean_emulators/datasets.py
alxmrs: Ok! So, I just found out by breaking tests that step needs to be in this position! Instead of changing it, I've just added a comment explaining why this is there.

src/ocean_emulators/datasets.py
alxmrs: Happy to rename this later on if I misunderstood; will proceed with landing.

src/ocean_emulators/datasets.py
jder: Sorry, I don't understand the comment. Why does time need to be before step for normalization to work?

src/ocean_emulators/datasets.py
jder: Yep! What you landed LGTM. Thanks!

PR #213, opened 2025-04-15
Removed the Lazy Loader from the codebase.

### LINKED ISSUE #207: Delete lazy loader (for now)

alxmrs:
Fixes #207.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -5,7 +5,7 @@ import numpy as np
 import torch
 import xarray as xr
 from einops import rearrange
-from jaxtyping import Float, Integer
+from jaxtyping import Float
 from torch.utils.data import Dataset
 from xarray_einstats.einops import rearrange as xr_rearrange  # noqa: F401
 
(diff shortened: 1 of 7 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -20,149 +20,10 @@
     PrognosticMask,
     PrognosticVarNames,
 )
-from ocean_emulators.utils.data import Normalize, mask, unflatten_masks
```
jder, line 23: 🐑 looks like we could delete mask and unflatten_masks now?

jder (approved):
Thanks! Nothing makes me happier than deleting code. Glad we built this and learned a bunch from it and excited to move forward.

src/ocean_emulators/datasets.py
alxmrs: That's a good point. Thanks for pointing that out.

alxmrs: I decided to keep `unflatten_masks` because it's nice to have round-trip tests. `mask`, however, is out.

alxmrs:
I definitely learned a lot about the codebase from writing this!

PR #214, opened 2025-04-15
Adding types and type-specific variable names to InferenceDataset 

alxmrs:
This addresses a number of flaky mypy errors in datasets.py.


jder (approved):
To be honest, it drives me a bit crazy that this is the only way to have reasonable types here. Compared to the previous version, it's now much easier to introduce a bug where you accidentally refer to an old/wrong version of `data_in`. But short of rewriting the whole thing as a series of pipes I think this is probably best.

PR #215, opened 2025-04-16
Torch Loader works with compact data.

### LINKED ISSUE #204: Apply level compaction to torch loader

alxmrs:
This PR depends on #211. Fixes #204.

Here is the WandB profile results: https://wandb.ai/m2lines/ocean-emulators/reports/Compaction-and-Vectorization-on-Data-Load-Time--VmlldzoxMjg1MDE4NQ

Conclusions: compaction is at least as fast as before, but often predictably faster.

### DIFF
--- a/scripts/clone_data.py
+++ b/scripts/clone_data.py
@@ -11,13 +11,16 @@
 #   "numcodecs>=0.15",
 #   "distributed",
 #   "tenacity",
+#   "ocean-emulators",
 # ]
+#
+# [tool.uv.sources]
+# ocean-emulators = { path = "../" }
 # ///
 
 import argparse
 import os
 import pathlib
-from collections import defaultdict
 
 import dask
 import dask.diagnostics
@@ -25,38 +28,14 @@ import xarray as xr
 from dask.distributed import LocalCluster
 from tenacity import retry
 
+from ocean_emulators.utils.data import compact_dataset
+
 DATA_ROOT = "https://nyu1.osn.mghpcc.org/m2lines-pubs/Samudra/"
 
 
 @retry
 def robust_open_dataset(target: str, **kwargs) -> xr.Dataset:
     return xr.open_dataset(target, **kwargs)
-
-
-def compact_dataset(ds: xr.Dataset) -> xr.Dataset:
-    data = ds.copy()
-
-    var_groups = defaultdict(list)
-    for key in data.keys():
-        if "_lev_" in (k := str(key)):
-            base_name = k.split("_lev_")[0]
-            var_groups[base_name].append(k)
-
-    def _parse_level(x) -> float:
-        return float(x.split("_lev_")[1].replace("_", "."))
-
-    for base_var, vars_ in var_groups.items():
-        sorted_vars = sorted(vars_, key=_parse_level)
-        levels = [_parse_level(var) for var in sorted_vars]
-        if hasattr(data, "lev"):
-            levels = data.lev.values
-        da = xr.concat([data[var] for var in sorted_vars], dim="lev").assign_coords(
-            lev=("lev", levels)
-        )
-        data[base_var] = da
-        data = data.drop_vars(vars_)
-
-    return data
 
 
 def main(args: argparse.Namespace) -> None:
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -22,7 +22,7 @@ from ocean_emulators.constants import (
     PrognosticMask,
     PrognosticVarNames,
 )
-from ocean_emulators.utils.data import DataSource, LoadStats
+from ocean_emulators.utils.data import DataSource, LoadStats, conditional_rearrange
 from ocean_emulators.utils.device import get_device, using_gpu
 from ocean_emulators.utils.logging import elapsed
 
@@ -184,11 +184,28 @@ class InferenceDataset(Dataset):
             data_in_ds = data_in_src.normalize()
         else:
             data_in_ds = data_in_src.data
-        data_in_np: np.ndarray = (
-            data_in_ds.to_array()
-            .transpose("window_dim", "time", "variable", "lat", "lon")
-            .to_numpy()
-        )
+
+        if data_in_src.is_compact:
+            data_in_np: np.ndarray = (
+                conditional_rearrange(
+                    data_in_ds,
+                    "window_dim (time variable lev)=var lat lon",
+                    concat_dim="var",
+                )
+                .rename({"var": "variable"})
+                .to_numpy()
+            )
+        else:
+            data_in_np = (
+                data_in_ds.to_array()
+                .transpose("window_dim", "time", "variable", "lat", "lon")
+                .to_numpy()
+            )
+            data_in_np = rearrange(
+                data_in_np,
+                "window_dim time variable lat lon -> "
+                "window_dim (time variable) lat lon",
+            )
         data_in: torch.Tensor = torch.from_numpy(data_in_np).float()
         data_in = torch.where(self.wet, data_in, self.masked_fill_value)
         if not self.normalize_before_mask:
@@ -213,11 +230,22 @@ class InferenceDataset(Dataset):
             data_in_boundary_ds = data_in_boundary_src.normalize()
         else:
             data_in_boundary_ds = data_in_boundary_src.data
-        data_in_boundary_np: np.ndarray = (
-            data_in_boundary_ds.to_array()
-            .transpose("window_dim", "time", "variable", "lat", "lon")
-            .to_numpy()
-        )
+        if data_in_boundary_src.is_compact:
+            data_in_boundary_np: np.ndarray = (
+                conditional_rearrange(
+                    data_in_boundary_ds,
+                    "window_dim time (variable lev)=var lat lon",
+                    concat_dim="var",
+                )
+                .rename({"var": "variable"})
+                .to_numpy()
+            )
+        else:
+            data_in_boundary_np = (
+                data_in_boundary_ds.to_array()
+                .transpose("window_dim", "time", "variable", "lat", "lon")
+                .to_numpy()
+            )
         data_in_boundary: torch.Tensor = torch.from_numpy(data_in_boundary_np).float()
         data_in_boundary = torch.where(
             self.wet_surface, data_in_boundary, self.masked_fill_value
@@ -240,19 +268,32 @@ class InferenceDataset(Dataset):
             label_ds = label_src.normalize()
         else:
             label_ds = label_src.data
-        label_np: np.ndarray = (
-            label_ds.to_array()
-            .transpose("window_dim", "time", "variable", "lat", "lon")
-            .to_numpy()
-        )
-        label: torch.Tensor = torch.from_numpy(label_np).float()
+        if label_src.is_compact:
+            label_np: np.ndarray = (
+                conditional_rearrange(
+                    label_ds,
+                    "window_dim (time variable lev)=var lat lon",
+                    concat_dim="var",
+                )
+                .rename({"var": "variable"})
+                .to_numpy()
+            )
+        else:
+            label_np = (
+                label_ds.to_array()
+                .transpose("window_dim", "time", "variable", "lat", "lon")
+                .to_numpy()
+            )
+            label: torch.Tensor = torch.from_numpy(label_np).float()
         label = torch.where(self.wet, label, self.masked_fill_value)
         if not self.normalize_before_mask:
             label = self._prognostic_src.normalize_with(label, variable_axis=2)
         label = rearrange(
-            label,
-            "window_dim time variable lat lon -> window_dim (time variable) lat lon",
-        )
+                label,
+                "window_dim time variable lat lon -> "
+                "window_dim (time variable) lat lon",
+            )
+
         return label
 
     def get_coords_dict(self):
@@ -618,27 +659,42 @@ class TorchTrainDataset(Dataset):
     def __getitem__(self, idx: int):
         start_time = time.perf_counter()
         TD = TrainData(self.num_prognostic_channels)
+        x_indexes = [self._get_x_index(idx, step) for step in range(self.steps)]
+        x_index = xr.concat(x_indexes, dim="step")
 
-        for step in range(self.steps):
-            x_index = self._get_x_index(idx, step)
+        if self._prognostic_src.is_compact:
+            prognostic_all = torch.from_numpy(
+                conditional_rearrange(
+                    self._prognostic_src.data.isel(time=x_index),
+                    "(variable lev)=var time lat lon",
+                    concat_dim="var",
+                )
+                .rename({"var": "variable"})
+                .to_numpy()
+            )
+        else:
             prognostic_all = torch.from_numpy(
                 self._prognostic_src.data.isel(time=x_index)
                 .to_array()
-                .transpose("time", "variable", "lat", "lon")
+                .transpose(
+                    "step", "time", "variable", "lat", "lon"
+                )
                 .to_numpy()
             )
-            boundary = torch.from_numpy(
-                self._boundary_src.data.isel(time=x_index)
-                .to_array()
-                .transpose("time", "variable", "lat", "lon")
-                .to_numpy()
-            )
+        boundary = torch.from_numpy(
+            self._boundary_src.data.isel(time=x_index)
+            .isel(time=self.hist, drop=True)
+            .to_array()
+            .transpose("step", "variable", "lat", "lon")
+            .to_numpy()
+        )
 
-            input_, label = self._get_input_and_label(prognostic_all, boundary)
+        input_, label = self._get_input_and_label(prognostic_all, boundary)
 
+        for i in range(input_.shape[0]):
             TD.insert(
-                input_=input_,
-                label=label,
+                input_=input_[i],
+                label=label[i],
             )
 
         TD.load_stats = LoadStats(time.perf_counter() - start_time)
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -1,4 +1,3 @@
-import dataclasses
 import logging
 import time
 from typing import Any
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -59,11 +59,9 @@ class InferenceDataset(Dataset):
         self.device = get_device()
 
         self.hist = hist
-        self.src = src
         data = src.data
 
         self.num_prognostic_channels = (hist + 1) * len(prognostic_var_names)
-        data = src.data
         self._prognostic_src = src.filter(prognostic_var_names, prefix="prognostic")
         self._boundary_src = src.filter(boundary_var_names, prefix="boundary")
         self._times = data.time
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -1,3 +1,4 @@
+import dataclasses
 import logging
 import time
 from typing import Any
@@ -59,6 +60,8 @@ class InferenceDataset(Dataset):
         self.device = get_device()
 
         self.hist = hist
+        self.src = src
+        data = src.data
 
         self.num_prognostic_channels = (hist + 1) * len(prognostic_var_names)
         data = src.data
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -716,12 +716,13 @@ class TorchTrainDataset(Dataset):
             tensor: torch.Tensor,
             src: DataSource,
             mask: torch.Tensor,
+            variable_axis: int = 2,
         ) -> torch.Tensor:
             if self.normalize_before_mask:
-                tensor = src.normalize_with(tensor, variable_axis=1).float()
+                tensor = src.normalize_with(tensor, variable_axis=variable_axis).float()
             tensor = torch.where(mask, tensor, self.masked_fill_value)
             if not self.normalize_before_mask:
-                tensor = src.normalize_with(tensor, variable_axis=1).float()
+                tensor = src.normalize_with(tensor, variable_axis=variable_axis).float()
             return tensor
 
         prognostic_steps = normalize_and_mask(
@@ -729,7 +730,7 @@ class TorchTrainDataset(Dataset):
         )
         if boundary_steps is not None:
             boundary_steps = normalize_and_mask(
-                boundary_steps, self._boundary_src, self.wet_surface
+                boundary_steps, self._boundary_src, self.wet_surface, variable_axis=1
             )
 
         # Flatten time and variable dimensions
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -59,9 +59,9 @@ class InferenceDataset(Dataset):
         self.device = get_device()
 
         self.hist = hist
-        data = src.data
 
         self.num_prognostic_channels = (hist + 1) * len(prognostic_var_names)
+        data = src.data
         self._prognostic_src = src.filter(prognostic_var_names, prefix="prognostic")
         self._boundary_src = src.filter(boundary_var_names, prefix="boundary")
         self._times = data.time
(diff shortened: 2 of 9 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -321,15 +322,27 @@ def _get_prognostic(self, x_index):
...
-            .transpose("window_dim", "time", "variable", "lat", "lon")
-            .to_numpy()
-        )
-        data_in_np = rearrange(
-            data_in_np,
-            "window_dim time variable lat lon -> window_dim (time variable) lat lon",
-        )
+        if src_in.is_compact:
+            data_in_np: np.ndarray = (
+                conditional_rearrange(
```
alxmrs, line 187: Once #165 is landed, we'll no longer need this `conditional_rearrange()` function.

src/ocean_emulators/utils/data.py
```diff
@@ -59,9 +94,12 @@ def map(
             tuple[xr.Dataset, xr.Dataset, xr.Dataset],
         ],
         *,
-        suffix: str,
+        suffix: str | None = None,
```
alxmrs, line 120: This is unrelated to compaction; I just thought it was nice-to-have.

src/ocean_emulators/utils/data.py
```diff
@@ -37,20 +38,54 @@ class DataSource:
     means: xr.Dataset
     stds: xr.Dataset
 
+    @cached_property
+    def is_compact(self) -> bool:
+        """Check if the data source is compact."""
+        return all(
+            v in d
+            for v in ["so", "uo", "vo", "thetao"]
```
jder: Is there something in constants.py we could use instead of hard-coding these here?

src/ocean_emulators/utils/data.py
```diff
@@ -37,20 +38,54 @@ class DataSource:
...
+            for d in [self.data, self.means, self.stds]
+        )
+
     def filter(
         self,
         var_names: PrognosticVarNames | BoundaryVarNames,
         *,
         prefix: str,
     ) -> Self:
         """Filter the data source to only include the specified variables."""
```
jder: Maybe we could use a comment or expanded doc string about the behavior of this function (or class?) with compact data? Looking at the code it seems like it expects that callers might pass in var_names with some (but not all) levels? Presumably we require that the set of levels is the same across all vars?

src/ocean_emulators/utils/data.py
```diff
@@ -59,9 +94,12 @@ def map(
...
         ],
         *,
-        suffix: str,
+        suffix: str | None = None,
     ) -> Self:
         """Map the function over the data source."""
+        if suffix is None:
+            suffix = func.__qualname__
+
         data, means, stds = func(self.data.copy(), self.means.copy(), self.stds.copy())
```
jder, line 126: So given there's no change here, this means that map bodies (and map_data bodies) must be aware that data may or may not be compact, right? Maybe this is too big of a change, but would it be possible to somehow hide the compact-vs-not distinction as part of this class's API?

src/ocean_emulators/utils/data.py
```diff
@@ -150,6 +211,30 @@ def from_config(cls, cfg: TrainConfig | EvalConfig, *, use_dask: bool) -> Self:
         )
 
 
+def conditional_rearrange(
+    data: xr.Dataset, pattern: str, except_dim="lev", concat_dim="variable"
+) -> xr.DataArray:
+    assert except_dim in pattern, f"{except_dim} must be in the pattern."
+    data_ = data.copy()
```
jder: I don't think you need this copy.

src/ocean_emulators/utils/data.py
```diff
@@ -150,6 +211,30 @@ def from_config(cls, cfg: TrainConfig | EvalConfig, *, use_dask: bool) -> Self:
         )
 
 
+def conditional_rearrange(
+    data: xr.Dataset, pattern: str, except_dim="lev", concat_dim="variable"
+) -> xr.DataArray:
```
jder, line 271: I think a doc comment would help with understandability here.

src/ocean_emulators/utils/data.py
```diff
@@ -370,8 +455,11 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
 
 def validate_data(src: DataSource) -> DataSource:
     """Validate the data such that we have the correct format for training."""
+    if src.is_compact:
+        src_ = src.map_data(with_lat_lon_coords)
+        return compute_anomalies(src_, ("hfds_anomalies",))
```
jder: Can we pull this from TensorMap rather than hard-coding?

jder (commented):
OK, sorry, I didn't get a chance to look carefully at everything but I have to run so leaving a few comments here and will return to this tonight or tomorrow.

src/ocean_emulators/utils/data.py
alxmrs: Found a better way.

src/ocean_emulators/utils/data.py
alxmrs: I kind of have an idea of what that could look like, but I actually like the status quo right now. Some advice that I really like is to "pull ifs up and to push fors down"[1]. The way `is_compact` is used now means we change the `map` calls instead of having a more complicated `map` call that has two sections. Or, am I misunderstanding how this would change? In any case: the `validate_data` function is a good example of this, at a higher level, we do A vs B based on whether the data is compact. 


[1]: https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

src/ocean_emulators/datasets.py
```diff
@@ -58,6 +58,8 @@ def __init__(
         self.device = get_device()
 
         self.hist = hist
+        self.src = src
+        data = src.data
```
jder: self.src appears to be unused, and data is assigned 2 lines below, so I don't think we want either of these lines of code?

src/ocean_emulators/utils/data.py
jder: I see the value of the general guideline but at the moment the fact that this class mostly exposes (rather than hides) the compactness of the data means the conditional_rearrange call is duplicated 4 times outside this class and it's pretty hard for me to tell if it's covering all the places it needs to. Anyone that accesses the .data member might need be edited depending on compactness. (For example, do we need to change InferenceDataset.get_coords_dict?) As opposed to, say, the `filter` method which works the same regardless of the compactness of the data. 

Anyway, if you want to leave it that's ok for this PR, but I do feel like we will end up wanting to centralize the knowledge of what structure the data has down the road.

src/ocean_emulators/utils/data.py
```diff
@@ -154,6 +226,52 @@ def from_config(cls, cfg: TrainConfig | EvalConfig, *, use_dask: bool) -> Self:
...
+    assert except_dim in pattern, f"{except_dim} must be in the pattern."
+
+    vars_with_dim = [v for v in data if except_dim in data[v].dims]
+    vars_without_dim = [v for v in data if except_dim not in data[v].dims]
+
+    data_with_dim = (
+        data[vars_with_dim]
+        .to_array()
+        .einops.rearrange(pattern, dask="allowed")
+        .drop_vars(concat_dim, errors="ignore")
```
jder, line 323: Why do we drop_vars?

src/ocean_emulators/utils/data.py
```diff
@@ -38,20 +40,64 @@ class DataSource:
     means: xr.Dataset
     stds: xr.Dataset
 
+    @cached_property
+    def is_compact(self) -> bool:
+        """Check if the data source is compact."""
+        standard_vars = list(DEFAULT_METADATA.keys())[:3]
```
jder: Reading the first 3 keys of a dictionary intended for another purpose seems pretty fragile to me, and for example seems like it will break if we're training on the `thetao_surface` variable set since that will exclude so/uo. How about checking for `_lev_` in the data var names? That or else passing is_compact into the constructor instead?

src/ocean_emulators/utils/data.py
```diff
@@ -38,20 +40,64 @@ class DataSource:
...
+        """
+        name = f"{prefix}[{self.name}]"
+        if self.is_compact:
+            parsed_var_names, levels = [], []
+            for mangled_var_name in var_names:
+                if "_" not in mangled_var_name:
+                    parsed_var_names.append(mangled_var_name)
+                    continue
+                tokens = mangled_var_name.split("_")
+                var_name, level = tokens[0], int(tokens[1])
```
jder, line 88: 🐑 Should we be asserting `var_name == "lev"` here?

jder (changes requested):
Thanks, generally looks good, left a few comments. What do you think about including a test that the compact data produces the same data as the non-compact for the torch loader?

src/ocean_emulators/utils/data.py
alxmrs: It's because of a weird xarray implementation detail where the concat dim is considered both a dimension and a variable. If I don't call this, then it causes errors when concating IIRC.

src/ocean_emulators/utils/data.py
alxmrs: I don't think so; `lev` will likely be a dimension on this var, but we want to know what the variable name will be.

src/ocean_emulators/utils/data.py
alxmrs: You bright up really good points. I see how this API has it both ways (between `filter` and `map`). I'll see if I can spot a pattern in how `conditional_rearrange` is used and create a better abstraction. If I can't, I'm happy to reconsider this later on as we learn more. I am worried that it may be too early to generalized right now.

alxmrs: Update! I had to change calls to `is_compact` just now to fix issues with test. I learned that `conditional_rearrange` isn't necessarily tied to compact data, and actually is tied to having a `lev` dimension. Now, the use of `is_compact` doesn't seem to have the external/internal difference. 

In any case, I think this shows that there is more discovery to be had in how these concepts are used.

alxmrs:
> What do you think about including a test that the compact data produces the same data as the non-compact for the torch loader?

I really like this idea, and I'm looking into a way of doing it that doesn't make our pytest configuration too complicated. Do you have any ideas for this? I'm still research at the moment. I think your new config changes make this kind of thing much easier to do, but the single `data_source` fixture makes this type of combination really hard. Maybe, the simplest think to do would be to make each data source into its own fixture, and then combine them into a "super" fixture. If you like that approach, then I think it's best to do that in a new PR.

src/ocean_emulators/utils/data.py
jder: Ah, right, sorry I misread.

jder:
> > What do you think about including a test that the compact data produces the same data as the non-compact for the torch loader?

> I really like this idea, and I'm looking into a way of doing it that doesn't make our pytest configuration too complicated.

I think probably I would avoid injecting fixtures for this and build two train_configs (and data_sources) by calling those fixtures as functions (or breaking their core logic out into a separate function) and then using make_loader on both. If you want to do some refactoring here in a separate PR to make that easy, that makes sense to me -- do you want to do that first? (To be clear, I'm fine either way.)

src/ocean_emulators/utils/data.py
```diff
@@ -38,20 +39,66 @@ class DataSource:
...
     stds: xr.Dataset
 
+    @cached_property
+    def is_compact(self) -> bool:
+        """Check if the data source is compact."""
+        return all(
+            "_" not in str(v)
+            for d in [self.data, self.means, self.stds]
+            for v in d.keys()
+            if "anom" not in str(v)
```
jder: This feels like this is going to bite us if we are doing this heuristically rather than having it be a property read from TensorMap, some clearer property of the data variables themselves (eg we mark them with `_lev_` in their name), or the compactness is passed in to the data source. For example, #224 in introduces some new data variables with underscores in their names.

alxmrs:
> Anyway, if you want to leave it that's ok for this PR, but I do feel like we will end up wanting to centralize the knowledge of what structure the data has down the road.

This is a fair point, but I want to lean putting this off till later mainly because I hope that soon, we will just use compact data and not have to support the non-standard Zarr store in the first place.

src/ocean_emulators/utils/data.py
alxmrs: Reading the tensormap sources now, it looks like they also partition on `"_"` to parse level variables. I agree heuristics are fragile, but I think this one may be good enough for now, until we can replace everything with compact data.

src/ocean_emulators/datasets.py
alxmrs: I found out that this is not quite true!

src/ocean_emulators/utils/data.py
```diff
@@ -156,6 +231,89 @@ def from_config(cls, cfg: TrainConfig | EvalConfig, *, use_dask: bool) -> Self:
...
+    # In the `concat` above, we put all the `data_without_dim` vars at the end. Some of
+    # these need to be moved to the front, and the rest stays at the back. Here, we
+    # compute a list of indices that will sort the data in the correct order.
+    order = np.concatenate(
+        (
+            np.roll(np.arange(n_center + n_front), n_front),  # Moves vars to the front
+            np.arange(n_center + n_front, n_front + n_center + n_back),  # rest of vars
+        )
+    )
+    order_da = xr.DataArray(order, dims=concat_dim)
```
alxmrs: @jder The test you asked for revealed a bug that this new block of code fixes: This rearrange also changes the normal order of variables. I don't think it would have affected any ML code, but now we can be sure the compact data is exactly equal to the data before. 

Thanks thinking of writing a test for this!

tests/test_datasets.py
```diff
@@ -393,6 +401,59 @@ def test_new_loaders__are_equal_to_v1_data_loader(train_config, loader_version):
             )
 
 
+@pytest.mark.parametrize("data_source", ["remote-om4"], indirect=True)
+def test_compact_loader__equals_flat_loader(
```
alxmrs, line 399: This was my KISS approach to have both remote and compact data in a test. It involves a bit of boilerplate, but I am happy that I don't have to boil the ocean to test these two data sources.

src/ocean_emulators/utils/data.py
jder: Ok, I think given #224 has landed there could now be a `sea_surface_fraction` in the data keys. How about looking for `_[0-9]+$` or something if you don't want to do a larger refactoring? (And probably the same thing needs to happen in filter below.)

src/ocean_emulators/utils/data.py
jder: What's the difference between `is_compact` and having a `lev` dimension?

src/ocean_emulators/utils/data.py
```diff
@@ -156,6 +234,90 @@ def from_config(cls, cfg: TrainConfig | EvalConfig, *, use_dask: bool) -> Self:
...
+    # In the `concat` above, we put all the `data_without_dim` vars at the end. Some of
+    # these need to be moved to the front, and the rest stays at the back. Here, we
+    # compute a list of indices that will sort the data in the correct order.
+    order = np.concatenate(
+        (
+            # Moves vars to the front
+            np.roll(np.arange((new_front := n_center + n_front)), n_front),
+            np.arange(new_front, new_front + n_back),  # rest of vars
+        )
+    )
```
jder, line 350: Not sure I'm reading this correctly, but is this equivalent to `order = [all_vars.index(v) for v in [vars_with_dim + vars_without_dim]]`? I'm a bit confused why we don't end up using the values of front/back right now (just their lengths).

src/ocean_emulators/utils/data.py
```diff
@@ -393,7 +559,11 @@ def _static_data_checks(data):
 
         src = src.map_data(_static_data_checks, suffix="static_data_checked")
 
-    def _rename(data, means, stds):
+    if src.is_compact:
+        src_ = src.map_data(with_lat_lon_coords)
+        return compute_anomalies(src_, anomalies_vars)
```
jder: :sheep: Might be slightly nicer to compute anomalies once at the end of this function and put the validation in an else rather than an early return?

tests/conftest.py
```diff
@@ -269,6 +270,32 @@ def decode(cls, da: xr.DataArray) -> tuple[Self, NDArray[np.uint]]:
...
+        if hasattr(data, "lev"):
+            levels = data.lev.values
+        da = xr.concat([data[var] for var in sorted_vars], dim="lev").assign_coords(
+            lev=("lev", levels)
+        )
+        data[base_var] = da
+        data = data.drop_vars(vars_)
+
+    return data
+
```
jder: (This is a good example of why I don't like a separate scripts directory; this is verbatim copied from there, I think?)

tests/test_utils_data.py
```diff
@@ -242,7 +242,7 @@ def data_init(hist: int):
         data_mean = data.mean() * 0.0
         data_std = data.std() * 0.0 + 1.0
         src = DataSource("test", data, data_mean, data_std)
-        val = validate_data(src)
+        val = validate_data(src, [])
```
jder: I think `tensor_map.boundary_var_names` might be less fragile than hard-coding []?

tests/test_stepper.py
```diff
@@ -64,7 +64,7 @@ def inf_data_init(hist: int):
         data_mean: xr.Dataset = data.mean() * 0.0
         data_std: xr.Dataset = data.std() * 0.0 + 1.0
         test_data = DataSource("test-data", data, data_mean, data_std)
-        val = validate_data(test_data)
+        val = validate_data(test_data, [])
```
jder: (same comment here about using tensor_map for boundary var names)

tests/test_datasets.py
```diff
@@ -393,6 +401,59 @@ def test_new_loaders__are_equal_to_v1_data_loader(train_config, loader_version):
...
+
+            x_not_equal = np.equal(x_orig, x_new) == False  # noqa: E712
+            y_not_equal = np.equal(y_orig, y_new) == False  # noqa: E712
+
+            x_not_equal_index = list(zip(*np.where(x_not_equal)))
+            y_not_equal_index = list(zip(*np.where(y_not_equal)))
+
+            assert not np.any(x_not_equal), (
+                f"{len(x_not_equal_index)} values differ: "
+                f"{x_orig[x_not_equal_index[0]]} != {x_new[x_not_equal_index[0]]}."
```
jder: So I think this is exactly identical to the code above except you've got a `[0]` here. Is this a meaningful difference or should we make a `compare_loaders` function which deduplicates this code?

tests/test_datasets.py
```diff
@@ -393,6 +401,59 @@ def test_new_loaders__are_equal_to_v1_data_loader(train_config, loader_version):
...
+                str(cache / src.name),
+            ]
+        )
+
+    flat_config = make_config(flat_data)
+    compact_config = make_config(compact_data)
+
+    with (
+        make_loader(flat_config, version=LoaderVersion.OM4_TORCH) as flat_loader,
+        make_loader(compact_config, version=LoaderVersion.OM4_TORCH) as compact_loader,
```
jder: Reading this again, I think this is broken -- we can't have both multiton scopes active at once. We need to use `with make_loader(flat…): original_samples = …` and then separately `with make_loader(compact…): new_samples = …`

tests/test_datasets.py
jder: It looks like this depends on the implementation details of `conftest.data_source`. If you don't want to refactor that code in order to reuse the relevant pieces, can you please put a comment here and in there pointing to the other copy and explaining how they share internals?

jder (commented):
Thanks! Just a few questions/comments from me. I'm ready to approve once you've got some benchmarks we can look at. I do think this is adding a lot of duplication, but I hope we can delete that (and conditional_rearrange and so on) after we remove the non-compact form of data.

src/ocean_emulators/utils/data.py
alxmrs: Thanks, these are great ideas.

src/ocean_emulators/utils/data.py
alxmrs: Fair point, I think these always co-occur in practice. When I replaced with `lev`, I think there was some corner case I uncovered, but it could have come from an unrelated change/cause.

tests/conftest.py
alxmrs: Yes, that's right -- and, a fair point.

tests/test_stepper.py
alxmrs: I'm passing in the argument in order to reduce dependency on tensor_map (as a gradual refactor).

src/ocean_emulators/utils/data.py
alxmrs: No, it's not equivalent to that. For one, `all_vars` doesn't include the (likely) generated combinations of new vars produced in a rearrage operation (e.g. `time (variable lev) lat lon`). For example, when I step in my IDE, `all_vars`, `vars_with_dim`, and `vars_without_dim` only ever include values `['uo', 'vo', 'thetao', 'so', 'zos']`; however, `data_with_dim` will have these variables cross the number of levels. 

This is confusing! So, to help, I'm adding an example in comments.

tests/test_datasets.py
alxmrs: A comment is a good idea; I decided to raise an assertion for stronger enforcement.

tests/test_datasets.py
alxmrs: I'd like to understand this better -- mind if we jump on a call? Right now, the only multiton we create is the TensorMap; the rest of the relevant values are injected in based on the config/params:

https://github.com/suryadheeshjith/Ocean_Emulator/blob/ff81c07df95dda17ba1dfb9971412c08c781c751/tests/test_datasets.py#L65

tests/test_stepper.py
jder: I like the fact that you're passing in the argument, I'm just suggesting that you pass in `tensor_map.boundary_var_names` (tensor_map is defined earlier on in this function) rather than `[]`.

tests/test_datasets.py
jder: I meant more that the cache paths + names used have to match `conftest.data_source` so I still think a comment in each pointing to the other is a good idea.

Looking at this again, I don't think we need either `flat_data` or `compact_data` (and so don't need `_compact` either) since we only ever use their name to build a config?

tests/test_utils_data.py
alxmrs: I see what you mean now! I didn't realize that we already init tensor map in the test.

tests/test_datasets.py
alxmrs: Very interesting! I see what you mean now; I'm computing a compact value that is, probably always, cached. Ok, I see what you mean now.

scripts/clone_data.py
```diff
@@ -11,20 +11,25 @@
...
 import pathlib
-from collections import defaultdict
 
 import dask
 import dask.diagnostics
 import xarray as xr
 from dask.distributed import LocalCluster
 from tenacity import retry
 
+from ocean_emulators.utils.data import compact_dataset
```
jder, line 31: <3

tests/test_datasets.py
```diff
@@ -361,36 +364,74 @@ def test_inference__data_is_not_zero(inference_loader_pair):
...
+    # fixture.
+    compact_source = dataclasses.replace(data_source, name="compact")
+    compact_config = make_config(compact_source)
+
+    with make_loader(flat_config, version=LoaderVersion.OM4_TORCH) as flat_loader:
+        original_samples = [extract_sample_arrays(sample) for sample in flat_loader]
+    with make_loader(compact_config, version=LoaderVersion.OM4_TORCH) as compact_loader:
+        new_samples = [extract_sample_arrays(sample) for sample in compact_loader]
+
+    assert_equal_samples(original_samples, new_samples)
```
jder, line 427: Beautiful 🤩

jder (commented):
Continues to LGTM! Thanks for the fixes, happy to merge when we've got the benchmarks sorted out!

jder:
@alxmrs did anything change here since my last review? Maybe we can chat live when you're around?

jder (approved):
Discussed live that we should merge this to avoid having it be on a branch forever and we'll time box follow ups on identifying the performance wins/losses here.

PR #219, opened 2025-04-17
Normalize prefill and fill value configuration

suryadheeshjith:
Add normalize pre fill boolean and fill value configuration for testing post fill configuration and multiple fill values

- [x] Add tests for pre-fill and post-fill configuration with multiple fill values


jder:
This is gonna conflict a lot with #211 -- you folks should probably discuss what order makes sense. (cc @alxmrs)

alxmrs:
@suryadheeshjith, do you mind if I merge mine first? I think it may make your code change much smaller in the end.

suryadheeshjith:
Go ahead

src/ocean_emulators/utils/data.py
```diff
@@ -126,7 +130,10 @@ def from_config(cls, cfg: TrainConfig | EvalConfig, *, use_dask: bool) -> Self:
             )
         else:
             data = xr.open_dataset(
-                root / cfg.data.data_path, chunks=chunks, consolidated=True
+                root / cfg.data.data_path,
+                chunks=chunks,
+                consolidated=True,
+                engine="zarr",
```
jder, line 132: Did this fix a bug of some kind?

src/ocean_emulators/utils/data.py
```diff
@@ -82,6 +82,10 @@ def slice(self, time: TimeConfig) -> Self:
         data = self.data.sel(time=time.time_slice)
         return dataclasses.replace(self, name=f"{time=}[{self.name}]", data=data)
 
+    def get_data(self) -> xr.Dataset:
+        """Get the data from the DataSource."""
+        return self.data
+
```
jder: Why do we need a method here? I think dataclasses are generally assumed to have public fields.

src/ocean_emulators/config.py
```diff
@@ -34,6 +34,8 @@ class DataConfig(BaseConfig):
     num_workers: int = 4
     hist: int = 1
     loader_version: str = str(LoaderVersion.OM4_EAGER.value)
+    normalize_pre_fill: bool = True
+    nan_fill_value: float = 0.0
```
jder: Should this be `masked_fill_value`? Not sure what NaN has to do with it.

src/ocean_emulators/config.py
```diff
@@ -34,6 +34,8 @@ class DataConfig(BaseConfig):
     num_workers: int = 4
     hist: int = 1
     loader_version: str = str(LoaderVersion.OM4_EAGER.value)
+    normalize_pre_fill: bool = True
```
jder: What about `normalize_before_mask`?

src/ocean_emulators/datasets.py
```diff
@@ -551,7 +601,7 @@ def __getitem__(self, idx: int):
             self._prognostic_src.data.isel(time=x_index)
             .to_array()
             .transpose(
-                "step", "variable", "time", "lat", "lon"
+                "step", "time", "variable", "lat", "lon"
```
jder: So this is no longer a "no-op" like the comment below says, right? This changed because you're passing around wet_without_history now?

jder (commented):
Just a few questions/comments from me. My main question is: what do you think about including `normalize_pre_fill`, `nan_fill_value ` and `wet_without_history` as part of the DataSource object and making a DataSource.mask_and_normalize method which holds the logic about normalization + masking? I think it'd let us do it once (maybe twice, once for tensors) rather than in a bunch of places and avoid having to pass the new config values around.

src/ocean_emulators/utils/data.py
suryadheeshjith: Yes, some of the data sources do not have the ".zarr" extension

src/ocean_emulators/config.py
suryadheeshjith: NaNs are what are present in the land cells, but I think masked_fill_value is more semantically meaningful

src/ocean_emulators/config.py
suryadheeshjith: Sure

src/ocean_emulators/utils/data.py
suryadheeshjith: Okay

src/ocean_emulators/config.py
jder: Ah, makes sense, thanks.

src/ocean_emulators/datasets.py
suryadheeshjith: Yepp

suryadheeshjith:
> Just a few questions/comments from me. My main question is: what do you think about including `normalize_pre_fill`, `nan_fill_value ` and `wet_without_history` as part of the DataSource object and making a DataSource.mask_and_normalize method which holds the logic about normalization + masking? I think it'd let us do it once (maybe twice, once for tensors) rather than in a bunch of places and avoid having to pass the new config values around.

I think that is a great idea, but I would rather not do this now since I would have to refactor every get functions from top to bottom. The data source class would not only handle the masking and normalizing, but also the reshape and everything that's in the function. For example: 

```
def _get_boundary(self, x_index):
        """
        This function returns the boundary condition for the current time step.

        With hist > 0, the boundary condition considered is always the last step of
        the input.
        """
        data_in_boundary_src = self._boundary_src.map_data(
            lambda ds: ds.isel(time=x_index).isel(time=self.hist)
        )
        if self.normalize_before_mask:
            data_in_boundary_ds = data_in_boundary_src.normalize()
        else:
            data_in_boundary_ds = data_in_boundary_src.data
        data_in_boundary_np: np.ndarray = (
            data_in_boundary_ds.to_array()
            .transpose("window_dim", "variable", "lat", "lon")
            .to_numpy()
        )
        data_in_boundary: torch.Tensor = torch.from_numpy(data_in_boundary_np).float()
        data_in_boundary = torch.where(
            self.wet_surface, data_in_boundary, self.masked_fill_value
        )
        if not self.normalize_before_mask:
            data_in_boundary = self._boundary_src.normalize_with(
                data_in_boundary, variable_axis=1
            )
        return data_in_boundary
```

jder:
> I think that is a great idea, but I would rather not do this now since I would have to refactor every get functions from top to bottom. 

OK, makes sense. Will re-review the other changes tomorrow morning. Thanks!

src/ocean_emulators/datasets.py
```diff
@@ -576,41 +624,54 @@ def __getitem__(self, idx: int):
...
         # normalize expects variables in third dimension
-        prognostic_steps = self._prognostic_src.normalize_with(
-            prognostic_steps, variable_axis=1
-        ).float()
+        if self.normalize_before_mask:
+            prognostic_steps = self._prognostic_src.normalize_with(
+                prognostic_steps, variable_axis=2
+            ).float()
+
+        # post-normalize, mask out values where there is no ocean
```
```suggestion
```

tests/test_datasets.py
```diff
@@ -385,10 +392,15 @@ def test_new_loaders__are_equal_to_v1_data_loader(train_config, loader_version):
 
 
 @pytest.fixture
-def traindataset_input():
+def dataset_input(normalize_before_mask: bool, masked_fill_value: float):
```
jder: Maybe `tiny_dataset_input`? Also might be nice to have defaults for the parameterizations here so that you can use this downstream without having to know to set these parameters? (ie maybe move the parameterize annotations from `test_train_dataset_no_input_change` to here?)

tests/test_datasets.py
```diff
@@ -442,23 +455,83 @@ def traindataset_input():
...
+            masked_fill_value=masked_fill_value,
+            long_rollout=True,
+        )
+        yield traindataset, torch_train_dataset, inference_dataset
+
 
+@pytest.mark.parametrize("normalize_before_mask", [True])
+@pytest.mark.parametrize("masked_fill_value", [0.0])
+def test_train_dataset_no_input_change(
+    dataset_input, normalize_before_mask, masked_fill_value
```
jder: 🐑 I think you can omit these two unused arguments. (But pytest confuses me so I could be wrong.)

tests/test_datasets.py
```diff
@@ -442,23 +455,83 @@ def traindataset_input():
...
+        td0_step0_input = td0.get_input(0)
+        td0_step0_label = td0.get_label(0)
+        inf_step0_input, inf_step0_label = inference_dataset[0]
+
+        assert td0_step0_input.shape == (6, 2, 2)
+        assert td0_step0_label.shape == (4, 2, 2)
+        assert inf_step0_input.shape == (1, 6, 2, 2)
+        assert inf_step0_label.shape == (1, 4, 2, 2)
+
+        if not normalize_before_mask:
```
jder: 🐑 Maybe slightly more clear to swap the if and else blocks? Might also be worth a comment that [0,0,0] is masked.

jder (approved):
Just a few nits and suggestions, but generally looks good to me, thanks!

suryadheeshjith:
AFAIK pytest requires you to have these parameters set downstream. They are required to setup the dataset. I am happy to be proven wrong though, I am still learning about pytest.

PR #210, opened 2025-04-15
pydantic-settings for config

jder:
Interested in high-level feedback about this approach.

<details><summary>--help output</summary>

```
usage: train.py [-h] --config CONFIG [--disk_mode bool] [--pin_mem bool] [--save_freq int] [--epochs int] [--batch_size int] [--learning_rate float] [--scheduler bool] [--loss str] [--finetune bool] [--resume_ckpt_path {str,null}] [--debug bool]
                [--test_using_ema bool] [--ema_decay float] [--faster_decay_at_start bool] [--backend {cpu,cuda,nccl,auto}] [--data_percent float] [--data_stride List[int]] [--steps List[int]] [--step_transition List[int]]
                [--inference_epochs List[int]] [--train TimeConfig] [--val TimeConfig] [--inference List[TimeConfig]] [--experiment JSON] [--experiment.base_name str] [--experiment.sub_name str] [--experiment.rand_seed int]
                [--experiment.base_output_dir str] [--experiment.gantry bool] [--experiment.cluster_data_dir str] [--experiment.wandb JSON] [--experiment.wandb.mode str] [--experiment.wandb.project str] [--experiment.wandb.entity str]
                [--experiment.wandb.group {str,null}] [--experiment.wandb.tags {List[str],null}] [--experiment.wandb.notes {str,null}] [--experiment.network str] [--experiment.prognostic_vars_key str] [--experiment.boundary_vars_key str]
                [--data JSON] [--data.data_path str] [--data.data_means_path str] [--data.data_stds_path str] [--data.scaling_residuals_file {str,null}] [--data.time_delta int] [--data.num_workers int] [--data.hist int] [--data.loader_version str]
                [--samudra JSON] [--samudra.ch_width List[int]] [--samudra.n_out int] [--samudra.dilation List[int]] [--samudra.n_layers List[int]] [--samudra.pred_residuals bool] [--samudra.last_kernel_size int] [--samudra.pad str]
                [--samudra.wet {Any,null}] [--samudra.core_block JSON] [--samudra.core_block.block_type str] [--samudra.core_block.kernel_size int] [--samudra.core_block.activation str] [--samudra.core_block.upscale_factor int]
                [--samudra.core_block.norm str] [--samudra.corrector JSON] [--samudra.corrector.non_negative_corrector_names {List[str],null}] [--samudra.down_sampling_block str] [--samudra.up_sampling_block str]

options:
  -h, --help            show this help message and exit
  --config CONFIG       Path to config YAML file
[...]

PR #220, opened 2025-04-21
Enable upgrade lints

jder:
On #210, @alxmrs was leaving comments about using `Union` rather than modern`|`, and there are a variety of related things like dict/list vs Dict/List and so on. I think either we should enforce these upgrade (`UPXXXXX`) lints with ruff (which is this PR) or else not worry about it. I'm honestly fine with either so opening this to have the discussion. 

This also turns on warnings about using `"".format` rather than f-strings, older import paths, inheriting from `object`, using `open(foo, "r")` rather than `open(foo)` and others. Full list is here https://docs.astral.sh/ruff/rules/#pyupgrade-up. 

All these fixes were automated (and therefore should be safe) except for one I will call out below.

### DIFF
--- a/src/ocean_emulators/aggregator/validate/snapshot.py
+++ b/src/ocean_emulators/aggregator/validate/snapshot.py
@@ -1,5 +1,3 @@
-from typing import Dict
-
 import torch
 
 from ocean_emulators.aggregator.plotting import plot_paneled_data
@@ -27,14 +25,14 @@ class SnapshotAggregator(ValidateSubAggregator):
         ),
     }
 
-    def __init__(self, metadata: Dict[str, Dict[str, str]] = {}, hist: int = 0):
+    def __init__(self, metadata: dict[str, dict[str, str]] = {}, hist: int = 0):
         """
         Args:
             metadata: Mapping of variable names their metadata that will
                 used in generating logged image captions.
             hist: Number of history steps to include in the snapshot.
         """
-        self._metadata: Dict[str, Dict[str, str]] = metadata
+        self._metadata: dict[str, dict[str, str]] = metadata
         self.hist = hist
 
     @torch.no_grad()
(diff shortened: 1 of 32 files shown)


src/ocean_emulators/utils/logging.py
```diff
@@ -208,9 +208,8 @@ def log_every(self, iterable, print_freq, header=None):
         total_time = time.perf_counter() - start_time
         total_time_str = str(datetime.timedelta(seconds=int(total_time)))
         logging.info(
-            "{} Total time: {} ({:.4f} s / it)".format(
-                header, total_time_str, total_time / len(iterable)
-            )
+            f"{header} Total time: {total_time_str} "
+            f"({total_time / len(iterable):.4f} s / it)"
```
jder, line 212: Manual re-wrapping on to multiple lines. (PS I don't like line length limits)

src/ocean_emulators/aggregator/validate/snapshot.py
```diff
@@ -27,14 +25,14 @@ class SnapshotAggregator(ValidateSubAggregator):
...
-    def __init__(self, metadata: Dict[str, Dict[str, str]] = {}, hist: int = 0):
+    def __init__(self, metadata: dict[str, dict[str, str]] = {}, hist: int = 0):
         """
         Args:
             metadata: Mapping of variable names their metadata that will
                 used in generating logged image captions.
             hist: Number of history steps to include in the snapshot.
         """
-        self._metadata: Dict[str, Dict[str, str]] = metadata
+        self._metadata: dict[str, dict[str, str]] = metadata
```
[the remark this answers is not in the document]
jder: This is actually a great point, but probably a separate PR. Will make an issue.

src/ocean_emulators/utils/logging.py
suryadheeshjith: I think this is reasonable

suryadheeshjith:
I like this 👍

suryadheeshjith (approved):

alxmrs:
Thank you for doing this!

PR #224, opened 2025-04-22
[Clean PR] Ocean Heat Corrector 

suryadheeshjith:
This PR replaces #165


configs/slurm_perlmutter_train_om4.yaml
```diff
@@ -49,7 +49,7 @@ experiment:
   prognostic_vars_key: thermo_all
   boundary_vars_key: tau_hfds_hfds_anom
 
-data: !include data/om4_internal.yaml
+data: !include data/om4_ohc.yaml
```
alxmrs, line 51: @jder it's really nice to see the new config system in use :)

src/ocean_emulators/constants.py
```diff
@@ -112,9 +140,13 @@
     "mask_18",
 ]
 
+RHO_0 = 1035.0  # DENSITY_OF_WATER_CM4 kg/m^3
+CP_SW = 3992.0  # SPECIFIC_HEAT_OF_WATER_CM4 J/kg/K
+SECONDS_PER_5DAY = 5 * 24 * 60 * 60  # 5 day average
```
alxmrs, line 144: 🐑 Not a suggestion, but this reminds me of pint: https://pint.readthedocs.io/en/stable/

https://xarray.dev/blog/introducing-pint-xarray

src/ocean_emulators/derived_variables.py
```diff
@@ -0,0 +1,81 @@
...
+from ocean_emulators.constants import CP_SW, RHO_0, TensorMap
+
+
+def compute_ocean_heat_content(T: Tensor, dz: Tensor) -> Tensor:
+    """Compute the heat content of the ocean.
+
+    Args:
+        T: Temperature tensor of shape (batch_size, depth, height, width) or
+                                        (batch_size, hist, depth, height, width)
+        dz: Depth tensor of shape (depth,)
```
alxmrs: If these are really consistent, then I think this is a good use case to use jaxtyping in the arguments.

src/ocean_emulators/derived_variables.py
```diff
@@ -0,0 +1,81 @@
...
+    total_HC_t = torch.nansum(HC_t, dim=depth_dim)
+
+    if mask:
+        total_HC_t = torch.where(surface_wet_mask, total_HC_t, torch.nan)
+
+    return total_HC_t
+
+
+def compute_global_ocean_heat_content(
+    T: Tensor, dz: Tensor, area_weighted_func: Callable
```
alxmrs: Same here: I think dimension info should be in the type; that way, we'll get static type checking (exercised at test time).

src/ocean_emulators/models/corrector.py
```diff
@@ -1,20 +1,99 @@
...
+    """Base class for tensor correction modules."""
+
+    def __init__(self, hist: int, tensor_map: TensorMap, normalize: Normalize):
         super().__init__()
-        self.corrector_config = corrector_config
-        self.non_negative_corrector_names: list[str] | None = (
-            corrector_config.non_negative_corrector_names
+        self.hist = hist
+        self.tensor_map = tensor_map
+        self.normalize = normalize
```
alxmrs, line 32: I'm happy that you're injecting these two objects instead of using a singleton!

src/ocean_emulators/models/corrector.py
```diff
@@ -1,20 +1,99 @@
...
+        self.num_prognostic_channels = len(self.tensor_map.prognostic_var_names)
+
+    def _flatten_hist(self, fts: HistChanneled) -> HistBatched:
```
alxmrs, line 35: I don't think these need to be private if they are meant to be distributed to subclasses. I'm a bit torn on it to be honest. WDYT?

src/ocean_emulators/models/corrector.py
```diff
@@ -25,55 +104,224 @@ def __init__(self, corrector_config, hist):
...
+
+
+class OceanHeatCorrector(BaseCorrector):
+    """
+    Applies a correction to potential temperature to conserve
+    ocean heat content.
+
+    Following this document - https://www.overleaf.com/project/67ed705406995df4c185e6b6
+
+    This class relies in input bounjdary conditions. Thus, we would
```
alxmrs:
```suggestion
    This class relies in input boundary conditions. Thus, we would
```

src/ocean_emulators/utils/data.py
```diff
@@ -471,3 +502,28 @@ def unnormalize_tensor_prognostic(
...
+        if data.ndim == 4:
+            assert data.shape[1] == self._boundary_mean_np.shape[0]
+            tensor_mean = tensor_mean.reshape([1, -1, 1, 1])
+            tensor_std = tensor_std.reshape([1, -1, 1, 1])
+        elif data.ndim == 5:
+            assert data.shape[2] == self._boundary_mean_np.shape[0]
+            tensor_mean = tensor_mean.reshape([1, 1, -1, 1, 1])
+            tensor_std = tensor_std.reshape([1, 1, -1, 1, 1])
+        else:
+            raise ValueError(f"Invalid data shape: {data.shape}")
```
alxmrs: Do you think something like this would work instead?

```suggestion
        expand_var_dim = [1] * data.ndim
        expand_var_dim[-3] = -1
        assert data.shape[-3] == self._boundary_mean_np.shape[0]
        tensor_mean = tensor_mean.reshape(expand_var_dim)
        tensor_std = tensor_std.reshape(expand_var_dim)
```

tests/test_corrector.py
```diff
@@ -0,0 +1,239 @@
...
+            prognostic_var_names=["var_0", "var_1"],
+            boundary_var_names=["var_2"],
+            wet_mask=wet_mask,
+            wet_mask_surface=wet_mask,
+        )
+
+    return normalize, tensor_map, wet_mask
+
+
+def test_relu_corrector(corrector_init):
```
alxmrs, line 78: These tests are great, thank you.

src/ocean_emulators/models/corrector.py
suryadheeshjith: I fixed this comment

suryadheeshjith:
My Jax type naming skills need work

src/ocean_emulators/derived_variables.py
```diff
@@ -0,0 +1,92 @@
...
+
+import torch
+
+from ocean_emulators.constants import (
+    CP_SW,
+    RHO_0,
+    Single3DHistPrognostic,
+    Single3DPrognostic,
+    SingleHistPrognostic,
+    SinglePrognostic,
```
alxmrs: 🐑 If these types are used once and just in the functions, then I think the definitions can live in the functions -- they don't have to be named and extracted into constants. In fact, jaxtyping can perform better type checking if type definitions are inline: https://docs.kidger.site/jaxtyping/api/array/#shape:~:text=Symbolic%20expressions%20are%20evaluated%20in%20two%20stages

src/ocean_emulators/derived_variables.py
```diff
@@ -0,0 +1,92 @@
...
+)
+
+
+def compute_ocean_heat_content(
+    T: Single3DPrognostic | Single3DHistPrognostic, dz: torch.Tensor
+) -> SinglePrognostic | SingleHistPrognostic:
+    """Compute the heat content of the ocean.
+
+    Args:
+        T: Temperature tensor of shape (batch_size, depth, height, width) or
```
alxmrs: Let's omit documenting the shape and just keep it in the type definition

src/ocean_emulators/derived_variables.py
alxmrs: Further, maybe the primary benefit of adding type information is that it it documents behavior. When we use aliases, we have to check another file to see the named order of dimensions.

src/ocean_emulators/eval.py
```diff
@@ -100,6 +100,7 @@ def __init__(self, cfg: EvalConfig) -> None:
         raw = DataSource.from_config(cfg, use_dask=True)
         self.src = validate_data(raw)
         self.data = self.src.data
+        self.static_data = self.src.get_static_data()
```
alxmrs: The more pythonic thing to do would be to make `static_data` a property; a getter method is more Java like.  I recommend using a `cached_property`, if possible.

src/ocean_emulators/utils/data.py
```diff
@@ -37,6 +38,7 @@ class DataSource:
     data: xr.Dataset
     means: xr.Dataset
     stds: xr.Dataset
+    static_data_vars: list[str] | None = None
```
alxmrs: I'm not convinced that `static_data_vars` should be part of this data class. I choose not to include prognostic and boundary vars here, for instance. Can we talk about this?

src/ocean_emulators/utils/data.py
```diff
@@ -83,6 +85,12 @@ def slice(self, time: TimeConfig) -> Self:
         data = self.data.sel(time=time.time_slice)
         return dataclasses.replace(self, name=f"{time=}[{self.name}]", data=data)
 
+    def get_static_data(self) -> xr.Dataset | None:
+        """Get the static data from the data source."""
+        if self.static_data_vars is None:
+            return None
+        return self.data[self.static_data_vars]
```
alxmrs: I don't think this is an essential operation on this data type. It seems just as easy to call
`self.src.data[cfg.static_data_vars]` as it would be to call `self.src.get_static_data()`.

src/ocean_emulators/utils/data.py
```diff
@@ -374,6 +385,20 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
 
 def validate_data(src: DataSource) -> DataSource:
     """Validate the data such that we have the correct format for training."""
+    static_data_vars = src.static_data_vars
```
alxmrs: IMO, `static_data_vars` should be an argument to `validate_data`. See, for example, how I've refactored validate data in this PR: https://github.com/suryadheeshjith/Ocean_Emulator/pull/215/files#diff-3e496072412d1e191c8f724220b196643bbb36e110f3452931a25815161a139bR493

src/ocean_emulators/utils/data.py
```diff
@@ -374,6 +385,20 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
...
+            for var in static_data_vars:
+                assert var in data.variables, (
+                    f"Static data variable {var} not found in data"
+                )
+                if "time" in data[var].dims:
+                    data[var] = data[var].isel(time=0)
+
+        return data, means, std
+
+    src = src.map(_static_data_checks, suffix="static_data_checked")
```
alxmrs: If you're only modifying `data`, then this could be done with a `.map_data()` call.

src/ocean_emulators/utils/data.py
```diff
@@ -374,6 +385,20 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
 
 def validate_data(src: DataSource) -> DataSource:
     """Validate the data such that we have the correct format for training."""
+    static_data_vars = src.static_data_vars
+
+    def _static_data_checks(data, means, std):
+        if static_data_vars is not None:
```
alxmrs: Recommendation: pull `ifs` up: https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

In other words, it would be better to not make a call to `.map` if `static_data_vars` is None rather than always make the call but sometimes do nothing.

src/ocean_emulators/eval.py
alxmrs: Ok, looking at the code below, I think a better fit would be to pass in the `cfg.data.static_data_vars` into the class, similar to `boundary_var_names` and then call: 
```suggestion
        self.static_data = self.src.data[static_data_vars]
```

src/ocean_emulators/train.py
```diff
@@ -163,6 +163,7 @@ def __init__(self, cfg: TrainConfig) -> None:
         raw = DataSource.from_config(cfg, use_dask=use_dask)
         self.src = validate_data(raw)
         self.data = self.src.data
+        self.static_data = self.src.get_static_data()
```
alxmrs: See comment above

src/ocean_emulators/utils/data.py
alxmrs: That we default this value to None indicates to me that it is not an essential property of a `DataSource`.

src/ocean_emulators/eval.py
suryadheeshjith: ~~I will also let @jder review this line~~

src/ocean_emulators/utils/data.py
suryadheeshjith: Ah makes sense

src/ocean_emulators/derived_variables.py
suryadheeshjith: Sure

src/ocean_emulators/derived_variables.py
suryadheeshjith: Ok

src/ocean_emulators/utils/data.py
suryadheeshjith: I wish I could do this but mypy complains

src/ocean_emulators/datasets.py
```diff
@@ -679,6 +676,29 @@ def _prep_prognostic_steps(
...
+        if not self.normalize_before_mask:
+            boundary_steps = self._boundary_src.normalize_with(
+                boundary_steps, variable_axis=2
+            ).float()
+
+        boundary_steps = rearrange(
+            boundary_steps,
+            "step time variable lat lon -> step (time variable) lat lon",
+        )
+        return boundary_steps
```
jder: I think _prep_boundary_steps is now identical to _prep_prognostic_steps except for wet vs wet_surface? Maybe we can merge the two?

src/ocean_emulators/derived_variables.py
```diff
@@ -0,0 +1,82 @@
+from collections.abc import Callable
+
+import torch
+from jaxtyping import Float
+
+from ocean_emulators.constants import CP_SW, RHO_0, Grid, TensorMap
+
+
+def compute_ocean_heat_content(
+    T: Float[Grid, "*batch depth"] | Float[Grid, "*batch depth hist"], dz: torch.Tensor
```
jder: These type annotations don't seem to match the code. The code below looks like it's expecting either 4 dimensions (depth at index 1) or 5 dimensions (depth at index 2). I think that would be `Float[Grid, "batch depth"]` (4 dimensions, depth at index 1) or `Float[Grid, "batch hist depth"]` (depth at index 2), right? (`*batch` means "zero or more batch dimensions")

src/ocean_emulators/utils/data.py
```diff
@@ -230,18 +231,18 @@ def spherical_area_weights(data: xr.Dataset) -> Grid:
...
 
     Args:
         time_config: Time configuration
-        time_delta: Time delta in days
         hist: Number of rollout steps
 
     Returns:
         num_steps: Number of rollout steps
     """
+    time_delta = 5  # Hardcoded
```
jder: 🐑 Maybe stick in constants.py with the other constants?

src/ocean_emulators/utils/data.py
jder: You could move the def into the if as well if you want mypy to chill.

src/ocean_emulators/models/corrector.py
```diff
@@ -25,55 +104,222 @@ def __init__(self, corrector_config, hist):
...
-                fts_reshaped, fill_value=0.0
+        fts[:, self.thetao_idx] = T_corrected
+
+        fts = self._normalize_fts_prognostic(fts)
+        fts = self._unflatten_hist(fts)
+
+        return fts
+
+
+class Corrector(torch.nn.Module):
```
jder: This kinda sounds like a base class to me. Maybe AllCorrectors or something?

src/ocean_emulators/models/corrector.py
```diff
@@ -25,55 +104,222 @@ def __init__(self, corrector_config, hist):
...
+                    normalize=self.normalize,
+                )
             )
 
-            # apply relu
-            indices = self.non_neg_indices.to(unnormalized_fts.device)
-            # unnormalized_fts_cloned = unnormalized_fts.clone()
-            unnormalized_fts[:, indices, :, :] = torch.relu(
-                unnormalized_fts[:, indices, :, :]
+        if hasattr(config, "ocean_heat_corrector") and config.ocean_heat_corrector:
```
jder:
```suggestion
        if config.ocean_heat_corrector:
```

src/ocean_emulators/utils/data.py
jder: I actually do think that we should include these here but probably we should do it with all the other TensorMap stuff.

src/ocean_emulators/utils/data.py
jder: Long term, I disagree with Alex -- cfg should not be floating around the program. It should be consumed at the start and `build()`ed into semantically helpful types like data_source which has the needed config info.

jder (approved):
Thanks for the clean PR! Just a couple of minor suggestions/questions from me.

src/ocean_emulators/utils/data.py
alxmrs: Haha, I disagree with myself long term too! I like your comment above about factoring static data at the same time as the TensorMap stuff. It's probably best to refactor holistically; I'm uncertain if TensorMap should be fused here or when we deep dive, a simpler/better pattern arises.

src/ocean_emulators/derived_variables.py
alxmrs: Jaxtyping should catch this in tests; I think this PR might need to modify the `addopts = "--jaxtyping-packages` line in the pyproject.toml?

alxmrs (approved):
LGTM, with the caveat of resolving the typing issue that Jesse pointed out (and maybe getting jaxtyping to programmatically check for us).

src/ocean_emulators/derived_variables.py
suryadheeshjith: Fixed, the line of code is already in pyproject.toml?

suryadheeshjith:
I am merging this for now, we will need to figure out why jaxtyping didn't pick up the error and also resolve the issue with data source and data configs in future PRs.

PR #227, opened 2025-04-24
Use dask for inference

### LINKED ISSUE #208: Resolve performance issue with torch loader & new inference code

jder:
Fixes memory issues with the torch loader, fixes #208. 

After this, the torch and eager loaders have the same peak memory usage during inference. The issue was that the non-dask version was using about 3x the memory of the dask version and getting OOM killed. This is due to a combination of the InferenceDataset not being written with eager evaluation in mind (eg slicing repeatedly) and due to what look like some pretty bad performance bugs in xarray. (It was allocating a ton of memory to lazily track reorderings of the data rather than eagerly shuffling the arrays -- kinda looks like 6x more than the underlying arrays but I could have misread and don't think it's worth digging into this more now.) (And to add insult to injury, the shuffle in question was the identity transform…)

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -164,6 +164,11 @@ class Trainer:
         self.src = validate_data(raw)
         self.data = self.src.data
 
+        # We use dask for inference since it has memory issues otherwise.
+        # TODO(jder): Could rewrite inference dataset like we did for TorchTrainDataset
+        # see https://github.com/suryadheeshjith/Ocean_Emulator/issues/208
+        self.inference_src = validate_data(DataSource.from_config(cfg, use_dask=True))
+
         self.metadata = construct_metadata(self.data)
         self.wet, self.wet_surface = extract_wet_mask(
             self.data, self.prognostic_var_names, cfg.data.hist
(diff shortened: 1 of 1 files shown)


src/ocean_emulators/train.py
```diff
@@ -164,6 +164,11 @@ def __init__(self, cfg: TrainConfig) -> None:
         self.src = validate_data(raw)
         self.data = self.src.data
 
+        # We use dask for inference since it has memory issues otherwise.
+        # TODO(jder): Could rewrite inference dataset like we did for TorchTrainDataset
+        # see https://github.com/suryadheeshjith/Ocean_Emulator/issues/208
+        self.inference_src = validate_data(DataSource.from_config(cfg, use_dask=True))
```
jder, line 170: PS @alxmrs I don't like that I have to remember to call validate_data here :)

suryadheeshjith (approved):
LGTM

PR #236, opened 2025-04-29
Fix wandb nested config logging

### LINKED ISSUE #228: Config is showing up flattened in wandb
!included configs show up as strings in wandb.

https://openathena.slack.com/archives/C08CYM42DT3/p1745520281909149?thread_ts=1745501514.022449&cid=C08CYM42DT3

jder:
The config system broke the nice logging of nested config structures to wandb. For example, see the "data" key in this run:

https://wandb.ai/m2lines/ocean-emulators/runs/2i57cuzt/overview

Screenshot 2025-04-29 at 11 21 59 AM

This PR uses pydantic serialization to correctly output nested keys, for example this run:

https://wandb.ai/m2lines/ocean-emulators/runs/twx4knvp/overview

Screenshot 2025-04-29 at 11 22 19 AM

Fixes #228

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -192,3 +192,6 @@ class EvalConfig(TopLevelConfig):
 
     def prepare_output_dirs(self) -> None:
         self.experiment.output_dir.mkdir(parents=True, exist_ok=True)
+
+
+AnyTopLevelConfig = TrainConfig | EvalConfig
--- a/src/ocean_emulators/utils/wandb.py
+++ b/src/ocean_emulators/utils/wandb.py
@@ -1,6 +1,5 @@
 import logging
 from collections.abc import Mapping
-from typing import Any
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -8,6 +7,7 @@ import torch
 import wandb
 from wandb.data_types import WBValue
 
+from ocean_emulators.config import AnyTopLevelConfig
 from ocean_emulators.utils.multiton import Multiton
 
 # Metrics supported by wandb -- probably there are more possible types too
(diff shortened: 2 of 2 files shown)


src/ocean_emulators/config.py
```diff
@@ -192,3 +192,6 @@ class EvalConfig(TopLevelConfig):
 
     def prepare_output_dirs(self) -> None:
         self.experiment.output_dir.mkdir(parents=True, exist_ok=True)
+
+
+AnyTopLevelConfig = TrainConfig | EvalConfig
```
suryadheeshjith, line 197: This could be TopLevelConfig instead of AnyTopLevelConfig

suryadheeshjith (approved):
LGTM!

src/ocean_emulators/config.py
jder: TopLevelConfig is already the class which TrainConfig and EvalConfig inherit from, but it doesn't include the `experiment` member. I tried a little to make that work by adding an `@abstract` `@property` for `experiment` to TopLevelConfig but mypy didn't like how the subclasses implemented the property with a raw member. Then I decided it was probably easier to do this but if you know how to make the other thing work happy to remove this guy altogether.

suryadheeshjith: Ah I see, that makes sense. What about just a different name?

jder: I actually like this name (it reminds me of [SwiftUI's AnyView](https://developer.apple.com/documentation/swiftui/anyview) which is a value which implements `View`). Do you have another suggestion? `TrainOrEvalConfig`?

suryadheeshjith: Alright, sounds good then!

PR #238, opened 2025-05-07
Proof of concept of a gin-config/Flask/FastAPI-like decorator interface for configuration.

alxmrs:
This `bind` decorator on the pydantic settings baseconfig is all that we need in order to map configured values from the YAML to user defined functions. Values are matched by type, and ties are resolved by type and name. In case of ambiguity, users can manually specify how to map params to access paths in the baseconfig tree.


jder:
@alxmrs how about this one? Should we close or save somewhere else and then close?

alxmrs:
Closing, and I have it stored on my local branch. If I somehow happened to lose it, the idea is captured in our design doc, so it shouldn't be hard to replicate.

PR #241, opened 2025-05-15
QoL improvements when working on using higher res training data

jder:
Small grab-bag of things I ran into while trying to train on higher-res data, I suggest anything you want to discuss I just break out into its own PR and we merge this with the easy stuff.

* added `static_data_vars: []` to various data configs -- this is so it overrides any default value of this when picking a dataset which doesn't have them. In particular, currently `uv run oe.train configs/train_om4.yaml --data=@configs/data/public.yaml` doesn't work because train_om4.yaml has a non-empty list here. After this change, it works. I suggest in general we make it our policy that all the configs/data/*.yamls have the same set of keys defined to avoid this problem (or perhaps the stronger "don't have default values for config in code").
* Rather than letting xarray interpret our date strings by magic with what ends up being `data.sel(time=slice("1975-10-31", "1985-10-31"))`, we instead explicitly parse those strings out of the config into the right cftime types (Julian dates at noon, which seems to be what our data uses). This pulls up the validation of format much earlier, but also lets us write some nice validation that you are selecting a reasonable range of time (and remove some existing parsing code). The mysterious errors that selecting small or empty ranges were a real headache when I was first getting started with this codebase and I think this will be helpful.
* Edit some jaxtyping annotations which assumed a particular grid.
* Support computing mean/stds on a DataSource (not used anywhere right now other than my notebook, but I like the direction of enhancing DataSource for use in data processing)
* Update our data processing script to save zarrs for means/stds which seems a little silly but is what the public data uses right now so it's what our configs say.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,8 +27,6 @@ jobs:
         with:
           path: .data_cache/
           key: ${{ runner.os }}-test-data-${{ hashFiles('tests/conftest.py', 'configs/*.test.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-test-data-
 
       - name: Install uv and set the python version
         uses: astral-sh/setup-uv@v5
--- a/configs/train_om4_halfdeg.yaml
+++ b/configs/train_om4_halfdeg.yaml
@@ -5,7 +5,7 @@ disk_mode: true
 pin_mem: true
 save_freq: 5
 epochs: 70
-batch_size: 1
+batch_size: 2
 learning_rate: 0.0002
 scheduler: true
 loss: mse
@@ -23,8 +23,9 @@ inference_times:
   - start: "1967-01-01"
     end: "1967-12-29"
 data_stride: [1]
-steps: [1]
+steps: [4]
 step_transition: []
+preemptible: false
 
 backend: auto
 
@@ -38,7 +39,7 @@ experiment:
     project: ocean-emulators
     entity: m2lines
   network: Samudra
-  prognostic_vars_key: thermo_dynamic_5
+  prognostic_vars_key: thermo_dynamic_all
   boundary_vars_key: tau_hfds_hfds_anom
 data:
   !include data/om4_ohc.yaml
--- a/configs/train_om4_halfdeg.yaml
+++ b/configs/train_om4_halfdeg.yaml
@@ -1,0 +1,68 @@
+# yaml-language-server: $schema=schemas/TrainConfig.json
+
+debug: false
+disk_mode: true
+pin_mem: true
+save_freq: 5
+epochs: 70
+batch_size: 1
+learning_rate: 0.0002
+scheduler: true
+loss: mse
+finetune: false
+resume_ckpt_path: null
+inference_epochs: [-1]
+data_percent: 1.0
+train_time:
+  start: "1958-01-03"
+  end: "1967-12-29"
+val_time:
+  start: "1967-01-01"
+  end: "1967-12-29"
+inference_times:
+  - start: "1967-01-01"
+    end: "1967-12-29"
+data_stride: [1]
+steps: [1]
+step_transition: []
+
+backend: auto
+
+experiment:
+  name: om4_samudra_halfdeg
+  rand_seed: 15
+  base_output_dir: .LOCAL
+  gantry: false
+  wandb:
+    mode: online
+    project: ocean-emulators
+    entity: m2lines
+  network: Samudra
+  prognostic_vars_key: thermo_dynamic_5
+  boundary_vars_key: tau_hfds_hfds_anom
+data:
+  !include data/om4_ohc.yaml
+
+
+samudra:
+  ch_width: [157, 200, 250, 300, 400]
+  n_out: 77
+  dilation: [1, 2, 4, 8]
+  n_layers: [1, 1, 1, 1]
+  pred_residuals: false
+  last_kernel_size: 3
+  pad: "circular"
+
+  core_block:
+    block_type: "conv_next_block"
+    kernel_size: 3
+    activation: "capped_gelu"
+    upscale_factor: 4
+    norm: "batch"
+
+  down_sampling_block: "avg_pool"
+  up_sampling_block: "bilinear_upsample"
+
+  corrector:
+    non_negative_corrector_names: null
+    ocean_heat_corrector: false
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -3,9 +3,8 @@ from pathlib import Path
 from typing import Annotated, Any, Literal
 
 import cftime
-from pydantic import PlainSerializer, PlainValidator, WithJsonSchema
-
 import torch
+from pydantic import PlainSerializer, PlainValidator, WithJsonSchema
 
 from ocean_emulators.config_base import BaseConfig, TopLevelConfig
 from ocean_emulators.constants import LoaderVersion
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -1,6 +1,9 @@
 from functools import cached_property
 from pathlib import Path
-from typing import Any, Literal
+from typing import Annotated, Any, Literal
+
+import cftime
+from pydantic import PlainSerializer, PlainValidator, WithJsonSchema
 
 import torch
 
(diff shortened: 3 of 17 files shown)


.github/workflows/test.yml
```diff
@@ -27,8 +27,6 @@ jobs:
         with:
           path: .data_cache/
           key: ${{ runner.os }}-test-data-${{ hashFiles('tests/conftest.py', 'configs/*.test.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-test-data-
```
alxmrs, line 31: Just curious: what does this change do?

configs/train_om4_halfdeg.yaml
```diff
@@ -0,0 +1,69 @@
...
+learning_rate: 0.0002
+scheduler: true
+loss: mse
+finetune: false
+resume_ckpt_path: null
+inference_epochs: [-1]
+data_percent: 1.0
+train_time:
+  start: "1958-01-03"
+  end: "1967-12-29"
```
alxmrs, line 18: Nice

.github/workflows/test.yml
jder: Before, this would restore data cached under the key `Linux-test-data` when there is a cache miss from the other key (computed based on file hashes). I don't think we want this since it means restoring some old data which doesn't match the code.

src/ocean_emulators/config.py
```diff
@@ -15,13 +18,56 @@ class WandBConfig(BaseConfig):
...
+
+    datetime: cftime.datetime
+
+    def __init__(self, s: str):
+        datetime = cftime.datetime.strptime(s, "%Y-%m-%d", calendar="julian")
+        datetime = datetime.replace(hour=12)
+        self.datetime = datetime
+
+    def __str__(self) -> str:
+        return self.datetime.strftime("%Y-%m-%d")
```
alxmrs, line 38: 🐑 I like this. And, I'm wondering if there is some way we can make it clear that these are datetimes at noon. Maybe, via the name of the class? Not that big of an issue, but it the potential for mistake jumps out at me. Might be nothing to fuss over, however.

alxmrs: Reading more of the code, I think this is fine.

src/ocean_emulators/utils/data.py
```diff
@@ -82,6 +83,22 @@ def map_data(
...
     def slice(self, time: TimeConfig) -> Self:
         """Slice the data source to only include the specified time slice."""
+        data_time_min = self.data.time.min().item()
+        data_time_max = self.data.time.max().item()
+        if time.start.datetime > data_time_max or time.end.datetime < data_time_min:
+            raise ValueError(
+                f"Time slice {time} is entirely outside the range of the data "
+                f"{data_time_min.strftime('%Y-%m-%d')} to "
+                f"{data_time_max.strftime('%Y-%m-%d')}"
+            )
```
alxmrs, line 93: Nice, thanks

src/ocean_emulators/utils/data.py
```diff
@@ -82,6 +83,22 @@ def map_data(
...
+
+        if time.start.datetime < data_time_min or time.end.datetime > data_time_max:
+            logger.warning(
```
alxmrs, line 96: Interesting corner case for a warning!

jder:
@alxmrs FYI I verified that the data is in fact in the Julian calendar (the 5-day timesteps correctly include Feb 29 on leap years).

src/ocean_emulators/utils/data.py
```diff
@@ -82,6 +83,22 @@ def map_data(
 
     def slice(self, time: TimeConfig) -> Self:
         """Slice the data source to only include the specified time slice."""
+        data_time_min = self.data.time.min().item()
+        data_time_max = self.data.time.max().item()
+        if time.start.datetime > data_time_max or time.end.datetime < data_time_min:
```
alxmrs, line 88: Did you add these due to an unexpected error (say, with the strange 3-digit date format)?

src/ocean_emulators/utils/data.py
```diff
@@ -115,6 +132,16 @@ def normalize_with(
         norm = norm.to(data.dtype)
         return norm
 
+    @classmethod
+    def from_dataset(cls, dataset: xr.Dataset) -> Self:
+        """Create a DataSource from the data, computing means and stds."""
+        assert "lev" not in dataset.dims, "We currently only support non-compact data"
+
+        means = dataset.map(lambda x: x.mean())
+        stds = dataset.map(lambda x: x.std())
```
alxmrs: Any reason why we can't just do the following?
```suggestion
        means = dataset.mean()
        stds = dataset.std()
```

src/ocean_emulators/utils/data.py
```diff
@@ -115,6 +132,16 @@ def normalize_with(
         norm = norm.to(data.dtype)
         return norm
 
+    @classmethod
+    def from_dataset(cls, dataset: xr.Dataset) -> Self:
+        """Create a DataSource from the data, computing means and stds."""
+        assert "lev" not in dataset.dims, "We currently only support non-compact data"
```
alxmrs, line 139: Is there a reason why calculating the mean and std wouldn't work for compact data? Or, is this PR just not tested for that use case? If the latter, would you mind putting a TODO here for me to check this later?

src/ocean_emulators/utils/data.py
```diff
@@ -115,6 +132,16 @@ def normalize_with(
         norm = norm.to(data.dtype)
         return norm
 
+    @classmethod
+    def from_dataset(cls, dataset: xr.Dataset) -> Self:
```
alxmrs: 🐑  optional: let users specify a name of the dataset, e.g. 
```suggestion
    def from_dataset(cls, dataset: xr.Dataset, *, name: str = 'dataset') -> Self:
```

...to be used as the data class name below. I could see how this would be helpful when reading the data class name in the debugger.

src/ocean_emulators/utils/data.py
```diff
@@ -125,7 +152,7 @@ def from_config(cls, cfg: TrainConfig | EvalConfig, *, use_dask: bool) -> Self:
             data = xr.open_dataset(
                 root / cfg.data.data_path,
                 engine="netcdf4",
-                chunks={"time": 1, "lat": 180, "lon": 360},
```
alxmrs, line 128: Nice

src/ocean_emulators/utils/data.py
```diff
@@ -232,30 +259,22 @@ def spherical_area_weights(data: xr.Dataset) -> Grid:
...
-        num_steps: Number of rollout steps
+        num_steps: Total number of rolled-out inferences which fit into the time range
     """
-    time_delta = TIME_DELTA
-    start_time_str = time_config.start
-    start_year, start_month, start_day = start_time_str.split("-")
-    start_time = cftime.DatetimeNoLeap(
-        int(start_year), int(start_month), int(start_day), 0, 0, 0
-    )
+    start_time = data_source.data.time.min().item()
```
alxmrs, line 274: Nice to see so much code in the red here.

alxmrs (approved):
A few minor questions about `from_dataset()`; but in general this PR is great! LGTM.

src/ocean_emulators/utils/data.py
jder: I added this because when I pointed an existing config to the new 10-year-only om4 data, I got a very hard-to-understand error somewhere in numpy about a window being of size zero. This was the ultimate cause.

src/ocean_emulators/utils/data.py
jder: Nope, thanks!

src/ocean_emulators/utils/data.py
jder: I think it's a little subtle because we don't want to compute the mean across the `lev` dimension (assuming we want to match current behavior) so the below code will not "just work" and so I wanted to make sure we actually tested it a little. Will add a todo.

src/ocean_emulators/utils/data.py
jder: Done, thanks

src/ocean_emulators/utils/data.py
alxmrs: Thanks!

src/ocean_emulators/utils/data.py
alxmrs: haha, I figured it was something like that! Thanks for preventing that from happening next time.

PR #243, opened 2025-05-16
Activation Checkpointing

jder:
I am looking into fitting the high res models on our GPUs, so I tried a couple options for activation checkpointing: checkpoint each ConvNeXtBlock and other top-level layers ("all") or checkpointing just simple scalings/nonlinearities ("simple"). 

Results are here: https://wandb.ai/m2lines/ocean-emulators/reports/Checkpointing-2025-05-16--VmlldzoxMjgwOTIyOA

tl;dr: "all" saves about 67% of GPU memory usage (!) but at a performance cost of about 10%. "simple" saves about 20% of GPU memory but has practically no performance degradation. There are probably strategies in between those two we could look at, too. (Perhaps also some wins to be had by using the extra space to pre-load future steps worth of data or similar.)

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -126,9 +126,11 @@ class SamudraConfig(BaseConfig):
 
     checkpointing: Checkpointing | None = Field(
         default=None,
-        description="Checkpointing strategy for the model; "
-        "'blocks' for recomputing each CoreBlock, "
-        "'simple' for recomputing only cheap layers like scales and nonlinearities",
+        description="""Strategy for storing activations for the model for use in
+        the backward pass. If not set, the model will store all activations in memory
+        (fast but lots of memory). If set to 'blocks', the model will recompute each
+        CoreBlock in the backward pass. If set to 'simple', the model will recompute
+        only cheap layers like scales and nonlinearities.""",
     )
 
 
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -105,7 +105,7 @@ class CorrectorConfig(BaseConfig):
 
 DownSamplingBlocks = Literal["avg_pool", "max_pool"]
 UpSamplingBlocks = Literal["bilinear_upsample", "transposed_conv"]
-Checkpointing = Literal["blocks", "simple"]
+Checkpointing = Literal["all", "simple"]
 
 
 class SamudraConfig(BaseConfig):
@@ -128,9 +128,10 @@ class SamudraConfig(BaseConfig):
         default=None,
         description="""Strategy for storing activations for the model for use in
         the backward pass. If not set, the model will store all activations in memory
-        (fast but lots of memory). If set to 'blocks', the model will recompute each
-        CoreBlock in the backward pass. If set to 'simple', the model will recompute
-        only cheap layers like scales and nonlinearities.""",
+        (fast but lots of memory). If set to 'all', the model will recompute each
+        top-level layer (CoreBlocks, scaling layers, etc.) in the backward pass.
+        If set to 'simple', the model will recompute only cheap layers like scales
+        and nonlinearities.""",
     )
 
 
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -4,9 +4,7 @@ from typing import Annotated, Any, Literal
 
 import cftime
 import torch
-from pydantic import PlainSerializer, PlainValidator, WithJsonSchema
-
-from pydantic import Field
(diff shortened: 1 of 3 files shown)


suryadheeshjith:
This is great, I am glad we are finally working on this!

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -121,7 +123,10 @@ def forward(self, fts: torch.Tensor) -> torch.Tensor:
                 fts = torch.nn.functional.pad(
                     fts, (0, 0, self.N_pad, self.N_pad), mode="constant"
                 )
-            fts = layer(fts)
+            if self.checkpoint_simple and not isinstance(layer, nn.Conv2d):
```
suryadheeshjith, line 128: Is this checkpointing everything including convnext blocks? I believe this is only avoiding checkpointing of a single Conv2d layer?

jder: In which mode are you talking about? In "blocks" mode we avoid saving activations within each convnext block (ie we have to recompute each full block on the backward pass). In "simple" mode we avoid saving activations for each non-conv2d layer inside the convnext blocks instead (so all the convolution results are saved).

suryadheeshjith: It might be better to rename simple to conv, and checkpoint all conv2d layers including the one in the Samudra model (not just the ones in conv/convnext blocks).

suryadheeshjith: simple is a bit vague

jder: By "simple" I meant we avoid saving activations for simple (ie cheap) layers, not that the strategy is simple. Open to other names!

I'm not sure what you mean by "checkpoint all conv2d layers" -- you mean you want to save the activations of all conv2d layers? That is what "simple" already does (since anything wrapped in `checkpoint` is *not* saved). Or you mean you want to avoid saving them?

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -121,7 +123,10 @@ def forward(self, fts: torch.Tensor) -> torch.Tensor:
...
+                fts = torch.utils.checkpoint.checkpoint(layer, fts, use_reentrant=False)
+            else:
+                fts = layer(fts)
```
alxmrs, line 131: This arrangement of conditionals makes the most sense to me -- what do you think?

```suggestion
            elif self.checkpoint_simple:
                fts = torch.utils.checkpoint.checkpoint(layer, fts, use_reentrant=False)
            else:
                fts = layer(fts)
```

jder: This is not equivalent -- in your updated code the layer is never invoked for Conv2D. I have a slight preference for keeping in the way it is over duplicating the line `fst = layer(fts)` in the first if block. WDYT?

alxmrs: I'm a little confused -- in the original code, isn't it also never invoked for Conv2D? isn't that what the `not isinstance(layer, nn.Conv2d)` does in the `if` statement?

alxmrs: oh! In the else block!

alxmrs: Ok, I follow and agree the original way is better.

alxmrs (approved):
LGTM. I don't have feedback as it looks like you create a simple way to plumb a config to two types of checkpointing. I learned a lot reading this PR and the docs for this new-to-me torch util. 

Great work on the performance improvement!

src/ocean_emulators/config.py
```diff
@@ -126,9 +126,11 @@ class SamudraConfig(BaseConfig):
 
     checkpointing: Checkpointing | None = Field(
         default=None,
-        description="Checkpointing strategy for the model; "
-        "'blocks' for recomputing each CoreBlock, "
-        "'simple' for recomputing only cheap layers like scales and nonlinearities",
+        description="""Strategy for storing activations for the model for use in
```
alxmrs, line 129: Nice, this is very helpful.

src/ocean_emulators/models/samudra.py
```diff
@@ -144,7 +165,10 @@ def forward_once(self, fts):
                 fts = torch.nn.functional.pad(
                     fts, (0, 0, self.N_pad, self.N_pad), mode="constant"
                 )
-            fts = layer(fts)
+            if self.checkpoint_blocks:
```
jder: Let's rename checkpoint_blocks since it also checkpoints non-blocks (eg upsamples)

PR #245, opened 2025-05-19
Apply same fix to benchmarks, too

jder:
Looks like this wasn't a perf regression, it was a failure. Sorry, I should have checked here after finding it in the other two places!


alxmrs (approved):
Ah! I missed it, too. Thank you!

alxmrs:
I was wondering why we weren't getting the normal performance alert. This makes sense.

PR #246, opened 2025-05-20
Switch to using EC2 for CI workers

mihasya:
This changeset switches the repo to use OpenAthena's EC2 account to perform CI checks.

**Before Merging**

- [x] update EC2 IAM policy to also match parent repo (this one)
- [x] add appropriate IAM Role as `AWS_ROLE` secret

**After Merging**

- [ ] make sure there isn't a significant impact to benchmark results, find a way to document if so.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -14,47 +14,40 @@ permissions:
   deployments: write # Allow to deploy to GitHub pages
 
 jobs:
-  create-runner:
+  start-aws-runner:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
     outputs:
-      label: ${{ steps.create-runner.outputs.label }}
+      mapping: ${{ steps.aws-start.outputs.mapping }}
+      instances: ${{ steps.aws-start.outputs.instances }}
     steps:
-      - id: create-runner
-        uses: Open-Athena/gce-github-runner@main
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
         with:
-          token: ${{ secrets.GH_SA_TOKEN }}
-          project_id: ${{ secrets.GCP_PROJECT_ID }}
-          service_account_key: ${{ secrets.GCP_SA_KEY }}
-          runner_service_account: ${{ secrets.GCP_RUNNER_SA_EMAIL }}
-          image_project: deeplearning-platform-release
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -113,3 +106,26 @@ jobs:
           comment-on-alert: true
           fail-on-alert: true
           alert-comment-cc-users: "@alxmrs,@jder"
+
+  stop-aws-runner:
```
jder, line 110: Ah, nice, this probably also fixes most cases of https://github.com/Open-Athena/ops/issues/8

.github/workflows/benchmarks.yml
```diff
@@ -14,47 +14,40 @@ permissions:
   deployments: write # Allow to deploy to GitHub pages
 
 jobs:
-  create-runner:
+  start-aws-runner:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
```
jder, line 21: Are these used given we have the GH_PAT already?

.github/workflows/test-gpu.yml
```diff
@@ -1,63 +1,43 @@
...
-      - main
-  workflow_dispatch:
-
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
-
+  push
+env:
+  AWS_REGION : "us-east-1"
```
jder, line 5: Why is this here but not in benchmarks? (Sorry for the duplication between these two…)

jder (approved):
Thanks, looks great! Do we have a successful run on this repo? If not, I think we can get it with a branch on this repo and then running the workflow from here https://github.com/suryadheeshjith/Ocean_Emulator/actions/workflows/test-gpu.yml.

jder:
Oh, sorry, I didn't notice you had auto-merge turned on. (We don't have "must resolve comments" enabled)

.github/workflows/test-gpu.yml
```diff
@@ -1,63 +1,43 @@
...
 name: Test GPU
-
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
-  workflow_dispatch:
```
jder, line 11: Let's restore the workflow_dispatch and pull_request triggers.

PR #249, opened 2025-05-22
Revert "Switch to using EC2 for CI workers"

mihasya:
Reverting while we figure out why EC2 runners are not registering correctly when running on the parent repo.

Reverts suryadheeshjith/Ocean_Emulator#246


jder (approved):
🙇

PR #250, opened 2025-05-22
Switch to using EC2 for CI workers, take 2

mihasya:
This changeset switches the repo to use OpenAthena's EC2 account to perform CI checks.

Unlike its predecessor, #246, this PR depends on a branch of `start-gha-ec2-runner` which contains a fix for a bug that prevented the runner machines from being detected by GHA.

We observe a modest improvement in benchmarks when running on AWS `g6.xlarge` VMs vs `g2-standard-8` Google Cloud VMs.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -12,13 +12,11 @@ on:
 permissions:
   contents: write # Allow updating content on GitHub pages
   deployments: write # Allow to deploy to GitHub pages
+  id-token: write   # This is required for interacting with AWS via OIDC
 
 jobs:
   start-aws-runner:
     runs-on: ubuntu-latest
-    permissions:
-      id-token: write
-      contents: read
     outputs:
       mapping: ${{ steps.aws-start.outputs.mapping }}
       instances: ${{ steps.aws-start.outputs.instances }}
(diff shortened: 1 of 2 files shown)


jder:
Triggered benchmark run here: https://github.com/suryadheeshjith/Ocean_Emulator/actions/runs/15211025480

.github/workflows/benchmarks.yml
```diff
@@ -12,49 +12,40 @@ on:
...
-          disk_size: "200GB"
-          accelerator: "count=1,type=nvidia-l4"
-          # shutdown_timeout: 1800 # uncomment to allow your CI machines to stay around after running for debugging purposes
+          role-to-assume: ${{ secrets.AWS_ROLE }}
+          role-session-name: github-actions-session
+          aws-region: us-east-1
+
+      - name: Create cloud runner
+        id: aws-start
+        uses: mihasya/start-aws-gha-runner@fix-worker-lookup
```
jder, line 33: Should we move this to the open-athena org before merging? Or you just plan to point back at upstream once the fix is merged?

jder (approved):
Thank you for pushing through all this!

.github/workflows/benchmarks.yml
mihasya: Yeah 100% planning to upstream the fix and then update this dep down the road.

PR #253, opened 2025-05-28
Try out 6g.2xlarge to see what that does to benchmark results

mihasya:
@jder can you trigger a manual benchmark run, please? I figure we can manually push to the `gh-pages` branch later to remove it.

Update: oh, I didn't realize I had perms to do this myself. Have kicked it off here https://github.com/suryadheeshjith/Ocean_Emulator/actions/runs/15302652916


jder:
@mihasya can we close this? Did you get the needed info?

mihasya:
Yup, saw an intuitive, modest perf bump from increasing the instance size. If we were interested in this line of inquiry, next steps would be to profile the benchmarks on Google and Amazon, but no need to do that through GHA.

PR #254, opened 2025-05-29
Remove pointer to patched version of GH Action

mihasya:
The `gha_runner` library has pushed a new release with our fix in it.

`start_aws_gha_runner` appears to pull in the latest version of the dependency, so we should no longer need to hardcode a patched version. Whether or not this PR builds successfully will tell us if the fix is being pulled in correctly, as we still have 37 runners registered (>30, the boundary condition for the bug)


jder (approved):
merci!

PR #255, opened 2025-06-05
Switch to torch loader and reduce default number of workers per GPU

jder:
I think the torch loader is a good default now, and I found 4 workers by default overloads the 8-GPU systems I've been using, resulting in very long waits at the start of an epoch. 1-worker-per-GPU seems to be too few on my other host, so I was thinking this might be a better default. (Though of course might need tuning.)

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -78,9 +78,9 @@ class DataConfig(BaseConfig):
     data_stds_path: str = "CM4_5daily_v0.4.0_stds"
     scaling_residuals_file: str | None = None
     static_data_vars: list[str] | None = None
-    num_workers: int = 4
+    num_workers: int = 2
     hist: int = 1
-    loader_version: str = str(LoaderVersion.OM4_EAGER.value)
+    loader_version: str = str(LoaderVersion.OM4_TORCH.value)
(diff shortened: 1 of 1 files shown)


src/ocean_emulators/config.py
```diff
@@ -78,9 +78,9 @@ class DataConfig(BaseConfig):
     data_stds_path: str = "CM4_5daily_v0.4.0_stds"
     scaling_residuals_file: str | None = None
     static_data_vars: list[str] | None = None
-    num_workers: int = 4
+    num_workers: int = 2
     hist: int = 1
-    loader_version: str = str(LoaderVersion.OM4_EAGER.value)
+    loader_version: str = str(LoaderVersion.OM4_TORCH.value)
```
alxmrs, line 83: 🚀

alxmrs (commented):
Defaulting to 2 workers does SGTM.

jder:
@alxmrs if you're happy with this, can you approve so we can merge?

alxmrs:
Something is weird, I don't see any way to approve in the GH interface

alxmrs:
Further, I thought I had approved sometime before

alxmrs (approved):
LGTM

PR #256, opened 2025-06-06
Deduplicate start/stop jobs, remove patched version of action

jder:
I resisted doing this because this is more moving pieces and also more code, but this is also at least the third time we've had them drift out of sync (in this case, one was using the mihasya patched version and the other not), so I think it's probably worth it.

### DIFF
--- a/.github/workflows/aws-runner-start.yml
+++ b/.github/workflows/aws-runner-start.yml
@@ -33,7 +33,6 @@ on:
 
 permissions:
   id-token: write # This is required for interacting with AWS via OIDC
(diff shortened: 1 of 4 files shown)


jder:
Benchmark run is here https://github.com/suryadheeshjith/Ocean_Emulator/actions/runs/15492375301

alxmrs (approved):
Did a brief review and this looks reasonable, but I would get @mihasya's input before a merge.

mihasya (approved):
This is an obvious and useful improvement 😎

PR #258, opened 2025-06-10
Complain if the validation or inference time ranges overlap with train

jder:
I did this by accident last week so let's prevent people from doing it.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -1,6 +1,6 @@
 from functools import cached_property
 from pathlib import Path
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, Literal, Self
 
 import cftime
 import torch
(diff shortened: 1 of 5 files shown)


src/ocean_emulators/config.py
```diff
@@ -68,6 +69,20 @@ class TimeConfig(BaseConfig):
     def time_slice(self) -> slice:
         return slice(self.start.datetime, self.end.datetime)
 
+    def overlaps(self, other: "TimeConfig") -> bool:
```
```suggestion
    def overlaps(self, other: Self) -> bool:
```

src/ocean_emulators/train.py
```diff
@@ -156,6 +156,18 @@ def __init__(self, cfg: TrainConfig) -> None:
 
         # Dataloaders
         logger.info(f"Loading data")
+        if cfg.train_time.overlaps(cfg.val_time):
```
alxmrs, line 159: Nice

src/ocean_emulators/config.py
alxmrs: (Optional nit)

tests/conftest.py
```diff
@@ -441,6 +447,7 @@ def train_config(
             "--backend",
             backend,
         ]
+        + extra_args
```
alxmrs: I really like this affordance!

alxmrs (approved):
LGTM, good idea.

PR #261, opened 2025-06-16
Notes on how to reproduce OM4 half-deg preprocessing.

alxmrs:
I've refreshed the contributing guide, adding a data engineering section. Here, I explain how to clone the data and how to start a coiled notebook to create the half-deg data. The docs indicate that this all may change soon -- and that is my plan. I am happy to refresh these docs overall.


CONTRIBUTING.md
```diff
@@ -382,4 +377,68 @@ uv run memray run src/ocean_emulators/train.py --config configs/train_om4.yaml -
...
+3. Connect to a cloud provider. Here are a few commands to set up the top three cloud providers:
+   ```shell
+   coiled setup aws
+   coiled setup gcp
+   coiled setup azure
+   ```
+
+4. Once you have coiled infra ready-to-go, spin up a notebook with a Pangeo docker image via the following script:
+   ```shell
+   ./scripts/.coiled-pangeo-notebook
```
jder: I think we should stop using the `.` prefix for scripts. How about `./scripts/coiled-pangeo-notebook.sh`? (Or just remove the script and document the command here)

CONTRIBUTING.md
```diff
@@ -382,4 +377,68 @@ uv run memray run src/ocean_emulators/train.py --config configs/train_om4.yaml -
...
+   coiled notebook start --container pangeo/pangeo-notebook --disk-size 256
+   ```
+   > Note: Coiled _does_ come with a `--sync` flag that is supposed to replicate your local environment with the remote
+   > notebook. I have found this doesn't work, likely, because my local environment contains a large data cache.
+   > You may try running the above command with the `--sync` flag added after switching into the `./notebooks` directory
+   > -- I bet that might work.
+
+5. Last, but not least, copy the most up-to-date pre-processing notebook into the Jupyter Lab instance that just opened
+   up. This can be found in the `./notebooks/` directory with the name "YYYY-MM-DD-data_preprocess.ipynb". We recommend
+   selecting the latest / most recent version.
```
jder: What do you think about checking this in as a .py file with [cell markers](https://code.visualstudio.com/docs/python/jupyter-support-py#_jupyter-code-cells) so we can use version control on it? And then get rid of the date? (There is an "import notebook to script" command in VS Code which you can use to make it.)

configs/schemas/BlockConfig.json
```diff
@@ -0,0 +1,46 @@
+{
```
jder: These should be git ignored, can you remove them please?

jder (approved):
Just a few suggestions, thanks!

CONTRIBUTING.md
alxmrs: Yeah, I agree it'd be nice to remove them. I was going for consistency, but this is a pattern worth breaking.

CONTRIBUTING.md
alxmrs: Any idea if we can import a .py file as a notebook in Jupyter lab? This would be difficult to run as a script bc it requires dependencies managed by docker.

configs/schemas/BlockConfig.json
alxmrs: gitcha, will do.

CONTRIBUTING.md
jder: I don't see an obvious way of turning a py file into a jupyter lab notebook, but looks like you can
`coiled run --container pangeo/pangeo-notebook my-script.py` https://docs.coiled.io/user_guide/cli-jobs.html

jder: (If you want to just leave it as a notebook for now that's ok too)

alxmrs: I'll leave it as a notebook for now -- the near-term plan is to re-write it as a script in some new execution framework (Cubed, Xarray-Beam, etc.).

PR #262, opened 2025-06-17
Support absolute URLs for file paths, remove '*' support

jder:
At the moment, we don't support `root` being a URL but this seems like a step in the right direction.

I also don't think we need the "*" support, and defaulting to `debug: false` in eval is consistent with the other configurations.


jder:
Apparently tests were relying on using .nc files, so I re-added that. That can't work if you pass `consolidated=True` but that seems to be redundant anyway (the default behavior of consolidated=None is to try to read consolidated metadata first and fall back to non-consolidated).

configs/eval_om4.yaml
```diff
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=schemas/EvalConfig.json
 
-debug: true
+debug: false
```
alxmrs, line 3: Nice, thanks.

jder:
@alxmrs if you have a sec, can you review this unblocking PR given the discussion we had this morning?

src/ocean_emulators/utils/data.py
```diff
@@ -228,32 +230,18 @@ def from_dataset(cls, dataset: xr.Dataset, *, name: str = "dataset") -> Self:
...
-                chunks=chunks,
-                consolidated=True,
-                engine="zarr",
-            )
-
-        means = xr.open_dataset(
-            root / cfg.data.data_means_path,
-            engine="netcdf4" if cfg.data.data_means_path.endswith(".nc") else "zarr",
-            chunks=chunks,
+        data = _open_dataset(cfg.data.data_path, root, chunks)
```
alxmrs, line 235: 🐑 🐑 🐑 
An idea to further reduce repetition: 
```suggestion
        open_ = functools.partial(_open_dataset, root=root, chunks=chunks)
        data, means, stds = open_(cfg.data.data_path), open_(cfg.data.data_means_path), open_(cfg.data.data_stds_path)
```

src/ocean_emulators/utils/data.py
```diff
@@ -737,3 +725,30 @@ def _parse_level(x) -> float:
...
     return data
+
+
+def _open_dataset(
+    path: str, root: Path, chunks: dict[str, int] | None = None
+) -> xr.Dataset:
+    """Open a dataset with appropriate engine and parameters based on path type.
+
+    Args:
+        path: Path to the dataset, or absolute url (eg s3://...)
```
alxmrs:
```suggestion
        path: Path to the dataset, or absolute url (e.g. "s3://...")
```

src/ocean_emulators/utils/data.py
```diff
@@ -737,3 +725,30 @@ def _parse_level(x) -> float:
...
+        root: root directory to prepend to relative paths
+        chunks: Optional chunking parameters for dask
+
+    Returns:
+        xarray.Dataset
```
alxmrs:
```suggestion
        An `xarray.Dataset` found at the `path` location.
```

src/ocean_emulators/utils/data.py
alxmrs: Reading the `_open_dataset` function below, I almost think it would be better suited to be a function that addresses the data location, like `parse_location`, since the chunks and netcdf/zarr logic is otherwise the same.

alxmrs: Combined with `functools.partial`, I think this could make for similarly succinct open call, while addressing the data location annoyance.

jder: I agree that we should break out the location logic, which we can do as discussed last week into another type. I try to avoid `functools.partial` because it defeats type checkers and in this case saves only minor and local duplication.

jder:
@alxmrs can you approve this please?

alxmrs (approved):
LGTM

PR #267, opened 2025-06-18
Created an icechunk repo config ADT

alxmrs:
It will either create or open an icechunk repo from a storage location, or do noting when not configured.

This commit also allows Zarr v3

### DIFF
--- a/src/ocean_emulators/config_schema.py
+++ b/src/ocean_emulators/config_schema.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import pydantic
 import yaml
 
-from ocean_emulators.config import EvalConfig, TrainConfig
+from ocean_emulators.config import EvalConfig, IceChunkRepoConfig, TrainConfig
 
 
 def get_pydantic_models(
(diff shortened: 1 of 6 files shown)


src/ocean_emulators/config_schema.py
```diff
@@ -139,6 +139,7 @@ def main():
     # Get all available models
     models = get_pydantic_models(TrainConfig)
     models.update(get_pydantic_models(EvalConfig))
+    models.update(get_pydantic_models(IceChunkRepoConfig))
```
alxmrs, line 142: @jder I wasn't sure how to generate this config schema without modifying the script. Is there a better way?

jder: It should automatically get detected by the get_pydantic_models call above (which walk the TrainConfig and EvalConfig to find sub-configs). If not, it's probably a bug in those.

alxmrs:
Closing this PR because we found another decent model for storing checkpoints.

PR #278, opened 2025-06-23
Adding SPDL adapter

alxmrs:
This PR follows the SPDL migration guide to build an equivalent pipeline to what we have today: https://facebookresearch.github.io/spdl/main/migration/pytorch.html

I include a test to prove that the data with this flag turned on is equivalent to not using the pipeline. After this release, I plan on experimenting tuning this pipeline with the aid of the library's performance logging utilities and documentation recommendations.

### DIFF
--- a/.github/workflows/test-gpu.yml
+++ b/.github/workflows/test-gpu.yml
@@ -45,7 +45,7 @@ jobs:
         run: uv sync --locked --dev --all-extras
 
       - name: Run pytest
-        run: uv run --locked pytest -n 2 -v -m "not manual and cuda"
+        run: uv run --locked pytest -v -m "not manual and cuda"
 
   stop-aws-runner:
     uses: ./.github/workflows/aws-runner-stop.yml
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -19,7 +19,7 @@ import numpy as np
 import torch
 import torch.nn as nn
 import xarray as xr
-from spdl.pipeline import Pipeline  # type: ignore
+from spdl.pipeline import Pipeline
 from torch.utils.data import (
     ConcatDataset,
     DataLoader,
@@ -83,6 +83,7 @@ from ocean_emulators.utils.loss import (
 )
 from ocean_emulators.utils.train import (
     CheckpointPaths,
+    SizedPipeline,
     as_spdl_pipeline,
     collate_inference_data,
     collate_train_data,
@@ -518,7 +519,7 @@ class Trainer:
         metric_logger.add_meter("lr", SmoothedValue(window_size=1, fmt="{value:.6f}"))
         header = f"Training Epoch: [{epoch}]"
 
-        iterator: DataLoader | Pipeline = self.train_loader
+        iterator: DataLoader | SizedPipeline = self.train_loader
 
         if self.use_spdl:
             iterator = as_spdl_pipeline(
@@ -616,7 +617,7 @@ class Trainer:
         metric_logger = MetricLogger(delimiter="  ")
         header = f"One-Step Validation Epoch: [{epoch}]"
 
-        iterator: DataLoader | Pipeline = self.val_loader
+        iterator: DataLoader | SizedPipeline = self.val_loader
 
         if self.use_spdl:
             iterator = as_spdl_pipeline(
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -12,14 +12,14 @@ import warnings
 from collections import OrderedDict
 from collections.abc import Callable, Sequence
 from functools import partial
-from typing import Any, assert_never
+from typing import Any, assert_never, cast
 
 import dask
 import numpy as np
 import torch
 import torch.nn as nn
 import xarray as xr
-from spdl.pipeline import Pipeline, PipelineBuilder
+from spdl.pipeline import Pipeline  # type: ignore
 from torch.utils.data import (
     ConcatDataset,
     DataLoader,
@@ -83,6 +83,7 @@ from ocean_emulators.utils.loss import (
 )
 from ocean_emulators.utils.train import (
     CheckpointPaths,
+    as_spdl_pipeline,
     collate_inference_data,
     collate_train_data,
 )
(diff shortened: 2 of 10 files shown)


src/ocean_emulators/utils/logging.py
```diff
@@ -183,7 +189,9 @@ def log_every(self, data_loader: DataLoader["TrainData"], print_freq, header=Non
                 data_load_time.update(obj.load_stats.load_time_seconds)
             yield obj
             iter_time.update(time.perf_counter() - end)
-            if i % print_freq == 0 or i == len(data_loader) - 1:
+            if hasattr(data_loader, "__len__") and (
```
alxmrs: I don't think spdl pipelines have a `__len__`. So, this skips all the logging that depend on knowing the length of the iterator.

jder: This seems to totally remove the per-step reporting which very helpful to understand what's going on. Can we maybe pass a length instead?

src/ocean_emulators/utils/logging.py
```diff
@@ -202,10 +210,7 @@ def log_every(self, data_loader: DataLoader["TrainData"], print_freq, header=Non
             end = time.perf_counter()
         total_time = time.perf_counter() - start_time
         total_time_str = str(datetime.timedelta(seconds=int(total_time)))
-        logging.info(
-            f"{header} Total time: {total_time_str} "
-            f"({total_time / len(data_loader):.4f} s / it)"
-        )
+        logging.info(f"{header} Total time: {total_time_str} ")
```
jder: Did you mean to remove the per-step information unconditionally? I think we can track the number of steps (`i`) and use that.

src/ocean_emulators/utils/train.py
```diff
@@ -1,8 +1,10 @@
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from itertools import tee
 from pathlib import Path
 
 import torch
+import torch.utils.data
+from spdl.pipeline import Pipeline, PipelineBuilder  # type: ignore
```
jder: As mentioned in slack, we can fix this with something like 

```
[[tool.mypy.overrides]]
module = ["spdl.*"]
follow_untyped_imports = true
```

in pyproject.toml

tests/test_datasets.py
```diff
@@ -430,6 +443,19 @@ def make_config(src: DataSource):
     assert_equal_samples(original_samples, new_samples)
 
 
+@pytest.mark.parametrize("data_source", ["remote-om4"], indirect=True)
+def test_loader__is_same__with_spdl(train_config):
+    data_config = train_config.data.model_copy(update={"use_spdl": True})
+    spdl_config = train_config.model_copy(update={"data": data_config})
```
jder: How about passing a new arg to make_loader (like version, time_config etc) rather than doing config surgery? I think that helps move toward a world where we don't need to use config objects in non-integration tests.

src/ocean_emulators/utils/train.py
```diff
@@ -65,3 +67,27 @@ def ema_checkpoint_path(self) -> Path:
...
+        .add_source(range(len(dataset)))  # type: ignore
+        .pipe(
+            dataset.__getitem__,
+            concurrency=num_workers,
+            output_order="input",
+        )
+        .aggregate(batch_size)
+        .pipe(collate_fn)
+        .add_sink(prefetch_factor)
+        .build(num_threads=num_workers)
```
jder, line 105: Am I understanding correctly that the effect of this pipeline will be to call `__getitem__` across `num_workers` threads rather than `num_workers` processes? 

Also I think this produces data in order rather than shuffled?

jder (commented):
OK, some questions and nits from me in a quick review, sorry I need to run but happy to look again in more depth tonight/tomorrow or when that makes sense.

alxmrs:
It looks like SPDL is creating some sort of deadlock combined with our parallel pytest runner 😢  -- the tests have been running for ~5 hours. 

Once I fix this commit, I'm going to update the test config to try to address this.

src/ocean_emulators/utils/train.py
alxmrs: > Am I understanding correctly that the effect of this pipeline will be to call __getitem__ across num_workers threads rather than num_workers processes?

Yes, that is also my understanding. We're shifting to thread-based parallelism. The concurrency part should lets us combine that with true concurrency (IO bound stuff will use asyncio to pull) and increase resources there independently of the number of workers. 

> Also I think this produces data in order rather than shuffled?

Thanks for calling this out! I just fixed this and also implemented `drop_last`.

tests/test_datasets.py
alxmrs: Thanks, that sound good to me.

alxmrs: PTAL at some nuance with this in the `make_loader` function -- I allow the config to set this value in addition to explicitly passing it.

src/ocean_emulators/train.py
```diff
@@ -514,69 +518,88 @@ def train_one_epoch(self, epoch):
...
+        iterator: DataLoader | SizedPipeline = self.train_loader
+
+        if self.use_spdl:
+            iterator = as_spdl_pipeline(
+                self.train_dataset,
+                num_workers=self.num_workers,
+                batch_size=self.batch_size,
+                drop_last=True,
             )
+            iterator.start()
```
jder: It seems like it would be more clear if `self.train_loader` was populated with some object which hid the difference between spdl and the previous DataLoaders, rather than this code having to test `self.use_spdl` and ignore self.train_loader when it's true. That object could hold the length of the dataset rather than `setattr`ing the pipeline's `__len__` and could have a do-nothing `start` and `stop` methods for the original data loader. Maybe worth discussing live? Am I missing some constraint?

.github/workflows/test-gpu.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         run: uv sync --locked --dev --all-extras
 
       - name: Run pytest
-        run: uv run --locked pytest -n 2 -v -m "not manual and cuda"
+        run: uv run --locked pytest -v -m "not manual and cuda"
```
jder, line 48: Just noting we should revert this before merging.

tests/test_datasets.py
```diff
@@ -53,7 +55,8 @@ def make_loader(
     time_config: TimeConfig | None = None,
     drop_last: bool = True,
     version: LoaderVersion | None = None,
-) -> Generator[DataLoader, None, None]:
+    use_spdl: bool = False,
```
jder, line 59: I think like `time_config` and `version` we should default to `None`, which uses the config's value for `use_spdl`, that way passing `use_spdl=False` will disable it (I think it currently does nothing if it's set to true in the config?)

jder (changes requested):
One main suggestion from me, lemme know if it's easier to discuss live.

alxmrs:
Closing this in favor of #289.

PR #283, opened 2025-06-25
Tweaks while running half-degree experiments

jder:
The 3 commits are independent so probably easiest to review that way. (Happy to break into separate PRs but I think they're all pretty uncontroversial.)

* Updates config for the half-degree experiments to load remotely when desired and to avoid doing inference during training.
* Adds a flag to load xarray variables concurrently based on [this suggestion](https://github.com/pydata/xarray/issues/8965#issuecomment-2083787484). This helps a lot locally with small datasizes (3x improvement for me) but with larger batch sizes it seems to not help (maybe GIL contention?).
* Swaps to the the "spawn" multiprocessing type since the s3fs requires this when passing those stores around.
* Adds more logging statements so you can tell what's going on more easily
* Removes some logic we had about tweaking the number of workers based on the number of GPUs in use -- the worker counts are are *already* per GPU so this was using `num_gpus * num_gpus * num_workers` workers in total. Also allows us to use multiple workers for CPU jobs, for testing purposes.

### DIFF
--- a/configs/data/om4_halfdeg_remote.yaml
+++ b/configs/data/om4_halfdeg_remote.yaml
@@ -1,0 +1,6 @@
+# yaml-language-server: $schema=../schemas/DataConfig.json
+
+data_path: s3://emulators/sd5313/OM4_highres/om4_halfdeg.zarr
+data_means_path: s3://emulators/sd5313/OM4_highres/om4_halfdeg_means.zarr
+data_stds_path: s3://emulators/sd5313/OM4_highres/om4_halfdeg_stds.zarr
+static_data_vars: []
--- a/configs/data/om4_halfdeg_remote.yaml
+++ b/configs/data/om4_halfdeg_remote.yaml
@@ -1,5 +1,9 @@
 # yaml-language-server: $schema=../schemas/DataConfig.json
 
+# TODO(jder): the URLs below assume you have FSSPEC_S3_ENDPOINT_URL=https://nyu1.osn.mghpcc.org
+# as well as the appropriate AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY (or other S3 credentials) in your environment.
+# Once we have a clear story for launching this task remotely we should remove this implicit requirement.
+
 data_path: s3://emulators/sd5313/OM4_highres/om4_halfdeg.zarr
 data_means_path: s3://emulators/sd5313/OM4_highres/om4_halfdeg_means.zarr
 data_stds_path: s3://emulators/sd5313/OM4_highres/om4_halfdeg_stds.zarr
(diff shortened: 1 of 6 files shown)


configs/data/om4_halfdeg_remote.yaml
```diff
@@ -0,0 +1,6 @@
+# yaml-language-server: $schema=../schemas/DataConfig.json
+
+data_path: s3://emulators/sd5313/OM4_highres/om4_halfdeg.zarr
+data_means_path: s3://emulators/sd5313/OM4_highres/om4_halfdeg_means.zarr
+data_stds_path: s3://emulators/sd5313/OM4_highres/om4_halfdeg_stds.zarr
```
alxmrs, line 9: Do we have control or access to the global `emulators` bucket? Or, does this map to the OSN pod?

src/ocean_emulators/datasets.py
```diff
@@ -735,3 +740,22 @@ def _get_x_index(self, idx: int, step: int) -> xr.DataArray:
 
         window_index = idx + step * (self.hist + 1) * self.stride
         return self.rolling_indices.isel(window=window_index, drop=True)
+
+
+def concurrent_compute(
+    *datasets: xr.Dataset,
+) -> None:
```
alxmrs, line 747: A pretty good, albeit scrappy, way to concurrently load the data!

alxmrs: An aside: we should consider comparing this approach vs just using dask

tests/test_trainer.py
```diff
@@ -78,23 +78,11 @@ def test_checkpoint(trainer_pair: TrainPair, caplog):
                 "1975-09-01",
             ],
         ),
-        (
```
alxmrs, line 81: Nice

alxmrs (approved):
LGTM, The only note is a comment about another (performance) experiment to try with remote data.

configs/data/om4_halfdeg_remote.yaml
jder: Hrm, yeah, fair question. This presumes you have FSSPEC_S3_ENDPOINT_URL set to https://nyu1.osn.mghpcc.org and the appropriate credentials for the OSN emulators bucket in your environment. I could add these as config parameters, but whether we want to do that or keep them as env vars kinda depends on what we want to do for machine setup + orchestration.

jder: I added a comment at least.

src/ocean_emulators/datasets.py
jder: Seems worth trying. FWIW the main problem with Dask was the CPU overhead of doing whatever graph stuff it was doing which this has none of. On the other hand the torch loader now does very little before converting to numpy so it might be that dask would have minimal overhead here.

PR #284, opened 2025-06-27
Add a "Location" to support S3 and local paths

jder:
This lets us:
* Add structured info for locations (eg s3 endpoint URLs)
* Still use raw strings most of the time
* Get some type safety to know if we've correctly resolved these against the root before opening.
* (In a future PR) detect the need for "spawn" multiprocessing (required for S3 but not for local files) which has a performance cost.

Also removes the "gantry" boolean because I think we can use `data_root:/` if we want that


jder:
I should probably add some tests for the Location types but thought I'd get your feedback first!

src/ocean_emulators/config.py
```diff
@@ -185,17 +200,13 @@ def nets_dir(self) -> Path:
...
-            return Path("/")
-        else:
-            if self.cluster_data_dir is None:
-                raise ValueError(
-                    "cluster_data_dir must be set, try"
-                    " --experiment.cluster_data_dir=path/to/data"
-                )
-            else:
-                return Path(self.cluster_data_dir)
+    def resolved_data_root(self) -> ResolvedLocation:
```
alxmrs, line 203: I like this runtime validation and type enforcement!

src/ocean_emulators/utils/data.py
```diff
@@ -230,24 +228,16 @@ def from_dataset(cls, dataset: xr.Dataset, *, name: str = "dataset") -> Self:
...
-            chunks,
-        )
-        stds = _open_dataset(
-            cfg.data.data_stds_path,
-            root,
-            chunks,
-        )
+        data = root.resolve(cfg.data.data_location).open(chunks)
+        means = root.resolve(cfg.data.data_means_location).open(chunks)
+        stds = root.resolve(cfg.data.data_stds_location).open(chunks)
```
alxmrs, line 235: This is slick!

src/ocean_emulators/utils/data.py
```diff
@@ -230,24 +228,16 @@ def from_dataset(cls, dataset: xr.Dataset, *, name: str = "dataset") -> Self:
...
 
         dask = "with_dask" if use_dask else "without_dask"
 
         return cls(
-            name=f"{cfg.experiment.name}-{cfg.experiment.data_dir.name}-{dask}",
+            name=f"{cfg.experiment.name}-{root}-{dask}",
```
alxmrs: I love how succinct this is now.

src/ocean_emulators/utils/location.py
```diff
@@ -0,0 +1,132 @@
...
+
+class ResolvedLocation(ABC):
+    """A location which is ready to be opened or resolved against."""
+
+    @abstractmethod
+    def open(self, chunks: dict[str, int] | None = None) -> xr.Dataset:
+        pass
+
+    @abstractmethod
+    def resolve(self, location: "Location") -> "ResolvedLocation":
```
alxmrs, line 42: 🐑  Does this work here?
```suggestion
    def resolve(self, location: "Location") -> Self:
```

src/ocean_emulators/utils/location.py
```diff
@@ -0,0 +1,132 @@
...
+        pass
+
+    def __truediv__(self, other: "Location") -> "ResolvedLocation":
```
alxmrs, line 45: 🐑 
```suggestion
    def __truediv__(self, other: "Location") -> Self:
```

src/ocean_emulators/utils/location.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def open(self, chunks: dict[str, int] | None = None) -> xr.Dataset:
+        # TODO(jder): could consider passing credentials here
+        # rather than relying on the environment
+
+        return xr.open_dataset(
+            self.url(),
+            backend_kwargs={"storage_options": {"endpoint_url": self.endpoint_url}},
+            engine="zarr",
+            chunks=chunks,
```
alxmrs, line 74: I believe `chunks=None` and `chunks={}` are different behaviors and I think we want the latter -- unless: do you mean to defer opening the Zarr data?

src/ocean_emulators/utils/location.py
```diff
@@ -0,0 +1,132 @@
...
+    endpoint_url: str | None = None
+    bucket: str
+    path: str
+
+    def open(self, chunks: dict[str, int] | None = None) -> xr.Dataset:
+        # TODO(jder): could consider passing credentials here
+        # rather than relying on the environment
+
+        return xr.open_dataset(
+            self.url(),
```
alxmrs, line 71: I love this interface!

src/ocean_emulators/utils/location.py
```diff
@@ -0,0 +1,132 @@
...
+            backend_kwargs={"storage_options": {"endpoint_url": self.endpoint_url}},
+            engine="zarr",
+            chunks=chunks,
+        )
+
+    def url(self) -> str:
+        path = quote(self.path.lstrip("/"))
```
alxmrs, line 78: TIL about the `quote` function.

src/ocean_emulators/utils/location.py
```diff
@@ -0,0 +1,132 @@
...
+        pass
+
+    def __truediv__(self, other: "Location") -> "ResolvedLocation":
+        return self.resolve(other)
+
+
+class S3Location(ResolvedLocation, BaseModel):
+    """An S3 bucket, assuming credentials in your environment.
+
+    For example:
```
alxmrs, line 52: Thanks for including the yaml config example, that's useful.

src/ocean_emulators/utils/location.py
```diff
@@ -0,0 +1,132 @@
...
+      path: /path/to/data
+    ```
+    """
+
+    type: Literal["local"] = "local"
+    path: Path
+
+    def open(self, chunks: dict[str, int] | None = None) -> xr.Dataset:
+        engine = "netcdf4" if self.path.suffix == ".nc" else "zarr"
+        return xr.open_dataset(self.path, engine=engine, chunks=chunks)
```
alxmrs, line 111: See the other `chunks` note above. Maybe at the call sight we pass `chunks={}`? I need to double check.

src/ocean_emulators/utils/location.py
```diff
@@ -0,0 +1,132 @@
...
+      type: local
+      path: /path/to/data
+    ```
+    """
+
+    type: Literal["local"] = "local"
+    path: Path
+
+    def open(self, chunks: dict[str, int] | None = None) -> xr.Dataset:
+        engine = "netcdf4" if self.path.suffix == ".nc" else "zarr"
```
alxmrs, line 110: I like how this is adapted for how we use the system. I think it's still poignant in that NetCDF is a local fs format anyway.

alxmrs:
Your approach is quite type-safe. I like how you create abstractions from the same concept based around how they're used. I think you do a good job at building affordances around how we care they're used rather than the varying components for how data locations could be specified. Good job and I look forward to seeing this PR fully-baked.

alxmrs:
Very rusty, well done.

src/ocean_emulators/utils/location.py
jder: Yeah, I almost did that but we can't promise that resolve returns the class of `self` -- if you have `self: LocalLocation` and `location: S3Location` then `self.resolve(location)` returns a S3Location, not a LocalLocation (in this case, Self would resolve to a LocalLocation). All we can promise is that it's *some* `ResolvedLocation`, not necessarily the same type as the instance you're calling it on.

src/ocean_emulators/utils/location.py
jder: Yeah, at the callsite we pass `{}` when we want dask. This is the same default as xarray.open_dataset, I think?

src/ocean_emulators/utils/location.py
jder: (Sorry, I ninja-edited my comment when I went to actually go do this and realized we can't)

jder:
@alxmrs I added some tests and rebased on the latest main, do you want to take another look?

src/ocean_emulators/utils/location.py
alxmrs: Yeah, I think you're right about the default -- further, I'm happy this is at the callsite. Thanks for confirming.

CONTRIBUTING.md
```diff
@@ -349,7 +349,7 @@ uv run pytest --benchmark-only --benchmark-cprofile="tottime_per" --benchmark-cp
 (Please consult [these docs](https://pytest-benchmark.readthedocs.io/en/latest/usage.html#:~:text=%2D%2Dbenchmark%2Dcprofile%3DCOLUMN)
 to see all available values for the `--benchmark-cprofile` flag.)
 
-This will generate a `.prof` file located in the `` directory (by default). You can visualize this trace with `snakeviz`
+This will generate a `.prof` file located in the `.` directory (by default). You can visualize this trace with `snakeviz`
```
alxmrs, line 352: nice, thanks

configs/data/cm4.yaml
```diff
@@ -1,8 +1,8 @@
 # yaml-language-server: $schema=../schemas/DataConfig.json
 
-data_path: cm4_piControl_ocean_200yr_full_chunked.zarr
-data_means_path: 2024-11-12-cm4-piControl-ocean-200yr-dataset-stats/centering.nc
-data_stds_path: 2024-11-12-cm4-piControl-ocean-200yr-dataset-stats/scaling-full-field.nc
+data_location: cm4_piControl_ocean_200yr_full_chunked.zarr
+data_means_location: 2024-11-12-cm4-piControl-ocean-200yr-dataset-stats/centering.nc
+data_stds_location: 2024-11-12-cm4-piControl-ocean-200yr-dataset-stats/scaling-full-field.nc
```
alxmrs, line 5: As a random aside (for a future PR): I think storing the data + means + std in an Xarray DataTree would be a good way to go. See: https://github.com/Open-Athena/hermes/blob/main/tests/test_samudra.py

configs/data/om4_halfdeg_remote.yaml
```diff
@@ -1,10 +1,23 @@
 # yaml-language-server: $schema=../schemas/DataConfig.json
 
-# TODO(jder): the URLs below assume you have FSSPEC_S3_ENDPOINT_URL=https://nyu1.osn.mghpcc.org
-# as well as the appropriate AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY (or other S3 credentials) in your environment.
-# Once we have a clear story for launching this task remotely we should remove this implicit requirement.
+# TODO(jder): we assume you have appropriate AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY (or other S3 credentials) in your environment.
```
alxmrs: I've been putting these throughout the codebase, WDYT?

```suggestion
# NB(jder): we assume you have appropriate AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY (or other S3 credentials) in your environment.
```

alxmrs: Taking a leaf from your book ;)

configs/data/om4_halfdeg_remote.yaml
```diff
@@ -1,10 +1,23 @@
...
+
+data_location:
+  type: s3
+  bucket: emulators
+  path: sd5313/OM4_highres/om4_halfdeg.zarr
+  endpoint_url: https://nyu1.osn.mghpcc.org
```
alxmrs, line 10: Really nice!

configs/train_default.test.yaml
```diff
@@ -42,11 +42,11 @@ experiment:
   network: Samudra
   prognostic_vars_key: thermo_dynamic_5
   boundary_vars_key: tau_hfds
-  cluster_data_dir: .data_cache/
+  data_root: .data_cache/
```
alxmrs, line 45: `data_root` is so much clearer + accurate.

tests/test_location.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def test_valid_absolute_path(self):
+        """Test that absolute paths are accepted."""
+        loc = UnresolvedLocation(path="/absolute/path/data.zarr")
+        assert loc.path == "/absolute/path/data.zarr"
+
+    def test_invalid_absolute_url(self):
+        """Test that absolute URLs are rejected."""
+        with pytest.raises(ValidationError, match="Absolute urls are not supported"):
+            UnresolvedLocation(path="s3://bucket/path")
```
alxmrs, line 32: Nice to see a test for this, thanks.

tests/test_location.py
```diff
@@ -0,0 +1,277 @@
...
+        assert resolved.path == Path("/base/path/subdir/file.zarr")
+
+    def test_resolve_resolved_location(self):
+        """Test resolving a ResolvedLocation returns it unchanged."""
+        base = LocalLocation(path=Path("/base/path"))
+        other = LocalLocation(path=Path("/other/path"))
+
+        resolved = base.resolve(other)
+
+        assert resolved is other
```
alxmrs, line 81: Nice `__eq__` test

tests/test_location.py
```diff
@@ -0,0 +1,277 @@
...
+            ds.to_netcdf(nc_path)
+
+            # Test opening
+            loc = LocalLocation(path=nc_path)
+            opened_ds = loc.open()
+
+            assert isinstance(opened_ds, xr.Dataset)
+            assert "temperature" in opened_ds.data_vars
+
+    def test_open_with_chunks(self):
```
alxmrs, line 123: 🐑  optional nit: might be nice to add a test to check chunks with values, e.g. `chunks=dict(x=2)`, etc. 

Though, I do think that such a feature works.

alxmrs: You can assert that `ds.chunks` equates to what you expect.

tests/test_location.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def test_url_generation(self):
+        """Test URL generation for S3Location."""
+        loc = S3Location(bucket="test-bucket", path="data/test.zarr")
+        assert loc.url() == "s3://test-bucket/data/test.zarr"
+
+    def test_url_generation_with_special_chars(self):
+        """Test URL generation with special characters."""
+        loc = S3Location(bucket="test-bucket", path="data/test file.zarr")
+        assert loc.url() == "s3://test-bucket/data/test%20file.zarr"
```
alxmrs, line 161: Nice! Great test.

tests/test_location.py
```diff
@@ -0,0 +1,277 @@
...
+
+            # Test opening with chunks
+            loc = LocalLocation(path=zarr_path)
+            opened_ds = loc.open(chunks={})
+
+            assert isinstance(opened_ds, xr.Dataset)
+            assert "temperature" in opened_ds.data_vars
+
+
+class TestS3Location:
```
alxmrs, line 141: 🐑  Including an explicit test of the `url()`, specifically of `self.path.lstrip("/")` might be nice. Even just prepending a `/` to one of the tests blow should do it -- and, maybe this exists but I've missed it.

tests/test_location.py
```diff
@@ -0,0 +1,277 @@
...
+
+        resolved = base.resolve(unresolved)
+        assert isinstance(resolved, S3Location)
+
+        assert resolved.endpoint_url == "https://s3.example.com"
+
+    def test_resolve_resolved_location(self):
+        """Test resolving a ResolvedLocation returns it unchanged."""
+        base = S3Location(bucket="test-bucket", path="base/path")
+        other = LocalLocation(path=Path("/other/path"))
```
alxmrs, line 197: I think this may exercise the `url()` test that I mentioned above.

tests/test_location.py
```diff
@@ -0,0 +1,277 @@
...
+
+        unresolved = UnresolvedLocation(path="data/test.zarr")
+        model = TestModel(location=unresolved)
+        assert model.location is unresolved
+
+
+class TestLocationIntegration:
+    """Integration tests for the location system."""
+
+    def test_mixed_location_types(self):
```
alxmrs, line 244: It's cool how we can mix these!

tests/test_location.py
```diff
@@ -0,0 +1,277 @@
...
+        assert s3_resolved.path == "base/subdir/file.zarr"
+
+        # Test local resolution
+        local_resolved = base_local.resolve(unresolved)
+        assert isinstance(local_resolved, LocalLocation)
+        assert local_resolved.path == Path("/local/base/subdir/file.zarr")
+
+    def test_path_joining_edge_cases(self):
+        """Test edge cases in path joining."""
+        # Test with trailing slash on base
```
alxmrs, line 262: Ah, it looks like you already got address the thing I called out! I should have kept scrolling.

alxmrs (approved):
Great work, LGTM

configs/data/om4_halfdeg_remote.yaml
jder: How is "NB" different from just a normal comment? I meant "TODO" insofar as we should at some point resolve that assumption.

alxmrs: Ah, I see how it is an action item to be addressed, I didn't understand it that way at first. 

I think I distinguish NB from a normal comment in that it's tagged with a username. But, that's not a long standing convention or anything, just a pattern I've come to like.

jder: Gotcha, I updated the comment to make it clear what there is to be done.

PR #285, opened 2025-06-27
Put OSN Bucket keys in environment variables.

### LINKED ISSUE #287: Remove notebooks from `main` branch
They can be archived at the [special branch Surya made](https://github.com/LaureZanna/Ocean_Emulator/tree/fixed-branch/Notebooks) but they are a constant source of diff noise, aren't linted/formatted/etc and IMO not needed in the main branch. I think we should remove them and (if needed) replace them with normal code and/or python code with [cell markers](https://code.visualstudio.com/docs/python/jupyter-support-py#_jupyter-code-cells).

alxmrs:
I plan to scrub the access keys that I checked in. For good measure, I also plan to invalidate these keys and get new ones for myself. 

This PR also adds a script that converts jupyter notebooks into Python scripts. It fixes #287.


jder (commented):
It looks like every image in the notebooks are considered a secret - does adding a line or re-running one of them cause a false positive? The .secrets.baseline also seems to include detection of the API keys you removed -- does it need to be manually regenerated somehow?

.secrets.baseline
```diff
@@ -0,0 +1,946 @@
...
+    {
+      "path": "detect_secrets.filters.heuristic.is_templated_secret"
+    }
+  ],
+  "results": {
+    "notebooks/2025-06-12-data_preprocess.ipynb": [
+      {
+        "type": "Hex High Entropy String",
+        "filename": "notebooks/2025-06-12-data_preprocess.ipynb",
+        "hashed_secret": "6c15f3a5d4f8fc804f7e6d4244f27176a9b37fea",
```
alxmrs: @jder I believe that this is not a secret, but rather refers to a URL in the notebook that contains a git commit hex string.

alxmrs: As a sanity check, I did try to regenerate the script, but the only thing that changed was the last generation time value.

alxmrs:
> It looks like every image in the notebooks are considered a secret - does adding a line or re-running one of them cause a false positive?

Ah! I was wondering what those values were -- I didn't realize they were from images. I do expect this to cause false positives.

alxmrs:
Following up from the discussion in, https://github.com/LaureZanna/Ocean_Emulator/issues/287, I think I'll omit solving the false-positive issue for images in notebooks and instead implement 287, which will get rid of this source of noise to begin with. Good shout on that idea which we've wanted to do anyway.

ryan-williams (approved):

PR #286, opened 2025-06-27
rm extra nb diffs: `s/001B/001b/g`

ryan-williams:
Targets #285, removes extra diffs in the notebooks there (`s/001B/001b/g`)

See cleaned up diff vs. main: [`main..rw/285-lint`](https://github.com/LaureZanna/Ocean_Emulator/compare/main..rw/285-lint).


alxmrs:
Thanks!

PR #288, opened 2025-07-03
Updated `concurrent_compute` logic to create a thread pool only once.

alxmrs:
I think that it's a good idea to try to load all the data variables concurrently. However, before this PR, we create and then tear down the executor way too many times (per loop per getitem). This change will allow to to see how effective we can concurrently load Xarray data.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -597,7 +597,6 @@ class TorchTrainDataset(Dataset):
         self.stride: int = stride
         self.normalize_before_mask: bool = normalize_before_mask
         self.masked_fill_value: float = masked_fill_value
-        self.concurrent_compute: bool = executor is not None
         self._executor = executor
 
         self.num_prognostic_channels: int = (hist + 1) * len(prognostic_var_names)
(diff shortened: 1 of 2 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -595,7 +597,8 @@ def __init__(
         self.stride: int = stride
         self.normalize_before_mask: bool = normalize_before_mask
         self.masked_fill_value: float = masked_fill_value
-        self.concurrent_compute: bool = concurrent_compute
+        self.concurrent_compute: bool = executor is not None
+        self._executor = executor
```
jder, line 600: 🐑 I don't see a need for both variables, let's just use `if self._executor is not None` below or even `if self._concurrent_compute_executor` if you want to be clear what it's for?

jder (approved):
Thanks for this! Silly of me not to notice this when I wrote it :)

src/ocean_emulators/datasets.py
alxmrs: SGTM

alxmrs:
Thanks for the review, and no worries! I imagine you were hyper-focused on experimentation.

PR #289, opened 2025-07-03
Added an SPDL-based DataLoader that handles concurrent IO separate from CPU processing.

alxmrs:
This PR is an improvement on #278 that uses learnings from a [prototype that I made in a one-off repo](https://github.com/LaureZanna/Ocean_Emulator/issues/281#issuecomment-3028790272). My bet is that this will improve data loading performance by separating the IO-bound parts of our TorchTrainDataset from the CPU-bound parts. This new loader provides an affordance for controlling the concurrency of these two aspects of the pipeline separately. Once this feature is landed, in a future task, I will work to tune these two parameters (which is dependent on the particulars of the VM in which we run the pipeline). 

Furthermore, this improves on #278 in that I got the memory model right s.t. it doesn't make a zombie thread that can't be killed. CI in this PR completes in a reasonable time frame, where as the former PR hangs indefinitely.


src/ocean_emulators/datasets.py
```diff
@@ -744,18 +776,131 @@ def _get_x_index(self, idx: int, step: int) -> xr.DataArray:
...
+            PipelineBuilder()
+            .add_source(self._source())  # type: ignore
+            .pipe(self._with_steps, output_order="input")  # type: ignore
+            .pipe(
+                self._download,  # type: ignore
+                concurrency=self.io_workers,
+                executor=self.executor,
+                output_order="input",
+            )
+            .aggregate(self.steps)
```
alxmrs, line 936: IIUC, This should download all the steps concurrently. Since I pass in the same executor that is passed in, I also think a shared threadpool will be used to download the components of each Variable in these steps.

tests/test_datasets.py
```diff
@@ -119,12 +121,26 @@ def make_loader(
...
-            data,
-            batch_size=cfg.batch_size,
-            drop_last=drop_last,
-            collate_fn=collate_fn,
-        )
+        if use_spdl_loader:
+            assert version == LoaderVersion.OM4_TORCH, (
+                "SPDL Loader only works for the Torch Dataset!"
+            )
+            assert isinstance(data, ConcatDataset), "Must not be an inference dataset!"
```
alxmrs, line 127: These asserts aren't really necessary, but I find that they make the type checker happy.

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,120 @@ def load_variable_data(var: xr.Variable) -> None:
             futures.append(executor.submit(load_variable_data, var))
 
     wait(futures)
+
+
+T = TypeVar("T")
+WithSteps = tuple[int, T]
+
+
+class SpdlTorchDataLoader(DataLoader):
```
alxmrs: TODO(alxmrs): Add docstring to class.

src/ocean_emulators/datasets.py
```diff
@@ -638,50 +651,66 @@ def __init__(
     def __len__(self) -> int:
         return self.size
 
-    @elapsed(level=logging.DEBUG)
-    def __getitem__(self, idx: int):
-        start_time = time.perf_counter()
-        TD = TrainData(self.num_prognostic_channels)
+    def perform_io(self, x_index, *, executor=None) -> tuple[xr.Dataset, xr.Dataset]:
```
jder: How about `load_prognostic_and_boundary`?

src/ocean_emulators/datasets.py
```diff
@@ -638,50 +651,66 @@ def __init__(
...
-                    )
-                    .rename({"var": "variable"})
-                    .to_numpy()
-                )
-            else:
-                prognostic_all = torch.from_numpy(
-                    prognostic_selected.to_array()
-                    .transpose("time", "variable", "lat", "lon")
-                    .to_numpy()
+    def prognostic_to_tensor(self, prognostic_selected):
```
jder, line 660: Type annotations, please (and for boundary_to_tensor)

src/ocean_emulators/datasets.py
jder: Also, looks like this is always passed `self.executor`? So might as well use that directly rather than passing it?

src/ocean_emulators/datasets.py
```diff
@@ -638,50 +651,66 @@ def __init__(
...
+            boundary = self.boundary_to_tensor(boundary_selected)
+
+            input_, label = self.join_and_partition_by_time(prognostic, boundary)
             TD.insert(input_=input_, label=label)
 
         TD.load_stats = LoadStats(time.perf_counter() - start_time)
         return TD
 
-    def _get_input_and_label(
+    def join_and_partition_by_time(
```
jder: 🐑 I prefer the old name, which is about the goal/interface of the method rather than the implementation/details of it.

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,175 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        ...     batch_size=8
+        ... )
+        >>> for batch in loader:
+        ...     # Process batch
+        ...     pass
+    """
+
+    def __init__(
+        self,
+        dataset: ConcatDataset,
```
jder, line 843: It seems surprising that we are depending on using a `ConcatDataset` and ignoring all datasets in that `ConcatDataset` other than the first. At a minimum we should assert there is only one dataset (perhaps in the caller and take a TorchTrainDataset here instead?). Between this and duplicate sampler stuff it kinda smells like we have the wrong abstraction somewhere, WDYT?

src/ocean_emulators/train.py
```diff
@@ -794,27 +797,59 @@ def init_data_loaders(self, cur_step: int) -> None:
...
+            train_sampler = (
+                RandomSampler
+                if self.distributed is None
+                else functools.partial(DistributedSampler, shuffle=True)
+            )
+            val_sampler = (
+                RandomSampler
+                if self.distributed is None
+                else functools.partial(DistributedSampler, shuffle=False)
+            )
```
jder: It puts us in an awkward state to make additional samplers here which are not the same as `self.train_sampler` / `self.val_sampler` (which are still created above) given the latter are used in `self.run`, even when `use_spdl` is true.

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,175 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        collate_fn: Callable,
+        io_workers: int,
+        cpu_workers: int,
+        batch_size: int,
+        drop_last: bool = False,
+        sampler: type[Sampler] | partial[DistributedSampler] | None = None,
```
jder, line 849: `Callable[[list[T]], Sampler[T]]` or something might be a nicer annotation here?

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,175 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        )
+        self.first_dataset: TorchTrainDataset = first_dataset
+        self.steps = self.first_dataset.steps
+        self.executor = self.first_dataset.executor
+        self._sampler_klass = sampler
+        self.io_workers = io_workers
+        self.cpu_workers = cpu_workers
+        self.timeout = timeout
+        if pin_memory:
+            warnings.warn(f"Pinning memory is currently not supported in {self!r}.")
```
jder: 🐑 I think this should probably be an error?

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,175 @@ def load_variable_data(var: xr.Variable) -> None:
...
+
+    def _source(self) -> Iterable[int]:
+        """Generated indexes to load, using the input sampler."""
+        items = list(range(len(self)))
+        samples: Iterable = items
+        if self._sampler_klass is not None:
+            samples = self._sampler_klass(items)
+        if self.drop_last:
+            samples = itertools.islice(samples, len(items) - 1)
```
jder: Should this only happen if the last batch is incomplete?

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,175 @@ def load_variable_data(var: xr.Variable) -> None:
...
+
+    def _process_traindata(
+        self, downloaded: list[WithSteps[tuple[xr.Dataset, xr.Dataset]]]
+    ) -> TrainData:
+        """Create a TrainData by aggregating across all steps across an index."""
+        train_data = TrainData(self.first_dataset.num_prognostic_channels)
+
+        # Ensure that the pipeline-level aggregation respects the steps of train data!
+        steps = [step for step, _ in downloaded]
+        assert steps == list(range(self.steps))
```
jder, line 911: Why are we threading the steps through all these stages rather than having a `self.steps()` or whatever that each stage can access?

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,175 @@ def load_variable_data(var: xr.Variable) -> None:
...
+            )
+            train_data.insert(input_, label)
+
+        return train_data
+
+    def _build_pipeline(self) -> spdl.pipeline.Pipeline[TrainData]:
+        """Wire together steps into an SPDL pipeline."""
+        total_workers = self.io_workers + self.cpu_workers + 2
+        # Below, we ignore type information because this pipeline can't infer that the
+        # return type is `TrainData`.
```
jder, line 925: Yeah, the type annotations for PipelineBuilder are pretty unhelpful. Can you fix with eg `PipelineBuilder[int, TrainData]()….pipe[int, WithSteps[xr.DataArray](…)`?

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,175 @@ def load_variable_data(var: xr.Variable) -> None:
...
+                executor=self.executor,
+                output_order="input",
+            )
+            .aggregate(self.steps)
+            .pipe(self._process_traindata, concurrency=self.cpu_workers)
+            .aggregate(self.batch_size or 1)
+            .pipe(self.collate_fn, concurrency=self.cpu_workers)
+            # Later, we can add a step to efficiently dispatch data to GPU.
+            .add_sink(6)
+            .build(num_threads=total_workers)
```
jder, line 946: I don't really understand the various magic numbers in this function, can you leave an explanation? Why + 2? (Why `num_workers + 2` in the caller?) Why do we want as many threads as we have concurrency when io_workers is from a separate thread pool?

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,175 @@ def load_variable_data(var: xr.Variable) -> None:
             futures.append(executor.submit(load_variable_data, var))
 
     wait(futures)
+
+
+T = TypeVar("T")
+WithSteps = tuple[int, T]
+
+
+class SpdlTorchDataLoader(DataLoader):
```
jder: Why is this class a torch `DataLoader`? It doesn't seem like it really meets the contract of that class (eg the `__iter__` return type, unsupported `__init__` arguments). It also doesn't seem to use anything from the superclass?

src/ocean_emulators/train.py
jder: (And I continue to find functools.partial more subtle and less type-safe than eg `lambda dataset: DistributedSampler(dataset, shuffle=False)`)

src/ocean_emulators/datasets.py
```diff
@@ -759,3 +788,175 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        )
+
+    # NB(alxmrs): This doesn't return the same type as the base data loader type, but I
+    # don't think this should be a problem, so we ignore type checking.
+    def __iter__(self) -> PipelineIterator[TrainData]:  # type: ignore
+        """Iterate across batches of examples."""
+        pipeline = self._build_pipeline()
+
+        # TIL that the __exit__ condition will still be called even though we use
+        # a `return` instead of a `yield`.
```
jder: I was very suspicious of this and so dug in a little. It's true that `__exit__` is called but it happens at the time of the `return` -- ie the pipeline is stopped before the iterator is ever iterated. Turns out this was passing the test because the pipeline silently (!) makes the iterator empty after the pipeline is stopped and `assert_equal_samples` uses `zip` without `strict=True` so `[]` will compare as equal to anything.

jder (changes requested):
Some feedback from me. My overall impression is that this doesn't seem to fit the existing abstractions very well and so left some suggestions about how to resolve. I am still worried about the immaturity of the library given the deadlocks you were getting and the pipeline typing stuff. Overall I don't think this is something we should merge without good evidence this is the a win, performance-wise so I vote we get that evidence first before resolving these comments.

src/ocean_emulators/datasets.py
alxmrs: !!

alxmrs: Ah, foiled by Google's AI Overviews!

src/ocean_emulators/datasets.py
alxmrs: Both SGTM.

src/ocean_emulators/datasets.py
alxmrs: I've address this, too, as it was related to the pipeline correctness. 

TODO(alex): Address the sampler interface -- you point out a good alternative.

alxmrs: wrt the sampler -- an alternative interface is an interesting consideration, esp given that I'm still not sure the best way to implement `drop_last`.

alxmrs:
Closing for now; this experiment may prove promising in the end, however, it requires changing our DDP training scheme to fully work, which is a bigger refactor than we originally bargained for.

PR #291, opened 2025-07-08
Ignore all hidden files by default, ignore ./playpen

jder:
I was running into problems with .mypy_cache, figured this is probably easier. `playpen` is something I'm experimenting with and can remove if it doesn't work out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 __pycache__
-.*
-/playpen
+.DS_Store
+.ipynb_checkpoints
+.mypy_cache
+.cache
 download-beaker.sh
 jobs
 saved_nets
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,14 +1,13 @@
 __pycache__
-.DS_Store
-.ipynb_checkpoints
+.*
```
alxmrs: Lol, nice! I haven't seen this pattern before, but I like it. 

Won't this ignore `.python-version`, `.pre-commit-config`, `.github/`, and `.secrets-baseline` though? Is that OK bc it's already checked in?

jder: Yes, gitignore only applies to `add` so things already checked in are ok. But you raise a good point, eg if someone makes a new file in `.github/something` it won't get added without -f. That's probably not worth it. I changed it so it just adds the new things I want to ignore.

jder:
Looks like tests don't run on forks, lemme re-open as a branch

PR #294, opened 2025-07-08
Ignore some more things

jder:


alxmrs (approved):
Very reasonable. LGTM. Thanks for ignoring the mypy stuff.

PR #295, opened 2025-07-09
Fix performance regression by moving towards a more buildy world with DataContainer

jder:
Fixes the performance regression from [here](https://github.com/LaureZanna/Ocean_Emulator/commit/32f3043d20bb43b24a1a097004f5c0f34f013075#commitcomment-160869800) by detecting if the data source supports forking or not. 

Doing this required a bit of code movement so I took the opportunity to move us closer towards a world where creating the trainer is `trainer = cfg.build()` by moving a bunch of the stuff in `Trainer.__init__` into `DataConfig.build` and creating a new class which is the result of that method, `DataContainer`. This also reduces some of the duplication between train/eval/tests.

Last commit is some drive-by fixes for warnings that were being printed in tests.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -7,7 +7,8 @@ import torch
 from pydantic import Field, PlainSerializer, PlainValidator, WithJsonSchema
 
 from ocean_emulators.config_base import BaseConfig, TopLevelConfig
-from ocean_emulators.constants import LoaderVersion
+from ocean_emulators.constants import BoundaryVarNames, LoaderVersion
+from ocean_emulators.utils.data import DataContainer, DataSource, validate_data
 from ocean_emulators.utils.location import LocalLocation, Location, ResolvedLocation
 from ocean_emulators.utils.profiler import Profiler
 
@@ -115,6 +116,72 @@ class DataConfig(BaseConfig):
     normalize_before_mask: bool = True
     masked_fill_value: float = 0.0
     concurrent_compute: bool = False
+
+    def build(
+        self,
+        data_root: ResolvedLocation,
+        boundary_var_names: BoundaryVarNames,
+    ) -> DataContainer:
+        loader_version = LoaderVersion(self.loader_version)
+        use_dask = loader_version != LoaderVersion.OM4_TORCH
+
+        data_location = data_root.resolve(self.data_location)
+        means_location = data_root.resolve(self.data_means_location)
+        stds_location = data_root.resolve(self.data_stds_location)
+
+        source = DataSource.from_locations(
+            data_location=data_location,
+            means_location=means_location,
+            stds_location=stds_location,
+            use_dask=use_dask,
+        )
+        source = validate_data(source, boundary_var_names, self.static_data_vars)
+
+        if use_dask:
+            data_using_dask = source
+        else:
+            data_using_dask = DataSource.from_locations(
+                data_location=data_location,
+                means_location=means_location,
+                stds_location=stds_location,
+                use_dask=True,
+            )
+            data_using_dask = validate_data(
+                data_using_dask, boundary_var_names, self.static_data_vars
+            )
+        scaling_residuals_location = (
+            data_root.resolve(self.scaling_residuals_file)
+            if self.scaling_residuals_file is not None
+            else None
+        )
+        scaling_residuals = (
+            scaling_residuals_location.open()
+            if scaling_residuals_location is not None
+            else None
+        )
+        static_data = (
+            source.data[self.static_data_vars]
+            if self.static_data_vars is not None
+            else None
+        )
+
+        supports_fork = all(
+            location is None or location.supports_fork
+            for location in [
+                data_location,
+                means_location,
+                stds_location,
+                scaling_residuals_location,
+            ]
+        )
+        return DataContainer(
+            source,
+            data_using_dask,
+            loader_version,
+            supports_fork,
+            scaling_residuals,
+            static_data,
+        )
 
 
 BlockType = Literal["conv_next_block", "conv_block"]
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -138,27 +138,27 @@ class DataConfig(BaseConfig):
         source = validate_data(source, boundary_var_names, self.static_data_vars)
 
         if use_dask:
-            data_using_dask = source
+            # If we're already using dask, we don't need a second source
+            source_using_dask = source
         else:
-            data_using_dask = DataSource.from_locations(
+            # If we're not using dask for the main source, create a separate one
+            source_using_dask = DataSource.from_locations(
                 data_location=data_location,
                 means_location=means_location,
                 stds_location=stds_location,
                 use_dask=True,
             )
-            data_using_dask = validate_data(
-                data_using_dask, boundary_var_names, self.static_data_vars
+            source_using_dask = validate_data(
+                source_using_dask, boundary_var_names, self.static_data_vars
             )
-        scaling_residuals_location = (
-            data_root.resolve(self.scaling_residuals_file)
-            if self.scaling_residuals_file is not None
-            else None
-        )
-        scaling_residuals = (
-            scaling_residuals_location.open()
-            if scaling_residuals_location is not None
-            else None
-        )
+
+        if self.scaling_residuals_file is not None:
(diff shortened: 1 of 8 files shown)


src/ocean_emulators/train.py
```diff
@@ -139,6 +134,20 @@ def __init__(self, cfg: TrainConfig) -> None:
...
+        )
+
+        self.mp_context: BaseContext | None
+        if cfg.data.num_workers > 0:
+            if self.data_container.supports_fork:
+                self.mp_context = multiprocessing.get_context("fork")
+            else:
+                self.mp_context = multiprocessing.get_context("spawn")
+        else:
+            self.mp_context = None
```
jder: This stanza is the only (intentional) behavior change -- everything else is a refactor.

src/ocean_emulators/utils/data.py
```diff
@@ -226,24 +229,37 @@ def from_dataset(cls, dataset: xr.Dataset, *, name: str = "dataset") -> Self:
...
-        stds = root.resolve(cfg.data.data_stds_location).open(chunks)
-
-        dask = "with_dask" if use_dask else "without_dask"
+        data = data_location.open(chunks)
+        means = means_location.open(chunks)
+        stds = stds_location.open(chunks)
 
         return cls(
-            name=f"{cfg.experiment.name}-{data}-{dask}",
+            name=f"{data_location}-{use_dask}",
```
jder, line 246: I guess this is also a change -- I think it was a bug that we were trying to render the dataset? I dropped the experiment name because I don't think that's going to be helpful during debugging (it's global) and would have required more threading.

src/ocean_emulators/config.py
```diff
@@ -116,6 +117,72 @@ class DataConfig(BaseConfig):
...
+        means_location = data_root.resolve(self.data_means_location)
+        stds_location = data_root.resolve(self.data_stds_location)
+
+        source = DataSource.from_locations(
+            data_location=data_location,
+            means_location=means_location,
+            stds_location=stds_location,
+            use_dask=use_dask,
+        )
+        source = validate_data(source, boundary_var_names, self.static_data_vars)
```
alxmrs, line 138: Could this go in the `if` block on line 140?

src/ocean_emulators/eval.py
```diff
@@ -91,20 +89,14 @@ def __init__(self, cfg: EvalConfig) -> None:
...
-        self.data_dir = cfg.experiment.resolved_data_root
-        self.data_location = cfg.data.data_location
-        self.data_means_location = cfg.data.data_means_location
-        self.data_stds_location = cfg.data.data_stds_location
-        self.scaling_residuals_file = cfg.data.scaling_residuals_file
-
-        raw = DataSource.from_config(cfg, use_dask=True)
-        self.src = validate_data(
-            raw, self.boundary_var_names, cfg.data.static_data_vars
+        self.data_container = cfg.data.build(
```
alxmrs, line 92: Nice, I'm happy to see the shift to `build` here.

src/ocean_emulators/utils/data.py
alxmrs: This change LGTM -- makes sense to associate with the data location now that such a concept exists.

src/ocean_emulators/utils/data.py
```diff
@@ -226,24 +229,37 @@ def from_dataset(cls, dataset: xr.Dataset, *, name: str = "dataset") -> Self:
...
             data=data,
             means=means,
             stds=stds,
         )
 
 
+@dataclasses.dataclass
+class DataContainer:
+    source: DataSource
+    source_using_dask: DataSource
```
alxmrs, line 256: Do we have both sources for the spawning behavior? Or, is this because the inference loader needs dask?

src/ocean_emulators/utils/data.py
```diff
@@ -226,24 +229,37 @@ def from_dataset(cls, dataset: xr.Dataset, *, name: str = "dataset") -> Self:
...
-            name=f"{cfg.experiment.name}-{data}-{dask}",
+            name=f"{data_location}-{use_dask}",
             data=data,
             means=means,
             stds=stds,
         )
 
 
+@dataclasses.dataclass
+class DataContainer:
```
alxmrs, line 254: 🐑  DataContainer is a decent name -- any other ideas, out of curiosity? `ConfiguredData`? `ResolvedData`?

src/ocean_emulators/train.py
alxmrs: 🐑  I prefer this style, as it has one less branch, but I see the merits of what's already here, too.

```suggestion
        self.mp_context: BaseContext | None = None
        if cfg.data.num_workers > 0:
            if self.data_container.supports_fork:
                self.mp_context = multiprocessing.get_context("fork")
            else:
                self.mp_context = multiprocessing.get_context("spawn")
```

src/ocean_emulators/config.py
jder: No, we want to do this unconditionally. The if/else block on 140 might create a second data source.

src/ocean_emulators/utils/data.py
jder: Yeah, because the infernece loader needs dask.

src/ocean_emulators/utils/data.py
jder: I kinda just want to call it Data, but I think probably we'll end up splitting this into more specific types which capture different aspects (eg the loader version seems separate from the sources, same with history count, etc)

src/ocean_emulators/train.py
```diff
@@ -139,6 +134,20 @@ def __init__(self, cfg: TrainConfig) -> None:
...
+            data_root=cfg.experiment.resolved_data_root,
+            boundary_var_names=self.boundary_var_names,
+        )
+
+        self.mp_context: BaseContext | None
+        if cfg.data.num_workers > 0:
+            if self.data_container.supports_fork:
+                self.mp_context = multiprocessing.get_context("fork")
+            else:
+                self.mp_context = multiprocessing.get_context("spawn")
```
alxmrs, line 147: 🐑 A similar style preference:

```suggestion
            context_type = "fork" if self.data_container.supports_fork else "spawn"
            self.mp_context = multiprocessing.get_context(context_type)
```

What you have here is fine, too.

src/ocean_emulators/utils/data.py
alxmrs: I support just calling it `Data` -- it makes sense that a `DataConfig` `build`s a `Data` object.

alxmrs: `Data` is consistent post this PR, and we can always re-name it later when we do split it.

src/ocean_emulators/utils/data.py
alxmrs: Gotcha, that makes sense.

src/ocean_emulators/config.py
alxmrs: Thanks, makes sense now that I see how this is used outside of configuration.

src/ocean_emulators/utils/data.py
```diff
@@ -4,14 +4,15 @@
...
+from typing import TYPE_CHECKING, Literal, Self
 
 import numpy as np
 import torch
 import xarray as xr
 from einops import rearrange
 
-from ocean_emulators.config import EvalConfig, TimeConfig, TrainConfig
+if TYPE_CHECKING:
+    from ocean_emulators.config import TimeConfig
```
alxmrs, line 15: Is this pattern so in order to prevent circular imports just for the sake of type checking?

alxmrs (approved):
Some minor nits wrt naming and conditional logic preferences. Overall, I like this refactoring. It provides good affordances for the mp_context feature.

src/ocean_emulators/utils/data.py
jder: Yes, that's right.

src/ocean_emulators/utils/data.py
jder: I went to do this rename and not having some other word here makes this kinda awkward. Eg Trainer's self.data is already a thing (it's data_container.source.data) and so I think self.data_container is probably the best answer there for now, and so probably also here.

jder:
Thanks for the quick review! Pushed up some changes and will merge when green.

PR #296, opened 2025-07-10
Improvements from trying to understand the model

jder:
* Makes the model summary include grid sizes by delaying calling it until we have data to run a forward pass (this will probably slow down benchmarks but we can ignore)
  * This required making TrainData a typing.Mapping to satisfy the summary function (though this still isn't perfect, it somehow can't figure out how much memory it uses)
* When debug = true, include all the layers not just top ones
* Rename and add type annotations to the core samudra model.

### DIFF
--- a/src/ocean_emulators/models/samudra.py
+++ b/src/ocean_emulators/models/samudra.py
@@ -1,4 +1,3 @@
-from collections.abc import Sequence
 from typing import assert_never
 
 import numpy as np
@@ -158,7 +157,7 @@ class Samudra(BaseModel):
             skip_inputs.append(torch.zeros_like(fts))
         count = 0
         for layer in self.layers:
-            crop: Sequence | np.ndarray = fts.shape[2:]
+            crop: torch.Size | np.ndarray = fts.shape[2:]
             if isinstance(layer, nn.Conv2d):
                 fts = torch.nn.functional.pad(
                     fts, (self.N_pad, self.N_pad, 0, 0), mode=self.pad
--- a/src/ocean_emulators/models/samudra.py
+++ b/src/ocean_emulators/models/samudra.py
@@ -1,3 +1,4 @@
+from collections.abc import Sequence
 from typing import assert_never
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


src/ocean_emulators/models/samudra.py
```diff
@@ -150,14 +151,14 @@ def __init__(self, config: SamudraConfig, hist, wet, area_weights, static_data):
         self.corrector = Correctors(config.corrector, hist, area_weights, static_data)
         self.num_steps = int(len(config.ch_width) - 1)
 
-    def forward_once(self, fts):
+    def forward_once(self, fts: torch.Tensor) -> torch.Tensor:
         fts_input = fts.clone().detach()
-        temp: list[torch.Tensor] = []
+        skip_inputs: list[torch.Tensor] = []
```
alxmrs, line 155: much better name!

src/ocean_emulators/models/samudra.py
```diff
@@ -150,14 +151,14 @@ def __init__(self, config: SamudraConfig, hist, wet, area_weights, static_data):
...
         for i in range(self.num_steps):
-            temp.append(torch.zeros_like(fts))
+            skip_inputs.append(torch.zeros_like(fts))
         count = 0
         for layer in self.layers:
-            crop = fts.shape[2:]
+            crop: Sequence | np.ndarray = fts.shape[2:]
```
alxmrs: Is this really a `np.ndarray` and not a `torch.Tensor`?

alxmrs: or a `torch.Size` (which IIUC is a `tuple`)?

src/ocean_emulators/utils/logging.py
```diff
@@ -208,11 +209,13 @@ def log_every(self, data_loader: DataLoader["TrainData"], print_freq, header=Non
...
 
 
-def get_model_summary(model: torch.nn.Module, num_input_channels: int):
+def get_model_summary(model: torch.nn.Module, data: Mapping | None, debug: bool):
     model_parameters = filter(lambda p: p.requires_grad, model.parameters())
     params = sum([np.prod(p.size()) for p in model_parameters])
     logging.info(f"Number of parameters: {params}")
-    logging.info(summary(model))
+    depth = 10 if debug else 2
+    # we pass verbose = 0 because we log the summary ourselves
```
alxmrs, line 217: Thanks, the comment is helpful.

alxmrs (approved):
Short and sweet. A comment about types, but this doesn't get in the way of a LGTM.

src/ocean_emulators/models/samudra.py
jder: Yeah, I said sequence rather than torch.Size but made it more specific. 

It is in fact assigned an np.ndarray below. (Probably we should restructure this a little to avoid this union typing)

alxmrs: Thanks I missed that it was assigned to a `np.ndarray` later on! That addresses my main concern. I think that Sequence is a good fit to abstract over `tuple`, etc.

alxmrs: The more specific type is a bit clearer, too. Happy with either approach.

PR #297, opened 2025-07-10
Fix bug which ignored loader version in tests

jder:
Our benchmarks showed a surprising jump in one of them after #295:
Screenshot 2025-07-10 at 10 26 37 AM

Turns out we were always building a torch loader, not an eager loader. Running the benchmarks locally after this change again shows a big performance gap between eager and torch loaders.


alxmrs (approved):

alxmrs:
Great catch!

PR #298, opened 2025-07-11
First pass at setting up one-command training with skypilot.

alxmrs:
Wrote entrypoint for performing training runs with skypilot. The `sky` command is documented within the `train.sky.yaml` file and contributing guide.

I did not successfully do a full training run, but that is because I hit data engineering errors (no static_data!).

<details>
<summary>(Latest error trace related to data engineering)</summary

Ultimately, hitting this error seems like a good sign because the error is in our code, not in the provisioning system.
```
AssertionError: : Static data variable sea_surface_fraction not found in dataStatic data variable sea_surface_fraction not found in data
```

</details<


train.sky.yaml
```diff
@@ -0,0 +1,84 @@
...
+  # Secrets
+  # As of v0.9.3, `sky` supports passing in secrets via a dotenv file! So, you can specify these with `--env-file`.
+  # To pass in secrets, we recommend setting environment variable and then ONLY passing in the name via `--env`
+  WANDB_API_KEY: null  # e.g. `--env WANDB_API_KEY`
+  # Run `sky check aws -v` to verify that AWS storage is properly set up.
+  # If you still hit permission issues, consider making a dedicated SkyPilot IAM via these docs:
+  # https://docs.skypilot.co/en/latest/cloud-setup/cloud-permissions/aws.html#dedicated-skypilot-iam-user
+
+  CONFIG: null
+  EXPERIMENT_NAME: null
```
alxmrs: Question for the reviewer: which config system do we want this to be driven from? the .env / CLI, or the config file?

train.sky.yaml
```diff
@@ -0,0 +1,84 @@
...
+  MASTER_ADDR=$(echo "$SKYPILOT_NODE_IPS" | head -n1)
+  echo "Starting distributed training, head node: $MASTER_ADDR"
+
+  # Explicit check for torchrun
+  if ! command -v torchrun >/dev/null 2>&1; then
+      echo "ERROR: torchrun command not found" >&2
+      exit 1
+  fi
+
+  uv run torchrun \
```
alxmrs, line 71: `uv` comes pre-installed on skypilot machines.

.skyignore
```diff
@@ -0,0 +1,55 @@
...
+.github/
+notebooks/
+scripts/
+tests/
+# just in case! Ignore data files
+*.zarr
+*.nc
+# Copied from .gitignore on 2025-07-15
+# .skyignore overrides the .gitignore default!
+# See - https://docs.skypilot.co/en/latest/examples/syncing-code-artifacts.html#exclude-uploading-files
```
jder: Might be nicer to let it use the existing `.gitignore`, is there anything we can't put in there which we need to prevent sky from uploading?

configs/data/om4_sky_local.yaml
```diff
@@ -0,0 +1,17 @@
...
+  type: local
+  path: /data/OM4.zarr
+
+data_means_location:
+  type: local
+  path: /data/OM4_means.zarr
+
+data_stds_location:
+  type: local
+  path: /data/OM4_means.zarr
```
jder: See https://github.com/Open-Athena/Ocean_Emulator/pull/309, these should be eg `data_location: "OM4.zarr"`

train.sky.yaml
```diff
@@ -0,0 +1,84 @@
...
+resources:
+  # Optional; if left out, automatically pick the cheapest cloud.
+  cloud: lambda
+  # 8x NVIDIA A100 GPU
+  accelerators: A100:8
+  disk_tier: best
+#  disk_size: 1_000 # in GiBs
+
+# FIXME: Remember to pass environment variables with `--env` or `--env-file`
+# FIXME: All `null` values must be specified in order for this script to run!
```
jder: What would we fix for these?

train.sky.yaml
jder: My vote would be to use the sky Python SDK and drive it from our existing config system but if that's a pain we can use env vars here.

train.sky.yaml
```diff
@@ -0,0 +1,84 @@
...
+envs:
+  # Secrets
+  # As of v0.9.3, `sky` supports passing in secrets via a dotenv file! So, you can specify these with `--env-file`.
+  # To pass in secrets, we recommend setting environment variable and then ONLY passing in the name via `--env`
+  WANDB_API_KEY: null  # e.g. `--env WANDB_API_KEY`
```
jder, line 36: It looks like secrets should go in a `secrets` dict and are handled differently (eg obscured from UI)? https://docs.skypilot.co/en/latest/running-jobs/environment-variables.html#user-specified-environment-variables-and-secrets and their recommendation is to set them in your env and use `--secret WANDB_API_KEY`

train.sky.yaml
```diff
@@ -0,0 +1,84 @@
...
+    --master_addr=$MASTER_ADDR \
+    --master_port=8008 \
+    --node_rank=${SKYPILOT_NODE_RANK} \
+    -m ocean_emulators.train ${CONFIG} \
+    --data=@${DATA_CONFIG} \
+    --experiment.data_root=/data/ \
+    --experiment.name=${EXPERIMENT_NAME} \
+    --experiment.base_output_dir=/outputs  \
+    --experiment.wandb.mode=online \
+    --samudra.checkpointing=all
```
jder, line 83: 🐑 If we want to keep using the sky yaml/cli, we should maybe consider turning on the env var resolution stuff in the config system so we don't need to thread things through here. (ie setting EXPERIMENT_NAME will be enough). Others here like data_root and base_output_dir are maybe worth moving somewhere that we can have a sky-specific config file for?

jder (approved):
Awesome stuff! I'm sure we'll iterate on the exact structure here as we figure out what we want to do with this. We should resolve the secrets thing before merging but otherwise I think we can merge + iterate from here.

.skyignore
alxmrs: I wanted to ignore files we use for maintaining the project (`scripts/`, `tests/`, `.github/`, etc.) but are not needed for training runs. These are not too consequential to upload, so happy to just use the .gitignore. WDYT?

jder: Yeah, my vote would be to let it upload those and have a single list in .gitignore, but this is a 🐑-level opinion.

train.sky.yaml
alxmrs: I like this idea! I think it's a bigger lift than I want to invest for now. I'll let the experiment name come from the config, but have data and secrets be driven by this YAML's env vars. I think this makes sense bc for the most part, data location won't change as fast as the rest of the configuration.

train.sky.yaml
alxmrs: This changed a lot in recent versions! It used to go in `--secrets`, but in the 0.9.3 and nightly versions, this option is combined.

configs/data/om4_sky_local.yaml
alxmrs: Thanks again for this fix!

.skyignore
alxmrs: Given Ryan's recent comment related to personal/global .gitignore configs, I think that opinion is popular. Will go with just using .gitignore.

PR #308, opened 2025-07-18
Port `test-gpu.yml` to `Open-Athena/ec2` runner

ryan-williams:
Use https://github.com/Open-Athena/ec2-gha/pull/3 (self-terminating GHA EC2 runner)

- 2x speedup:
  - [14mins][Before] → [7mins][After]
  - Previously waited an extra 7mins shutting down instance, before ✅
- Less boilerplate (smaller "start ec2" block, no "stop ec2" block)
- Reusable across repos (e.g. [mamba#771]).
- Pattern should scale to other clouds, Lambda, etc.

[Before]/[After] gif:

![gha]

[Open-Athena/Ocean_Emulator#308]: https://github.com/Open-Athena/Ocean_Emulator/pull/308
[mamba#771]: https://github.com/state-spaces/mamba/pull/771

[Before]: https://github.com/Open-Athena/Ocean_Emulator/actions/runs/16361834629
[After]: https://github.com/Open-Athena/Ocean_Emulator/actions/runs/16362266185?pr=308

[gha]: https://gist.githubusercontent.com/ryan-williams/64629445740677f2d8fcb79946d20d3a/raw/9a26e9d7a90e9c5567063119181f0a4b42296db6/448e9a7a-d978-44ff-8c9c-e86e7260975f.gif


.github/workflows/test-gpu.yml
```diff
@@ -11,16 +11,14 @@ on:
...
-  start-aws-runner:
-    uses: ./.github/workflows/aws-runner-start.yml
-    secrets:
-      AWS_ROLE: ${{ secrets.AWS_ROLE }}
-      GH_SA_TOKEN: ${{ secrets.GH_SA_TOKEN }}
 
+jobs:
+  ec2:
+    uses: Open-Athena/ec2/.github/workflows/runner.yml@main
+    secrets: inherit
```
jder: Is the only way you can tell what secrets this needs now to grep the runner.yml for `secrets.`? If so, I kind prefer being explicit here. Alternatively we could consider setting these up at an [org level](https://github.com/organizations/Open-Athena/settings/secrets/actions), perhaps with more explicit names like OA_ORG_GH_SA_TOKEN or something. (IIUC AWS_ROLE is not actually secret, we might change it to a normal variable sometime)

jder (approved):
Thanks! Just one question/suggestion from me.

alxmrs (approved):

.github/workflows/test-gpu.yml
ryan-williams: Done:

```diff
-    secrets: inherit
+    secrets:
+      GH_SA_TOKEN: ${{ secrets.GH_SA_TOKEN }}
```

jder:
Hey @ryan-williams just wanted to check on the state of this PR. What's left before we can merge this?

ryan-williams:
@jder merging this now is reasonable. The only reason to wait would be because it uses [ec2-gha@v2], which I've only just sent for review at https://github.com/Open-Athena/ec2-gha/pull/3.

It empirically works with this repo's GPU tests, so if you're comfortable with that, we can merge this without waiting on that PR.

[ec2-gha@v2]: https://github.com/Open-Athena/ec2-gha

.github/workflows/test-gpu.yml
```diff
@@ -9,18 +9,20 @@ on:
...
+  contents: read # Required for actions/checkout (normally set by default, but must explicitly specify when defining a custom `permissions` block.
+
 jobs:
-  start-aws-runner:
-    uses: ./.github/workflows/aws-runner-start.yml
+  ec2:
+    uses: Open-Athena/ec2-gha/.github/workflows/runner.yml@v2
+    with:
+      ec2_instance_type: g4dn.xlarge
+      ec2_image_id: ami-00096836009b16a22  # Deep Learning OSS Nvidia Driver AMI GPU PyTorch
```
jder: FYI when you update this there will be a change here to a new AMI.

jder (approved):

.github/workflows/test-gpu.yml
ryan-williams: I'd actually pushed a previous merge here that missed this! I've corrected it now (I force-pushed over the previous bad merge, that also brings in latest `main`, so this thread is now orphaned 🫠 ty!)

jder (approved):
Thanks!

PR #309, opened 2025-07-18
Avoid LocalLocation relative footgun

jder:
Previously, we permitted you to make *relative* local locations as in:

```yaml
data_location:
  type: local
  path: foo/bar
```

But the rule is that if you give a ResolvedLocation (type: local is a ResolvedLocation) then it's *not* resolved relative to the data_root. So this is actually relative to the cwd rather than data_root. This is just confusing so we are now forbidding relative local locations; instead you want:

```yaml
data_location: "foo/bar"
```

This means we can no longer express "resolved paths which are relative to cwd" but this seems preferable to having this footgun.

### DIFF
--- a/src/ocean_emulators/utils/location.py
+++ b/src/ocean_emulators/utils/location.py
@@ -118,7 +118,7 @@ class LocalLocation(ResolvedLocation, BaseModel):
         if not self.path.is_absolute():
             raise ValueError(
                 "Locations with type: local must be absolute. "
-                "For relative paths,use a string instead of a structured location. "
+                "For relative paths, use a string instead of a structured location. "
                 "i.e. 'my/relative/path' instead of "
                 "{type: 'local', path: 'my/relative/path'}"
             )
--- a/src/ocean_emulators/utils/location.py
+++ b/src/ocean_emulators/utils/location.py
@@ -117,7 +117,7 @@ class LocalLocation(ResolvedLocation, BaseModel):
     def validate_path(self) -> Self:
         if not self.path.is_absolute():
             raise ValueError(
-                "Locations with type: local must be absolute. "
+                "Locations with type: 'local' must be absolute. "
                 "For relative paths, use a string instead of a structured location. "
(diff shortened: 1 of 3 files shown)


tests/test_location.py
```diff
@@ -78,7 +83,7 @@ def test_resolve_resolved_location(self):
 
         resolved = base.resolve(other)
 
-        assert resolved is other
+        assert resolved == other
```
jder, line 86: This was a drive-by from another iteration of this PR, but left it since the test was too strict given these are all pure value types.

src/ocean_emulators/utils/location.py
```diff
@@ -113,6 +113,17 @@ class LocalLocation(ResolvedLocation, BaseModel):
     type: Literal["local"] = "local"
     path: Path
 
+    @model_validator(mode="after")
+    def validate_path(self) -> Self:
+        if not self.path.is_absolute():
```
alxmrs, line 118: Thanks! This makes more sense to me. We can always `$(realpath ...)` it on the command line; definitely makes it easier to configure in provisioning scrips.

src/ocean_emulators/utils/location.py
```diff
@@ -113,6 +113,17 @@ class LocalLocation(ResolvedLocation, BaseModel):
...
+            raise ValueError(
+                "Locations with type: local must be absolute. "
```
alxmrs: 🐑 
```suggestion
                "Locations with type: 'local' must be absolute. "
```

src/ocean_emulators/utils/location.py
```diff
@@ -113,6 +113,17 @@ class LocalLocation(ResolvedLocation, BaseModel):
...
+                "For relative paths, use a string instead of a structured location. "
```
alxmrs, line 121: I like that there is an escape hatch to get the old behavior.

alxmrs (approved):
LGTM, thanks for the fix!

PR #310, opened 2025-07-18
MSE Dynamic Loss

jder:
Adds a new loss function to dynamically re-weight loss scaling to make each channel contribute equally to the loss.

### DIFF
--- a/configs/train_om4.yaml
+++ b/configs/train_om4.yaml
@@ -11,7 +11,7 @@ scheduler: true
 loss: mse
 finetune: false
 resume_ckpt_path: null
-inference_epochs: [-1]
+inference_epochs: []
 data_percent: 1.0
 train_time:
   start: "1975-01-03"
@@ -46,9 +46,7 @@ experiment:
   network: Samudra
   prognostic_vars_key: thermo_dynamic_all
   boundary_vars_key: tau_hfds_hfds_anom
-data:
-  !include data/om4_ohc.yaml
-
+data: !include data/om4_ohc.yaml
 
 samudra:
   ch_width: [157, 200, 250, 300, 400]
--- a/src/ocean_emulators/aggregator/validate/main.py
+++ b/src/ocean_emulators/aggregator/validate/main.py
@@ -1,6 +1,5 @@
 import torch
 
-from ocean_emulators.aggregator.loss import LossAggregator
 from ocean_emulators.aggregator.train import TrainAggregator
 from ocean_emulators.aggregator.validate.map import MapAggregator
 from ocean_emulators.aggregator.validate.reduced import MeanAggregator
@@ -31,7 +30,6 @@ class ValidateAggregator(TrainAggregator):
         }
         self._aggregators = val_aggregators
         self.normalize = Normalize.get_instance()
-        self._loss_scaling = LossAggregator.get_instance().loss_scale
         self.hist = hist
         self.num_prognostic_channels = num_prognostic_channels
         self.wet = wet
@@ -96,38 +94,5 @@ class ValidateAggregator(TrainAggregator):
         for agg_label in self._aggregators:
             for k, v in self._aggregators[agg_label].get_logs(label=agg_label).items():
                 logs[f"{label}/{k}"] = v
-        # TODO(jder): we have an implicit assumption here that
-        # the superclass actually only returns float values;
-        # would be nice move that to a separate, maybe standalone, function.
-        logs.update(
-            self._get_loss_scaled_mse_components(
-                validation_metrics=logs,  # type: ignore[arg-type]
-                label=label,
-            )
-        )
+
         return logs
-
-    def _get_loss_scaled_mse_components(
-        self,
-        validation_metrics: dict[str, float],
-        label: str,
-    ):
-        """
-        Account for different scales and units between variables using a
-        custom loss scaling dictionary. Then compute the fractional contributions
-        of each variable to the total loss.
-        """
-        scaled_squared_errors = {}
-
-        for var in self._loss_scaling:
-            rmse_key = f"{label}/mean/weighted_rmse/{var}"
-            if rmse_key in validation_metrics:
-                scaled_squared_errors[var] = (
-                    validation_metrics[rmse_key] / self._loss_scaling[var]
-                ) ** 2
-        scaled_squared_errors_sum = sum(scaled_squared_errors.values())
-        fractional_contribs = {
-            f"{label}/mean/mse_fractional_components/{k}": v / scaled_squared_errors_sum
-            for k, v in scaled_squared_errors.items()
-        }
-        return fractional_contribs
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -293,7 +293,12 @@ class ProfilerConfig(BaseConfig):
 # See backend.py for how these are turned into concrete devices
 TrainBackendConfig = Literal["cpu", "cuda", "nccl", "auto"]
 LossType = Literal[
-    "mse", "mse_diff_weighted", "mse_cos_weighted", "mse_residual_scaled", "mse_mae"
+    "mse",
+    "mse_diff_weighted",
+    "mse_cos_weighted",
+    "mse_residual_scaled",
+    "mse_mae",
+    "mse_dynamic",
 ]
 
 
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -329,7 +329,7 @@ class TrainConfig(TopLevelConfig):
     data_stride: list[int] = [1]
     steps: list[int] = [4]
     step_transition: list[int] = []
-    inference_epochs: list[int] = [-1]
+    inference_epochs: list[int] | None = [-1]
     train_time: TimeConfig = TimeConfig(
         start=JulianDate("0151-01-06"), end=JulianDate("0306-01-01")
     )
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -329,7 +329,7 @@ class TrainConfig(TopLevelConfig):
     data_stride: list[int] = [1]
     steps: list[int] = [4]
     step_transition: list[int] = []
-    inference_epochs: list[int] | None = [-1]
+    inference_epochs: list[int] = [-1]
     train_time: TimeConfig = TimeConfig(
         start=JulianDate("0151-01-06"), end=JulianDate("0306-01-01")
     )
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -299,6 +299,7 @@ LossType = Literal[
     "mse_residual_scaled",
     "mse_mae",
     "mse_dynamic",
+    "mse_dynamic_no_limit",
 ]
 
 
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -551,16 +551,6 @@ class Trainer:
 
             self.optimizer.step()
 
-            if update := getattr(self.loss_fn, "update", None):
-                with torch.no_grad():
-                    # TODO: could avoid a second forward pass here
-                    single_step_data = TrainData(data.num_prognostic_channels)
-                    # Each entry in data is one step in a rollout.
-                    input, label = data[0]
-                    single_step_data.insert(input, label)
-                    pred = self.model(single_step_data)
-                    update(pred[0], label)
-
             lr = (
                 self.optimizer.param_groups[-1]["lr"]
                 if self.scheduler is None
@@ -591,14 +581,27 @@ class Trainer:
                         "data_wait_time"
                     ].value,
                 }
-            if loss_per_channel := getattr(
-                self.loss_fn, "loss_scale_per_channel", None
-            ):
-                metrics.update(
-                    **self.loss_aggregator.get_channel_loss_scale_dict(
-                        label="train", loss_per_channel=loss_per_channel()
+
+                if loss_per_channel_fn := getattr(
+                    self.loss_fn, "loss_scale_per_channel", None
+                ):
+                    loss_per_channel = loss_per_channel_fn()
+                    unscaled_loss_per_channel = TO.loss_per_channel / loss_per_channel
+                    unscaled_loss = torch.mean(unscaled_loss_per_channel)
+
+                    metrics.update(
+                        {
+                            **self.loss_aggregator.get_channel_loss_scale_dict(
+                                label="train", loss_per_channel=loss_per_channel
+                            ),
+                            **self.loss_aggregator.get_channel_loss_dict(
+                                label="train",
+                                loss_per_channel=unscaled_loss_per_channel,
+                                loss_name="loss_unscaled",
+                            ),
+                            "train/batch/loss_unscaled": unscaled_loss,
+                        }
                     )
-                )
 
             if (it_time := metric_logger.meters["iter_time"]).count > 0:
                 metrics["train/batch/iter_time"] = it_time.value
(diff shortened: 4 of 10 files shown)


configs/train_om4.yaml
```diff
@@ -11,7 +11,7 @@ scheduler: true
 loss: mse
 finetune: false
 resume_ckpt_path: null
-inference_epochs: [-1]
+inference_epochs: []
```
jder, line 14: I made this work so we can just disable it altogether.

src/ocean_emulators/aggregator/validate/main.py
```diff
@@ -31,7 +30,6 @@ def __init__(
         }
         self._aggregators = val_aggregators
         self.normalize = Normalize.get_instance()
-        self._loss_scaling = LossAggregator.get_instance().loss_scale
```
jder, line 34: LossAggregator had some loss scaling mechanisms that were never used. I deleted them and that let me delete the aggregator itself.

src/ocean_emulators/utils/logging.py
```diff
@@ -215,7 +215,10 @@ def get_model_summary(model: torch.nn.Module, data: Mapping | None, debug: bool)
     logging.info(f"Number of parameters: {params}")
     depth = 10 if debug else 2
     # we pass verbose = 0 because we log the summary ourselves
-    logging.info(summary(model, input_data=[data], depth=depth, verbose=0))
+    if data is not None:
+        logging.info(summary(model, input_data=[data], depth=depth, verbose=0))
+    else:
+        logging.info(summary(model, depth=depth, verbose=0))
```
jder, line 221: This was a bug which broke eval. Made #317

src/ocean_emulators/train.py
```diff
@@ -365,9 +383,9 @@ def __init__(self, cfg: TrainConfig) -> None:
         self.profiler = cfg.profiler.build(self.output_dir, self.device)
 
         assert self.tensor_map is not None
-        self.loss_aggregator = LossAggregator.init_instance()
 
-        self.init_inference_stores()
+        if self.inference_epochs:
+            self.init_inference_stores()
```
jder, line 391: Previously if you had no epoch to inference at, this would break.

src/ocean_emulators/config.py
```diff
@@ -293,7 +293,13 @@ def build(self, output_dir: Path, device: torch.device) -> Profiler:
...
 TrainBackendConfig = Literal["cpu", "cuda", "nccl", "auto"]
 LossType = Literal[
-    "mse", "mse_diff_weighted", "mse_cos_weighted", "mse_residual_scaled", "mse_mae"
+    "mse",
+    "mse_diff_weighted",
+    "mse_cos_weighted",
+    "mse_residual_scaled",
+    "mse_mae",
+    "mse_dynamic",
+    "mse_dynamic_no_limit",
```
alxmrs, line 302: I was hoping for "no cap"

alxmrs: (not a serious suggestion)

src/ocean_emulators/train.py
```diff
@@ -72,6 +78,7 @@
     handle_warnings,
 )
 from ocean_emulators.utils.loss import (
```
alxmrs, line 81: What do you think about promoting `utils.loss` into `ocean_emulators.loss`?

src/ocean_emulators/train.py
```diff
@@ -238,6 +245,7 @@ def __init__(self, cfg: TrainConfig) -> None:
         self.nets_dir = cfg.experiment.nets_dir
         self.network = cfg.experiment.network
 
+        self.loss_fn: Callable[[torch.Tensor, torch.Tensor], torch.Tensor]
         # Loss function
         if cfg.loss == "mse":
```
alxmrs: 🐑  optional nit: this could be converted to a `match` expression. Though, I totally understand if you'd rather not, say, to minimize the diff.

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
...
     return combined.mean(dim=(0, 2, 3))
+
+
+N_WINDOW = 25
+
+
+class MseDynamic:
+    """A loss function that scales each channel contribute equally to the loss.
+
+    See: https://m2lines.slack.com/files/U086XUETAJ0/F0952H4G3GF/dynamic_loss_weighting.pptx
```
alxmrs: How about we check this into an `assets/` or `docs/` folder?

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
     mae = F.l1_loss(pred, target, reduction="none")
     combined = (mse + mae) / 2
     return combined.mean(dim=(0, 2, 3))
+
+
+N_WINDOW = 25
```
alxmrs: Could this live in the `constants.py` file?

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
...
+        )
+        if should_limit:
+            vars: Float[torch.Tensor, " var"] = stds.pow(2)
+            self.limits: Float[torch.Tensor, " var"] | None = 1.0 / vars
+        else:
+            self.limits = None
+
+    def __call__(
+        self,
+        pred: Float[torch.Tensor, "batch hist*var lat lon"],
```
alxmrs, line 107: Nice use of computed typing with `hist*var`

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
...
+        target: Float[torch.Tensor, "batch hist*var lat lon"],
+    ) -> Float[torch.Tensor, " hist*var"]:
```
alxmrs, line 109: What is the function of the space prefix? I believe that is intentional, but, what is it for?

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
...
+        loss_with_history_channels: Float[torch.Tensor, " hist*var"] = decomposed_mse(
+            pred, target, self.wet
+        )
+        scaled_loss_including_history_dimension: Float[torch.Tensor, "hist var"] = (
```
alxmrs, line 113: Happy about these jaxtyping annotations, makes it much easier to follow.

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
     mae = F.l1_loss(pred, target, reduction="none")
     combined = (mse + mae) / 2
     return combined.mean(dim=(0, 2, 3))
+
+
+N_WINDOW = 25
+
+
+class MseDynamic:
+    """A loss function that scales each channel contribute equally to the loss.
```
alxmrs: I'm a little confused by this sentence, is this what you mean?

```suggestion
    """A loss function that scales each channel contribution equally to the loss.
```

alxmrs: I think additional description of what this does (and why it might be wanted) would be helpful.

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
...
+            loss_with_history_channels.reshape(self.per_channel_scale.shape[0], -1)
+            * self.per_channel_scale.unsqueeze(1)
+        )
+        return scaled_loss_including_history_dimension.reshape(-1)
+
+    def update(
+        self,
+        pred: Float[torch.Tensor, "batch hist*var lat lon"],
+        target: Float[torch.Tensor, "batch hist*var lat lon"],
+    ):
```
alxmrs: I think this method could benefit from a docstring (or maybe a docstring and more descriptive name).

alxmrs:
```suggestion
    ) -> None:
```

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
...
+        mse_loss = decomposed_mse(pred, target, self.wet)
+        mse_loss = torch.where(mse_loss == 0, 1e-8, mse_loss)
+        new_target_weights_with_history: Float[torch.Tensor, " hist*var"] = (
+            1.0 / mse_loss
+        )
+        # Reshape from channels * history to channels by averaging
```
alxmrs: :sheep: I was a little confused until I parsed this more in detail

```suggestion
        # Reshape from channels * history to channels by averaging (along the `hist` dimension)
```

src/ocean_emulators/utils/loss.py
alxmrs: After reading how it's used in the class below, I'd be happy to see this as a static class member, too.

alxmrs: Furthermore, I think this constant could benefit by having a doctring, if only to explain the dimensions of the window. In my understanding, it's approximately timesteps or a window of history looking back? 

A modeling question I have: If I understand the units correctly, do we want a single look back window per every variable type? Put another way: would it ever make sense to use different window sizes to address differing time scales? To answer my own question (to check my understanding): the window can be fixed because we take aggregate statistics over the window, and the variances across each variable will naturally address the difference of each time scale, applied per each time scale.

src/ocean_emulators/utils/loss.py
alxmrs: This makes it clear to me that there is no return, as opposed to the return type merely being omitted.

src/ocean_emulators/utils/loss.py
alxmrs: Thinking aloud -- does this window function as a sort of "learning rate"?

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
...
+
+        if get_world_size() > 1:
+            all_reduce_mean(new_target_weights)
+
+        self.per_channel_scale = (
+            self.per_channel_scale * (N_WINDOW - 1) + new_target_weights
+        ) / N_WINDOW
+
+    def loss_scale_per_channel(self) -> Float[torch.Tensor, " var"]:
+        return self.per_channel_scale
```
alxmrs: If we're exposing this to be read as a public method, then should the `per_channel_scale` member variable be private? On the other hand, why add this function, and not merely access the public member variable?

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,83 @@ def decomposed_mse_mae(
...
+
+    # new methods for saving and loading state
+    def state_dict(self) -> dict[str, torch.Tensor]:
```
alxmrs, line 151: I could see these two methods being packaged into a protocol (happy to have in a later PR).

alxmrs (commented):
Initial thoughts on the PR. I can't wait for dynamic weighting to land, this is quite promising.

src/ocean_emulators/train.py
jder: I don't see much difference, what's the motivation?

src/ocean_emulators/utils/loss.py
jder: My 2 cents is that it's fine to reference things in slack, GitHub, google drive, etc. If it was a markdown file we'd be editing over time I would feel differently but it's for reference and a binary blob.

src/ocean_emulators/utils/loss.py
jder: It is necessary because of weird interaction with other linters: https://docs.kidger.site/jaxtyping/faq/#flake8-or-ruff-are-throwing-an-error

src/ocean_emulators/utils/loss.py
jder: I moved this to a static class variable and added a doc string. For the modeling question, I would ask Adam if you want to discuss it.

src/ocean_emulators/utils/loss.py
jder: Expanded a bit.

src/ocean_emulators/utils/loss.py
jder: I was trying to be consistent with eg `state_dict` which is what Torch uses elsewhere and is a zero-arg function. I will add the underscore.

jder:
@alxmrs thanks for the review! I think I addressed most of these, PTAL!

src/ocean_emulators/utils/loss.py
jder: Added a docstring.

src/ocean_emulators/train.py
alxmrs: I see loss as a primary concern to the modeling rather than a utility.

src/ocean_emulators/utils/loss.py
alxmrs: My feeling is that it's often hard to remember to externalize links like this when we eventually go to open source this model + code. I like front-loading this sort of documentation. 

As a compromise: what about uploading the docs to Google Drive and pasting a Google Drive link (that we can then make open later on)?

src/ocean_emulators/utils/loss.py
alxmrs: Happy to discuss later. I'll think more about how it affects the modeling. In any case, it's an interesting approach and I'm glad it works well!

Thanks for making the other changes.

src/ocean_emulators/utils/loss.py
alxmrs: Thanks for the pointer! That makes sense.

src/ocean_emulators/utils/loss.py
alxmrs: Thanks!

src/ocean_emulators/train.py
alxmrs: Happy to consider this later; not blocking landing this PR.

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,88 @@ def decomposed_mse_mae(
     mae = F.l1_loss(pred, target, reduction="none")
     combined = (mse + mae) / 2
     return combined.mean(dim=(0, 2, 3))
+
+
+class MseDynamic:
+    """A loss function that scales each channel to contribute equally to the loss.
+
+    This uses a rolling estimate of the loss of each channel to scale each
+    channel's loss, discouraging the model from focusing on only a few channels.
```
alxmrs, line 80: Thanks, this explanation is really helpful!

src/ocean_emulators/utils/loss.py
```diff
@@ -67,3 +71,88 @@ def decomposed_mse_mae(
...
+
+    See: https://openathena.slack.com/archives/C08CYM42DT3/p1752275713570969
+    """
+
+    N_WINDOW = 25
+    """Rolling window size to average over. (~number of steps)"""
```
alxmrs, line 86: nit: I think it's conventional to go in the other order.

```suggestion
    """Rolling window size to average over. (~number of steps)"""
    N_WINDOW = 25
```

alxmrs (approved):
Thanks for the changes, this PR LGTM! Excited to have this feature in main. 

(Sorry for posting the review comments one by one)

src/ocean_emulators/utils/loss.py
jder: If we end up open sourcing this code (eg we may not if the experiments don't work out), I don't think we'd want to necessarily release that PPT as well. We're going to have to come back and review/remove/change when that happens. If we're developing in the open then of course this is easier, but for now I don't think we should spend time getting ready to open source each code change.

src/ocean_emulators/utils/loss.py
jder: I agree it is weird, but the [relevant PEP](https://peps.python.org/pep-0257/#what-is-a-docstring) says:

> String literals occurring immediately after a simple assignment at the top level of a module, class, or __init__ method are called “attribute docstrings”.

And according to [various folks on SO](https://stackoverflow.com/questions/8820276/docstring-for-variable), this is what Sphinx, IntelliJ, VS Code, etc, support.

jder:
Thanks for the re-review! I extracted out #326 and replied to your other comments. Planning to merge when green but happy to discuss if open threads afterwards.

PR #313, opened 2025-07-23
Fix EMA state restore when resuming training from a checkpoint

### LINKED ISSUE #282: Resuming from a checkpoint does not seem to restore EMA state correctly
When I restarted [this run](https://wandb.ai/m2lines/ocean-emulators/runs/c5u2f4uu) the ema_cur_decay somehow got reset. Skimming the code I see no good reason for this so would be good to see if we can repro and test/fix.

Image

jder:
This fixes a silly bug which overwrote all EMA state after resuming from a checkpoint. It also now declares the two kinds of checkpoints to be for different uses:

* EMA checkpoints: these bake in the averaged weights into the model itself, and so are useful for inference that wants to start from that averaged model. 
* Non-EMA checkpoints: these now save the averaged model weights separately from the main model weights so we can resume training without starting that averaging over again. 

These maybe could be renamed "inference checkpoints" and "training checkpoints" or something if this turns out to be a useful distinction. 

I ended up not adding a test that training continues exactly as it would have without checkpointing/restoring because I think there are other bugs that are not critical to solve right now like rand seeds not being saved. (I started a [discussion here](https://openathena.slack.com/archives/C0884476QSC/p1753229734048809))

Fixes #282


src/ocean_emulators/utils/ema.py
```diff
@@ -45,7 +45,7 @@ class EMATracker:
 
     def __init__(
         self,
-        model: Samudra | nn.parallel.DistributedDataParallel,
+        model: torch.nn.Module,
```
alxmrs, line 48: Nice

src/ocean_emulators/utils/ema.py
```diff
@@ -81,6 +81,28 @@ def __init__(
```
alxmrs, line 80: ^ This `__init__` starts out with a call to `super()`, but it doesn't look like it is inheriting anything

src/ocean_emulators/utils/ema.py
```diff
@@ -81,6 +81,28 @@ def __init__(
...
+            and all_equal(self.num_updates, other.num_updates)
+            and self._faster_decay_at_start == other._faster_decay_at_start
+            and self._module_name_to_ema_name == other._module_name_to_ema_name
+            and all(
+                all_equal(self._ema_params[k], other._ema_params[k])
+                for k in self._ema_params
+            )
+            and all(
+                all_equal(a, b)
+                for a, b in zip(self._stored_params, other._stored_params, strict=True)
```
alxmrs, line 101: Glad you're using `strict=True`, that's a useful lesson learned.

src/ocean_emulators/utils/ema.py
```diff
@@ -154,37 +176,54 @@ def restore(self, parameters: Iterable[nn.Parameter]):
...
         Returns:
             The EMA tracker.
         """
         ema = cls(model, float(state["decay"]), state["faster_decay_at_start"])
         ema.num_updates = state["num_updates"]
         ema._module_name_to_ema_name = state["module_name_to_ema_name"]
+        if ema_params := state.get("ema_params"):
+            assert ema_params.keys() == ema._ema_params.keys(), (
+                "EMA parameters keys do not match. "
+                "This is likely due to a mismatch between the model and checkpoint."
```
alxmrs, line 225: Thanks, this is a helpful error message.

alxmrs (approved):
Bugfix looks very reasonable. I'll now try to look into why the GPU tests might be failing.

PR #316, opened 2025-07-25
AGENTS.md

jder:
I keep wanting to put things in here so made a first cut at it so we can grow it over time.


AGENTS.md
```diff
@@ -0,0 +1,124 @@
+# AGENTS.md
+
+This file provides guidance to automated agents when working with code in this repository.
+
+## Overview
+
+Ocean Emulator is a PyTorch-based machine learning project for training and evaluating ocean model emulators. It implements a ConvNeXt U-Net neural network architecture.
```
alxmrs:
```suggestion
Ocean Emulator is a PyTorch-based machine learning project for training and evaluating models that emulate ocean physics. Currently, it implements a ConvNeXt U-Net neural network architecture.
```

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Development Commands
```
alxmrs, line 9: For the agents file, can we just direct them to read the CONTRIBUTING guide?

alxmrs: Like, can we direct them to read even specific sections of the contributing guide?

alxmrs: I also like the idea of adding `<details>` sections to the contributing guide – similar to the TL;DR sections – that have specific notes for agents (if we can direct them to read that document).

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+uvx pre-commit run --all-files
+```
+
+### Training, Evaluation, and Visualization
+
+Please ask if you need to train or evaluate a model.
+
+For vizualization or other long-running tasks:
+
+* Run with `PYTHONUNBUFFERED=1 uv run ... > /tmp/logfile.txt 2>&1`
```
alxmrs, line 54: :sheep: think it would be useful to instruct them to use `nohup`?

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+   * Handles OM4 and CM4 ocean model data via Zarr format
+   * Supports time-based train/validation splits
+   * Variables include temperature, salinity, u/v velocities, sea surface height, and heat
+   * Data normalization and preprocessing
+
+3. **Training Loop** (`src/ocean_emulators/train.py`)
+   * Distributed training support via PyTorch DDP
+   * Checkpointing with model state and optimizer
+   * Weights & Biases integration for experiment tracking
+   * Learning rate scheduling and warmup
```
alxmrs, line 77:
```suggestion
   * Learning rate scheduling and warmup
   * Initializes data loaders and training datasets.
```

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+   * Ocean heat content (OHC) analysis
+   * ENSO metrics and basin-specific analysis
+   * Visualization tools for maps, time series, PDFs
+   * Aggregator pattern for metric collection
+
+5. **Configuration System** (`src/ocean_emulators/config.py`)
+   * YAML-based configuration with JSON schema validation
+   * Hierarchical configs with `!include` directives
+   * Pydantic models for type safety
+   * Command-line overrides supported
```
alxmrs, line 91: :sheep: I can see some benefit of adding the `utils/` dir to this list.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Key Design Patterns
+
+1. **Multiton Pattern**: Used for managing global state in tests via `MultitonScope`
```
alxmrs, line 95: I'd leave a note that the word "multiton" is our own invention that is a play on the term "Singleton", except we can instantiate them a few times.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+2. **Factory Pattern**: For creating network blocks dynamically
+3. **Configuration-Driven**: All major components configured via YAML
```
alxmrs, line 97: :sheep: maybe we can say that the `build()` methods in our configuration systems are a light-weight version of the Visitor Pattern.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+├── config.py             # Configuration classes
+├── config_schema.py      # JSON schema generation
+├── datasets.py           # Data loading
+├── models/               # Neural network architectures
+├── aggregator/           # Metric aggregation
+└── utils/                # Utilities for distributed training, logging
+
+configs/                  # YAML configuration files
+tests/                    # Comprehensive test suite
+scripts/                  # Data download and preprocessing
```
alxmrs, line 115: :sheep: What do you think about adding somewhere (e.g. in a section at the top) instructions on how to use `uv` to add dependencies? On the other hand, I see how this might be something we would want to supervise.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+### Important Considerations
+
+1. **Data Format**: Uses Zarr format for efficient array storage
+2. **Distributed Training**: Supports multi-GPU via PyTorch DDP
+3. **Memory Management**: Includes profiling tools (memray, py-spy, scalene)
```
alxmrs: What do you think about calling this consideration "performance oriented" or "performance mindset"? I do think a value in this codebase is to write performant code, not merely correct. To me, "memory mgmt" doesn't quite capture this.

alxmrs (approved):
I like this idea! I am new to writing agents files and I haven't read too much about them. Here are some initial thoughts. At a high level, it would be cool to have the agent read docs in our project like any other (human) contributor. I could see some value in extracting out some of these notes for a more general audience. However, I don't know what the end effect on the quality of SWE agents would be to take that approach, so I'm happy to merge this and iterate on this over time.

AGENTS.md
jder: We can direct them to do that (this PR was originally just that) but it is filled with a lot of irrelevant information for agents, so I think it is better to have a smaller focused doc with the info they need plus anything specific to them. (We could alternatively strip down CONTRIBUTING to look more like this one and just have the agent-specific stuff in here.)

AGENTS.md
jder: I don't think they quit their shells, at least not in my testing.

AGENTS.md
jder: Visitor is a pattern for having polymorphism in how you act on an separate/fixed structure, so I don't think it is. It's a builder pattern :)

AGENTS.md
jder: I don't think there's anything project-specific about that so I think it probably will get it right.

PR #318, opened 2025-07-28
Hopefully fix data race in cache reading/writing

### LINKED ISSUE #257: Data cache (probably) causes data corruption with multi-process pytest
See spurious failure here: https://github.com/suryadheeshjith/Ocean_Emulator/actions/runs/15472999851/job/43562135411?pr=255 

And discussion here: https://openathena.slack.com/archives/C08CYM42DT3/p1749143812784849

Could resolve with our own file-based lock, finding the bug in xarray/dask/zarr, or going back to 1 process.

jder:
Breaks out the test change from #313 and fixes the bug (see eg [this actions run](https://github.com/Open-Athena/Ocean_Emulator/actions/runs/16573115391/job/46880287790)) that exposed.

We now ensure that only one process (of possibly-many pytest processes) write to a given data source cache at once. We also ensure that processes that would like to `open` a given cache wait for any writers of that cache to finish. (They also wait for other readers, though this is unnecessary.) This locking is also hoisted upwards such that only one process will try to create the remote/compact/mock data in the case when it is not cached, which should hopefully reduce the chance of that failing. 

Fixes #257


alxmrs (approved):
Thanks for tracking down this bug and providing a fix. Weird that this started only causing problems recently.

jder:
> Thanks for tracking down this bug and providing a fix. Weird that this started only causing problems recently.

I think it has been causing problems since we turned on multiple pytest proceses, see https://github.com/Open-Athena/Ocean_Emulator/issues/257

PR #319, opened 2025-07-28
Sky Train: Cloning data faster via setup section instead of file mount

alxmrs:
Since `file_mount` options don't let you pass arguments to rclone, this PR opts to manually `rclone` data during the `setup` stage of the job. This allows us to make faster file transfers with the `--transfers` flag.

It also has required that I map the `data/` dir to home and not root.

### DIFF
--- a/configs/data/om4_sky_local.yaml
+++ b/configs/data/om4_sky_local.yaml
@@ -12,6 +12,9 @@ data_stds_location:
   type: local
   path: OM4_means.zarr
 
+# More workers than default; needed for faster data_iter time.
+num_workers: 4
+
 static_data_vars:
   - sea_surface_fraction
   - hfgeou
--- a/configs/data/om4_sky_local.yaml
+++ b/configs/data/om4_sky_local.yaml
@@ -12,9 +12,6 @@ data_stds_location:
   type: local
   path: OM4_means.zarr
 
-# More workers than default; needed for faster data_iter time.
-num_workers: 4
-
 static_data_vars:
   - sea_surface_fraction
   - hfgeou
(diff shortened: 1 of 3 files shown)


train.sky.yaml
```diff
@@ -48,20 +47,37 @@ envs:
...
+  mkdir ~/data/
+
+  RCLONE_CONF=$(rclone config file | awk '/Configuration file is stored at:/ {getline; print}')
+
+  cat << EOF > $RCLONE_CONF
+  [s3-source]
+  type = s3
+  provider = AWS
+  env_auth = true
+  EOF
```
jder: I think `rclone config create s3-source s3 provider=AWS env_auth=true` will do this for you rather than having to sniff the config location.

train.sky.yaml
```diff
@@ -48,20 +47,37 @@ envs:
...
+  # The current method of copying (in the setup section, below) is faster.
 
   /outputs:
     source: s3://${OUTPUTS_BUCKET}
     mode: MOUNT_CACHED
 
 workdir: .
 
+setup: |
+  mkdir ~/data/
```
jder, line 60: Should this be `./data`? Not sure what the expected pwd vs home directory setup is.

configs/data/om4_sky_local.yaml
```diff
@@ -12,6 +12,9 @@ data_stds_location:
   type: local
   path: OM4_means.zarr
 
+# More workers than default; needed for faster data_iter time.
+num_workers: 4
```
jder: 🐑 I had to do this too, we might just change the default back to 4…

jder (approved):
Just some minor suggestions from me, excited to try this!

train.sky.yaml
alxmrs: Thanks!

train.sky.yaml
alxmrs: I think both of these go to the same place, reading the logs, but I'm not totally sure. So long as they are consistent, I'm indifferent as to where the data directory should go. The file mounts default to putting them in the root directory, but I have a hard time writing to that location with the permissions of the setup step.

PR #324, opened 2025-07-30
Initializing EMA after wrapping model with DDP

alxmrs:
I encountered a bug where I hit a key error during the EMATracker call phase where it was unable to find the key for any layer.
```
 File "/home/ubuntu/sky_workdir/src/ocean_emulators/utils/ema.py", line 129, in __call__
    ema_name = self._module_name_to_ema_name[key]
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  KeyError: 'module.layers.0.skip_module.weight'
```

I found that this was caused by wrapping the `DDP` component onto the model -- doing so adds a `module` prefix to the weight names. Thus, the names of the weights don't match the model setup and EMATracker call during training.

In this PR, I make a simple fix, which is to just move the DDP wrapper code above the part where we set up the EMATracker.

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -310,6 +310,13 @@ class Trainer:
             cfg.resume_ckpt_path, cfg, finetune=cfg.finetune
         )
 
+        # Modify DDP setup based on device
+        if self.distributed is not None:
+            self.model = nn.parallel.DistributedDataParallel(
+                nn.SyncBatchNorm.convert_sync_batchnorm(self.model),
+                device_ids=[self.distributed.gpu],
+            )
+
         self.num_batches_seen = 0
         if cfg.resume_ckpt_path is not None:
             if cfg.finetune:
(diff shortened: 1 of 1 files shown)


jder:
Thanks, sorry about that. I already fixed this in #310 (we could break it out) in a [slightly different way](https://github.com/Open-Athena/Ocean_Emulator/pull/310/commits/16a53a22bbdc7c090b62a5d7c361a0f946542cef) because I was iterating on that branch. I think with the fix in this PR, resuming from a checkpoint into a distributed model won't work because load_checkpoint only works with non-DDP models since it intentionally removes "module." from loaded key names. (I assume to make it possible to train in a distributed fashion and then eval either way.) Really it would be nice to get rid of that prefix removal and have a principled way to create and access the underlying (non-DDP) module. (Or use a framework which would get us that for free)

alxmrs:
No problem. Sounds good to break the fix out. That makes sense, your approach does sound better than this PR's.

jder:
Broke it out into #326

alxmrs:
Closing in favor of the alternative PR (https://github.com/Open-Athena/Ocean_Emulator/pull/326).

PR #325, opened 2025-08-04
Fix UnresolvedLocation JSON schema to permit string

jder:
This was previously causing red squigglies in VS Code. It's annoying that this annotation and the two `@model_*` annotations on this class are all required -- I spent some time looking for a more direct way to express "this type serializes to/from a string" in the pydantic docs but everything I tried was broken or more complicated than this.


alxmrs (approved):
LGTM! Any idea what causes the error in VS code? I don't see anything similar in PyCharm.

PR #327, opened 2025-08-05
Standard deviations should be stds, not means. 😬

alxmrs:


jder (approved):
Sorry, should have caught this in review! Nice find, makes me wonder how we could have automatically detected this. I guess if we merged these all together into a single zarr file it would be harder to do this.

alxmrs:
I am wondering, too. It was a silly copy+paste error on my part, and boy was it hard for me to notice. 

Yeah - an Xarray/Zarr DataTree would be a good fit for this kind of thing. This seems better than where my mind was going, which is writing some sort of unit test to cover this config (that approach's problem is that mis-configuration is IID, and how much of the config do we deeply cover?).

alxmrs:
One initial idea for how I could catch things like these in the future -- I bet I could ask Claude for a review, paying attention to possible silly mistakes! I wouldn't be surprised if that worked nearly all the time :)

PR #328, opened 2025-08-05
Add .git to .gitignore for the sake of skypilot

alxmrs:
Without this, it seems that skypilot attempts to upload the whole .git folder. Adding this makes it skip doing so. To me, this is a bug in skypilot; I'll file an issue over there and use this as a workaround for the time being.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.git
 __pycache__
 .DS_Store
 .ipynb_checkpoints


alxmrs:
xref: https://github.com/skypilot-org/skypilot/issues/6513

jder (approved):
Seems fine to me. You might remember I was advocating for ignoring `.*` but we didn't move forward because of the `.github` folder which we want to be tracked -- I realized we can unignore that with `!.github` in the gitignore so that's also an option which might be more likely to catch things we don't want synced.

alxmrs:
Ah, that's a good shout! I do like the idea of a whitelist than a blacklisit in the first place. That comment will make a good issue IMO :)

PR #333, opened 2025-08-06
Stop complaining about long lines

jder:
This rule seems low-value to me, and we already enforce `ruff format` which will wrap lines when reasonable. If ruff format is OK with the line length I don't think we should be wrapping things by hand.


alxmrs (approved):
Reading title in my email, I thought you might have taken Robin to Disneyland in the summer. 

LGTM.

PR #334, opened 2025-08-06
Add a scheduler which can run a cosine schedule followed by a flat tail

jder:
To try extending the behavior seen greatly improving loss at low learning rates from [this run](https://openathena.slack.com/archives/C08CYM42DT3/p1753308638843249?thread_ts=1752275713.570969&cid=C08CYM42DT3).


alxmrs (approved):
Excellent interface, thanks for adding this! I'm really liking the build methods.

PR #336, opened 2025-08-08
Early stochastic depth dropout experiments

alxmrs:
After reviewing three relevant papers (see comments for citations), I've provided affordances for adding early stochastic depth dropout in our modeling baseline experiments. Based on my reading of this paper, in particular (_Dropout reduces underfitting_, https://arxiv.org/pdf/2303.01500), I believe adding early drop out that is appropriate for our ConvNext UNet model. This should lead to lower losses on our half-degree baseline run.


src/ocean_emulators/models/modules/blocks.py
```diff
@@ -129,6 +131,15 @@ def forward(self, fts: torch.Tensor) -> torch.Tensor:
...
                 fts = layer(fts)
+
+        # Apply stochastic depth to non-residual blocks (adapted from 1603.09382)
+        if hasattr(self, "drop_path"):
+            # For non-residual blocks, drop_path will either return the transformed features
+            # or zero them out, effectively making this layer an identity when dropped
+            transformation = fts - input_fts  # Get the transformation
+            dropped_transformation = self.drop_path(transformation)
+            fts = input_fts + dropped_transformation
+
```
alxmrs, line 142: I'm not sure if we want to add this now, but I did for consistency.

src/ocean_emulators/models/modules/dropout.py
```diff
@@ -0,0 +1,83 @@
...
+      https://arxiv.org/abs/1603.09382
+    - "A ConvNet for the 2020s" (2201.03545)
+      https://arxiv.org/abs/2201.03545
+    """
+
+    def __init__(
+        self,
+        drop_prob: float = 0.0,
+        early_epochs: int = 0,
+        schedule: Literal["early_only", "late_only", "constant"] = "early_only",
```
alxmrs: Consider extracting this literal into a type in the config file

src/ocean_emulators/models/modules/dropout.py
```diff
@@ -0,0 +1,83 @@
...
+    def get_current_drop_prob(self) -> float:
+        """Calculate dropout probability based on current epoch.
+
+        Implements epoch-based scheduling as described in:
+        "Early Dropout: Dropping layers to reduce underfitting" (2303.01500)
+        """
+        if self.base_drop_prob == 0.0 or self.early_epochs == 0:
+            return 0.0
+
+        if self.schedule == "early_only":
```
alxmrs: Nit: replace with match expression

src/ocean_emulators/utils/dropout.py
```diff
@@ -0,0 +1,92 @@
...
+        if self.config.per_stage_multipliers:
+            # Ensure we don't go out of bounds
+            multiplier_index = min(
+                layer_index, len(self.config.per_stage_multipliers) - 1
+            )
+            stage_multiplier = self.config.per_stage_multipliers[multiplier_index]
+
+        return self.config.drop_path_rate * stage_multiplier
+
+    def create_drop_path(self, layer_index: int) -> "EarlyDropPath | None":
```
alxmrs: Normal type annotation?

src/ocean_emulators/config.py
```diff
@@ -203,6 +210,57 @@ class CorrectorConfig(BaseConfig):
...
+        Late dropout with constant rate:
+        >>> config = StochasticDepthConfig(
+        ...     drop_path_rate=0.1, early_dropout_epochs=30,
+        ...     dropout_schedule="late_only"
+        ... )
+    """
+
+    # Early dropout settings (2303.01500)
+    drop_path_rate: float = 0.0  # Peak drop rate during early training, e.g. 0.1-0.2
+    early_dropout_epochs: int = 0  # Apply for first N epochs (0 = disabled)
```
jder, line 240: I think it would be more clear to have early_dropout_epochs be a member of a struct EarlyDropoutSchedule. (And then you can have either late_dropout_epochs or early_no_dropout_epochs be a member of LateDropoutSchedule.)

src/ocean_emulators/config.py
```diff
@@ -203,6 +210,57 @@ class CorrectorConfig(BaseConfig):
...
+    dropout_schedule: DropoutSchedule = "early_only"
+
+    # Optional: Linear decay within early period (2303.01500)
+    linear_decay_to_zero: bool = (
+        False  # Decay from drop_path_rate to 0 over early_dropout_epochs
```
jder, line 245: I assume this is not over `early_dropout_epochs` if we're using late dropout?

src/ocean_emulators/config.py
```diff
@@ -203,6 +210,57 @@ class CorrectorConfig(BaseConfig):
...
+    )
+
+    # Optional: per-stage rates for spatial variation (inspired by 2201.03545)
+    per_stage_multipliers: list[float] | None = None  # e.g., [0.5, 1.0, 1.5, 2.0]
```
jder, line 249: Should this be `per_layer_multipliers`? Stage sounds to me like it's a time thing?

src/ocean_emulators/utils/dropout.py
```diff
@@ -0,0 +1,97 @@
...
+        if self.config.drop_path_rate == 0.0 or self.config.early_dropout_epochs == 0:
+            return 0.0
+
+        # Apply per-stage multiplier if specified
+        stage_multiplier = 1.0
+        if self.config.per_stage_multipliers:
+            # Ensure we don't go out of bounds
+            multiplier_index = min(
+                layer_index, len(self.config.per_stage_multipliers) - 1
+            )
```
jder, line 36: Would it be better to complain if the size of per_stage_multipliers doesn't match the number of layers when building into the StochasticDepthManager? (You can pass the known number of layers into build)

src/ocean_emulators/utils/dropout.py
```diff
@@ -0,0 +1,97 @@
...
+        return self.config.drop_path_rate * stage_multiplier
+
+    def create_drop_path(self, layer_index: int) -> "ScheduledDepthDropout | None":
+        """Create a DropPath module for a specific layer.
+
+        Returns None if dropout is disabled for this configuration.
+        """
+        # Early return for better performance when dropout is disabled
+        if not self.is_enabled():
+            return None
```
jder, line 48: I would bet this is not a performance improvement :)

src/ocean_emulators/models/modules/dropout.py
```diff
@@ -0,0 +1,124 @@
...
+import torch
+import torch.nn as nn
+
+from ocean_emulators.config import DropoutSchedule
+
+logger = logging.getLogger(__name__)
+
+
+class ScheduledDepthDropout(nn.Module):
+    """Epoch-aware Drop Path for (early) dropout experiments.
```
jder, line 12:
```suggestion
    """Epoch-aware Drop Path for dropout experiments.
```

src/ocean_emulators/utils/dropout.py
```diff
@@ -0,0 +1,97 @@
...
+
+class StochasticDepthManager:
+    """Centralized manager for applying stochastic depth to models.
+
+    Handles all dropout rate calculations, layer registration, and
+    epoch tracking in one place.
+    """
+
+    def __init__(self, config: "StochasticDepthConfig"):
+        self.config = config
```
jder, line 18: Generally my goal with the MyConfig.build pattern is that you don't keep the config object around later. It's raw data which is designed to be ergonomic for setting from CLI/YAML and so has few validity constraints. Can we pass (maximally-validated) values here instead?

src/ocean_emulators/utils/dropout.py
```diff
@@ -0,0 +1,97 @@
...
+        self.drop_path_modules: list[ScheduledDepthDropout] = []
+
+    def calculate_drop_rate(self, layer_index: int) -> float:
+        """Calculate dropout rate for a specific layer.
+
+        Implements stage-wise dropout rate calculation with optional per-stage
+        multipliers as described in the papers.
+        """
+        if self.config.drop_path_rate == 0.0 or self.config.early_dropout_epochs == 0:
```
jder, line 27: I think there's a bug here related to my comment below about the name `early_dropout_epochs` -- if `self.config.early_dropout_epochs` is 0 and `self.config.dropout_schedule` is `late_dropout` then it's *always* enabled, not *never* enabled.

src/ocean_emulators/models/modules/dropout.py
```diff
@@ -0,0 +1,124 @@
...
+                    decay_factor = 1.0 - (self.current_epoch / self.early_epochs)
+                    return self.base_drop_prob * decay_factor
+                else:
+                    # Constant rate during early period
+                    return self.base_drop_prob
+
+            case "late_only":
+                if self.current_epoch < self.early_epochs:
+                    return 0.0  # No dropout during early period
+                return self.base_drop_prob
```
jder, line 98: Seems weird that linear_decay does nothing in this case?

src/ocean_emulators/models/modules/factory.py
```diff
@@ -31,10 +31,25 @@
 }
 
 
-def create_block(block_type: str, **kwargs) -> nn.Module:
+def create_block(
+    block_type: str,
+    dropout_manager=None,  # StochasticDepthManager | None
```
jder, line 36: Why is this commented out?

src/ocean_emulators/models/modules/factory.py
```diff
@@ -31,10 +31,25 @@
...
         raise ValueError(f"Unknown block type: {block_type}")
-    return BLOCK_REGISTRY[block_type](**kwargs)
+
+    block = BLOCK_REGISTRY[block_type](**kwargs)
+
+    # Apply dropout if supported (2303.01500)
+    if dropout_manager is not None and isinstance(block, (ConvNeXtBlock, ConvBlock)):
+        drop_path = dropout_manager.create_drop_path(layer_index)
+        if drop_path is not None:
+            block.drop_path = drop_path
```
jder, line 50: I think blocks are always one of `ConvNeXtBlock, ConvBlock`, so I think we can/should instead pass `drop_path` directly into `create_block` which can pass it along via kwargs to `CoreBlock.__init__`?

src/ocean_emulators/utils/dropout.py
```diff
@@ -0,0 +1,97 @@
...
+        # Register for epoch tracking
+        self.drop_path_modules.append(drop_path)
+
+        return drop_path
+
+    def register_model(self, model: nn.Module) -> None:
+        """Register all dropout modules in the model for epoch tracking.
+
+        This scans the model for any EarlyDropPath modules that might not have
+        been registered during creation and adds them to the tracking list.
```
jder, line 74: Why would these have not been registered? That seems like a bug?

src/ocean_emulators/utils/dropout.py
```diff
@@ -0,0 +1,97 @@
...
+            # Ensure we don't go out of bounds
+            multiplier_index = min(
+                layer_index, len(self.config.per_stage_multipliers) - 1
+            )
+            stage_multiplier = self.config.per_stage_multipliers[multiplier_index]
+
+        return self.config.drop_path_rate * stage_multiplier
+
+    def create_drop_path(self, layer_index: int) -> "ScheduledDepthDropout | None":
+        """Create a DropPath module for a specific layer.
```
jder, line 42:
```suggestion
        """Create a ScheduledDepthDropout module for a specific layer.
```

jder: (We say `path` in a bunch of places but maybe that's out of date?)

jder (commented):
Ok, I only got part way through this review but sending these comments along since I have to run. Generally looks reasonable to me, though!

There are a few issues with late_dropout -- do we need that? Generally I'm not sure how to think about reviewing code which we might not want depending on some experiments we're about to run. Maybe it makes more sense to do some experiments and then get a review on the code we want to keep? I'm honestly not sure.

src/ocean_emulators/models/modules/dropout.py
```diff
@@ -0,0 +1,124 @@
...
+        keep_prob = 1 - current_drop_prob
+
+        # Create random tensor with shape (batch_size, 1, 1, ...) to broadcast properly
+        # This works for any number of dimensions
+        mask_shape = (x.shape[0],) + (1,) * (x.ndim - 1)
+        random_tensor = keep_prob + torch.rand(
+            mask_shape, dtype=x.dtype, device=x.device
+        )
+        binary_mask = torch.floor(random_tensor)
+        return x.div(keep_prob) * binary_mask
```
jder, line 124: I was a little surprised this didn't end up using torch.Dropout somewhere. Is that a significantly different thing? I see torch also has conv-specific dropout layers like Dropout2d.

src/ocean_emulators/config.py
alxmrs: Maybe, `per_level_multiplier`? I want to make it clear that it refers to both sides of the u-net. I'll double check if "stage" comes from the ConvNext paper; if so, then I think it'd be best to keep the same terminology.

src/ocean_emulators/utils/dropout.py
alxmrs: That's a good idea for an early error to throw!

src/ocean_emulators/utils/dropout.py
alxmrs: haha! Claude thought so, and I don't mind returning early. I'll think critically about it in the next review pass.

src/ocean_emulators/utils/dropout.py
alxmrs: Sure, that SGTM. This separation of concerns makes sense, thanks for the explanation.

tests/test_dropout.py
jder, line 1: I went to review this today and I don't think we need 800 lines of test to build confidence in the correctness of this class. For example, I don't think any of the tests in `TestScheduledDepthDropout` have much value and maybe even be encoding incorrect behavior (set_epoch(-1) should probably throw an error if anything). I think an integration test (perhaps using a full model/config) and a dozen examples of some config and expected epoch -> dropout mappings is probably sufficient?

src/ocean_emulators/models/modules/dropout.py
alxmrs: Just the other day, I encountered an example of this type of dropout in the wild, see: https://github.com/NVIDIA/torch-harmonics/blob/00e6f65542e88df6af3f80cd4a1ee8c8950e412d/torch_harmonics/examples/models/s2unet.py

Specifically, see the `DropPath` feature here.

alxmrs: That is to say: This is different than torch.Dropout2d IIUC, both can be used at once.

alxmrs: In addition: this codebase makes me think that this CC generated implementation could be a lot simpler.

alxmrs:
This is a nice direction, but ultimately stale. Closing for now.

Issue #337, opened 2025-08-08
Skypilot runs don't have git hashes


jder:
It looks like skypilot doesn't have a git checkout on the remote host so it doesn't record the git hash. Compare:

non-skypilot: https://wandb.ai/m2lines/ocean-emulators/runs/c5u2f4uu/overview

<img width="773" height="66" alt="Image" src="https://github.com/user-attachments/assets/2258f148-f483-4e56-8494-883fa83d699c" />

skypilot: https://wandb.ai/m2lines/ocean-emulators/runs/rfy2paom/overview

Can we record this somehow?

alxmrs:
This may be my bad: I added `.git/` to the gitignore for faster syncs.

jder:
Fixed by #338, see for example https://wandb.ai/m2lines/ocean-emulators/runs/k5szfus3/overview

PR #338, opened 2025-08-08
Un-ignoring .git folder for skypilot syncs

alxmrs:
I think this could address #337.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.git
 __pycache__
 .DS_Store
 .ipynb_checkpoints


jder (approved):
Was this causing the skypilot runs to be slow before? I'm fine with this revert but if there's a cheaper way to log the git info that works for me too.

alxmrs:
It wasn't especially slow, especially compared to the time it takes to transfer the data from s3. This mainly improved the ability to sync my local workdir when coding on the go (e.g. tethered to my laptop in a car), where I'd often experience shaky internet issues that would cause problems for the experiment launch. In retrospect, this is probably not the situation we should optimize for 😉 -- better to have the git commit in the wandb logs.

PR #339, opened 2025-08-08
Vizualization code

jder:
This is an import and reorganization of the code from [this notebook](https://github.com/Open-Athena/Ocean_Emulator/blob/fixed-branch/Notebooks/notebooks/2025-04-08-Samudra_Eval_OM4.ipynb). The goal was to get it to a state we could check it in and iterate from there. This PR:

* Moves that code from a notebook to python code in src/ocean_emulators/viz/core.py
* Creates a new TopLevelConfig for visualization in src/ocean_emulators/viz/config.py (see CONTRIBUTING.md for an example invocation)
* Does a minimal amount of deduplication, cleanup, and typing of the code there to get it passing CI and to allow the config to configure all the main knobs. Also added some TODOs about clear things we might want to do next.
* Includes a tool to compare old and new viz output -- this is how I verified that the previous notebook and this code produce nearly identical results (modulo some FP error due to slight changes in ordering, I think).
* Includes pointers to the basin data, now in the emulator pod, which was created by the new `notebooks/regrid_basins.py` code, which I included for posterity but don't think we really need to maintain per se.

I would not try to review core.py too deeply but skimming the structure is good; same with notebooks/regrid_basins.py. Everything else is fair game :)


notebooks/regrid_basins.py
```diff
@@ -0,0 +1,551 @@
...
+
+from pathlib import Path
+from typing import Dict, Any
+
+import numpy as np
+import xarray as xr
+from scipy.spatial import KDTree
+
+BASINS_PATH = Path("/Users/jder/oa/data/basins/")
+HIGH_RES_DATA_PATH = Path("/Users/jder/oa/data/half_deg_10y/")
```
alxmrs, line 26: Maybe we can make them relative to the current machine?

notebooks/regrid_basins.py
```diff
@@ -0,0 +1,551 @@
...
+
+# Define basin names and their IDs
+BASIN_INFO: Dict[str, Dict[str, Any]] = {
```
alxmrs, line 29: 🐑 I could see using a `TypedDict` on the inner dictionary as a sort of light-weight data class.

notebooks/regrid_basins.py
```diff
@@ -0,0 +1,551 @@
...
+    )
+    print(
+        f"Target lat range: {target_grid.lat.min().values:.2f} to {target_grid.lat.max().values:.2f}"
+    )
+
+    return target_grid
+
+
+# %%
+def regrid_combined_basins(source_ds, target_grid):
```
alxmrs, line 132: This is a cool implementation! Great use of kdtrees. In the past, I've always seen this library used for regridding: https://xesmf.readthedocs.io/en/stable/

Any reason why we should roll our own regridding system?

src/ocean_emulators/utils/compare.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # Convert to numpy arrays and compare
+        arr1 = np.array(img1)
+        arr2 = np.array(img2)
+
+        if np.array_equal(arr1, arr2):
+            return True
+
+        # If different and diff_dir is provided, create difference image
+        if diff_dir:
```
alxmrs, line 38: Neat img diff!

src/ocean_emulators/utils/compare.py
```diff
@@ -0,0 +1,174 @@
...
+            composite[:, :width] = arr1
+            composite[:, width + 10 : width * 2 + 10] = arr2
+            composite[:, width * 2 + 20 :] = diff_enhanced
+
+            # Save the composite image
+            diff_path = diff_dir / f"diff_{Path(file1).name}"
+            Image.fromarray(composite).save(diff_path)
+
+        return False
+    except Exception as e:
```
alxmrs, line 57: Is there a more specific error you expect to get?

src/ocean_emulators/utils/compare.py
```diff
@@ -0,0 +1,174 @@
...
+    matching_count = 0
+
+    # Compare common files
+    print(f"Comparing {len(common_files)} common files...")
+    for f in common_files:
+        file1 = dir1 / f
+        file2 = dir2 / f
+
+        # Special handling for PNG files
+        if str(f).lower().endswith(".png"):
```
alxmrs: 🐑 I suspect we're always producing pngs. It looks like `compare_images` is agnostic to the image format. A small nit: maybe we could check the mime type of the file and make this work for all image files? https://docs.python.org/3/library/mimetypes.html

src/ocean_emulators/utils/compare.py
```diff
@@ -0,0 +1,174 @@
...
+    print(f"Common files: {len(common_files)}")
+    print(f"Matching files: {matching_count}")
+
+    # Print present but different files first
+    if content_different:
+        print(f"\nFiles with different content ({len(content_different)}):")
+        for f in content_different[:10]:
+            print(f"  - {f}")
+        if len(content_different) > 10:
+            print(f"  ... and {len(content_different) - 10} more")
```
alxmrs: This abbreviation is nice.

src/ocean_emulators/utils/compare.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Then print missing files
+    if missing_in_dir2:
+        print(f"\nFiles only in {dir1.name} ({len(missing_in_dir2)}):")
+        for f in missing_in_dir2[:10]:
+            print(f"  - {f}")
+        if len(missing_in_dir2) > 10:
+            print(f"  ... and {len(missing_in_dir2) - 10} more")
```
alxmrs: I think this could be extracted into a function. Even a private function that's defined within this existing func would be nice.

src/ocean_emulators/viz/__main__.py
```diff
@@ -0,0 +1,3 @@
+from ocean_emulators.viz.config import VizConfig, main
+
+main(VizConfig.from_yaml_and_cli())
```
alxmrs, line 3: Sweet

src/ocean_emulators/viz/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class VizConfig(TopLevelConfig):
+    base_output_dir: Path
+    name: str
+    dataset_name: str
+    runs: list[VizRunConfig]
+    data_root: Location | None = None
+    groundtruth_location: Location
+    basins_location: Location
```
alxmrs, line 34: Will these always be local locations?

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+import sys
+import warnings
+from collections.abc import Iterable
+from copy import deepcopy
+from subprocess import PIPE, STDOUT, Popen
+from typing import Any
+
+import cartopy.crs as ccrs  # type: ignore
+import cartopy.feature as cfeature  # type: ignore
+import cmocean as cm  # type: ignore
```
alxmrs, line 15: I have found this imports to also be a pain wrt to typing. And, sometimes with package managers. Great going getting this to work.

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+        var_list = {
+            "vo": r"$v$ $( m/s )$",
+            "uo": r"$u$ $( m/s )$",
+            "thetao": r"$T$ $( ^\circ C )$",
+            "tos": r"$T$ $( ^\circ C )$",
+            "so": r"$so$ $( psu )$",
+            "zos": r"$zos$ $( m )$",
+            "KE": r"$KE$ $( J/m^2 )$",
+            "OHC": r"$OHC$ $Anomaly$ $( ZJ )$",
+        }
```
alxmrs, line 88: 🐑  Maybe this could go in a place for constants?

alxmrs: I'm undecided if it should go in project constants or a viz-specific place.

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+        if not os.path.isdir(self.temp_path):
+            os.makedirs(self.temp_path)
+
+        self.salinity_path = os.path.join(output_path, f"Salinity")
+        if not os.path.isdir(self.salinity_path):
+            os.makedirs(self.salinity_path)
+
+        self.pdfs_path = os.path.join(output_path, f"PDFs")
+        if not os.path.isdir(self.pdfs_path):
+            os.makedirs(self.pdfs_path)
```
alxmrs, line 109: :sheep: A tiny/private helper function might be nice here.

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+            pacific_mask0["lat"] >= -32
+        )  # TODO: include this -32 masking in the basin data
+        pacific_mask = process_mask(data, pacific_mask)
+        indian_ocean_mask0 = basins["basin_indian"]
+        indian_ocean_mask = indian_ocean_mask0.where(indian_ocean_mask0["lat"] >= -32)
+        indian_ocean_mask = process_mask(data, indian_ocean_mask)
+        southern_ocean_mask0 = basins["basin_southern"]
+        southern_ocean_mask = process_mask(data, southern_ocean_mask0)
+        arctic_mask0 = basins["basin_arctic"]
+        arctic_ocean_mask = process_mask(data, arctic_mask0)
```
alxmrs, line 139: ah, yes. the "seven seas".

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+        # Compute profile means
+        with ProgressBar():
+            print("Ground truth " + dataset_name)
+            profile_groundtruth = profile_mean(data).load()
+
+            for k in pred_dict.keys():
+                print(k)
+                pred_dict[k]["profile_prediction"] = profile_mean(
+                    pred_dict[k]["ds_prediction"]
+                ).load()
```
alxmrs, line 160: Print statements and progress bars in a constructor feels a little off to me. Is there a better place for this? What is your thinking here?

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+        self.thetao_mae_metrics()
+        self.sst_mae_metrics()
+        self.pdf_plots_short()
+        self.enso_plots()
+        self.ohc_maps()
+        self.sst_mean_maps()
+        self.sst_time_snapshot_maps()
+        self.salinity_mean_map()
+        self.salinity_snapshot_maps()
+        self.movies()
```
alxmrs, line 201: I like how the details about each of these methods are commented.

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+                OHC_pred.time.data,
+                np.arange(OHC_pred.size) * coeffs_OHC_pred_trend[0]
+                + coeffs_OHC_pred_trend[1],
+                c=self.clist[i],
+                ls="--",
+            )
+            # ax[0].annotate(f'{coeffs_OHC_pred_trend[0]:.2e}',
+            #          xy=(pos.get_xdata()[-1], pos.get_ydata()[-1]),
+            #          xytext=(pos.get_xdata()[-2], pos.get_ydata()[-2]),
+            #          fontsize=9, color=self.clist[i])
```
alxmrs, line 731: Omit dead code?

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+
+        OHC.plot(ax=ax, label=self.dataset_name, c="k")  # type: ignore
+        coeffs_OHC_trend = np.polyfit(np.arange(OHC.size), OHC, 1)
+        (pos,) = ax.plot(
+            OHC.time.data,
+            np.arange(OHC.size) * coeffs_OHC_trend[0] + coeffs_OHC_trend[1],
+            c="k",
+            ls="--",
+        )
+        # ax[0].annotate(f'{coeffs_OHC_trend[0]:.2e}',
```
alxmrs, line 745: Same here

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+            1,
+        )
+        (pos,) = ax[0].plot(
+            OHC_truth_upper[Days_to_Eq:].time.data,
+            np.arange(OHC_truth_upper[Days_to_Eq:].size) * coeffs_OHC_ground_trend[0]
+            + coeffs_OHC_ground_trend[1],
+            c="k",
+            ls="--",
+        )
+        # ax[0].annotate(f'{coeffs_OHC_ground_trend[0]:.2e}',
```
alxmrs, line 862: omit?

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+            )
+            (pos,) = ax[0].plot(
+                self.pred_dict[k]["OHC_pred_upper"][Days_to_Eq:].time.data,
+                np.arange(self.pred_dict[k]["OHC_pred_upper"][Days_to_Eq:].size)
+                * self.pred_dict[k]["coeffs_OHC_pred_trend_upper"][0]
+                + self.pred_dict[k]["coeffs_OHC_pred_trend_upper"][1],
+                c=self.clist[i],
+                ls="--",
+            )
+            # ax[0].annotate(f'{self.pred_dict[k]["coeffs_OHC_pred_trend_upper"][0]:.2e}',
```
alxmrs, line 908: omit?

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+            print(
+                self.pred_dict[k]["name"],
+                " upper_trend_pred: ",
+                self.pred_dict[k]["upper_trend_pred"],
+            )
+            self.pred_dict[k]["total_trend_pred"] = self.pred_dict[k][
+                "upper_trend_pred"
+            ]
+
+        # ax[0].legend()
```
alxmrs, line 928: omit?

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+            np.arange(OHC_truth_mid[Days_to_Eq:].size), OHC_truth_mid[Days_to_Eq:], 1
+        )
+        (pos,) = ax[1].plot(
+            OHC_truth_mid[Days_to_Eq:].time.data,
+            np.arange(OHC_truth_mid[Days_to_Eq:].size) * coeffs_OHC_ground_trend[0]
+            + coeffs_OHC_ground_trend[1],
+            c="k",
+            ls="--",
+        )
+        # ax[1].annotate(f'{coeffs_OHC_ground_trend[0]:.2e}',
```
alxmrs, line 950: omit?

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+            )
+            (pos,) = ax[1].plot(
+                self.pred_dict[k]["OHC_pred_mid"][Days_to_Eq:].time.data,
+                np.arange(self.pred_dict[k]["OHC_pred_mid"][Days_to_Eq:].size)
+                * self.pred_dict[k]["coeffs_OHC_pred_trend_mid"][0]
+                + self.pred_dict[k]["coeffs_OHC_pred_trend_mid"][1],
+                c=self.clist[i],
+                ls="--",
+            )
+            # ax[1].annotate(f'{self.pred_dict[k]["coeffs_OHC_pred_trend_mid"][0]:.2e}',
```
alxmrs, line 998: omit?

I think I'll stop comment about these...

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+        # Save or display the plot
+        plt.savefig(
+            os.path.join(self.salinity_path, "SeaSurfaceSalinity_Global_map.png"),
+            bbox_inches="tight",
+            dpi=600,
+        )
+        # plt.show()
+
+    def salinity_snapshot_maps(self):
+        # TODO(jder): this is a copy-paste of self.plot_sst/plot_diff_sst but with minor changes
```
alxmrs, line 2776: Is there a todo here? I'm happy for this note to hang around.

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+            if p_status != 0:
+                print("No ffmpeg found")
+                return None
+            else:
+                # parse version number
+                if copyright := re.search(
+                    "ffmpeg version (.+?) Copyright", str(output)
+                ):
+                    return copyright.group(1).replace(" ", "")
+                return None
```
alxmrs, line 3055: Glad there are videos here!

alxmrs: Any idea if we should use the standard python ffmpeg wrapper instead? https://github.com/kkroening/ffmpeg-python

alxmrs: (This also applies below to the subprocess code)

src/ocean_emulators/viz/core.py
```diff
@@ -0,0 +1,4005 @@
...
+                    fig, ax, pp = self.render_frame(fi, timesteps[fi])
+                    frame_save(
+                        fig,
+                        fi,
+                        odir=odir,
+                        frame_pattern=self.frame_pattern,
+                        dpi=self.dpi,
+                    )
+
+            # Needs more testing! Slower than the sequential counterpart
```
alxmrs, line 3316: Ah, the parallel python blues.

alxmrs (approved):
A heroic effort! You've spent enough time on this and the PR already accomplishes our goals for automated visualization during experiments. Consider my feedback, sheep and non-sheep alike as optional. 

Great work, I'm excited to try this out.

notebooks/regrid_basins.py
jder: I'm not sure what that means, sorry. There's a relative path that would work on all machines?

notebooks/regrid_basins.py
jder: I actually started with that package but it is a pain (only available on conda or as a docker image, not pypi) and given the trivial nature of this discrete class regridding I didn't think we needed it. WDYT?

src/ocean_emulators/utils/compare.py
jder: The goal is to just avoid halting the process midway through since it takes a while.

src/ocean_emulators/viz/config.py
jder: basins_location is not by default. Probably groundtruth will be either a LocalLocation or an UnresolvedLocation.

jder:
@alxmrs thanks for the review, I think I addressed most of them. I think I will skip the comments in core.py as you offered -- there are about a zillion cleanups we could do in there, but I think the first step is actually making it possible to run a single stage (or maybe even image) at once so it's (much) easier to test, then we can start refactoring without waiting for 30+ minutes to verify nothing changed unintentionally.

PR #340, opened 2025-08-08
A simple kill switch implementation.

alxmrs:
Adding a `target_epochs` argument to all LR schedulers that allows us to simulate long LR schedules (that actually stop short).

### DIFF
--- a/tests/test_schedule.py
+++ b/tests/test_schedule.py
@@ -1,9 +1,21 @@
 import torch
 
 from ocean_emulators.utils.schedule import (
+    CosineSchedulerConfig,
     CosineWithTailSchedulerConfig,
     CosineWithWarmupConfig,
 )
+
+
+def simulate_lr_history(optimizer, scheduler, total_epochs):
+    lr_history = []
+    # Need to call step after optimizer.step() to avoid warnings
+    # For testing, we'll simulate optimizer steps
+    for epoch in range(total_epochs):
+        lr_history.append(optimizer.param_groups[0]["lr"])
+        optimizer.step()  # Simulate optimizer step
+        scheduler.step()
+    return lr_history
 
 
 def test_cosine_with_tail_holds_constant_lr():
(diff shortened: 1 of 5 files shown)


jder:
I was thinking more about this -- is it really just about LR schedule vs run length? If so maybe we add an optional `target_epochs` to the cosine scheduler config instead?

alxmrs:
Sorry for the late reply. I'm not sure how this would fit into the lr scheduler. Can you explain more? 

My intention is to not interrupt with the normal LR schedule while also ending the training runs early in order to iterate through many configurations.

jder:
I'm saying that the epochs config is already a "kill after N epochs" switch. Rather than adding a second one I'm suggesting you instead add a config option to the lr scheduler which tells it how long the decay should be over. 

So rather than:

epochs = 70
kill_after_n_epochs = 5

You'd have:

epochs = 5
lr_scheduler = {type: cosine, decay_epochs: 70}

(Where decay_epochs = None means the current behavior)

alxmrs:
Thanks for the explanation -- I now understand your suggestion. I'm happy to implement that! Now that I've seen how to implement a new lr scheduler (#343), I see how I could implement this.

tests/test_schedule.py
```diff
@@ -49,3 +57,44 @@ def test_cosine_with_tail_holds_constant_lr():
...
+        assert last_lr >= lr, "Learning rate needs to be monotonically decreasing"
+        last_lr = lr
+
+    assert len(lr_history) == total_epochs
+
+    for short_lr, base_lr in zip(lr_history, base_lr_history):
+        # TODO(alxmrs): Is this an OK tolerance?
+        assert abs(short_lr - base_lr) < 1e-3, (
+            "Shortened schedule should be the same shape as standard LR schedule"
+        )
```
alxmrs: @jder It seems like this alternative method does not produce exactly the same schedule -- this tolerance looks pretty coarse to me. WDYT about going with this implementation vs the original, given this?

jder: It seemed logically impossible to me that this method would not produce the same schedule (since the schedule has no way of knowing how long we're gonna run it for), so I grabbed the code and did some digging. The problem is that the optimizer was being shared between the baseline and test schedule simulations, and so it maintained some state between those two. Pushed up a fix, probably also needs to be applied to the other test.

alxmrs: ah, that's surprising, thanks for digging into this! will TAL now.

alxmrs: Thanks, this is a great contribution. I took the liberty of pulling the repeated optimizer setup into the "simulate" function.

jder (approved):
Thanks! Sorry, thought I was still waiting for you on this.

alxmrs:
No problem; it wasn't blocking anything.

PR #341, opened 2025-08-11
New baseline model: Wide Samudra (training and eval config)

alxmrs:
Since I'm about to fork off experiments from these configurations, I thought it should be merged into main.


configs/eval_om4.yaml
jder, line 1: I think these changes should be in eval_om4_halfdeg

configs/data/om4_sky_local.yaml
```diff
@@ -12,6 +12,5 @@ data_stds_location:
   type: local
   path: OM4_stds.zarr
 
-static_data_vars:
-  - sea_surface_fraction
-  - hfgeou
+concurrent_compute: true
+num_workers: 4
```
jder: like I said in slack, I'm not sure why we need a separate sky file (this one also is invalid, you should have precommit errors if you merge main)

configs/train_om4_halfdeg.yaml
```diff
@@ -70,7 +71,7 @@ samudra:
     block_type: "conv_next_block"
     kernel_size: 3
     activation: "capped_gelu"
-    upscale_factor: 4
+    upscale_factor: 2
```
jder, line 74: What's the intuition behind changing this?

alxmrs: The intuition is that upscale factor makes each level of the U Net 4x or 2x times bigger (IIUC, by making it deeper). Since I make the channels significantly wider, I need to slow this factor increase so that the model will fit into GPU memory. Using an upscale factor of 2 with the configured widths makes the model as big as it can be while fitting into GPU memory (considering the overhead of the data and optimizer, etc.).

configs/data/om4_sky_local.yaml
alxmrs: I didn't realize that the `public.yaml` is basically the same thing! Thanks for pointing that out.

alxmrs: I missed that comment in chat, thanks for reiterating.

configs/eval_om4_halfdeg.yaml
jder, line 1: Oh, sorry, I thought I had already made a halfdeg file. Thanks, though, I think this is better in the end.

configs/train_om4_halfdeg.yaml
jder: ok, so this is about fitting in GPU memory, got it. So perhaps we'd turn this back up to 4 if we accumulate gradients across batches or use bigger GPUs.

jder (approved):
Thanks!

configs/eval_om4_halfdeg.yaml
alxmrs: I agree, thanks for suggesting this.

PR #342, opened 2025-08-11
Avoid chunking lat/lon

jder:
This was causing us to write 4x the number of chunks we need; I'm pretty sure from experiments and xarray docs that omitting the lat/lon vars here will make single large chunks.


alxmrs (approved):
LGTM!

PR #343, opened 2025-08-11
Implementing Warmup + Cosine Decay LR schedule from ConvNeXt paper.

alxmrs:
Here is a Warmup implementation. This was added to match the LR schedule of the original ConvNeXt paper. Following the citations, warmup is theorhetical grounded and empirically verified to help the loss curves of networks with _large batch sizes_ -- see this paper: https://arxiv.org/pdf/1706.02677

We have very small batch sizes, though each sample is fairly large (~80 MB). This technique is not likely to lead to a modeling improvement; however, I've queued up a few jobs to test if this improves anything. While this might not make it into the final baseline Samudra config, I could see it being valuable for the next iteration of the model.

### DIFF
--- a/src/ocean_emulators/utils/schedule.py
+++ b/src/ocean_emulators/utils/schedule.py
@@ -63,7 +63,7 @@ class CosineWithWarmupConfig(BaseModel):
         return torch.optim.lr_scheduler.SequentialLR(
             optimizer,
             schedulers=[warmup, cosine],
-            milestones=[epochs - self.warmup_epochs],
+            milestones=[self.warmup_epochs],
         )
 
 
--- a/src/ocean_emulators/utils/schedule.py
+++ b/src/ocean_emulators/utils/schedule.py
@@ -40,7 +40,7 @@ class CosineWithTailSchedulerConfig(BaseModel):
 
 
 class CosineWithWarmupConfig(BaseModel):
-    """Cosine scheduler which goes to head_lr for the first head_epochs."""
+    """Cosine scheduler which goes from warmup_lr to the default lr for the first warmup_epochs."""
 
     type: Literal["cosine_with_warmup"] = "cosine_with_warmup"
 
--- a/src/ocean_emulators/utils/schedule.py
+++ b/src/ocean_emulators/utils/schedule.py
@@ -50,6 +50,9 @@ class CosineWithWarmupConfig(BaseModel):
     def build(
         self, optimizer: torch.optim.Optimizer, epochs: int
     ) -> torch.optim.lr_scheduler.LRScheduler:
+        assert len(optimizer.param_groups) == 1, (
+            "There can only be one parameter group for the optimizer."
+        )
         warmup = torch.optim.lr_scheduler.LinearLR(
             optimizer,
             start_factor=self.warmup_lr / optimizer.param_groups[0]["lr"],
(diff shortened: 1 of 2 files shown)


src/ocean_emulators/utils/schedule.py
```diff
@@ -39,4 +39,34 @@ def build(
         )
 
 
-SchedulerConfig = CosineSchedulerConfig | CosineWithTailSchedulerConfig
+class CosineWithWarmupConfig(BaseModel):
+    """Cosine scheduler which goes to head_lr for the first head_epochs."""
```
jder:
```suggestion
    """Cosine scheduler which goes from warmup_lr to the default lr for the first warmup_epochs."""
```

src/ocean_emulators/utils/schedule.py
```diff
@@ -39,4 +39,34 @@ def build(
...
+
+    warmup_lr: float = 1e-6
+    warmup_epochs: int = 5
+
+    def build(
+        self, optimizer: torch.optim.Optimizer, epochs: int
+    ) -> torch.optim.lr_scheduler.LRScheduler:
+        warmup = torch.optim.lr_scheduler.LinearLR(
+            optimizer,
+            start_factor=self.warmup_lr / optimizer.param_groups[0]["lr"],
```
jder, line 58: Assert there's just one group?

src/ocean_emulators/utils/schedule.py
```diff
@@ -39,4 +39,34 @@ def build(
...
+            end_factor=1.0,  # Reaches full LR
+            total_iters=self.warmup_epochs,
+        )
+
+        cosine = torch.optim.lr_scheduler.CosineAnnealingLR(
+            optimizer, T_max=epochs - self.warmup_epochs
```
jder, line 67: Assert warmup_epochs <= epochs?

tests/test_schedule.py
```diff
@@ -49,3 +52,46 @@ def test_cosine_with_tail_holds_constant_lr():
...
+
+    assert max(lr_history) == target_lr, "The peak LR should be the target LR"
+    last_lr = -1
+    for i, lr in enumerate(lr_history):
+        if i <= warmup_epochs:
+            assert last_lr <= lr, "During warmup, LR should increase"
+
+        if i > warmup_epochs:
+            # Monotonically decreasing, with error tolerance for equality.
+            assert last_lr > lr or abs(last_lr - lr) < 1e7, (
```
jder: I don't understand why there's an error tolerance for equality here. Shouldn't they be monotonically decreasing? If `lr == last_lr + 1e8` that seems broken.

tests/test_schedule.py
```diff
@@ -49,3 +52,46 @@ def test_cosine_with_tail_holds_constant_lr():
...
+                "During cosine annealing, LR should decrease"
+            )
+
+        last_lr = lr
+
+    # Ensure that after warmup, the LR does actually decrease
+    assert (top := lr_history[warmup_epochs + 1]) > (last := lr_history[-1]), (
+        "LR must decrease after warmup"
+    )
+    assert top / 20 > last, "LR must decrease by a significant amount after warmup"
```
jder, line 94: This (and the previous cosine_with_tail test) both make me want to switch over to snapshot tests, maybe with Syrupy, rather than these fickle assertions. WDYT?

jder (approved):
Thanks, just a handful of 🐑 suggestions from me.

tests/test_schedule.py
alxmrs: Yeah, that's very reasonable. Well, luckily, when I delete this extra logic, the tests all still pass. Maybe something was broken (that should have been broken) and has since been fixed.

tests/test_schedule.py
alxmrs: syrupy is new to me! That does sound reasonable to me -- maybe in a follow up PR?

PR #344, opened 2025-08-13
First cut of viz bugs and allowing running a subset of images

jder:
* adds a logger, reuses existing logging setup logic
* creates a VizStep pydantic type which gives us autocomplete and validation of specifying a given step we want to run, rename all steps to make it easy to produce this list automatically
* allow specifying an allow/block list of steps to run on the CLI (both is nice because "everything but movies" is what I often want to run)

and then a couple actual bug fixes:
* change hard-coded "CM4" to the dataset name
* fix the blank OHC maps which were hard-coded to use the range +/- 0.05 ZJ when the actual values were ~1e-17 ZJ.


src/ocean_emulators/viz/config.py
```diff
@@ -36,6 +63,9 @@ class VizConfig(TopLevelConfig):
     groundtruth_time_range: TimeConfig = Field(
         description="Dates from the rollout (not same as eval *input* dates; these are the dates the output is produced for during eval)"
     )
+    steps: list[VizStep] | None = None  # None means all
+    not_steps: list[VizStep] = Field(default_factory=lambda: [])  # None means none
```
alxmrs: These comments are helpful. When both defaults are set to None, what takes precedence (is this a mutually exclusive setting?)?

src/ocean_emulators/viz/config.py
```diff
@@ -62,15 +92,35 @@ def build(self, default_root: ResolvedLocation) -> Viz:
...
 
-    # TODO(jder): would be nice to specify which plots to make,
-    # parallelize, re-run just needed plots on errors, etc.
-    viz.run()
+    steps = [s for s in cfg.steps or _all_steps() if s not in cfg.not_steps]
+    logger.info(f"Running steps: {', '.join(steps)}")
+
+    # TODO(jder): could use a ProcessPoolExecutor here, but steps currently
+    # are not exactly independent (some write to pred_dict which others read,
+    # there's some appending to a metrics file.)
```
alxmrs, line 118: Darn!

alxmrs (approved):
LGTM

PR #346, opened 2025-08-13
Omit the data config path from sky training.

alxmrs:
Now that we prefer to use `data/public.yaml`, there's no need to specify this in the sky training config. The status quo led to an error when deploying training runs (the sky config has since been deleted).


jder (approved):
:shipit:

PR #348, opened 2025-08-14
minor tweaks to skypilot setup

jder:
* Fixes numba bug with print being defined inside a function rather than a module.
* Fixes sqlalchemy required by skypilot
* Some docs and convenience updates

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ dependencies = [
   "s3fs>=2025.5.1",
   "scipy>=1.15.2",
   "skypilot[aws,lambda]>=0.10",
+  "sqlalchemy>=2",                # skypilot requires sqlalchemy 2.0 but doesn't include that requirement
   "torch~=2.2.1",
   "torchinfo>=1.8",
   "wandb>=0.19.8",
--- a/train.sky.yaml
+++ b/train.sky.yaml
@@ -6,6 +6,8 @@
 # ```
 # sky check -v
 # ```
+# You will generally need to create a dedicated IAM user for skypilot, and put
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
   "s3fs>=2025.5.1",
   "scipy>=1.15.2",
   "skypilot[aws,lambda]>=0.10",
+  "sqlalchemy>=2",                # skypilot requires sqlalchemy 2.0 but doesn't include that requirement
```
alxmrs, line 41: Oh thank you. So happy to have this fix

train.sky.yaml
```diff
@@ -10,7 +10,7 @@
 # To create or re-use a cluster:
 # ```
 # export WANDB_API_KEY=<my-key>  # Get your key at https://wandb.ai/authorize
-# sky launch -c fomo-cluster train.sky.yaml --env-file <my-env>.env --env WANDB_API_KEY
+# uv run sky launch train.sky.yaml --env-file <my-env>.env --env WANDB_API_KEY
 # ```
 # As the job progresses, skypilot will provide commands to locally monitor each stage.
```
alxmrs, line 17: Nit: I recommend adding an example of using `sky exec` so the next person doesn't repeat my mistake with queuing additional experiments.

jder: As discussed, let's figure out how to queue first.

jder:
(marking this for auto merge once you approve)

alxmrs (approved):
Thanks so much!

train.sky.yaml
alxmrs: Makes sense to me! I just posted in the skypilot slack to ask how to do this. It's unclear from the docs.

PR #352, opened 2025-08-22
Eval + Viz Skypilot

jder:


skypilot/train.sky.yaml
```diff
@@ -35,19 +35,19 @@ envs:
...
   # https://docs.skypilot.co/en/latest/cloud-setup/cloud-permissions/aws.html#dedicated-skypilot-iam-user
 
   CONFIG: null
   OUTPUTS_BUCKET: oa-fomo-outputs
   DATA_BUCKET: oa-fomo
-  # We assume that data in this directory is named in the same pattern as
-  # the $DATA_CONFIG yaml.
-  DATA_ROOT_PATH: null # data/om4_onedeg_compact  # NOTE: path cannot end in a slash `/`.
+  DATA_ROOT_PATH: null # path to copy from DATA_BUCKET into the local data root eg data/om4_onedeg
+  NAME: null # wandb train experiment name (and used for downstream tasks)
```
alxmrs, line 46: Nice, thanks -- this is a good affordance.

skypilot/train.sky.yaml
```diff
@@ -61,12 +61,15 @@ workdir: .
...
   rclone config create s3-source s3 provider=AWS env_auth=true
 
-  rclone copy --progress s3-source:${DATA_BUCKET}/${DATA_ROOT_PATH}/OM4_means.zarr ~/data/OM4_means.zarr --ignore-existing
-  rclone copy --progress s3-source:${DATA_BUCKET}/${DATA_ROOT_PATH}/OM4_stds.zarr ~/data/OM4_stds.zarr --ignore-existing
-  rclone copy --progress s3-source:${DATA_BUCKET}/${DATA_ROOT_PATH}/OM4.zarr ~/data/OM4.zarr --transfers=32 --ignore-existing
+  mkdir -p ~/data/
+  rclone copy --progress s3-source:${DATA_BUCKET}/${DATA_ROOT_PATH} ~/data --ignore-existing --transfers=32
+
+  mkdir -p ~/basins/
+  rclone copy --progress s3-source:${DATA_BUCKET}/basins ~/basins
```
alxmrs, line 69: What are basins used for? Are these needed for every phase?

skypilot/viz.sky.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+workdir: .
+
+setup: |
+  set -ex
+
+  rclone config create s3-source s3 provider=AWS env_auth=true
+
+  mkdir -p ~/data/
+  rclone copy --progress s3-source:${DATA_BUCKET}/${DATA_ROOT_PATH} ~/data --ignore-existing --transfers=32
```
alxmrs, line 66: Just want to confirm -- this copies the means, stds, and base OM4 data?

skypilot/viz.sky.yaml
```diff
@@ -0,0 +1,86 @@
...
+  set -ex
+
+  DATA_ROOT=$(realpath ~/data)
+  BASINS_ROOT=$(realpath ~/basins)
+
+  uv run -m ocean_emulators.viz "${CONFIG}" \
+    --data_root="${DATA_ROOT}" \
+    --base_output_dir=/outputs/ \
+    --name="${NAME}" \
+    --basins_location="{\"type\": \"local\", \"path\": \"${BASINS_ROOT}/${BASIN_PATH}\"}" \
```
alxmrs, line 83: Seems like basins are only used here. Is there a benefit to including it in every compute stage's set up? If not, maybe we should reconsider what's in the shared set up.

If the basin data is small / the data transfer is quick, then I can see this not mattering.

src/ocean_emulators/utils/writer.py
```diff
@@ -65,7 +67,7 @@ def write(self):
             coords=coords,
         )
         ds.attrs["model_path"] = str(self.model_path)
-        ds = ds.chunk({"time": 1})
+        ds = ds.chunk({"time": self.time_chunk_size})
```
alxmrs, line 70: Nice!

src/ocean_emulators/stepper.py
```diff
@@ -69,11 +69,16 @@ def inference(
                     "output_dir and model_path must be provided if save_zarr is True"
                 )
             coords = dataset.get_coords_dict()
+            if num_model_steps_forward > 0:
+                chunk_size = num_model_steps_forward
+            else:
+                chunk_size = 20
```
alxmrs, line 75: This indeed seems like a good default chunk size for the time dimension.

alxmrs (commented):
Early thoughts on the PR, I know it's still WIP.

skypilot/train.sky.yaml
jder: Yeah I keep forgetting to set this and it's Bad (tm) if you forget.

skypilot/train.sky.yaml
jder: They're only used for viz, but it's annoying to have this only in the viz setup because then you have to use `launch` to have that happen, which means you have to wait and do it after your other jobs are done (because otherwise it breaks the other data mounts)

skypilot/viz.sky.yaml
jder: Yes!

skypilot/viz.sky.yaml
jder: Yeah, it's very tiny, just a single 6-channel boolean mask of the earth.

src/ocean_emulators/utils/writer.py
jder: Thanks! Yeah this makes viz much faster.

skypilot/train.sky.yaml
alxmrs: Awesome, that's a great reason.

alxmrs (approved):
Can't wait to start using this!

PR #361, opened 2025-09-02
Extract Globe padding Conv2d into its own module.

alxmrs:
A minor refactor given our Samudrax hackathon project.

I welcome any suggestions about how to name this module :)


alxmrs:
This was a nice thought, but I don't think factoring it this was is essential. Closing for now.

PR #364, opened 2025-09-03
Added simple 2d patch embedding module for the FOMO encoder.

### LINKED ISSUE #362: Create gridded patch embedding
For the encoder, input data need to be encoded into spatial, latent patches before feeding into the Perceiver. 

See: https://github.com/microsoft/aurora/blob/main/aurora/model/patchembed.py

alxmrs:
Soon, we will add a Perceiver-based encoder module to the project. Before applying the Perceiver module, the first step we take on is to break up the input into ViT-style patches. This PR introduces a simple patch embedding mechanism in prep for the rest of the encoder.

Fixes #362.

### DIFF
--- a/src/ocean_emulators/models/modules/patchembed.py
+++ b/src/ocean_emulators/models/modules/patchembed.py
@@ -45,9 +45,7 @@ class PatchEmbed2d(nn.Module):
         self.embed_dim: int = embed_dim
         self.n_channels = len(input_vars) * (1 + hist)
 
-        self.n_patches = patch_dim = (
-            self.n_channels * self.patch_size[0] * self.patch_size[1]
-        )
+        patch_dim = self.n_channels * self.patch_size[0] * self.patch_size[1]
 
         # While we could perform a patch embedding and linear projection in one step with a convolution, this
         # implementation is much clearer. I don't expect the additional computational cost to be arduous.
--- a/src/ocean_emulators/models/modules/patchembed.py
+++ b/src/ocean_emulators/models/modules/patchembed.py
@@ -66,17 +66,12 @@ class PatchEmbed2d(nn.Module):
 
         # Ensure patch_size is appropriate for the data.
         assert H % self.patch_size[0] == 0, f"{H} %  {self.patch_size[0]} != 0."
-        assert (h := (H // self.patch_size[0]) ** 2) and h >= 16, (
-            f"{h=}: A picture is worth 16x16 words!"
-        )
         assert W % self.patch_size[1] == 0, f"{W} %  {self.patch_size[1]} != 0."
-        assert (w := (W // self.patch_size[1]) ** 2) and w >= 16, (
-            f"{w=}: A picture is worth 16x16 words!"
-        )
 
         x = self.patches(x)
         x = self.norm_patches(x)
         x = self.linear(x)
         x = self.norm_embedding(x)  # (batch, patch_dim, embed_dim)
         x = x.transpose(1, 2)  # (batch, embed_dim, patch_dim)
+
         return x
--- a/src/ocean_emulators/models/modules/patchembed.py
+++ b/src/ocean_emulators/models/modules/patchembed.py
@@ -12,7 +12,7 @@ from ocean_emulators.constants import Input
 class PatchEmbed2d(nn.Module):
     """A patch embedding for Samudra's flattened data (the channel dim is a cross of variable, level, and time).
 
-    Arguments:
+    Args:
         input_vars (list[str]): list of input variable names. For input, this is typically the target prognostic
           and boundary variable names.
         patch_size (int): the size of the patches to embed. Patches must evenly divide the input grid.
--- a/src/ocean_emulators/models/modules/patchembed.py
+++ b/src/ocean_emulators/models/modules/patchembed.py
@@ -1,12 +1,12 @@
 # Sources inspired by the following implementations:
 # - https://github.com/microsoft/aurora/blob/main/aurora/model/patchembed.py
 # - https://github.com/lucidrains/vit-pytorch
-
+import torch
 from einops.layers.torch import Rearrange
 from jaxtyping import Float
 from torch import nn
 
-from ocean_emulators.constants import Array, Input
+from ocean_emulators.constants import Input
 
 
 class PatchEmbed2d(nn.Module):
@@ -56,7 +56,9 @@ class PatchEmbed2d(nn.Module):
         self.linear = nn.Linear(patch_dim, embed_dim)
         self.norm_embedding = norm(embed_dim) if norm is not None else nn.Identity()
 
-    def forward(self, x: Input) -> Float[Array, "*batch {self.embed_dim} patch_dim"]:
+    def forward(
+        self, x: Input
+    ) -> Float[torch.Tensor, "*batch {self.embed_dim} patch_dim"]:
         B, V, H, W = x.shape
 
         # V is a cross product of variable, level (encoded in vars), and time (has history).
--- a/src/ocean_emulators/models/modules/patchembed.py
+++ b/src/ocean_emulators/models/modules/patchembed.py
@@ -37,6 +37,9 @@ class PatchEmbed2d(nn.Module):
         if isinstance(patch_size, int):
             self.patch_size: tuple[int, int] = (patch_size, patch_size)
         else:
+            assert isinstance(patch_size, tuple) and len(patch_size) == 2, (
+                "We only ever consider patches for the spatial dimensions (lat and lon)!"
+            )
             self.patch_size = patch_size
 
         self.embed_dim: int = embed_dim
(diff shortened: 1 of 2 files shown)


src/ocean_emulators/models/modules/patchembed.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(
+        self,
+        input_vars: list[str],
+        patch_size: int | tuple[int, int] = 4,
+        embed_dim: int = 1024,
+        hist: int = 1,
+        norm: type[nn.Module] | None = nn.LayerNorm,
+    ) -> None:
+        """Embed TrainData arrays into patches.
```
alxmrs: This short description feels not quite right -- any better ideas? I'm trying not to say "perform patch embedding" because that feels like defining a word with a different variant of that word.

src/ocean_emulators/models/modules/patchembed.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Args:
+            input_vars (list[str]): list of input variable names. For input, this is typically the target prognostic
+              and boundary variable names.
+            patch_size (int): the size of the patches to embed. Patches must evenly divide the input grid. Further, the
+              grid dimension divided by the patch size must be greater than 16 pixels.
```
jder: Why 16?

src/ocean_emulators/models/modules/patchembed.py
```diff
@@ -0,0 +1,82 @@
...
+        patch_dim = self.n_channels * self.patch_size[0] * self.patch_size[1]
+
+        # While we could perform a patch embedding and linear projection in one step with a convolution, this
+        # implementation is much clearer. I don't expect the additional computational cost to be arduous.
+        self.patches = Rearrange(
+            "b v (h ph) (w pw) -> b (h w) (ph pw v)",
+            ph=self.patch_size[0],
+            pw=self.patch_size[1],
+        )
+        self.norm_patches = norm(patch_dim) if norm is not None else nn.Identity()
```
jder: Do all nn.Module constructors take a dimension like this? If not maybe document the requirement above or else hard code LayerNorm here?

src/ocean_emulators/models/modules/patchembed.py
jder: I don't think you need something here if the class has a docstring.

src/ocean_emulators/models/modules/patchembed.py
```diff
@@ -0,0 +1,82 @@
...
+
+from einops.layers.torch import Rearrange
+from jaxtyping import Array, Float
+from torch import nn
+
+from ocean_emulators.constants import Input
+
+
+class PatchEmbed2d(nn.Module):
+    """A patch embedding for Samudra's flattened data (the channel dim is a cross of variable, level, and time)."""
```
jder: I was surprised after reading this that it's just a linear embedding. Perhaps you could include the word "projection" or "linear" in the name and/or description here? 

Also, see comment below but maybe we could have this class not know anything about variables, levels, or time and be responsible only for turning a set of channels * x * y into an embedding dim * num_patches? (I see aurora's implementation knows about some of this but that's because it wants to handle batches which have different variables which we're not doing here.)

src/ocean_emulators/models/modules/patchembed.py
```diff
@@ -0,0 +1,82 @@
...
+        if isinstance(patch_size, int):
+            self.patch_size: tuple[int, int] = (patch_size, patch_size)
+        else:
+            assert isinstance(patch_size, tuple) and len(patch_size) == 2, (
+                "Patch sizes must only span spatial dimensions (lat and lon)!"
+            )
+            self.patch_size = patch_size
+
+        self.embed_dim: int = embed_dim
+        self.n_channels = len(input_vars) * (1 + hist)
```
jder: Seems like it might be preferable to pass n_channels in directly rather than doing this computation in here? Seems weird to have this class know about the input var names and the concept of history just to compute this number (and makes it less flexible if we, say, decide to include some metadata which is not replicated over history). Or maybe I'm not understanding how you're expecting this to be used such that this is the desired interface?

If you do want to keep this, I would remove the default for hist in the constructor -- seems like it's pretty clearly a bug if anyone ever uses that default value.

src/ocean_emulators/models/modules/patchembed.py
```diff
@@ -0,0 +1,82 @@
...
+        self.patches = Rearrange(
+            "b v (h ph) (w pw) -> b (h w) (ph pw v)",
+            ph=self.patch_size[0],
+            pw=self.patch_size[1],
+        )
+        self.norm_patches = norm(patch_dim) if norm is not None else nn.Identity()
+        self.linear = nn.Linear(patch_dim, embed_dim)
+        self.norm_embedding = norm(embed_dim) if norm is not None else nn.Identity()
+
+    def forward(self, x: Input) -> Float[Array, "*batch {self.embed_dim} patch_dim"]:
```
jder: I was surprised at the output shape here. I was expecting maybe `batch patch_dim embed_dim` or `batch nx_patches ny_patches embed_dim` if we're going to want to add positional embeddings afterwards. Maybe it will become clear after I see how this is used?

src/ocean_emulators/models/modules/patchembed.py
```diff
@@ -0,0 +1,82 @@
...
+        B, V, H, W = x.shape
+
+        # V is a cross product of variable, level (encoded in vars), and time (has history).
+        assert V == self.n_channels
+
+        # Ensure patch_size is appropriate for the data.
+        assert H % self.patch_size[0] == 0, f"{H} %  {self.patch_size[0]} != 0."
+        assert (h := (H // self.patch_size[0]) ** 2) and h >= 16, (
```
jder: It looks like this tests for H / patch_size[0] > 4? Why 4? If we want to check the total area, shouldn't we be testing h * w against 16, not h^2 and separately w^2?

tests/test_patchembed.py
```diff
@@ -0,0 +1,121 @@
...
+
+from ocean_emulators.config import TrainConfig
+from ocean_emulators.constants import BOUNDARY_VARS, PROGNOSTIC_VARS
+from ocean_emulators.datasets import TrainData
+from ocean_emulators.models.modules.patchembed import PatchEmbed2d
+
+from .test_datasets import make_loader
+
+
+def test_makes_patches():
```
jder, line 11: If you take my suggestion to get rid of history and input_vars, I think we probably only need 1 or 2 tests here. (Which might also help with speed -- alternatively you could use smaller grid sizes so our poor CI machines aren't actually doing convolutions over high res data.)

tests/test_patchembed.py
```diff
@@ -0,0 +1,121 @@
...
+            # The forward pass of our base model manages a rollout across steps, where the model
+            # only ever needs to predict one step. There are two possible inputs in this step rollout
+            # system: the initial state, or a "merged" state, where we merge our latest prognostic
+            # prediction with the boundary forcings for that step.
+            # Thus, to ensure our patch encoder works as expected (for the one-step prediction part),
+            # we only need to test that the patch embedding works for either the initial input or the
+            # merged input.
+            initial_input = td.get_initial_input()
+            patches = patch_embed(initial_input)
+            assert patches.shape == (1, 1024, 4050)
```
jder: I like the idea of doing a test on real data but hard-coding these numbers seems a bit brittle, can we rely on the previous tests to cover shape correctness?

jder (approved):
Thanks, just a bunch of 🐑 from me, mostly related to not understanding how it's intended to be used.

src/ocean_emulators/models/modules/patchembed.py
alxmrs: Mainly due to this paper: https://arxiv.org/abs/2010.11929

src/ocean_emulators/models/modules/patchembed.py
alxmrs: I think I'm going to get rid of this constraint altogether. With our default patch size and relatively large grid, this is not a condition we're likely to hit anyway.

src/ocean_emulators/models/modules/patchembed.py
alxmrs: It was originally `batch patch embed`, and then I transposed it to match what Aurora was doing. When I move to using this (currently writing this in parallel), I may rm the transpose step.

I'm indifferent to either scheme -- it's a matter of making sure the shapes on consistent at the end of the day.

src/ocean_emulators/models/modules/patchembed.py
alxmrs: I like this idea a lot! I've just changed it, and now the patch embedding interface is really succinct. Thanks for the suggestion.

tests/test_patchembed.py
alxmrs: I've just asserted on shapes we absolutely control. WDYT of that alternative? If not, do you have any suggestions for what are appropriate asserts?

PR #366, opened 2025-09-04
Viz fixes

jder:
A few things that let me run viz again on a larger set of runs:

* Use AWS; it's faster/zero-cost to sync data and has large cheap non-GPU machines.
* Copy data down to the box first (MOUNT_COPY has a [10G fixed local cache](https://github.com/skypilot-org/skypilot/blob/eff22067df3d112890264d3f66167ec66771f778/sky/data/mounting_utils.py#L399) which is not nearly large enough for our predictions zarrs)
* Fix ordering of steps (this is both because I like it to be in a consistent order and also because there are currently inter-step dependencies)

### DIFF
--- a/skypilot/viz.sky.yaml
+++ b/skypilot/viz.sky.yaml
@@ -135,5 +135,5 @@ run: |
     --base_output_dir=/outputs/ \
     --name="${NAME}" \
     --basins_location="{\"type\": \"local\", \"path\": \"${BASINS_ROOT}/${BASIN_PATH}\"}" \
-    --runs="${RUNS}"
+    --runs="${RUNS}" \
     ${ARGS}
--- a/skypilot/viz.sky.yaml
+++ b/skypilot/viz.sky.yaml
@@ -20,11 +20,8 @@
 # export SKYPILOT_DEBUG=true
 # ```
 resources:
-  # Optional; if left out, automatically pick the cheapest cloud.
-  cloud: lambda
-  # 1x NVIDIA A100 GPU -- Viz needs no GPUs but lambda only has GPU machines, we do this so you can share a cluster with eval.
-  accelerators: A100:1
-  disk_tier: best
+  infra: aws/us-east-2
+  disk_size: 1TB
 
 # NOTE: Remember to pass environment variables with `--env` or `--env-file`
 # NOTE: All `null` values must be specified in order for this script to run!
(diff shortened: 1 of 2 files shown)


skypilot/viz.sky.yaml
```diff
@@ -10,8 +10,11 @@
 # To create or re-use a cluster:
 # ```
 # export WANDB_API_KEY=<my-key>  # Get your key at https://wandb.ai/authorize
-# sky launch -c fomo-cluster viz.sky.yaml --env-file <my-env>.env --env WANDB_API_KEY
+# sky launch -c fomo-cluster viz.sky.yaml --env-file <my-env>.env --env WANDB_API_KEY --env RUNS=[{"location": "/inputs/my_eval_job/predictions.zarr"}, …]
 # ```
+# Any location starting with /inputs will automatically be copied from the OUTPUTS_BUCKET
+# in order to access previous jobs' outputs.
```
alxmrs, line 16: Nice! It's cool that it can cache data like this.

skypilot/viz.sky.yaml
```diff
@@ -65,6 +67,58 @@ setup: |
...
+  for run in runs:
+      location = run['location']
+
+      # Only process locations under /inputs/
+      if not location.startswith('/inputs/'):
+          print(f'Skipping {location} - not under /inputs/')
+          continue
+
+      # Extract the path after /inputs/
+      path_after_inputs = location[8:]  # Remove '/inputs/' prefix
```
alxmrs: nit: let's use `removeprefix()`

https://docs.python.org/3/library/stdtypes.html#str.removeprefix

skypilot/viz.sky.yaml
```diff
@@ -65,6 +67,58 @@ setup: |
...
+      s3_source = f's3-source:{os.environ["OUTPUTS_BUCKET"]}/{top_dir}'
+
+      # Construct the local destination path
+      local_dest = f'/inputs/{top_dir}'
+
+      # Execute rclone copy command
+      cmd = ['rclone', 'copy', '--progress', s3_source, local_dest, '--ignore-existing', '--transfers=32']
+      print(f'Copying {s3_source} to {local_dest}')
+      subprocess.run(cmd, check=True)
+  EOPYTHON
```
alxmrs, line 120: TIL about using EOPYTHON in a bash script -- thanks, that's nifty.

skypilot/viz.sky.yaml
```diff
@@ -81,5 +135,5 @@ run: |
     --base_output_dir=/outputs/ \
     --name="${NAME}" \
     --basins_location="{\"type\": \"local\", \"path\": \"${BASINS_ROOT}/${BASIN_PATH}\"}" \
-    --runs="${RUNS}"
+    --runs="${RUNS}" \
```
alxmrs, line 138: Nice catch.

alxmrs (approved):
LGTM. Just one python nit.

PR #368, opened 2025-09-05
First pass at integrating an encoder into Samudra

alxmrs:
This PR introduces the ability to turn on and configure an encoder at the beginning of the Samudra model. It updates the linear patch embedding system to use Perceivers.


alxmrs:
This PR has been implemented via #378 and #379. Closing.

PR #378, opened 2025-09-08
The linear patch embedder is now our encoder.

### LINKED ISSUE #360: Add Perceiver Encoder (spike)
Time box exploration to figure out what it takes to add Perceivers to Samudra. 

Timebox: 4 days

Add a Perceiver encoder layer to Samudra, see if I can run a training run (on 1° data).
- Define a positional encoding scheme
- Figure out how to divide input into latent patches
- Default to propagating depth information to the latent space (i.e. keep the same depth levels as the input)

alxmrs:
This creates a tested, isolated encoder module out of our linear patch embedder. Here, we make use of Perceiver transformers to encode the input space into a latent representation. This embedding is created along the input channels (a whole column of the ocean) -- it still keeps spatial representation together along spatial patches.

Fixes #360.


src/ocean_emulators/models/modules/encoder.py
```diff
@@ -0,0 +1,90 @@
...
+          can be found here: https://github.com/lucidrains/perceiver-pytorch#usage.
+    """
+
+    def __init__(
+        self,
+        n_channels: int,
+        patch_size: int | tuple[int, int],
+        embed_dim: int,
+        perceiver_depth: int,
+        **perceiver_kwargs: dict[str, Any],
```
```suggestion
        **perceiver_kwargs: Any,
```
The annotation here is on the values, not on the type of the resulting dict. Alternatively, do we need to expose this? Some of these shouldn't be passed (eg the ones we're passing ourselves).

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -0,0 +1,90 @@
...
+            max_freq=1.0,
+            depth=perceiver_depth,
+            input_axis=2,  # Number of positional dims before token dim
+            input_channels=self.n_channels,  # input_dim
+            num_classes=embed_dim,  # output_dim
+            **perceiver_kwargs,
+        )
+        self.norm_embedding = nn.LayerNorm(embed_dim)
+
+    def forward(self, x: Input) -> Float[torch.Tensor, "*batch h w {self.embed_dim}"]:
```
```suggestion
    def forward(self, x: Input) -> Float[torch.Tensor, "batch h w {self.embed_dim}"]:
```

The code below looks like it works for exactly 1 batch dimension, but `*batch` means "zero or more"

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -0,0 +1,90 @@
...
+
+from ocean_emulators.constants import Input
+
+
+class PerceiverEncoder(nn.Module):
+    """A perceiver-based encoder for Samudra's flattened data (a whole column of the ocean, with history).
+
+    Args:
+        n_channels (int): the number of input channels (roughly:  time x variable x (surface + depths)).
+        patch_size (int): the size of the patches to embed. Patches must evenly divide the input grid.
```
jder: Maybe either say this is in order height, width or else just force square sizes for now?

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -0,0 +1,90 @@
...
+        self.norm_embedding = nn.LayerNorm(embed_dim)
+
+    def forward(self, x: Input) -> Float[torch.Tensor, "*batch h w {self.embed_dim}"]:
+        _, V, H, W = x.shape
+
+        # V is a cross product of variable, level (encoded in vars), and time (has history).
+        assert V == self.n_channels
+        # Ensure patch_size is appropriate for the data.
+        assert H % self.patch_size[0] == 0, f"{H} %  {self.patch_size[0]} != 0."
+        assert W % self.patch_size[1] == 0, f"{W} %  {self.patch_size[1]} != 0."
```
```suggestion
        assert H % self.patch_size[0] == 0, f"{H} % {self.patch_size[0]} != 0."
        assert W % self.patch_size[1] == 0, f"{W} % {self.patch_size[1]} != 0."
```

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Perceiver experiment ideas:
+        # 1. leave it as it is: treating each pixel as a token -- i.e. all channels (includes depths) per pixel
+        # 2. change to original plan, where each float is its own token
+        # 3. Add a third dim -- ph pw d v -- so each spatial position is a token
+        x = rearrange(
+            x,
+            "b v (h ph) (w pw) -> (b h w) ph pw v",
+            pw=self.patch_size[0],
+            ph=self.patch_size[1],
```
jder: Here we have the patch *width* being at index 0 but above (and below) we're treating index 0 as the height. I vote for making them square.

tests/test_encoder.py
```diff
@@ -1,62 +1,70 @@
...
+        perceiver_depth=2,
     )
 
     patches = patch_embed(x)
 
-    assert patches.shape == (1, 1024, 4050)
+    assert patches.shape == (1, 1, 2, 4)
 
 
+@pytest.mark.skip(reason="Computationally expensive!")
```
jder: Maybe just delete? Or mark as manual? (Not sure when to skip vs use manual)

jder (changes requested):
Thanks! Generally looks good, just a few changes.

src/ocean_emulators/models/modules/encoder.py
alxmrs: Yeah, that's a fair point -- in some sense these are implementation details. I thought we might want to have this be config-driven at first, but it probably is best to manage this via code.

src/ocean_emulators/models/modules/encoder.py
alxmrs: Thanks, great catch! I added a test to catch this.

tests/test_encoder.py
alxmrs: I'll just delete this.

jder (approved):
Thanks!

PR #379, opened 2025-09-09
Updated our configuration system to accommodate multiple models, including FOMO v0. 

alxmrs:
This change makes space for us to develop and configure non-Samudra models. I've created a basic implementation of FOMOv0, but this is intentionally a placeholder. The focus of the PR is on the config system changes. To that end, all models are constructed with our builder pattern. 

For the benefit of working with both FOMOv0 and Samudra, this PR factors out the UNet implementation into its own module. We've also updated all our config files to follow this new interface.


src/ocean_emulators/config.py
```diff
@@ -197,32 +205,76 @@ class BlockConfig(BaseConfig):
     upscale_factor: int = 4
     norm: NormType = "batch"
 
+    def build(self) -> Callable[..., CoreBlock]:
+        def create_block(*args, **kwargs) -> CoreBlock:
```
alxmrs: Taking a leaf out of @gonzalobenegas's book :)

gonzalobenegas: Hey Alex I have absolutely no idea what you meant :)

alxmrs: This sort of pattern stuck out to me in your bloom filter implementation.

src/ocean_emulators/config.py
```diff
@@ -347,7 +492,7 @@ class TrainConfig(TopLevelConfig):
     # Config components
     experiment: ExperimentConfig
     data: DataConfig
-    samudra: SamudraConfig
+    model: SamudraConfig | FOMOConfig
```
alxmrs: using `BaseModelConfig` causes schema validations errors.

src/ocean_emulators/config.py
```diff
@@ -197,37 +210,107 @@ class BlockConfig(BaseConfig):
     upscale_factor: int = 4
     norm: NormType = "batch"
 
+    def build(self) -> Callable[..., CoreBlock]:
```
jder: 🐑 what do you think about adding some type safety here? Something like

```
class CoreBlockBuilder(Protocol):
    def __call__(self, in_channels: int, out_channels: int): ... # … etc
```

And then you can `def build(self) -> CoreBlockBuilder:`

src/ocean_emulators/config.py
```diff
@@ -197,37 +210,107 @@ class BlockConfig(BaseConfig):
...
+        def create_block(*args, **kwargs) -> CoreBlock:
+            activation = ACTIVATION_REGISTRY[self.activation]
+            Block = BLOCK_REGISTRY[self.block_type]
```
jder: 🐑 🐑 could also match on self.block_type and get rid of the registries?

src/ocean_emulators/config.py
```diff
@@ -197,37 +210,107 @@ class BlockConfig(BaseConfig):
...
+            non_negative_corrector_names=self.non_negative_corrector_names,
+            ocean_heat_corrector=self.ocean_heat_corrector,
+            hist=hist,
+            area_weights=area_weights,
+            static_data=static_data,
+        )
+
+
+class EncoderConfig(BaseConfig):
+    patch_size: int | tuple[int, int] = Field(
```
jder: Can `tuple[int, int]` be expressed on the command line? ie does `--encoder.patch_size="[4, 6]"` or something work? If not I suggest we make this into a list or dataclass with width/height members.

src/ocean_emulators/config.py
```diff
@@ -197,37 +210,107 @@ class BlockConfig(BaseConfig):
...
+    def build(
+        self,
+        pad: str,
+        checkpointing: Checkpointing | None,
+        override_in_channels: int = 0,
+    ) -> UNetBackbone:
+        ch_width = self.ch_width.copy()
+        # This is needed for Samudra magic.
+        if override_in_channels != 0:
+            ch_width[0] = override_in_channels
```
jder: I don't understand when we *wouldn't* enter into this block. ie should this be `in_channels: int` (no default) and always get passed?

One 🐑 suggestion: we could get rid of this "overriding" stuff by having the config be `out_width` instead of `ch_width` in the config. (This makes the items in that list correspond to the same blocks as dilation/n_layers.) Then we can push the in_channels on to the front here before passing along to UNetBackbone. (Happy to talk live if that made no sense)

src/ocean_emulators/config.py
jder: (Could also do this for create_upsampling_block and friends)

src/ocean_emulators/config.py
```diff
@@ -239,6 +322,68 @@ class SamudraConfig(BaseConfig):
...
+                pad=self.pad,
+                checkpointing=self.checkpointing,
+                override_in_channels=in_channels,
+            ),
+            corrector=corrector,
+            pos_channels=self.pos_channels,
+            hist=hist,
+            wet=wet,
+            static_data=static_data,
+        )
```
jder, line 443: This is beautiful 😍

src/ocean_emulators/config.py
jder: Yeah, you might consider making a `AnyModelConfig` like I did for `AnyTopLevelConfig`.

src/ocean_emulators/eval.py
```diff
@@ -120,20 +119,20 @@ def __init__(self, cfg: EvalConfig) -> None:
         # Model
         logger.info(f"Getting model {cfg.experiment.network}")
         if "Samudra" == cfg.experiment.network:
-            if cfg.samudra.ch_width[0] != self.num_in:
+            assert isinstance(cfg.model, SamudraConfig)
```
jder: OK, bear (🐑 ) with me but can we:
* Get rid of cfg.experiment.network
* Add an abstract `def build(self, in_channels: int, out_channels: int, hist: int, wet: Grid, area_weights: Grid, static_data: xr.Dataset | None)` to BaseModel. 
* Remove in_channels and out_channels from BaseModelConfig.
* Replace all this code with `model = cfg.model.build(self.num_in, self.num_out, hist…).to(self.device)`
* Adjust the builds for SamudraConfig and EncoderConfig to match the new signature (I think this is ~a no-op)
* (Same in train.py)

src/ocean_emulators/models/corrector.py
```diff
@@ -246,7 +246,8 @@ class Correctors(torch.nn.Module):
 
     def __init__(
         self,
-        config,
+        non_negative_corrector_names: list[str] | None,
+        ocean_heat_corrector: bool,
```
jder, line 250: <3

src/ocean_emulators/models/fomo.py
```diff
@@ -0,0 +1,49 @@
+import torch
+import xarray as xr
+from torch import nn
+
+from ocean_emulators.constants import Grid
+from ocean_emulators.models.base import BaseModel
+from ocean_emulators.models.modules.unet_backbone import UNetBackbone
+
+
+class FOMOv0(BaseModel):
```
jder: I suggest we get rid of the version number, who knows how/if we're going to want to version this.

src/ocean_emulators/models/fomo.py
```diff
@@ -0,0 +1,49 @@
...
+            wet=wet,
+            hist=hist,
+            pred_residuals=pred_residuals,
+            last_kernel_size=last_kernel_size,
+            pad=pad,
+            static_data=static_data,
+        )
+        # TODO(alxmrs): Properly wire up the encoder with the processor.
+        self.layers = [encoder, processor]
+        assert isinstance(processor, UNetBackbone)
```
jder: Can we add it as a type constraint?

src/ocean_emulators/config.py
```diff
@@ -197,37 +210,107 @@ class BlockConfig(BaseConfig):
...
+        override_in_channels: int = 0,
+    ) -> UNetBackbone:
+        ch_width = self.ch_width.copy()
+        # This is needed for Samudra magic.
+        if override_in_channels != 0:
+            ch_width[0] = override_in_channels
+
+        DownsamplingBlock = DOWNSAMPLE_REGISTRY[self.down_sampling_block]
+
+        def create_upsampling_block(in_channels, out_channels) -> nn.Module:
```
```suggestion
        def create_upsampling_block(in_channels: int, out_channels: int) -> nn.Module:
```

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -15,22 +15,23 @@ class PerceiverEncoder(nn.Module):
...
     Args:
-        n_channels (int): the number of input channels (roughly:  time x variable x (surface + depths)).
+        in_channels (int): the number of input channels (roughly:  time x variable x (surface + depths)).
         patch_size (int | tuple[int, int]): the size of the patches to embed. Patches must evenly divide the input grid.
           If a tuple is supplied, then it represents the (height, width) of the patches to embed.
         embed_dim (int): size of the latent dimension.
         perceiver_depth (int): depth of the perceiver module core.
     """
 
+    # TODO(alxmrs): Implement checkpointing
```
```suggestion
    # TODO(alxmrs): Implement gradient checkpointing
```
I assume that's what you mean?

src/ocean_emulators/models/modules/unet_backbone.py
```diff
@@ -0,0 +1,185 @@
...
+                        pads[1] // 2,
+                        pads[1] - pads[1] // 2,
+                        pads[0] // 2,
+                        pads[0] - pads[0] // 2,
+                    ]
+                    fts = nn.functional.pad(fts, pads)
+                    fts += skip_inputs[int(2 * self.num_steps - count - 1)]
+                    count += 1
+
+        return fts
```
jder, line 192: I'm assuming none of this changed other than extracting the last conv?

configs/train_default.test.yaml
```diff
@@ -52,21 +52,24 @@ data:
...
   pad: "circular"
 
-  core_block:
-    block_type: "conv_next_block"
-    kernel_size: 3
-    activation: "capped_gelu"
-    upscale_factor: 2
-    norm: "batch"
+  unet:
+    ch_width: [2, 2, 2]
```
jder: This is different from before. Was this a drive-by bug fix? I'm confused how it worked at all before.

jder (approved):
ok, this is amazing, thank you so much for doing all this cleanup. Feels way better to me now. I left you many 🐑 but would be perfectly happy if they were follow-ups or we made tickets about them for next clean-up week. If you do want to do some of the bigger ones, would love to take another look before merging. 

One thing I will ask: there's a lot of moving parts here, can you run an epoch or whatever of a real training run with the baseline model and make sure the outputs/losses look sane?

src/ocean_emulators/models/fomo.py
alxmrs: I was on the fence about this. Happy to get rid of it.

src/ocean_emulators/models/modules/encoder.py
alxmrs: Yess.

configs/train_default.test.yaml
alxmrs: You know what? I didn't even notice that I made this fix! I copied from one test config to the other, assuming it was consistent. 

Lol, I don't know how it worked before either.

src/ocean_emulators/config.py
alxmrs: I actually prefer a dictionary lookup to a match statement in this case.

src/ocean_emulators/config.py
alxmrs: I like this suggestion a lot. Though, I like the name `ch_width`; but I agree the `in_channels` and `ch_width` settings are redundant.

src/ocean_emulators/eval.py
alxmrs: 🫡

src/ocean_emulators/config.py
jder: Why? I prefer a match statement because (a) the compiler will check we cover all the cases (since block_type is a literal) and (b) it will store type safety we currently don't have about how Block(…) is called below.

src/ocean_emulators/models/modules/unet_backbone.py
alxmrs: No, the bones of the network are the same. Some notable changes are we're using the factor methods created in `build()`. And yes, the last Conv2d I'm pulling aside as a "Samudra decoder" layer. 

I did just change this in order to fix some type issues (see the bottom of the init).

src/ocean_emulators/config.py
alxmrs: It's all the config system that you envisioned!

src/ocean_emulators/config.py
alxmrs: You're right. I like dictionaries because they functioned like match statements before that feature was added to the language. We don't need a separate factory file and build system, which is a flavor of factories anyway.

alxmrs: Writing a few of these, it also feels like using match cases is way more code than simply using a dictionary. Added addition effort for error handling.

alxmrs: On the other hand (!) using match cases has caught a lot of typing issues! That settles the argument in my mind.

jder: Huh, I'm surprised it's more code. I was thinking it'd be about the same?

src/ocean_emulators/config.py
```diff
@@ -197,37 +216,171 @@ class BlockConfig(BaseConfig):
...
     norm: NormType = "batch"
 
+    def build(self) -> CoreBlockBuilder:
+        match self.activation:
+            case "relu":
+                activation: type[nn.Module] = ReLU
+            case "capped_gelu":
+                activation = CappedGELU
+            case _:
+                raise ValueError(f"Activation {self.activation!r} not supported.")
```
```suggestion
               assert_never(self.activation)
```
This will cause a type error if you forget a case. (Will also need to import from typing)

jder: (same for the below matches)

src/ocean_emulators/config.py
```diff
@@ -197,37 +216,171 @@ class BlockConfig(BaseConfig):
...
+                        out_channels=out_channels,
+                        dilation=dilation,
+                        n_layers=n_layers,
+                        pad=pad,
+                        checkpoint_simple=checkpoint_simple,
+                        kernel_size=self.kernel_size,
+                        upscale_factor=self.upscale_factor,
+                        norm=self.norm,
+                        activation=activation,
+                    )
```
jder, line 263: Oh, I see why it's more code. Yeah, I was expecting to assign `Block = ConvBlock` in the first case and then call as before at the end of the function. Either's fine with me.

src/ocean_emulators/config.py
```diff
@@ -197,37 +216,171 @@ class BlockConfig(BaseConfig):
...
 
+    def build(
+        self,
+        in_channels: int,
+        pad: str,
+        checkpointing: Checkpointing | None,
+    ) -> UNetBackbone:
+        assert len(self.ch_width) == len(self.dilation) == len(self.n_layers), (
+            "`ch_width`, `dilation`, and `n_layers` must have the same length."
+        )
```
jder, line 341: 💯

src/ocean_emulators/config.py
```diff
@@ -238,6 +391,91 @@ class SamudraConfig(BaseConfig):
...
+        out_channels: int,
+        hist: int,
+        wet: Grid,
+        area_weights: Grid,
+        static_data: xr.Dataset | None,
+    ) -> BaseModel:
+        pass
+
+
+class SamudraConfig(BaseModelConfig):
```
jder, line 406: I think probably you'll want to do what I did with the Location subclasses and add a `type: Literal['samudra'] = 'samudra'` so that the config will show `type: samudra` -- otherwise we're inferring in the config that you have a Samudra because there's a unet member, which is slightly weird.

src/ocean_emulators/eval.py
```diff
@@ -118,40 +117,24 @@ def __init__(self, cfg: EvalConfig) -> None:
...
 
-        self.model = model
         if cfg.ckpt_path is None:
             raise ValueError(
                 "ckpt_path must be set; try --ckpt_path=path/to/checkpoint"
             )
         self.load_checkpoint(cfg.ckpt_path)
 
-        self.network = cfg.experiment.network
+        self.network = self.model.__class__.__name__
```
jder, line 137: Is this used?

src/ocean_emulators/config.py
```diff
@@ -197,37 +216,171 @@ class BlockConfig(BaseConfig):
...
+                case "transposed_conv":
+                    return TransposedConvUpsample(
+                        in_channels=in_channels, out_channels=out_channels
+                    )
+                case _:
+                    raise ValueError(
+                        f"Unsupported `up_sampling_block`: {self.up_sampling_block!r}."
+                    )
+
+        ch_width = [in_channels] + self.ch_width.copy()
```
jder, line 356: Just my 2 cents but I suggested a different name for the `ch_width` in config because I thought it was confusing to have the same name in two places with closely-related-but-different semantics.

src/ocean_emulators/models/base.py
```diff
@@ -58,7 +55,7 @@ def forward(
                 pred = (
                     input_tensor[
                         :,
-                        : self.num_prognostic_channels,
+                        : self.out_channels,
```
jder, line 58: This is one place where I think the prognostic name was better. Seems like maybe 50-50

jder (approved):
Ship it! Thanks for all the changes, this is really great. All my comments are 🐑  and half of them are just arguing with myself :D

src/ocean_emulators/eval.py
alxmrs: It's used in some notebooks. I was on the fence whether to delete this or not.

jder: Ah, ok. yeah, either way is fine with me.

src/ocean_emulators/config.py
alxmrs: The assert never stmt caught an error that I missed!! :)

PR #380, opened 2025-09-10
Upgrade torch version to 2.8

jder:
Seems like a reasonable first step before getting into performance tuning.


.github/workflows/aws-runner-start.yml
```diff
@@ -18,6 +18,11 @@ on:
         required: false
         type: string
         default: "/home/ubuntu"
+      aws_root_device_size:
+        description: "Root device size in GB"
+        required: false
+        type: number
+        default: 200
```
jder, line 25: Added this because the default AMI above was running out of disk space.

src/ocean_emulators/models/modules/activations.py
```diff
@@ -12,7 +12,7 @@ def __init__(self, **kwargs):
         :param kwargs: passed to torch.nn.ReLU
         """
         super().__init__()
-        self.add_module("relu", torch.nn.ReLU())
+        self.relu = torch.nn.ReLU(**kwargs)
```
jder, line 15: This is to help the type checker (it thought that self.relu could be a tensor or a module but in fact it can only be a module) but also a drive-by bugfix to pass kwargs along.

.github/workflows/benchmarks.yml
```diff
@@ -25,9 +25,6 @@ jobs:
     needs: start-aws-runner
     runs-on: ${{ fromJSON(needs.start-aws-runner.outputs.instances) }}
 
-    env:
-      # see note about machine image above
-      CUDA_VERSION: 12.1
```
jder, line 30: This is/was unused

tests/test_trainer.py
```diff
@@ -73,13 +74,14 @@ def test_checkpoint_inference(trainer_pair: TrainPair, caplog):
...
     model = trainer.model
+    assert isinstance(model, BaseModel)
     out = model.forward_once(X[0][0].to(trainer.device))
 
     with tempfile.TemporaryDirectory() as tmpdir:
         trainer.save_checkpoint(1, Path(tmpdir) / "test.pt")
         trainer.load_checkpoint(Path(tmpdir) / "test.pt")
 
-    out2 = trainer.model.forward_once(X[0][0].to(trainer.device))
+    out2 = model.forward_once(X[0][0].to(trainer.device))
```
jder, line 84: Again just to help the type checker

src/ocean_emulators/models/modules/activations.py
alxmrs: No problem, happy to see drive-by bug fixes.

alxmrs (approved):
LGTM, thanks for doing this!

PR #384, opened 2025-09-15
FOMO module components agree in dimension.

alxmrs:
I'm following an `in_channel`/`out_channel` pattern for all Modules now.
I've modified the config system to `build()` with these two values as
inputs. The FOMO model now accepts a tunable `embedding_dim` parameter, 
which define the size of the latent channels of the processor.

The trainer test now includes testing a FOMO config.

### DIFF
--- a/configs/train_om4_fomo.yaml
+++ b/configs/train_om4_fomo.yaml
@@ -35,7 +35,7 @@ step_transition: []
 backend: auto
 
 experiment:
-  name: local_om4_samudra
+  name: local_om4_fomo
   rand_seed: 15
   base_output_dir: .LOCAL
   wandb:
--- a/configs/train_om4_fomo.yaml
+++ b/configs/train_om4_fomo.yaml
@@ -53,6 +53,9 @@ model:
   last_kernel_size: 3
   pad: "circular"
 
+  encoder:
+    perceiver_depth: 4
+
   processor:
     ch_width: [200, 250, 300, 400]
     dilation: [1, 2, 4, 8]
@@ -62,7 +65,7 @@ model:
       block_type: "conv_next_block"
       kernel_size: 3
       activation: "capped_gelu"
-      upscale_factor: 4
+      upscale_factor: 2
       norm: "batch"
 
     down_sampling_block: "avg_pool"
--- a/configs/train_om4_fomo.yaml
+++ b/configs/train_om4_fomo.yaml
@@ -5,7 +5,7 @@ disk_mode: true
 pin_mem: true
 save_freq: 5
 epochs: 70
-batch_size: 4
+batch_size: 2
 learning_rate: 0.0002
 scheduler: {type: cosine}
 loss: mse
@@ -52,12 +52,15 @@ model:
   pred_residuals: false
   last_kernel_size: 3
   pad: "circular"
+  checkpointing: "all"
+  embedding_dim: 64
 
   encoder:
-    perceiver_depth: 4
+    perceiver_depth: 2
+    patch_size: [4, 8]
 
   processor:
-    ch_width: [200, 250, 300, 400]
+    ch_width: [64, 50, 40, 25]
     dilation: [1, 2, 4, 8]
     n_layers: [1, 1, 1, 1]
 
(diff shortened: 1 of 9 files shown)


src/ocean_emulators/config.py
```diff
@@ -333,13 +332,20 @@ class UNetBackboneConfig(BaseConfig):
     def build(
         self,
         in_channels: int,
+        out_channels: int | None,
```
jder: Do UNets usually have the same number of input and output channels? Kinda feels reasonable to me to remove `out_channels`, assume symmetry and to use a decoder afterwards if you want to change the number of channels as we do in Samudra. (I don't really understand why the original unet/samudra model has a final convolution from b channels to b channels, but seems like we should conservatively keep it that way when we're using it within the FOMO model?)

src/ocean_emulators/config.py
```diff
@@ -333,13 +332,20 @@ class UNetBackboneConfig(BaseConfig):
...
         checkpointing: Checkpointing | None,
     ) -> UNetBackbone:
         assert len(self.ch_width) == len(self.dilation) == len(self.n_layers), (
             "`ch_width`, `dilation`, and `n_layers` must have the same length."
         )
 
+        # This replicates the default behavior of Samudra.
+        if out_channels is None:
+            out_channels = (
+                self.ch_width[1] if len(self.ch_width) > 1 else self.ch_width[0]
```
jder: This doesn't make sense to me, why would we want to use `self.ch_width[1]`? Was this code from before the ch_width prepending of in_channels moved? (Again this seems confusing to have the same name for two slightly different meanings.) I think we always want `self.ch_width[0]` here, or do what I suggested above and don't pass out_channels at all.

src/ocean_emulators/config.py
```diff
@@ -432,6 +438,7 @@ def build(
             pad=self.pad,
             unet=self.unet.build(
                 in_channels=total_in_channels,
+                out_channels=None,  # `out_channels` is set from the unet config (via `ch_width[1]`).
```
jder: (Comment has the same problem as above?)

jder (changes requested):

src/ocean_emulators/config.py
alxmrs: I think this is a plan ol' bug! Yes, I think I was confused about if `in_channels` were appended to `ch_widths`.

src/ocean_emulators/config.py
alxmrs: How about I call this param `in_out_channels`?  Not sure why there's a final b to b channel, but I agree we should conservatively keep it.

jder: SGTM

alxmrs: No, it makes sense to be `in_channels` and the out channel is simply the first `ch_width`.

alxmrs: With `in_channels`, it's kind of like we add include a CNN "encoder" to the U-Net, but not a CNN decoder.

alxmrs: I'd like to chat about this for a bit. My bone to pick with not including a decoder module is that we have to use some sort of "reflection" to peek into the output dimension of the UNet backbone, we can't specify that cleanly what the decoder input dimension should be.

jder: I don't have strong feelings about whether there's a decoder in the unet or not, so long as we keep the same behavior as the original model. FWIW I don't think "reflection" is bad, you can build() the unet and then ask it for out_channels() before build()ing the decoder.

configs/train_om4_fomo.yaml
```diff
@@ -45,23 +45,30 @@ experiment:
...
+  checkpointing: "all"
+  embedding_dim: 64
+
+  encoder:
+    perceiver_depth: 3
+    patch_size: [4, 8]
 
   processor:
-    ch_width: [200, 250, 300, 400]
+    ch_width: [64, 50, 40, 25]
```
jder: Not relevant to the point of this PR but why are these tiny and decreasing?

jder (approved):
Thanks!

configs/train_om4_fomo.yaml
alxmrs: Oh, this is me trying to get the model to train! It's quite big now due to the perceivers, so I'm trying to make the embedding dimension and processor to work on much smaller channels. That it is decreasing is a mistake that I've made. Thanks for catching this.

PR #385, opened 2025-09-15
Updating the Contrib guide with skypilot examples.

alxmrs:
Parsing how to run each new skypilot script from the docstrings.


jder (approved):
Thanks!

PR #386, opened 2025-09-17
Always print the model summary during training.

alxmrs:
I find it really helpful to iterate and debug models to have the model summary. I think we should always print it.

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -526,7 +526,8 @@ class Trainer:
             data.to(self.device)
 
             if self.num_batches_seen == 0:
-                get_model_summary(self.model, data, self.debug)
+                # We always want to print the detailed model summary for training.
+                get_model_summary(self.model, data, False)
 
             TO: TrainBatchOutput = Stepper.train_batch(self.model, data, self.loss_fn)
             TO.loss.backward()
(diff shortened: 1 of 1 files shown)


jder:
Why is setting `debug = true` not sufficient? It is a ton of output and a bit annoying when you're working on other things.

alxmrs:
`debug=true` in the configs changes the training strategy to something less performant IIRC. I always want to print the model summary because it is a fixed additional log output, but it really helps me use CC during model debugging.

jder:
Setting debug = true does not change the training strategy to something less performant. Can you clarify why you can't set debug = true while debugging with CC? I would love a solution which does not make this output unavoidable for all users all the time.

alxmrs:
Maybe I could factor out model summaries as it's own config option? Currently, `debug=true` seems to mess with the training and validation loops: 

https://github.com/Open-Athena/Ocean_Emulator/blob/86163fdf6a5b9ab4ec20dd0551efb48bca60af20/src/ocean_emulators/train.py#L522

https://github.com/Open-Athena/Ocean_Emulator/blob/86163fdf6a5b9ab4ec20dd0551efb48bca60af20/src/ocean_emulators/train.py#L648

alxmrs:
Happy to close this and use debug for the time being.

PR #390, opened 2025-09-19
Decoder now using linear projection to unpatchify.

alxmrs:
This seems to be the standard ViT approach to reverse patches. I've also kept the simple Conv2d because I think it has some benefits (shared weights compared to a nn.Linear; better spatial reasoning; channel mixing is nice).

### DIFF
--- a/src/ocean_emulators/models/fomo.py
+++ b/src/ocean_emulators/models/fomo.py
@@ -52,8 +52,6 @@ class FOMO(BaseModel):
         )
 
     def forward_once(self, fts: torch.Tensor) -> torch.Tensor:
-        _, _, H, W = fts.shape
-
         for layer in self.layers:
             fts = layer(fts)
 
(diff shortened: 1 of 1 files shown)


src/ocean_emulators/models/fomo.py
```diff
@@ -37,15 +38,39 @@ def __init__(
...
+        self.patch_size = encoder.patch_size
+
         # Placeholder decoder is a non-globe aware Conv2d.
         layers = [
             encoder,
             processor,
             nn.Conv2d(processor.out_channels, out_channels, last_kernel_size),
         ]
         self.layers = nn.ModuleList(layers)
+        self.unpatch = nn.Linear(
```
alxmrs, line 55: Is `self.project` a better name for this?

adam-subel: I think upatch maybe more clear in this instance, a bit awkward, but I get what it means. It is a projection, but that feels less tied to this specific step of the process

adam-subel (approved):
Looks good!

PR #391, opened 2025-09-19
Essential hparam to tune in the encoder: `latent_dim`.

alxmrs:
I've added the ability to configure the `latent_dim` of the encoder's perceiver. This is essential to decrease (or increase) the memory complexity of the encoder.


adam-subel (approved):
Looks good!

PR #392, opened 2025-09-22
First round of perf improvements 

jder:
3 small fixes from profiling:

* Request async data transfers from CPU to GPU memory.
* Add the [required](https://docs.pytorch.org/docs/stable/data.html#memory-pinning) `pin_memory` method on our custom data type so those transfers can be a DMA 
* Remove a number of places we waited unnecessarily for GPU results to be ready (to check against NaN) which caused our CPU to wait there for the GPU to catch up. This isn't very helpful yet but is a step towards freeing up the CPU for other things (eg data decompression).

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -334,12 +334,8 @@ class TrainData:
     def to(self, device: torch.device) -> None:
         for step in self.td_dict:
             self.td_dict[step] = (
-                self.td_dict[step][0].to(
-                    device, non_blocking=True, memory_format=torch.channels_last
-                ),
-                self.td_dict[step][1].to(
-                    device, non_blocking=True, memory_format=torch.channels_last
-                ),
+                self.td_dict[step][0].to(device, non_blocking=True),
+                self.td_dict[step][1].to(device, non_blocking=True),
             )
 
     def pin_memory(self):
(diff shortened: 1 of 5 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -334,10 +334,18 @@ def __len__(self) -> int:
     def to(self, device: torch.device) -> None:
         for step in self.td_dict:
             self.td_dict[step] = (
-                self.td_dict[step][0].to(device),
-                self.td_dict[step][1].to(device),
+                self.td_dict[step][0].to(device, non_blocking=True),
+                self.td_dict[step][1].to(device, non_blocking=True),
```
alxmrs, line 338: Nice

src/ocean_emulators/datasets.py
```diff
@@ -334,10 +334,18 @@ def __len__(self) -> int:
...
             )
 
+    def pin_memory(self):
+        for step in self.td_dict:
+            self.td_dict[step] = (
+                self.td_dict[step][0].pin_memory(),
+                self.td_dict[step][1].pin_memory(),
+            )
+        return self
```
alxmrs, line 347: Definitely useful to have.

alxmrs (approved):
LGTM! Thanks for doing this.

PR #398, opened 2025-09-22
Naive decoder implementation.

alxmrs:
Created a simple decoder that mirrors our encoder.

### DIFF
--- a/src/ocean_emulators/models/modules/decoder.py
+++ b/src/ocean_emulators/models/modules/decoder.py
@@ -5,6 +5,18 @@ from torch import nn
 
 
 class PerceiverDecoder(nn.Module):
+    """A perceiver-based decoder that maps processed latent predictions into a whole column of the ocean.
+
+    Args:
+        in_channels (int): the number of input channels (typically, the output of our UNet backbone).
+        out_channels (int): size of our output channels (roughly: variables x depths).
+        patch_size (int | tuple[int, int]): the size of the patches to embed. Patches must evenly divide the input grid.
+          If a tuple is supplied, then it represents the (height, width) of the patches to embed.
+        perceiver_depth (int): depth of the perceiver module core.
+        perceiver_latent_dim (int): latent dimension of the perceiver module core. The `N` of the Perceiver's `O(M*N)`
+          complexity, where the `M` corresponds to the size of the input data.
+    """
+
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,62 @@
...
+        x = rearrange(x, "b l h w -> b h w l")
+        x = self.norm_patches(x)
+        x = self.perceiver(x)
+        x = self.norm_embedding(x)
+        x = rearrange(
+            x,
+            "b (h w c) -> b c h w",
+            c=self.out_channels,
+            h=self.grid_size[0],
+            w=self.grid_size[1],
```
alxmrs, line 58: These values could also be calculated based on the `patch_size[0] * h`, but we wouldn't be able to get `h` in the init of this module.

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,61 @@
...
+        self.in_channels = in_channels
+        self.out_channels = out_channels
+        self.grid_size = grid_size
+
+        grid_area = grid_size[0] * grid_size[1]
+
+        self.norm_patches = nn.LayerNorm(self.in_channels)
+        self.perceiver = Perceiver(
+            num_freq_bands=4,
+            max_freq=10.0,  # Depending on patch size and grid, consider values ranging from 3-10.
```
alxmrs, line 37: If we wanted to accept the patch size, we could automatically calculate this value (according to CC).

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,61 @@
...
+            latent_dim=perceiver_latent_dim,
+            num_classes=out_channels * grid_area,
+            weight_tie_layers=True,  # share weights of cross-attn blocks
+            self_per_cross_attn=2,  # ratio of self-attn (latent, small) and cross-attn (input, big) blocks
+        )
+        self.norm_embedding = nn.LayerNorm(out_channels * grid_area)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = rearrange(x, "b l h w -> b h w l")
+        x = self.norm_patches(x)
```
alxmrs, line 50: These norms are kinda misnamed

PR #407, opened 2025-09-30
Use OSN bucket for skypilot data, other minor fixes

jder:
NB stacked on #406, merge that first.

I copied the halfdeg data to a new location in the OSN bucket so we can use the same naming scheme as our skypilot configs were already expecting, then pointed to them. Tested via a [training run](https://wandb.ai/m2lines/ocean-emulators/runs/6dtu2st0), an [eval run](https://wandb.ai/m2lines/ocean-emulators/runs/cnmoqz0c?nw=nwuseroa_jder) and a [viz run](https://drive.google.com/drive/folders/1t3I819t0uiysWQCzukvEENuUA2lhJy_H).

Viz currently runs on AWS so no clear we should pull data from OSN but consistency seemed better than that minor optimization. 

Closes #369

PR #406, opened 2025-09-30
More common config defaults, fix typo bug

jder:
Fixes #372, some minor tweaks to make running jobs we typically run easier.

PR #408, opened 2025-09-30
Periodic padding

jder:
NB Stacked on #407 (and so #406).

We have had a clear artifact in our outputs at lon=0. I believe there were 2 causes ([thread](https://openathena.slack.com/archives/C09FXFDR2BF/p1758734811767269)):
* On 1 degree data: since there are 4 ConvNext blocks and each has a 2x downsample before them, that gives a final image width of 360/(2^4) which is 22.5. So, the last downsample block (from 45 -> 22) will end up sampling from implicit zero padding in the far-right column (which is lon=359). And on the upscale from 22 -> 45 we pad with zeros. These are both non-periodic.
* In any dataset, the bilinear upsample was non-periodic i.e. `[[1, 2, 3], ...]` gets upsampled to `[[1.0, 1.25, 1.75, 2.25, 2.75, 3.0], ...]` rather than circularly to eg `[[1.0, 1.5, 2.0, 2.5, 3.0, 2.0], …]`. 

This PR resolves the latter and appears to resolve the artifact for us on half-degree data. 
Before (from [om4_samudra_halfdeg_wide_dmse_no_cap](https://wandb.ai/m2lines/ocean-emulators/runs/k5szfus3?nw=nwuseroa_jder)):
SST_map_snapshot_t_0_before

After (from [periodic_test](https://wandb.ai/m2lines/ocean-emulators/runs/6dtu2st0?nw=nwuseroa_jder) and [periodic_eval](https://wandb.ai/m2lines/ocean-emulators/runs/cnmoqz0c?nw=nwuseroa_jder)):
SST_map_snapshot_t_0_after

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -28,6 +28,7 @@ from ocean_emulators.models.modules import (
     TransposedConvUpsample,
     UNetBackbone,
 )
+from ocean_emulators.models.modules.blocks import PeriodicBilinearUpsample
 from ocean_emulators.utils.data import DataContainer, DataSource, validate_data
 from ocean_emulators.utils.location import LocalLocation, Location, ResolvedLocation
 from ocean_emulators.utils.profiler import Profiler
@@ -322,7 +323,7 @@ class EncoderConfig(BaseConfig):
 
 
 DownSamplingBlocks = Literal["avg_pool", "max_pool"]
-UpSamplingBlocks = Literal["bilinear_upsample", "transposed_conv"]
+UpSamplingBlocks = Literal["bilinear_upsample", "transposed_conv", "periodic_upsample"]
 Checkpointing = Literal["all", "simple"]
 
 
@@ -354,6 +355,8 @@ class UNetBackboneConfig(BaseConfig):
                     return TransposedConvUpsample(
                         in_channels=in_channels, out_channels=out_channels
                     )
+                case "periodic_upsample":
+                    return PeriodicBilinearUpsample()
                 case _:
                     assert_never(self.up_sampling_block)
 
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -28,7 +28,7 @@ from ocean_emulators.models.modules import (
     TransposedConvUpsample,
     UNetBackbone,
 )
-from ocean_emulators.models.modules.blocks import PeriodicBilinearUpsample
+from ocean_emulators.models.modules.blocks import ZonallyPeriodicBilinearUpsample
 from ocean_emulators.utils.data import DataContainer, DataSource, validate_data
 from ocean_emulators.utils.location import LocalLocation, Location, ResolvedLocation
 from ocean_emulators.utils.profiler import Profiler
@@ -323,7 +323,9 @@ class EncoderConfig(BaseConfig):
 
 
 DownSamplingBlocks = Literal["avg_pool", "max_pool"]
-UpSamplingBlocks = Literal["bilinear_upsample", "transposed_conv", "periodic_upsample"]
+UpSamplingBlocks = Literal[
+    "bilinear_upsample", "transposed_conv", "zonally_periodic_upsample"
+]
 Checkpointing = Literal["all", "simple"]
 
 
@@ -355,8 +357,8 @@ class UNetBackboneConfig(BaseConfig):
                     return TransposedConvUpsample(
                         in_channels=in_channels, out_channels=out_channels
                     )
-                case "periodic_upsample":
-                    return PeriodicBilinearUpsample()
+                case "zonally_periodic_upsample":
+                    return ZonallyPeriodicBilinearUpsample()
                 case _:
                     assert_never(self.up_sampling_block)
 
--- a/src/ocean_emulators/models/modules/blocks.py
+++ b/src/ocean_emulators/models/modules/blocks.py
@@ -46,6 +46,45 @@ class BilinearUpsample(torch.nn.Module):
         return self.upsampler(x)
 
 
+class PeriodicBilinearUpsample(torch.nn.Module):
+    """Bilinear upsampling that enforces periodicity along the x/longitude axis."""
+
+    def __init__(self, upsampling: int | tuple[int, int] = 2):
+        super().__init__()
+        if isinstance(upsampling, tuple):
+            if tuple(upsampling) != (2, 2):
+                raise ValueError("PeriodicBilinearUpsample only supports 2x upsampling")
+            self.scale_h, self.scale_w = upsampling
+        else:
+            if upsampling != 2:
+                raise ValueError("PeriodicBilinearUpsample only supports 2x upsampling")
+            self.scale_h = self.scale_w = upsampling
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        # First upsample along latitude/height using standard "bilinear" interpolation.
+        # (This is just linear interpolation but pytorch requires us to use "bilinear" for
+        # 2d data.)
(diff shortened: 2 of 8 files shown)


src/ocean_emulators/config.py
```diff
@@ -322,7 +323,7 @@ def build(self, in_channels: int, out_channels: int) -> PerceiverEncoder:
 
 
 DownSamplingBlocks = Literal["avg_pool", "max_pool"]
-UpSamplingBlocks = Literal["bilinear_upsample", "transposed_conv"]
+UpSamplingBlocks = Literal["bilinear_upsample", "transposed_conv", "periodic_upsample"]
```
adam-subel: From a naming perspective, would labeling this "zonally_periodic_upsample" be more specific?

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -46,6 +46,45 @@ def forward(self, x):
         return self.upsampler(x)
 
 
+class PeriodicBilinearUpsample(torch.nn.Module):
```
adam-subel: Same comment as before, may be worth changing the name to clarify this is only zonally periodic

tests/test_periodic.py
```diff
@@ -0,0 +1,104 @@
...
+@pytest.mark.parametrize(
+    "upsampler_cls",
+    [
+        pytest.param(BilinearUpsample, id="bilinear"),
+        pytest.param(PeriodicBilinearUpsample, id="periodic"),
+    ],
+)
+@pytest.mark.parametrize("width", [8, 10, 16])
+@pytest.mark.parametrize("num_blocks", [1, 2, 3])
+def test_downscale_then_upscale_rotation_invariance(
```
adam-subel: Another nomenclature question. If I am not mistaken, this is a translation test not a rotation test, maybe the naming should adjust to reflect that

adam-subel (approved):
Looks good, I left a couple suggestions for naming, but I will let you decide if any of it is worth changing.

jder:
Makes sense to me, thanks for the review!

src/ocean_emulators/config.py
```diff
@@ -354,6 +357,8 @@ def create_upsampling_block(in_channels: int, out_channels: int):
                     return TransposedConvUpsample(
                         in_channels=in_channels, out_channels=out_channels
                     )
+                case "zonally_periodic_upsample":
+                    return ZonallyPeriodicBilinearUpsample()
```
alxmrs, line 361: ~I'm curious why these don't need `in_channels` and `out_channels`.~

It's cool that this sampling is independent of the inputs and outputs.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -46,6 +46,49 @@ def forward(self, x):
...
+                raise ValueError(
+                    "ZonallyPeriodicBilinearUpsample only supports 2x upsampling"
+                )
+            self.scale_h, self.scale_w = upsampling
+        else:
+            if upsampling != 2:
+                raise ValueError(
+                    "ZonallyPeriodicBilinearUpsample only supports 2x upsampling"
+                )
+            self.scale_h = self.scale_w = upsampling
```
alxmrs: 🐑  optional alternative
```suggestion
        if isinstance(upsampling, int):
            upsampling = (upsampling, upsampling)
        if tuple(upsampling) != (2, 2):
            raise ValueError(
                    "ZonallyPeriodicBilinearUpsample only supports 2x upsampling"
            )
        self.scale_h, self.scale_w = upsampling
```

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -46,6 +46,49 @@ def forward(self, x):
...
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        # First upsample along latitude/height using standard "bilinear" interpolation.
+        # (This is just linear interpolation but pytorch requires us to use "bilinear" for
+        # 2d data.)
```
alxmrs, line 65: Thanks, these notes are helpful.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -46,6 +46,49 @@ def forward(self, x):
...
+        right = torch.roll(up_height, shifts=-1, dims=-1)
+        midpoint = 0.5 * (left + right)
+
+        out = torch.empty(
+            (*up_height.shape[:-1], up_height.shape[-1] * self.scale_w),
+            dtype=up_height.dtype,
+            device=up_height.device,
+        )
+        out[..., ::2] = left
+        out[..., 1::2] = midpoint
```
alxmrs, line 83: A really elegant upsample! :)

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -46,6 +46,49 @@ def forward(self, x):
...
+        left = up_height
+        right = torch.roll(up_height, shifts=-1, dims=-1)
+        midpoint = 0.5 * (left + right)
+
+        out = torch.empty(
+            (*up_height.shape[:-1], up_height.shape[-1] * self.scale_w),
+            dtype=up_height.dtype,
+            device=up_height.device,
+        )
+        out[..., ::2] = left
```
alxmrs, line 82: Is the last dim always the channel dim? I thought it was dim=2 in data space? How does this apply to our CNNs, I think I'm a bit confused/curious.

tests/test_periodic.py
```diff
@@ -0,0 +1,104 @@
...
+import torch
+
+from ocean_emulators.models.modules.blocks import (
+    AvgPool,
+    BilinearUpsample,
+    ZonallyPeriodicBilinearUpsample,
+)
+
+
+def _pad_like_unet(feature: torch.Tensor, skip: torch.Tensor) -> torch.Tensor:
```
alxmrs, line 12: 🐑 What do you think of extracting the main unet pad block (above) into a function? I could go either way.

tests/test_periodic.py
```diff
@@ -0,0 +1,104 @@
...
+@pytest.mark.parametrize(
+    "upsampler_cls",
+    [
+        pytest.param(BilinearUpsample, id="bilinear"),
+        pytest.param(ZonallyPeriodicBilinearUpsample, id="zonally_periodic"),
+    ],
+)
+@pytest.mark.parametrize("width", [8, 10, 16])
+@pytest.mark.parametrize("num_blocks", [1, 2, 3])
+def test_downscale_then_upscale_translation_invariance(
```
alxmrs, line 36: 🐑 Seems fuzzable with hypothesis!

tests/test_periodic.py
```diff
@@ -0,0 +1,104 @@
...
+        # if the final plane would be a width of 1, there's no shift we can try
+        # below to confirm translation invariance
+        pytest.skip("Width is too small for the given number of blocks")
+    divisibility_condition = width % total_downscale == 0
+    batch, channels, height = 1, 3, 8
+    base_input = torch.arange(
+        batch * channels * height * width, dtype=torch.float32
+    ).reshape(batch, channels, height, width)
+
+    def run_pipeline(original: torch.Tensor, count) -> torch.Tensor:
```
alxmrs, line 56: What a fun use of recursion!

tests/test_periodic.py
alxmrs: This is a well written test. Thanks for illustrating the difference.

alxmrs (approved):
Small nits, in general, thanks for the fix!

src/ocean_emulators/models/modules/blocks.py
jder: The last dim is longitude, which is what we want to up-sample here.

PR #413, opened 2025-10-06
Readme and config updates based on Cody onboarding

jder:


jder:
@alxmrs I'm going to merge this after tests pass so that Amogh will have the updated info but please do take a look when you're back in case there's anything worth fixing up in a follow-up.

PR #417, opened 2025-10-07
Jesse decoder suggestions

jder:
This is meant to be an async handoff of some of the things I experimented with while you were away; I wouldn't really expect you to merge this so much as take the things which seem good to you. Will leave some comments below about motivation/results/context.


configs/train_om4_fomo.yaml
```diff
@@ -5,10 +5,10 @@ disk_mode: true
...
 save_freq: 5
 epochs: 70
-batch_size: 1
-learning_rate: 0.0002
-scheduler: {type: cosine}
-loss: mse
+batch_size: 4
+learning_rate: 0.0006
+scheduler: { type: cosine }
+loss: mse_dynamic_no_limit
```
jder, line 11: Most of the changes in this file are to make the model closer to the original model, while adding the new decoder and fitting into GPU memory.

configs/train_om4_fomo.yaml
```diff
@@ -44,32 +44,37 @@ experiment:
     entity: m2lines
     group: samudra-train-om4
   prognostic_vars_key: thermo_dynamic_all
-  boundary_vars_key: tau_hfds_hfds_anom
-data:
-  !include data/public.yaml
+  boundary_vars_key: tau_hfds
```
jder, line 47: Also simplified here and above in `steps` to try and reduce training memory usage/complexity.

configs/train_om4_fomo.yaml
```diff
@@ -44,32 +44,37 @@ experiment:
...
   pad: "circular"
   checkpointing: "all"
-  embedding_dim: 80
+  embedding_dim: 280
 
   encoder:
-    perceiver_depth: 2
-    patch_size: [12, 24]
+    perceiver_depth: 3
+    patch_size: [12, 20]
```
jder, line 59: I think with 24 this caused us to end up with too small of a final convnext block at the innermost one.

src/ocean_emulators/models/fomo.py
```diff
@@ -40,6 +45,20 @@ def __init__(
...
+                m,
+                nn.LayerNorm
+                | FeedForward
+                | nn.Linear
+                | Perceiver
+                | PerceiverEncoder
+                | UNetBackbone
+                | Attention,
+            ),
+        )
```
jder, line 60: This helped fit into GPU memory. I think checkpointing Attention is really all we need in here. In general we might consider removing the more complicated checkpointing thing we have and using this strategy instead.

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -48,16 +49,20 @@ def __init__(
         self.norm_patches = nn.LayerNorm(self.in_channels)
         self.perceiver = Perceiver(
             num_freq_bands=4,
-            max_freq=1.0,
+            max_freq=max(
+                *self.patch_size
+            ),  # This is not actually a "frequency" but a maximum of the width appears to be reasonable from looking at the code
```
jder, line 54: I spent some time reviewing the code here and I think this is badly named and is closer to "number of frequency spatial bands to identify". Perhaps if we swap to the FlashPerceiver we can remove this weirdness.

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -48,16 +49,20 @@ def __init__(
...
             depth=perceiver_depth,
             input_axis=2,  # Number of positional dims before token dim
             input_channels=self.in_channels,
             latent_dim=perceiver_latent_dim,
+            num_latents=perceiver_num_latents,
             num_classes=out_channels,
-            weight_tie_layers=True,  # share weights of cross-attn blocks
+            weight_tie_layers=False,  # share weights of cross-attn blocks
```
jder, line 61: Was just seeing if this worked, can probably put this back.

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -48,16 +49,20 @@ def __init__(
...
             self_per_cross_attn=2,  # ratio of self-attn (latent, small) and cross-attn (input, big) blocks
+            final_classifier_head=False,
```
jder, line 63: This I think is an important change. The default behavior is to mean all the latents together and then a layernorm (which doesn't make much sense to me) before a linear projection. So we turn this off and do our own mean + linear projection below. I think a good next thing to try would be to get rid of the mean and just do a linear projection from the num_latents x laten_dim embeddings directly.

src/ocean_emulators/models/modules/decoder.py
jder, line 1: A bunch of separate suggested changes in here:

* Removed the norm before & after the perceiver -- I didn't see this in aurora and it doesn't make much sense to me though maybe I missed it.
* Changed the decoder so each output patch is only based on decoding the embeddings for the patch it is trying to decode, not the whole ocean. 
* This results in us having to treat the patch embeddings (ie a vector of length N) as X tokens by Y input_channels. We don't have any such structure so I arbitrarily chose to make N tokens by 1 channel, but there is probably something more principled we should do here, or perhaps this is a signal that we're doing something weird. I tried using PerceiverIO instead (which this fits more naturally into), but ran out of memory (I think either because the implementation is inefficient or because I should have broken my decoding up into multiple sets of queries).

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -78,12 +83,9 @@ def forward(self, x: Input) -> Float[torch.Tensor, "batch {self.embed_dim} h w"]
             ph=self.patch_size[0],
             pw=self.patch_size[1],
         )
-        # This is applying a layer norm across all our data channels. I am not sure if this will have a positive or
-        # negative effect. It may make it easier for the Perceiver to process data across scales, but it destroys the
-        # physical relationships in the data and aggregates across depth level and history. We should be cautious here.
-        x = self.norm_patches(x)
```
jder, line 84: I agreed with this comment and got rid of this norm.

jder (commented):
Suggested next steps:

* I think the built-in Perceiver fourier encoding is probably unhelpful and we should prioritize adding real positional information somehow in both encoder + decoder and turning off the built-in encoding. 
* Might even make sense to pull some of these encoder changes over but leave the decoder linear and see how far we can get with that. 
* I think having a mini model (small input, perhaps synthetic data or just top-5 depths of 1-degree) we can train quickly would be very helpful in exploring what is helping and what is not. 

https://wandb.ai/m2lines/ocean-emulators/runs/g7utrp31/overview is the main run I did, which includes most of the below changes but not all (it doesn't include the norm-removal changes, for example, which I was experimenting with in the later v0_7_* runs to build some intuition about whether things were totally broken or not, but did not let them finish).

src/ocean_emulators/models/modules/encoder.py
alxmrs: Oh, thank you! Yeah, I wasn't sure what to make of how this argument was formulated. I was relying on CC for recommendations here, and boy am I glad to use something more principled instead.

PR #419, opened 2025-10-08
Improvements to our encoder

alxmrs:
Pulling in improvements from #417, focusing just on the encoder. These include: 
- Activation checkpointing (nice, clean implementation, @jder!).
- Training config wrangling
- Adding an essential config option for the encoder (`perceiver_num_latents`)
- Turning off the default linear projection scheme so that we can implement our own.

### DIFF
--- a/configs/train_om4_fomo.yaml
+++ b/configs/train_om4_fomo.yaml
@@ -52,16 +52,17 @@ model:
   last_kernel_size: 3
   pad: "circular"
   checkpointing: "all"
-  embedding_dim: 80
+  embedding_dim: 280
 
   encoder:
     perceiver_depth: 5
-    patch_size: [12, 24]
+    patch_size: [12, 12]
     perceiver_latent_dim: 64
+    perceiver_num_latents: 512
 
   processor:
-    ch_width: [100, 200, 300]
-    dilation: [1, 1, 1]
+    ch_width: [380, 480, 520]
+    dilation: [2, 4, 8]
     n_layers: [1, 1, 1]
 
     core_block:
--- a/configs/train_om4_fomo.yaml
+++ b/configs/train_om4_fomo.yaml
@@ -29,7 +29,7 @@ inference_times:
   - start: "2004-10-05"
     end: "2005-10-05" # "2014-10-05"
 data_stride: [1]
-steps: [1]
+steps: [4]
 step_transition: []
 
 backend: auto
--- a/configs/train_om4_fomo.yaml
+++ b/configs/train_om4_fomo.yaml
@@ -5,8 +5,8 @@ disk_mode: true
 pin_mem: true
 save_freq: 5
 epochs: 70
-batch_size: 1
-learning_rate: 0.0003
+batch_size: 4
+learning_rate: 0.0006
 scheduler: { type: cosine }
 loss: mse_dynamic_no_limit
 finetune: false
@@ -69,7 +69,7 @@ model:
       block_type: "conv_next_block"
       kernel_size: 3
       activation: "capped_gelu"
-      upscale_factor: 1
+      upscale_factor: 2
       norm: "batch"
 
     down_sampling_block: "avg_pool"
--- a/configs/train_om4_fomo.yaml
+++ b/configs/train_om4_fomo.yaml
@@ -62,7 +62,7 @@ model:
 
   processor:
     ch_width: [380, 480, 520]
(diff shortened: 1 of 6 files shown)


alxmrs:
Currently running a training run to verify that this encoder branch is good.

alxmrs:
WIP training run: https://wandb.ai/m2lines/ocean-emulators/runs/c0u72uv1 🤞 hope it goes through

alxmrs:
Fix and a successful (start of a) training run is here: https://wandb.ai/m2lines/ocean-emulators/runs/jkaaikov

configs/train_om4_fomo.yaml
```diff
@@ -29,7 +29,7 @@ inference_times:
   - start: "2004-10-05"
     end: "2005-10-05" # "2014-10-05"
 data_stride: [1]
-steps: [4]
+steps: [1]
```
adam-subel: Is this just setting recurrent steps to 1 for testing or are you trying to look at decreasing the rollouts in training?

configs/train_om4_fomo.yaml
```diff
@@ -44,32 +44,32 @@ experiment:
     entity: m2lines
     group: samudra-train-om4
   prognostic_vars_key: thermo_dynamic_all
-  boundary_vars_key: tau_hfds_hfds_anom
-data:
-  !include data/public.yaml
+  boundary_vars_key: tau_hfds
```
adam-subel: Same question here about dropping anoms, not super important, but just curious

adam-subel (commented):
Looks good, I did my best to go over details of the perceiver to make sure I understood what the design changes affect and everything made sense of me. My only questions were clarifying if the config is a new baseline or just for testing?

configs/train_om4_fomo.yaml
alxmrs: I'm not totally sure, I think the intention is for training -- cc: @jder.

configs/train_om4_fomo.yaml
alxmrs: Jesse and I discussed this (correct me if I'm wrong @jder), but we'd ideally like fewer vars for experimenting with the models before scaling up.

configs/train_om4_fomo.yaml
jder: This was for memory reasons. If it trains with 4 please go ahead and revert.

configs/train_om4_fomo.yaml
jder: I do think it would be helpful to build a mini model we can train quickly but this was just because I was annoyed at waiting for the anoms to get calculated all the time and we haven't done the work to precompute them. Fine to revert this, too. (Though once we add temporal embeddings, the model can compute this itself if it's helpful.)

src/ocean_emulators/models/fomo.py
```diff
@@ -56,6 +61,20 @@ def __init__(
...
+                m,
+                nn.LayerNorm
+                | FeedForward
+                | nn.Linear
+                | Perceiver
+                | PerceiverEncoder
+                | UNetBackbone
+                | Attention,
+            ),
+        )
```
jder: I know I wrote this but do you think we should somehow make this gated on the checkpointing configuration?

jder (approved):
Thanks for getting this landed!

configs/train_om4_fomo.yaml
alxmrs: I'll start a run with 4 steps now and see.

src/ocean_emulators/models/fomo.py
alxmrs: Good suggestion; will add this now.

configs/train_om4_fomo.yaml
alxmrs: Ah, makes sense. Reverted.

configs/train_om4_fomo.yaml
alxmrs: This does indeed train! https://wandb.ai/m2lines/ocean-emulators/runs/zghagqcp?nw=nwuseralxmrs

configs/train_om4_fomo.yaml
adam-subel: I want to add that I think it might be fine to more broadly drop using anoms (I honestly do not think it is that helpful compared to just passing hfds)

PR #420, opened 2025-10-09
Fix mutable args

### LINKED ISSUE #221: Fix mutable default arguments
In aggregator/validate/map.py and aggregator/validate/snapshot.py there is a mutable dict for a default. We should fix these and probably turn on ruff's [B006](https://docs.astral.sh/ruff/rules/mutable-argument-default/) rule and perhaps related ones to prevent this in the future.

>               Using a mutable default argument for 'metadata' can lead to unexpected behavior across instances. Consider setting the default to None and initializing within the method.

_Originally posted by @Copilot in https://github.com/suryadheeshjith/Ocean_Emulator/pull/220#discussion_r2052447164_

amogh-gulati:
closes #221

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,8 +79,7 @@ dev = [
 ]
 
 [tool.ruff]
-lint.select = [ "D", "E", "F", "I", "UP", "W" ]
-lint.extend-select = [ "B006" ]
+lint.select = [ "B006", "D", "E", "F", "I", "UP", "W" ]
 # F722: https://docs.kidger.site/jaxtyping/faq/#flake8-or-ruff-are-throwing-an-error
 lint.ignore = [
   "D1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,7 @@ dev = [
 
 [tool.ruff]
 lint.select = [ "D", "E", "F", "I", "UP", "W" ]
+lint.extend-select = [ "B006" ]
 # F722: https://docs.kidger.site/jaxtyping/faq/#flake8-or-ruff-are-throwing-an-error
 lint.ignore = [
   "D1",
--- a/src/ocean_emulators/aggregator/validate/map.py
+++ b/src/ocean_emulators/aggregator/validate/map.py
@@ -20,14 +20,18 @@ class MapAggregator(ValidateSubAggregator):
         ),
     }
 
-    def __init__(self, metadata: dict[str, dict[str, str]] = {}, hist: int = 0):
+    def __init__(
+        self, metadata: dict[str, dict[str, str]] | None = None, hist: int = 0
+    ):
         """
         Args:
             metadata: Mapping of variable names their metadata that will
                 used in generating logged image captions.
             hist: Number of history steps to include in the snapshot.
         """
-        self._metadata: dict[str, dict[str, str]] = metadata
+        self._metadata: dict[str, dict[str, str]] = (
+            {} if metadata is None else dict(metadata)
+        )
         self._n_batches = 0
         self._target_data: dict[str, torch.Tensor] = {}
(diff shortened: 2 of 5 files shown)


jder:
Thanks! I'm on vacation this week so maybe @alxmrs can review.

src/ocean_emulators/aggregator/validate/map.py
```diff
@@ -20,14 +20,18 @@ class MapAggregator(ValidateSubAggregator):
...
+        self, metadata: dict[str, dict[str, str]] | None = None, hist: int = 0
+    ):
         """
         Args:
             metadata: Mapping of variable names their metadata that will
                 used in generating logged image captions.
             hist: Number of history steps to include in the snapshot.
         """
-        self._metadata: dict[str, dict[str, str]] = metadata
+        self._metadata: dict[str, dict[str, str]] = (
```
alxmrs: 🐑 if the ruff formatting with the paren makes this multiline, I think it would be simpler to be more like an `if` statement with early returning. 

(Btw, "🐑" is from https://docs.google.com/document/d/1dfD7RKXbWnILxtr7iREnC0aWcnRUBaUi7ULGYYrTVzE/edit?usp=drivesdk, which means a sheepish suggestion. Feel free to ignore.)


For example, this seems simpler:
```
if metadata is None:
    metadata = {}
self._metadata = metadata
```

src/ocean_emulators/aggregator/validate/snapshot.py
```diff
@@ -25,14 +25,18 @@ class SnapshotAggregator(ValidateSubAggregator):
...
+        self, metadata: dict[str, dict[str, str]] | None = None, hist: int = 0
+    ):
         """
         Args:
             metadata: Mapping of variable names their metadata that will
                 used in generating logged image captions.
             hist: Number of history steps to include in the snapshot.
         """
-        self._metadata: dict[str, dict[str, str]] = metadata
+        self._metadata: dict[str, dict[str, str]] = (
```
alxmrs: 🐑 see above note

src/ocean_emulators/aggregator/validate/map.py
alxmrs: This may also help with the type inference (if that is why you use `dict` here).

alxmrs (approved):
A small suggestion. LGTM. Welcome to the Ocean Emulator Contributors!

alxmrs (approved):
A small nit, but LGTM! Welcome to the OA Ocean Emulators contributors!

pyproject.toml
```diff
@@ -80,6 +80,7 @@ dev = [
 
 [tool.ruff]
 lint.select = [ "D", "E", "F", "I", "UP", "W" ]
+lint.extend-select = [ "B006" ]
```
jder: I think we can add this to the list in `lint.select` above, no?

src/ocean_emulators/viz/core.py
```diff
@@ -3071,11 +3071,14 @@ def convert_gif(
             mpath,
             gpath="movie.gif",
             gif_palette=False,
-            resolution=[480, 320],
+            resolution=None,
```
jder: Alternatively could probably use a tuple here instead.

jder (approved):
Great, thanks @amogh-gulati!

jder:
(feel free to merge whenever you're ready)

src/ocean_emulators/aggregator/validate/map.py
amogh-gulati: yes formatter makes it multiline! used an if instead

src/ocean_emulators/viz/core.py
amogh-gulati: that seems more elegant yes

PR #423, opened 2025-10-15
Allow us to block ec2-related failures from merging

jder:
#420 was merged without GPU tests passing. This was permitted by our branch protection rules because the "run-tests" job in the GPU tests workflow was *skipped* (because one of its dependencies, the ec2 job which spins up a GPU for us) failed. Perhaps surprisingly, this skipped state [counts as success](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks) for the purpose of branch protection rules. 

So this defines a new job which explicitly fails when either the ec2-spin-up or GPU test themselves fail, which we'll add to our branch protection rules.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -24,7 +24,7 @@ jobs:
     secrets:
       GH_SA_TOKEN: ${{ secrets.GH_SA_TOKEN }}
 
-  test:
+  benchmark:
     needs: ec2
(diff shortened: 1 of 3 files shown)


jder:
(Before merging this I will update the branch protection rules with the new names and verify I can in fact merge it.)

ryan-williams (approved):

PR #424, opened 2025-10-15
Logging instance instead of global logger

### LINKED ISSUE #102: Use logging instances, not global loggers.

amogh-gulati:
closes #102

### DIFF
--- a/src/ocean_emulators/backend.py
+++ b/src/ocean_emulators/backend.py
@@ -2,6 +2,8 @@ import logging
 
 import torch
 
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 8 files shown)


jder (approved):
Thank you!

Issue #403, opened 2025-09-29
Resolve new data loading bottleneck


jder:
After https://github.com/Open-Athena/Ocean_Emulator/pull/392 maybe, our actual train step time is ~5s and data loading appears to be the bottleneck again see eg https://wandb.ai/m2lines/ocean-emulators/runs/6dtu2st0?nw=nwuseroa_jder

jder:
Ok, this appears to be some combination of loader-to-CPU transfer overhead (mmap/munmap is very slow) and the data loaders themselves falling behind. (Probably not helped by having 32 threads, but in fact I think the problem is the normalization/etc math happening which we should move to GPU).

jder:
Step one is #412

PR #425, opened 2025-10-17
Move data loader compute pieces to GPU

### LINKED ISSUE #412: Move data loading steps to GPU
It appears in #403 that in some setups we are just spending too much time on the CPU during data loading. Most of these steps could be moved to GPU, which we'll need for #377 anyway.

jder:
We were spending a ton of CPU time doing normalization, which was causing all kinds of bad knock-on performance effects due to (a) not having CPU time for the main process and (b) data loaders not keeping up with the GPU since they were busy doing math. 

This wraps the torch DataLoader with a new class which knows how to do some post-processing on the GPU before handing off the data. This keeps the I/O and xarray/zarr/etc python overhead in separate processes but lets us do the normalization on the main process on the GPU. Experiments on half-degree data shows this cuts train iteration time significantly.

To avoid having multiple paths here, this also deletes the old "eager" data loader. I'd also like to delete the inference data loader to reduce duplication caused by it but figured it could be a follow-up. 

Closes #412 and #403

### DIFF
--- a/src/ocean_emulators/constants.py
+++ b/src/ocean_emulators/constants.py
@@ -236,7 +236,6 @@ def construct_metadata(data: xr.Dataset) -> dict[str, dict[str, str]]:
 
 
 class LoaderVersion(enum.Enum):
-    OM4_EAGER = "om4-eager"
     OM4_TORCH = "om4-torch"
 
 
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -416,7 +416,7 @@ class TorchTrainDataset(Dataset[RawTrainData]):
         executor: ThreadPoolExecutor | None = None,
     ):
         super().__init__()
-        self.id = self.__class__.__name__ + "_" + str(id(self))
+        self.id = f"{self.__class__.__name__}_{str(id(self))}"
         self.device = get_device()
 
         self.hist: int = hist
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -635,6 +635,7 @@ def concurrent_compute(
     wait(futures)
 
 
+@final
 class TrainDataLoader:
     """Wrapper around a torch DataLoader that handles GPU post-processing.
 
@@ -653,11 +654,11 @@ class TrainDataLoader:
     def __init__(
         self,
         dataloader: torch.utils.data.DataLoader[RawTrainData],
-        datasets: dict[TorchTrainDataset.Id, TorchTrainDataset],
+        datasets: list[TorchTrainDataset],
         device: torch.device,
     ):
         self._dataloader = dataloader
-        self._datasets = datasets
+        self._datasets = {dataset.id: dataset for dataset in datasets}
         self._device = device
 
     def __iter__(self):
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -2,7 +2,7 @@ import logging
 import time
 from concurrent.futures import wait
 from concurrent.futures.thread import ThreadPoolExecutor
-from typing import Any
+from typing import TypeAlias, final
 
 import numpy as np
 import torch
@@ -13,7 +13,6 @@ from torch.utils.data import Dataset
 from xarray_einstats.einops import rearrange as xr_rearrange  # noqa: F401
 
 from ocean_emulators.constants import (
-    Boundary,
     BoundaryVarNames,
     Example,
     GridMask,
@@ -295,12 +294,38 @@ class InferenceDatasets(Dataset):
         return (self.datasets[idx], self.lengths[idx])
 
 
+class RawTrainData:
+    def __init__(self, dataset_id: "TorchTrainDataset.Id"):
+        self.dataset_id: TorchTrainDataset.Id = dataset_id
+        self.raw_data: list[tuple[torch.Tensor, torch.Tensor]] = []
+        self.load_stats: LoadStats | None = None
+
+    def insert(self, all_prognostic: torch.Tensor, all_boundary: torch.Tensor):
+        self.raw_data.append((all_prognostic, all_boundary))
+
+    def to(self, device: torch.device):
+        self.raw_data = [
+            (
+                all_prognostic.to(device, non_blocking=True),
+                all_boundary.to(device, non_blocking=True),
+            )
+            for all_prognostic, all_boundary in self.raw_data
+        ]
+
+    def pin_memory(self):
+        self.raw_data = [
+            (all_prognostic.pin_memory(), all_boundary.pin_memory())
+            for all_prognostic, all_boundary in self.raw_data
+        ]
+        return self
+
+
 class TrainData:
     def __init__(self, num_prognostic_channels: int):
         self.td_dict: dict[int, Example] = {}
-        self.load_stats: LoadStats | None = None
         self.num_prognostic_channels = num_prognostic_channels
         self.steps = 0
+        self.load_stats: LoadStats | None = None
 
     def insert(self, input_: Input, label: Prognostic):
         self.td_dict[self.steps] = (input_, label)
@@ -350,7 +375,8 @@ class TrainData:
         return iter(self.td_dict)
 
 
-class TrainDataset(Dataset):
+@final
+class TorchTrainDataset(Dataset[RawTrainData]):
     """
     This class is used for training and validation.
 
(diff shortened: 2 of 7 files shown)


src/ocean_emulators/utils/train.py
```diff
@@ -16,22 +16,23 @@ def pairwise(iterable):
...
 
 
-def collate_train_data(data: Sequence[TrainData]) -> TrainData:
-    num_prognostic_channels = data[0].num_prognostic_channels
-    steps = len(data[0])
+def collate_raw_train_data(data: Sequence[RawTrainData]) -> RawTrainData:
+    batched_data = RawTrainData(data[0].dataset_id)
+    assert all(d.dataset_id == batched_data.dataset_id for d in data), (
+        "we don't support heterogenous batches yet"
+    )
```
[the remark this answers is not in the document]
jder: Yeah, I wasn't sure how far to go here. My feeling is this is a reasonable first step towards a new data loading infrastructure which would either do per-_item_ normalization or else enforce batch-wide uniform datasets but at the moment we only ever use stride=1 and so while this is slightly awkward it at least explodes immediately rather than doing anything bad if we use multiple datasets before fixing this. Open to (human) feedback if we should change something here.

src/ocean_emulators/constants.py
```diff
@@ -236,7 +236,6 @@ def construct_metadata(data: xr.Dataset) -> dict[str, dict[str, str]]:
 
 
 class LoaderVersion(enum.Enum):
-    OM4_EAGER = "om4-eager"
```
alxmrs, line 235: Why are we getting rid of the old loader? So we don't have to maintain backwards compatibility given the new RawTrainData interface?

src/ocean_emulators/datasets.py
```diff
@@ -604,6 +416,7 @@ def __init__(
         executor: ThreadPoolExecutor | None = None,
     ):
         super().__init__()
+        self.id = self.__class__.__name__ + "_" + str(id(self))
```
alxmrs: 🐑 nit: this is slightly faster/more canonical:
```suggestion
        self.id = f"{self.__class__.__name__}_{str(id(self))}"
```
(I don't know if `str` should be there or not)

src/ocean_emulators/datasets.py
```diff
@@ -636,26 +449,39 @@ def __init__(
             indices_da + stride * window_dim
         )
 
-        self.wet = wet.bool()
-        self.wet_surface = wet_surface.bool()
+        self.wet = wet.bool().to(self.device)
+        self.wet_surface = wet_surface.bool().to(self.device)
+
+        def flatten_to_device(means_or_stds: xr.Dataset) -> torch.Tensor:
```
alxmrs, line 455: Nice!

src/ocean_emulators/datasets.py
```diff
@@ -689,63 +515,97 @@ def __getitem__(self, idx: int):
...
-
+            TD.insert(prognostic_all, boundary)
         TD.load_stats = LoadStats(time.perf_counter() - start_time)
+
         return TD
 
+    def to_train_data(self, raw_train_data: RawTrainData) -> TrainData:
+        train_data = TrainData(self.num_prognostic_channels)
+        for prognostic_all, boundary_all in raw_train_data.raw_data:
+            input, label = self._get_input_and_label(
```
alxmrs, line 526: I see, interesting that we can do this late -- it makes sense how this could lead to a perf improvement.

src/ocean_emulators/datasets.py
```diff
@@ -689,63 +515,97 @@ def __getitem__(self, idx: int):
...
 
+        def normalize(
+            data: Float[torch.Tensor, "batch time var lat lon"],
+            means: Float[torch.Tensor, " var"],
+            stds: Float[torch.Tensor, " var"],
+            fill_nan: bool = True,
+            fill_value: float = 0.0,
+        ) -> Float[torch.Tensor, "batch time var lat lon"]:
+            """Normalize input data treated as torch Tensors."""
+            norm = (data - means.view(1, 1, -1, 1, 1)) / stds.view(1, 1, -1, 1, 1)
```
alxmrs, line 565: Nice use of `view`.

src/ocean_emulators/datasets.py
```diff
@@ -773,3 +633,65 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        datasets: dict[TorchTrainDataset.Id, TorchTrainDataset],
+        device: torch.device,
+    ):
+        self._dataloader = dataloader
+        self._datasets = datasets
+        self._device = device
+
+    def __iter__(self):
+        """Iterate over the dataloader, converting RawTrainData to TrainData."""
+        for raw_train_data in self._dataloader:
```
alxmrs, line 666: Just want to confirm: the random shuffling is handled by the _dataloader, right?

alxmrs: Looks to be the case given the `sampler` property.

src/ocean_emulators/datasets.py
```diff
@@ -773,3 +633,65 @@ def load_variable_data(var: xr.Variable) -> None:
             futures.append(executor.submit(load_variable_data, var))
 
     wait(futures)
+
+
+class TrainDataLoader:
```
alxmrs, line 639: I don't quite understand why this wrapper class is needed.

alxmrs: Is it so we can collect a dictionary of all the created datasets? Does that specifically handle the GPU loading or main/process thread concern?

src/ocean_emulators/constants.py
alxmrs: How will this affect our automated performance benchmarks?

src/ocean_emulators/train.py
```diff
@@ -732,78 +731,53 @@ def init_data_loaders(self, cur_step: int) -> None:
...
-                            masked_fill_value=self.normalize_fill_value,
-                            stride=stride,
-                        )
-                        for stride in self.data_stride
-                    ]
-                )
             case TorchTrainDataset.FLAG:
-                train_data = ConcatDataset(
+                train_data: torch.utils.data.Dataset[RawTrainData] = ConcatDataset(
                     [
```
alxmrs: 🐑  I don't mind the `register_dataset` function, but I am somewhat bothered that it acts via side effect. We are already creating the list of datasets via list comprehensions -- what if instead of the registration method, we stored the lists in a variable, then parsed the id:dataset dictionaries from the variables, then passed the list to `ConcatDataset`. It seems overall less complex than the side-effect-in-a-comprehension method. WDYT?

src/ocean_emulators/utils/data.py
```diff
@@ -162,13 +162,15 @@ def slice(self, time: "TimeConfig") -> Self:
         data = self.data.sel(time=time.time_slice)
         return dataclasses.replace(self, name=f"{time=}[{self.name}]", data=data)
 
+    # TODO(jder): delete this once we've de-duplicated InferenceDataset with TorchTrainDataset
```
alxmrs, line 165: Thanks for the TODOs -- is there an associated GH issue?

src/ocean_emulators/utils/train.py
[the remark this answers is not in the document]
alxmrs: Remind me, what is stride doing anyway? Will it always ever be 1 given our current roadmap? It makes sense to me that we would vary `step`, but I'm not clear on to what effect `stride` would have.

alxmrs (commented):
Somewhat of a big change, these are my thoughts after a first round of feedback. I'd like to understand your approach better. Glad to see the performance results from the NVIDIA meeting.

src/ocean_emulators/constants.py
jder: Right, from the PR description:

> To avoid having multiple paths here, this also deletes the old "eager" data loader. I'd also like to delete the inference data loader to reduce duplication caused by it but figured it could be a follow-up.

I didn't think it was worth it to support two different paths here.

jder: For benchmarks, I assume it will just stop updating the eager-related performance benchmarks? Seems ok to me.

src/ocean_emulators/utils/train.py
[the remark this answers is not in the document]
jder: Yeah, stride effectively reduces the total amount of data we're training on by that factor and changes our 5-day timestep to stride * 5 days. We don't have any plans to use it but it doesn't seem crazy to want it. Another option would be to just delete it, but we're going to want multiple datasets soon enough so not sure that saves us too much.

src/ocean_emulators/datasets.py
jder: Yes, that's right.

src/ocean_emulators/datasets.py
jder: There are other ways we could structure it, but the core issue is a need to do some stuff after the (torch) data loader produces the item. (Because it is after it produces the item that we are in the GPU-attached process.) What other structure did you have in mind?

src/ocean_emulators/utils/data.py
jder: No, there's no github issue for deleting InferenceDataset.

src/ocean_emulators/constants.py
alxmrs: Yeah, that also seems fine to me.

src/ocean_emulators/datasets.py
alxmrs: Happy to keep this, it's simple enough. I mainly want to take the opportunity to understand this change better.

src/ocean_emulators/datasets.py
```diff
@@ -773,3 +633,66 @@ def load_variable_data(var: xr.Variable) -> None:
...
+    """
+
+    def __init__(
+        self,
+        dataloader: torch.utils.data.DataLoader[RawTrainData],
+        datasets: list[TorchTrainDataset],
+        device: torch.device,
+    ):
+        self._dataloader = dataloader
+        self._datasets = {dataset.id: dataset for dataset in datasets}
```
alxmrs, line 661: Nice, I like that this is done inside the class. Thanks, this is sweet.

src/ocean_emulators/utils/train.py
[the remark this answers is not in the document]
alxmrs: This assert is good for now in my mind. When we need strides (or when the batch changes s.t. we need to rethink the shape), we can deal with figuring this out then.

alxmrs (approved):
Thanks for the performance improvement, and thanks for answering my questions. This LGTM.

src/ocean_emulators/datasets.py
jder: If it helps, I started by just adding the `to_train_data` in the core train loop, but many things in tests etc assume they can grab the final normalized/structured data right out of a data loader, so that's why I ended up with this data loading "facade" that does the GPU stuff before handing off the data.

PR #426, opened 2025-10-17
Multiple, configurable perceiver implementations, including with flash attention.

alxmrs:
This PR extracts the Perceiver into its own config. This allows us to specify which perceiver implementation to adopt in either the encoder or decoder (just the encoder for now). We add a flash attention perceiver as the second option beyond a "naive" perceiver. Last, users can select "auto" as an implementation option, and the config will pick the best implementation given the runtime environment.


pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
   "cftime>=1.6.4.post1",
   "dacite>=1.9.2",
   "dask>=2025.2",
-  "einops>=0.8.1",
+  "einops",
```
adam-subel: I am not familiar with the choice in dependencies, but just wanted to note that this loosens a requirement and wanted to check that was intentional.

src/ocean_emulators/config.py
```diff
@@ -287,16 +295,61 @@ def build(
...
+                    output_dim=out_channels,
+                    latent_dim=self.latent_dim,
+                    use_flash_attn=True,
+                    weight_tie_layers=True,  # share weights of cross-attn blocks during latent iteration
+                    self_per_cross_attn=2,  # ratio of self-attention (latent, small) per cross-attn (input, big) blocks
+                )
+            case "naive":
+                perceiver = NaivePerceiver(
+                    num_freq_bands=4,
+                    max_freq=1.0,
```
adam-subel: I notice that this is a departure from the previous code you shared for review. You had changes max_freq to be dependent on the encoder patch size. This is dropped now. Is that deemed to be uninportant?

adam-subel: I suppose similar comments on num_latents. Also, maybe I am mixing the order of changes across these two commits.

adam-subel (approved):
Looks good. I left a couple comments I think maybe based more on my own confusion on how this interacts with the previous pull request you shared.

skypilot/train.sky.yaml
```diff
@@ -61,6 +61,9 @@ workdir: .
 setup: |
   set -ex
 
+  # Ensure CUDA dependencies are installed
+  uv sync --extra cuda
```
jder, line 65: Should probably mention this somewhere in the docs, too?

pyproject.toml
```diff
@@ -77,6 +77,10 @@ dev = [
   "types-tqdm>=4.67.0.20250516",
   "xarrayutils>=2.0.1",
 ]
+cuda = [
+  "flash-attn>=2.2.5,<3",
+  "flash-perceiver>=0.2",
+]
```
jder: 🐑 Not that it really matters right now, but I think these should probably be a set of [optional dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies) rather than a [development dependency group](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups) since presumably people that pip install our package would want to run on GPUs.

pyproject.toml
```diff
@@ -117,3 +121,6 @@ check_untyped_defs = true
 [[tool.mypy.overrides]]
 module = [ "cftime.*", "scipy.*", "fsspec.*", "perceiver_pytorch.*" ]
 follow_untyped_imports = true
+
+[tool.uv]
+no-build-isolation-package = [ 'flash-attn' ]
```
jder: Rather than turning of build isolation completely, I suggest we do what the [uv docs give as an example for this package](https://docs.astral.sh/uv/concepts/projects/config/#augmenting-build-dependencies)

```suggestion
[tool.uv.extra-build-dependencies]
flash-attn = [{ requirement = "torch", match-runtime = true }]

[tool.uv.extra-build-variables]
flash-attn = { FLASH_ATTENTION_SKIP_CUDA_BUILD = "TRUE" }
```

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -44,23 +39,9 @@ def __init__(
...
-            latent_dim=perceiver_latent_dim,
-            num_latents=perceiver_num_latents,
-            num_classes=out_channels,
-            weight_tie_layers=True,  # share weights of cross-attn blocks
-            self_per_cross_attn=2,  # ratio of self-attn (latent, small) and cross-attn (input, big) blocks
-            final_classifier_head=False,
-        )
-        self.project = nn.Linear(perceiver_latent_dim, out_channels)
+        self.perceiver = perceiver
+        latent_dim: int = getattr(self.perceiver, "latent_dim")
```
jder: This (and the get/set/hasattrs) seem like a smell to me. I would vote in favor of continuing to pass in this information separately, perhaps something like perceiver_output_shape? Alternatively you could include the final mean + projection into the passed nn.Module before passing it in here.

pyproject.toml
jder: Yeah, is this conflicting with the new perceiver impl somehow? If so would be good to look into why this requirement was here in the first place.

src/ocean_emulators/config.py
```diff
@@ -287,31 +295,95 @@ def build(
...
-    perceiver_num_latents: int = Field(
+    num_latents: int = Field(
         default=512,
         description="The number of latent vectors in the Perceiver. This is the `M` dimension for the Perceiver's `O(M*N)` complexity",
     )
 
+    def build(self, in_channels: int, patch_size: tuple[int, int]) -> nn.Module:
+        # This is not really a "frequency" but a maximum of the width appears to be reasonable from looking at the code.
+        max_freq = max(*patch_size)
+        match self.implementation:
```
jder: Does this need to be configurable? Might be preferable to instead use the flash version iff we're using a CUDA backend and otherwise the non-flash one? (And yell if we're using a cuda backend but the flash implementation is not installed?) Put another way: what are we going to check in for this configuration value?

src/ocean_emulators/config.py
```diff
@@ -287,31 +295,95 @@ def build(
...
+            case "flash":
+                if not FLASH_ENABLED:
+                    raise ValueError(
+                        'The "flash" perceiver implementation does not work in this environment.'
```
```suggestion
                        'Configuration requested the flash perceiver, but the `flash_perceiver` package could not be imported'
```

src/ocean_emulators/config.py
```diff
@@ -287,31 +295,95 @@ def build(
...
+            case "naive":
+                perceiver = NaivePerceiver(
+                    num_freq_bands=4,
+                    max_freq=max_freq,
+                    depth=self.depth,
+                    input_axis=2,  # Number of positional dims before token dim
+                    input_channels=in_channels,
+                    latent_dim=self.latent_dim,
+                    num_latents=self.num_latents,
+                    num_classes=self.latent_dim,  # This is ignored!
```
```suggestion
```

jder (changes requested):
Sorry for the delay in reviewing this!

src/ocean_emulators/models/modules/encoder.py
alxmrs: This is the thing I was going to bring up with you! At the point where we're including the mean and projection in a Module we write, why wouldn't we simply just configure each implementation to do just that? They both support the same thing that we're implementing IIUC -- except, the naive flash attention mechanism also throws in a layer norm.

alxmrs: I agree that the set/get attrs is less than ideal. I'll play around with it to see if there is another clean approach (will see what your suggestions look like in code).

src/ocean_emulators/config.py
alxmrs: I do think it should be configurable. While I have read each implementation and they look equivalent, the devil is in the (implementation) details. There may be a scientific reason why we would want to switch one implementation for another, outside of performance. Further, I have found yet another Perceiver implementation (that is well starred on GH), which we may want to experiment with later. Thus, I think this is worthy of a parameter that we explicitly control.

pyproject.toml
alxmrs: I'm happy to pin the requirement now that I loosened it; I did at one point need einops, I remember adding this dep.

skypilot/train.sky.yaml
alxmrs: Good shout.

src/ocean_emulators/models/modules/encoder.py
alxmrs: I now have an impl that is simpler, but it includes a LayerNorm instead of a simple linear projection. WDYT?

skypilot/viz.sky.yaml
```diff
@@ -62,6 +62,9 @@ workdir: .
 setup: |
   set -ex
 
+  # Ensure CUDA dependencies are installed
+  uv sync --extra cuda
```
alxmrs: @jder is CUDA needed in this case? I am now thinking: no.

jder: Good point, yeah. I have actually been running this on AWS CPU nodes so would be nice to skip CUDA stuff.

src/ocean_emulators/config.py
jder: OK, then I think we should have an `auto` implementation which we use by default which does what I said above: if we're using a cuda backend, use "flash" (and error if it's not enabled) or else use "naive"

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
   "cftime>=1.6.4.post1",
   "dacite>=1.9.2",
   "dask>=2025.2",
-  "einops>=0.8.1",
+  "einops>=0.7",
```
jder, line 30: Why did this change from 0.8.1 to 0.7? Looks like the original requirements.txt included 0.8.0.

src/ocean_emulators/models/modules/encoder.py
jder: The reason I turned off the default behavior was to remove the LayerNorm. I don't really see why we'd want a LayerNorm there but I also don't have good evidence it's harmful.

src/ocean_emulators/config.py
```diff
@@ -287,31 +295,85 @@ def build(
...
         default=512,
         description="The number of latent vectors in the Perceiver. This is the `M` dimension for the Perceiver's `O(M*N)` complexity",
     )
 
+    def build(
+        self, in_channels: int, out_channels: int, patch_size: tuple[int, int]
+    ) -> nn.Module:
+        # This is not really a "frequency" but a maximum of the width appears to be reasonable from looking at the code.
+        max_freq = max(*patch_size)
+        # TODO(alxmrs,jder): Why compute the mean? Is it better to directly project from the num_latents x latent_dim?
```
jder: Can you either clarify this is about the final output projection or rm the comment?

jder (approved):
Thanks, just a few follow-ups.

pyproject.toml
alxmrs: I did it because of a version conflict with flash-perceivers: 

```
 × No solution found when resolving dependencies for split (python_full_version >= '3.13' and sys_platform == 'linux'):
  ╰─▶ Because only flash-perceiver<=0.2.0 is available and flash-perceiver==0.2.0 depends on einops>=0.7.0,<0.8.0, we can conclude that flash-perceiver>=0.2.0 depends on einops>=0.7.0,<0.8.0.
      And because ocean-emulators[cuda] depends on flash-perceiver>=0.2, we can conclude that ocean-emulators[cuda] depends on einops>=0.7.0,<0.8.0.
      And because your project depends on einops>=0.8 and your project requires ocean-emulators[cuda], we can conclude that your project's requirements are unsatisfiable
```

jder: ok, skimming the release notes that seems ok. Might also be worth an upstream PR bumping to `0.7<=0.9` since I doubt 0.8 actually breaks anything.

src/ocean_emulators/config.py
```diff
@@ -287,31 +295,89 @@ def build(
...
+        self, in_channels: int, out_channels: int, patch_size: tuple[int, int]
+    ) -> nn.Module:
+        # This is not really a "frequency" but a maximum of the width appears to be reasonable from looking at the code.
+        max_freq = max(*patch_size)
+
+        # TODO(alxmrs,jder): Each implementation takes the mean of the num_latents dim to produce the final output_dim.
+        #  Why compute the mean? Is it better to directly project from the num_latents x latent_dim?
+        if (
+            self.implementation == "auto" and FLASH_ENABLED
+        ) or self.implementation == "flash":
```
jder, line 328: @alxmrs I think the behavior here is that if you forget to include `--extras cuda` it will silently fall back to the naive implementation. Can you please change this so that "auto" always means "flash" if we're using a torch cuda backend (ie so you get an error if you try to use it on GPU without installing the cuda extras)

PR #428, opened 2025-10-21
gradient detach

### LINKED ISSUE #225: Detach graph for autoregressive outputs
This would allow increased batch sizes

amogh-gulati:
closes #225

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -529,6 +529,10 @@ class FOMOConfig(BaseModelConfig):
     processor: UNetBackboneConfig = UNetBackboneConfig()
     # decoder will go here.
     embedding_dim: int = 128
+    gradient_detach_interval: int = Field(
+        default=0,
+        description="""Interval for detaching gradients in autoregressive training. `0` means no detaching.""",
+    )
 
     def build(
         self,
@@ -556,6 +560,7 @@ class FOMOConfig(BaseModelConfig):
             wet=wet,
             static_data=static_data,
             checkpointing=self.checkpointing,
+            gradient_detach_interval=self.gradient_detach_interval,
         )
 
 
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -466,6 +466,11 @@ class BaseModelConfig(BaseConfig, abc.ABC):
         and nonlinearities.""",
     )
 
+    gradient_detach_interval: int = Field(
+        default=0,
+        description="""Interval for detaching gradients in autoregressive training. `0` means no detaching.""",
+    )
+
     @abc.abstractmethod
     def build(
         self,
@@ -485,10 +490,6 @@ class SamudraConfig(BaseModelConfig):
     pos_channels: int = Field(
         default=0,
         description="""Number of channels used for a learned positional embedding""",
-    )
-    gradient_detach_interval: int = Field(
-        default=0,
-        description="""Interval for detaching gradients in autoregressive training. `0` means no detaching.""",
     )
 
     def build(
@@ -529,10 +530,6 @@ class FOMOConfig(BaseModelConfig):
     processor: UNetBackboneConfig = UNetBackboneConfig()
     # decoder will go here.
     embedding_dim: int = 128
-    gradient_detach_interval: int = Field(
-        default=0,
-        description="""Interval for detaching gradients in autoregressive training. `0` means no detaching.""",
-    )
 
     def build(
         self,
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -488,7 +488,7 @@ class SamudraConfig(BaseModelConfig):
     )
     gradient_detach_interval: int = Field(
         default=0,
-        description="""Interval for detaching gradients in autoregressive training. 0 means no detaching.""",
+        description="""Interval for detaching gradients in autoregressive training. `0` means no detaching.""",
     )
 
     def build(
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -475,7 +475,6 @@ class BaseModelConfig(BaseConfig, abc.ABC):
         wet: Grid,
         area_weights: Grid,
         static_data: xr.Dataset | None,
-        gradient_detach_interval: int = 0,
     ) -> BaseModel:
         pass
 
@@ -487,6 +486,10 @@ class SamudraConfig(BaseModelConfig):
         default=0,
         description="""Number of channels used for a learned positional embedding""",
     )
+    gradient_detach_interval: int = Field(
+        default=0,
+        description="""Interval for detaching gradients in autoregressive training. 0 means no detaching.""",
+    )
 
     def build(
         self,
@@ -496,7 +499,6 @@ class SamudraConfig(BaseModelConfig):
         wet: Grid,
         area_weights: Grid,
         static_data: xr.Dataset | None,
-        gradient_detach_interval: int = 0,
     ) -> Samudra:
         corrector = None
         if self.corrector is not None:
@@ -518,7 +520,7 @@ class SamudraConfig(BaseModelConfig):
             hist=hist,
             wet=wet,
             static_data=static_data,
-            gradient_detach_interval=gradient_detach_interval,
+            gradient_detach_interval=self.gradient_detach_interval,
         )
 
 
@@ -536,7 +538,6 @@ class FOMOConfig(BaseModelConfig):
         wet: Grid,
         area_weights: Grid,
         static_data: xr.Dataset | None,
-        gradient_detach_interval: int = 0,
     ) -> FOMO:
         return FOMO(
             in_channels=in_channels,
@@ -643,7 +644,6 @@ class TrainConfig(TopLevelConfig):
     finetune: bool = False
     resume_ckpt_path: str | None = None
     debug: bool = False
-    gradient_detach_interval: int = 0  # 0 means no detaching
     test_using_ema: bool = True
     ema_decay: float = 0.999
     faster_decay_at_start: bool = True
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -475,6 +475,7 @@ class BaseModelConfig(BaseConfig, abc.ABC):
         wet: Grid,
         area_weights: Grid,
         static_data: xr.Dataset | None,
+        gradient_detach_interval: int = 0,
     ) -> BaseModel:
         pass
 
@@ -495,6 +496,7 @@ class SamudraConfig(BaseModelConfig):
         wet: Grid,
         area_weights: Grid,
         static_data: xr.Dataset | None,
+        gradient_detach_interval: int = 0,
     ) -> Samudra:
         corrector = None
         if self.corrector is not None:
@@ -516,6 +518,7 @@ class SamudraConfig(BaseModelConfig):
             hist=hist,
             wet=wet,
             static_data=static_data,
+            gradient_detach_interval=gradient_detach_interval,
         )
 
 
@@ -533,6 +536,7 @@ class FOMOConfig(BaseModelConfig):
         wet: Grid,
         area_weights: Grid,
         static_data: xr.Dataset | None,
+        gradient_detach_interval: int = 0,
     ) -> FOMO:
         return FOMO(
             in_channels=in_channels,
@@ -639,6 +643,7 @@ class TrainConfig(TopLevelConfig):
     finetune: bool = False
     resume_ckpt_path: str | None = None
     debug: bool = False
+    gradient_detach_interval: int = 0  # 0 means no detaching
     test_using_ema: bool = True
     ema_decay: float = 0.999
     faster_decay_at_start: bool = True
--- a/src/ocean_emulators/models/base.py
+++ b/src/ocean_emulators/models/base.py
@@ -22,7 +22,7 @@ class BaseModel(torch.nn.Module):
         last_kernel_size,
         pad,
         static_data,
-        gradient_detach_interval: int = 0,
+        gradient_detach_interval: int,
     ) -> None:
         super().__init__()
         assert last_kernel_size % 2 != 0, "Cannot use even kernel sizes!"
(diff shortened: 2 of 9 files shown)


src/ocean_emulators/train.py
```diff
@@ -224,6 +224,7 @@ def __init__(self, cfg: TrainConfig) -> None:
             wet=self.wet.to(self.device),
             area_weights=self.area_weights,
             static_data=self.static_data,
+            gradient_detach_interval=cfg.gradient_detach_interval,
```
alxmrs: If it's in cfg, I think this var is accessible to the build method (as a class member) and doesn't need to be passed in this way.

src/ocean_emulators/config.py
```diff
@@ -407,6 +407,7 @@ def build(
         wet: Grid,
         area_weights: Grid,
         static_data: xr.Dataset | None,
+        gradient_detach_interval: int = 0,
```
alxmrs: Why not make it a class member instead of an argument to the build method?

alxmrs: Further:  I think this could benefit from a `Field` so we can add a docstring.

src/ocean_emulators/config.py
```diff
@@ -427,6 +428,7 @@ def build(
...
```
alxmrs: If we make it a class member, then I think we can omit them for these args.

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,164 @@
...
+
+from ocean_emulators.config import SamudraConfig, UNetBackboneConfig
+from ocean_emulators.constants import TensorMap
+from ocean_emulators.datasets import TrainData
+from ocean_emulators.utils.data import DataSource, Normalize
+from ocean_emulators.utils.multiton import MultitonScope
+
+
+@pytest.fixture
+def samudra_setup():
```
alxmrs: I like this fixture!

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,164 @@
...
+            train_data.insert(input_tensor, label_tensor)
+
+        return create_samudra_model, train_data
+
+
+class TestSamudraGradientDetaching:
+    def test_samudra_forward_pass_no_detaching(self, samudra_setup):
+        """Test Samudra forward pass without gradient detaching."""
+        create_model, train_data = samudra_setup
+        model = create_model(gradient_detach_interval=0)
```
alxmrs: This is a nice interface.

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,164 @@
...
+    grad_count = sum(1 for p in model.parameters() if p.grad is not None)
+    total_params = sum(1 for _ in model.parameters())
+    assert grad_count > 0, "Model should have gradients after backward pass"
+    assert grad_count == total_params, (
+        f"Expected all {total_params} parameters to have gradients, got {grad_count}"
+    )
+
+
+if __name__ == "__main__":
+    _run_standalone_test()
```
alxmrs: I think we should rely on test runners (pytest) instead of making this a runnable script. How much additional value does this standalone test add? Could we just make it a `test_`?

alxmrs (changes requested):
I have an alternative proposal for how to configure this change. But, the core of the change LGTM. Thank you, and I like the tests.

src/ocean_emulators/config.py
amogh-gulati: both samudra and the base model dont have access to the entire config, I think only the required args are being shared

src/ocean_emulators/config.py
amogh-gulati: done!

tests/test_gradient_detaching.py
amogh-gulati: I was using that as a scratch pad, a test runner makes sense, yes!

src/ocean_emulators/config.py
```diff
@@ -418,6 +418,10 @@ class SamudraConfig(BaseModelConfig):
         default=0,
         description="""Number of channels used for a learned positional embedding""",
     )
+    gradient_detach_interval: int = Field(
+        default=0,
+        description="""Interval for detaching gradients in autoregressive training. 0 means no detaching.""",
```
```suggestion
        description="""Interval for detaching gradients in autoregressive training. `0` means no detaching.""",
```

src/ocean_emulators/models/base.py
```diff
@@ -22,6 +22,7 @@ def __init__(
         last_kernel_size,
         pad,
         static_data,
+        gradient_detach_interval: int = 0,
```
alxmrs: Let's not set a default value here. Ideally, we want one source of truth for the config. That it might thrown an error when we forget to set this is a good thing. We used to do this and it caused hard to track down bugs.

src/ocean_emulators/models/samudra.py
```diff
@@ -22,6 +22,7 @@ def __init__(
         hist: int,
         wet: Grid,
         static_data: xr.Dataset | None,
+        gradient_detach_interval: int = 0,
```
alxmrs: See note about not setting default vals.

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,161 @@
...
+        h, w = 8, 8
+        coords = {
+            "lev": [0],
+            "lat": (["y"], torch.linspace(-90, 90, h).numpy()),
+            "lon": (["x"], torch.linspace(-180, 180, w).numpy()),
+        }
+        data = xr.Dataset(
+            {
+                "thetao": (["lev", "y", "x"], torch.randn(1, h, w).numpy()),
+                "hfds": (["y", "x"], torch.randn(h, w).numpy()),
```
alxmrs: 🐑 nit: these array creation functions could replace `torch` with `np` to be more succinct.

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,161 @@
...
+            },
+            coords=coords,
+        )
+        ones = xr.Dataset(
+            {
+                "thetao": (["lev", "y", "x"], torch.ones(1, h, w).numpy()),
+                "hfds": (["y", "x"], torch.ones(h, w).numpy()),
```
alxmrs: Same here

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,161 @@
...
+                dilation=[1, 1],
+                n_layers=[1, 1],
+            ),
+            pos_channels=0,
+        )
+
+        def create_samudra_model(gradient_detach_interval=0):
+            # Set the gradient_detach_interval on the config object
+            config.gradient_detach_interval = gradient_detach_interval
+            return config.build(
```
alxmrs: 🐑 instead of changing the gradient detach attr by side effect, I think it's safe to simply move the config instantiation code inside of this function. 
```suggestion

        def create_samudra_model(gradient_detach_interval=0):
            config = SamudraConfig(
                unet=UNetBackboneConfig(
                    ch_width=[4, 8],
                    dilation=[1, 1],
                    n_layers=[1, 1],
                ),
                pos_channels=0,
                gradient_detach_interval=gradient_detach_interval,
            )
            return config.build(
```

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,161 @@
...
+        train_data = TrainData(num_prognostic_channels=1)
+        for step in range(4):
+            input_tensor = torch.randn(1, 2, h, w, requires_grad=True)
+            label_tensor = torch.randn(1, 1, h, w)
+            train_data.insert(input_tensor, label_tensor)
+
+        return create_samudra_model, train_data
+
+
+class TestSamudraGradientDetaching:
```
alxmrs: A lot of these tests are the same, but just have gradient detaching turned on. This is great: it means we can make the gradient detaching options (i.e. `[0, 1]`) into a fixture and then just pass the into the tests. This combinatorial approach should cut the number of test methods in half (and will let you succinctly add more combinations later -- ofc, these could make test time longer.)

alxmrs (commented):
A small note about default vals, and then I've added feedback to your test code.

tests/test_gradient_detaching.py
alxmrs: If you want to be really suave, you could do this, too:

```suggestion

        def create_samudra_model(gradient_detach_interval=0):
            return SamudraConfig(
                unet=UNetBackboneConfig(
                    ch_width=[4, 8],
                    dilation=[1, 1],
                    n_layers=[1, 1],
                ),
                pos_channels=0,
                gradient_detach_interval=gradient_detach_interval,
            ).build(
```

src/ocean_emulators/config.py
amogh-gulati: gotcha

src/ocean_emulators/models/base.py
amogh-gulati: yup this makes sense!

src/ocean_emulators/config.py
```diff
@@ -487,6 +487,10 @@ class SamudraConfig(BaseModelConfig):
         default=0,
         description="""Number of channels used for a learned positional embedding""",
     )
+    gradient_detach_interval: int = Field(
+        default=0,
+        description="""Interval for detaching gradients in autoregressive training. `0` means no detaching.""",
+    )
```
alxmrs: If it's going here and the FOMO model, then I think we could move this up to the BaseModelConfig -- I think it should live next to `checkpointing`, for example. This saves us from having to write the config twice.

src/ocean_emulators/models/base.py
```diff
@@ -22,6 +22,7 @@ def __init__(
         last_kernel_size,
         pad,
         static_data,
+        gradient_detach_interval: int,
```
alxmrs, line 25: Indeed, since this argument lives in `BaseModel`, it makes even more sense to make the config live in `BaseModelConfig`.

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,136 @@
...
+
+from ocean_emulators.config import SamudraConfig, UNetBackboneConfig
+from ocean_emulators.constants import TensorMap
+from ocean_emulators.datasets import TrainData
+from ocean_emulators.utils.data import DataSource, Normalize
+from ocean_emulators.utils.multiton import MultitonScope
+
+
+@pytest.fixture(params=[0, 1, 2])
+def gradient_detach_interval(request):
```
alxmrs, line 14: Are these quick tests? Do they finish fast? If so, then let's definitely test 0, 1, and 2. If they're slow, then I'd be hesitant to include all three cases. In this project, I've been guilty of promoting more test coverage through combinations like this at the cost of having tests take a long time. 

At a high level, this fixture looks great.

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,136 @@
...
+            return model, train_data
+
+    return _create_model_helper
+
+
+@pytest.fixture
+def samudra_setup(gradient_detach_interval, create_samudra_model):
+    """Parametrized fixture that uses the factory to create models."""
+    model, train_data = create_samudra_model(gradient_detach_interval)
+    return model, train_data, gradient_detach_interval
```
alxmrs: I appreciate what you could be trying to do here wrt removing repeated calls, but in my view, this is a bit too DRY. I think it's perfectly OK for each test to accept two fixtures (`create_samudra_model` and `gradient_detatch_interval`) and repeat the function invocation. Looking at the tests, it appears that we need to use too many of the intermediate values. That we accept arguments and immediately return them makes me think that we would be better off omitting this fixture.

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,136 @@
...
+from ocean_emulators.constants import TensorMap
+from ocean_emulators.datasets import TrainData
+from ocean_emulators.utils.data import DataSource, Normalize
+from ocean_emulators.utils.multiton import MultitonScope
+
+
+@pytest.fixture(params=[0, 1, 2])
+def gradient_detach_interval(request):
+    """Parametrized fixture for gradient detach intervals. `0` means no detaching."""
+    return request.param
```
alxmrs: 🐑 Looking at how this fixture is used, here's an idea: what if this returned tuples of the interval + the "name" of the interval, i.e. if the gradient is detached or not. It could look like this:
```suggestion
    return (i := request.param), "no detaching" if i == 0 else f"detaching every {i}"
```

I am not sure if this is a good fit, but I bet it would clean up some of the logic in your tests code.

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@pytest.fixture
+def samudra_setup(gradient_detach_interval, create_samudra_model):
+    """Parametrized fixture that uses the factory to create models."""
+    model, train_data = create_samudra_model(gradient_detach_interval)
+    return model, train_data, gradient_detach_interval
+
+
+class TestSamudraGradientDetaching:
```
alxmrs: 🐑  this is totally optional: What do you think about omitting the class? Most of the pytest code I've seen doesn't group things together in classes. Or, when it does, it's because a test file includes lots of tests. Since the tests now just include two functions with combinations, we could keep things really simple and just have the tests be two functions.

alxmrs: (totally a matter of taste, and I leave the decision up to you).

tests/test_gradient_detaching.py
```diff
@@ -0,0 +1,136 @@
...
+def gradient_detach_interval(request):
+    """Parametrized fixture for gradient detach intervals. `0` means no detaching."""
+    return request.param
+
+
+@pytest.fixture
+def create_samudra_model():
+    """Factory fixture for creating Samudra models with different gradient detach intervals."""
+
+    def _create_model_helper(gradient_detach_interval: int):
```
alxmrs, line 23: I like this use of currying.

alxmrs (commented):
A few more notes, but this is looking really sharp. Thanks for the change, and I'm happy we could discuss some conventions in this codebase that aren't really written down anywhere else. Tests are looking good! I love succinct combinatorial testing with pytest.

tests/test_gradient_detaching.py
amogh-gulati: this was helpful!

tests/test_gradient_detaching.py
amogh-gulati: all 6 of the tests only take 0.51s on my local setup

alxmrs (approved):
A nice, clean change. Excellent work, and thanks for your patience.

PR #436, opened 2025-10-29
Quiet many pyright warnings about generic types

jder:
Mini PR which makes these types not generic to quiet many pyright warnings. Previously, Grid, Prognostic, etc, etc were all generic over the type of the underlying array and so technically need to appear as `foo: Grid[T]` or whatever when they're used. This seems like unhelpful noise so instead I swapped to a duck-typed type.


jder:
Oh, this type doesn't represent "like an array" it represents "something which can be converted into an array" which is not the same thing. @alxmrs do you know of a type which represents "has the array interface"? I looked around and came up empty.

alxmrs:
Let me research this more, because this is a fun question. But for now, the two interfaces that are top-of-mind are via numpy: 
https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike

and then the official Python Array API; however, I haven't yet found a specific typing part of their docs: https://data-apis.org/array-api/2024.12/index.html

alxmrs:
Wait, isn't numpy's ArrayLike type similar to how `type` works in Python? 

> Oh, this type doesn't represent "like an array" it represents "something which can be converted into an array" which is not the same thing.

https://stackoverflow.com/questions/23197122/what-are-pythons-type-objects-exactly -- IIUC, the second main use of `type` is to create objects. And, in a weird sort of python way, isn't "something which can be converted into an array" the same as is "like an array"?

alxmrs:
The way I like of it, `int` is a data type, but also a function to convert a primitive into that data type, for example.

jder:
The problem is that "can be converted to an array" does not mean "has `.shape`, `.dtype`, etc" which is what the pre-commit failures are about. I am looking for a way to annotate the type, not a way to convert.`foo: int` does not mean "foo could be converted to an int" it means "foo is an int". The string `"12"` is not an int but can be converted to one with `int`.

alxmrs:
Ok! A solution just occurred to me: Let's use a new Protocol to define the Array spec. The API surface would be so small I think we can just vendor it. I have done a quick search and I can't find a default implementation either in the data api or numpy. I think that would address this tactical concern. WDYT?

alxmrs:
Searching further: `numpy.typing.NDArray` could be a good thing to try: https://numpy.org/doc/2.3/reference/typing.html#numpy.typing.NDArray

alxmrs:
There official Array API compatibility layer doesn't seem to have the specific protocol that we need: https://github.com/data-apis/array-api-compat/blob/35d333af3b7e718e2a147400db0b16a874ea42da/array_api_compat/common/_typing.py

jder:
Thanks. I don't think numpy.typing.NDArray is what we want either -- it is still generic over dtype and assumes its a numpy type. I tried changing to just torch Tensor and mypy seems to be happy so I suggest we just do that for now. (It looks like in practice we're assuming Tensor anyway right now so let's make it official and we can change our mind later if desired)

alxmrs:
Sounds reasonable to me! I enjoyed the discussion either way.

alxmrs (approved):
LGTM. Thanks Jesse.

PR #437, opened 2025-10-29
Position and scale encoding via Aurora implementation

alxmrs:
This PR adds MSFT Aurora as a dependency. It makes use of that model's formulation for position and scale encodings. These geometry-informed encodings are added right before passing the embeddings to the processor. Ideally, these help the processor better make sense of spatial relationships.


src/ocean_emulators/config.py
```diff
@@ -336,13 +330,15 @@ def build(
                 use_flash_attn=True,
                 weight_tie_layers=True,  # share weights of cross-attn blocks during latent iteration
                 self_per_cross_attn=2,  # ratio of self-attention (latent, small) per cross-attn (input, big) blocks
+                cross_rotary_emb_dim=0,  # Turn off rotary embeddings
```
alxmrs: It just occurs to me that we could include this feature _without_ turning these off -- it may provide a benefit.

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -62,6 +69,19 @@ def forward(self, x: Input) -> Float[torch.Tensor, "batch {self.embed_dim} h w"]
...
+            self.out_channels,  # aka "embed_dim"
+            self.lat,
+            self.lon,
+            self.patch_size,
+            pos_expansion=pos_expansion,
+            scale_expansion=scale_expansion,
+        )
+        pos_encoding = self.pos_embed(pos_encode[None, None, :].to(dtype=x.dtype))
+        scale_encoding = self.scale_embed(scale_encode[None, None, :].to(dtype=x.dtype))
+        x = x + pos_encoding.squeeze() + scale_encoding.squeeze()
```
jder: This gives spatial embeddings to the processor, not the encoder. I think giving them to the processor is a fine idea (since the processor currently assumes the gaussian grid can be treated uniformly which is surely a bad assumption), but wasn't the goal of this PR to provide information to the encoder about the spatial relationships in its input data?

jder: OK, I looked at the code more carefully and we already had positional embeddings as input to the perceiver which I think I had forgotten in the various iterations. So this seems like a different change than I was expecting but also seems fine.

src/ocean_emulators/models/modules/vendor/fourier.py
```diff
@@ -0,0 +1,174 @@
+"""Fourier positional encodings.
+
+Sources taken from MSFT's Auroral model:
+ - https://github.com/microsoft/aurora/blob/main/aurora/model/fourier.py
+ - https://github.com/microsoft/aurora/blob/main/aurora/area.py
+
+Copyright (c) Microsoft Corporation. Licensed under the MIT license.
+
+Modifications made by Open Athena AI Foundation contributors.
```
jder: These don't look like we've changed them significantly from the original sources; could we instead depend on the [pypi package](https://pypi.org/project/microsoft-aurora/)? (If not, can you leave a comment why not?)

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -28,6 +37,8 @@ def __init__(
         out_channels: int,
         patch_size: int | tuple[int, int],
         perceiver: nn.Module,
+        lat: torch.Tensor,
+        lon: torch.Tensor,
```
jder: Can you add either jaxtyping annotations or comments about the shape here, please?

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -62,6 +76,19 @@ def forward(self, x: Input) -> Float[torch.Tensor, "batch {self.embed_dim} h w"]
...
         # NB(alxmrs): This is includes a mean and LayerNorm before linear projection!
         x = self.perceiver(x)  # (B_H_W, PH, PW, V) -> (B_H_W, out_channels)
+
+        pos_encode, scale_encode = pos_scale_enc(
+            self.out_channels,  # aka "embed_dim"
+            self.lat,
+            self.lon,
+            self.patch_size,
+            pos_expansion=pos_expansion,
+            scale_expansion=scale_expansion,
```
jder, line 99: I think these pos_expansion and scale_expansion are not really very well tuned for our dataset (we'll never have patches the size of the earth, for example); maybe worth a comment here that we're just starting with Aurora's baseline?

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -62,6 +76,19 @@ def forward(self, x: Input) -> Float[torch.Tensor, "batch {self.embed_dim} h w"]
...
+        )
+        pos_encoding = self.pos_embed(pos_encode[None, None, :].to(dtype=x.dtype))
+        scale_encoding = self.scale_embed(scale_encode[None, None, :].to(dtype=x.dtype))
+        x = x + pos_encoding.squeeze() + scale_encoding.squeeze()
```
jder: I might be missing something but this looks broken to me. We're unsqueezing the encodings from `(H_W, out_channels)` to `(1, 1, H_W, out_channels)`, doing a linear layer to `(1, 1, H_W, out_channels)` and then squeezing away the two extra axes to `(H_W, out_channels)` and then trying to add it to `(B_H_W, out_channels)`. The squeeze/unsqueeze seems unnecessary and I think this would break with batch dimension > 1? I think to make this work you need to break the batch dimension out of x's `B_H_W` so we can broadcast across it. Or, alternatively we could keep x as `(B, H_W, ...)` all along assuming the perceiver implementation can handle that.

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -62,6 +76,19 @@ def forward(self, x: Input) -> Float[torch.Tensor, "batch {self.embed_dim} h w"]
         )
         # NB(alxmrs): This is includes a mean and LayerNorm before linear projection!
         x = self.perceiver(x)  # (B_H_W, PH, PW, V) -> (B_H_W, out_channels)
+
+        pos_encode, scale_encode = pos_scale_enc(
+            self.out_channels,  # aka "embed_dim"
```
jder, line 92: 🐑 Probably makes sense for this to be a separate parameter so we can control how much spatial information we supply. Then the linear layers can project from that size to the final out_channels size before final addition. If you don't want to do that now, I suggest we leave a comment that we might generalize in this way so it's not confusing why this is re-used for both this shape and the final shape.

src/ocean_emulators/train.py
```diff
@@ -224,6 +224,8 @@ def __init__(self, cfg: TrainConfig) -> None:
             wet=self.wet.to(self.device),
             area_weights=self.area_weights,
             static_data=self.static_data,
+            lat=torch.from_numpy(self.data.lat.values),
+            lon=torch.from_numpy(self.data.lon.values),
```
jder, line 229: Since I think this is the first time we're using the lat/lon data in a way which affects the model, can we confirm these are lat/lon _centers_ for the relevant datapoints? Maybe by inspecting the preprocessing steps?

jder (changes requested):
Thanks, generally makes sense but have a few things I think are worth resolving before merging.

src/ocean_emulators/models/modules/vendor/fourier.py
alxmrs: TIL there is a pypi package at all! Thanks, I'll do that :)

src/ocean_emulators/train.py
alxmrs: I just reviewed this with Adam yesterday. This step in our data processing ensures that we're working with cell centers: https://github.com/Open-Athena/Ocean_Emulator/blob/e48eb09cce36239b270dff6d9fa6a37617f056fb/notebooks/2025-06-16-data_preprocess.py#L268

src/ocean_emulators/models/modules/encoder.py
alxmrs: I like this idea, I'll consider tackling it later, see #451.

src/ocean_emulators/models/modules/encoder.py
alxmrs: Good shout, happy to comment for now and tune later.

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -28,6 +38,8 @@ def __init__(
         out_channels: int,
         patch_size: int | tuple[int, int],
         perceiver: nn.Module,
+        lat: Lat,
+        lon: Lon,
```
jder, line 42: 🐑 since lat and lon are usually dimensions, what do you think about making this a bit more explicit and calling them eg `cell_center_lats`? Or add some docs for these?

jder (approved):
Thanks!

jder:
Oh, can you also update the PR description (and so merge commit message) before merging?

src/ocean_emulators/models/modules/encoder.py
alxmrs: I'll add a doc comment. Good shout.

PR #438, opened 2025-10-29
Auto should always be `flash` if cuda is available.

### LINKED ISSUE #435: Make "auto" means "flash" if we're using a torch cuda backend
@alxmrs I think the behavior here is that if you forget to include `--extras cuda` it will silently fall back to the naive implementation. Can you please change this so that "auto" always means "flash" if we're using a torch cuda backend (ie so you get an error if you try to use it on GPU without installing the cuda extras)

_Originally posted by @jder in https://github.com/Open-Athena/Ocean_Emulator/pull/426#discussion_r2466912739_

alxmrs:
Fixes #435.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -328,7 +328,7 @@ class PerceiverConfig(BaseConfig):
         ) or self.implementation == "flash":
             if not FLASH_ENABLED:
                 raise ValueError(
-                    "`implementation==flash` but the flash attention dependencies were not installed! Please run `uv sync --extra cuda` and try again."
+                    "`implementation==flash` or flash was automatically chosen for `implementation==auto`, but the flash attention dependencies could not be imported. Please run `uv sync --extra cuda` or specify the `naive` attention implementation."
                 )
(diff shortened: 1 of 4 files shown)


src/ocean_emulators/config.py
```diff
@@ -324,8 +324,12 @@ def build(
         # TODO(alxmrs,jder): Each implementation takes the mean of the num_latents dim to produce the final output_dim.
         #  Why compute the mean? Is it better to directly project from the num_latents x latent_dim?
         if (
-            self.implementation == "auto" and FLASH_ENABLED
+            self.implementation == "auto" and torch.cuda.is_available()
         ) or self.implementation == "flash":
+            if not FLASH_ENABLED:
+                raise ValueError(
+                    "`implementation==flash` but the flash attention dependencies were not installed! Please run `uv sync --extra cuda` and try again."
```
```suggestion
                    "`implementation==flash` or flash was automatically chosen for `implementation==auto`, but the flash attention dependencies could not be imported. Please run `uv sync --extra cuda` or specify the `naive` attention implementation."
```

jder (approved):
Thanks!

PR #440, opened 2025-10-31
add LLC_mean_std calculation file

GHOpenonic:
This adds the mean and std calculation file for LLC data


LLC_mean_std.py
```diff
@@ -0,0 +1,156 @@
...
+                mean_da = da.mean(skipna=True, dim=['j_g', 'i', 'face']).persist()
+                std_da = ((da - mean_da)**2).mean(skipna=True, dim=['j_g', 'i', 'face'])**0.5 
+               
+            else: 
+                mean_da = da.mean(skipna=True, dim=['j', 'i', 'face']).persist()
+                std_da = ((da - mean_da)**2).mean(skipna=True, dim=['j', 'i', 'face'])**0.5 
+
+            # average over times
+            mean_da = mean_da.mean(dim='time_sampled')
+            std_da  = std_da.mean(dim='time_sampled')
```
jder: It looks like we're averaging over the stds of each timestep rather than averaging the variance and then sqrting afterwards. I'm not actually sure how big of a difference this makes or what is actually more principled to do here but I believe it is different from the previous preprocessing strategy.

GHOpenonic: that's correct. I'll do a quick run averaging across variance and sqrting after to compare outputs :)

jder: Thanks! @GHOpenonic would you mind updating this PR to delete the extra file at the top level directory? Then we can merge the one which is in notebooks/. 🙇

GHOpenonic: should be fixed! thanks for the reminder :)

jder (approved):

PR #447, opened 2025-11-05
Fix benchmarks

jder:
Mock data was float64; the previous pipeline must have converted which I must have dropped, so convert it early on.


alxmrs (approved):
Thanks, this LGTM.

PR #459, opened 2025-11-12
U/jder/quiet lock

jder:


jder:
Unfortunately this doesn't work https://github.com/orgs/community/discussions/12343

PR #460, opened 2025-11-13
Created a module to add 3d Cartesian coordinates to the input data.

### LINKED ISSUE #367: Create positional encoding scheme for location, area, relative time, and absolute time.
i.e. add geospatial positional encodings.

alxmrs:
Extracted a re-usable component from a recent experiment. Now, the config system includes a flag to append 3d cartesian coordinates to the input data before feeding it to each of our ML models. With #437, let's say this closes #367 for now (time encoding can come later).


src/ocean_emulators/models/modules/augment_input.py
```diff
@@ -0,0 +1,35 @@
...
+from ocean_emulators.constants import Lat, Lon
+
+
+class Add3dCoordinates(nn.Module):
+    """Add 3d Cartesian Coordinates on a unit sphere to the channel dimension.
+
+    This provides better pole handling than raw lat/lon coordinates.
+    """
+
+    def __init__(self, lat: Lat, lon: Lon):
```
jder, line 32: 🐑 I'm confused what the semantics of the `Lat` type are now. Most of our jaxtyping types are about shapes, but it seems like from the comment on `Lat` this is now about data. ie there's only one possible value which is of type `Lat`? (As opposed to `Grid` of which there are many values like `boundary_vars: Grid` or `prognostic_vars: Grid`?) 

It seems like it would be more clear if the *types* `Lat` and `Lon` are about shape and the variables `lat` and `lon` indicate by their name or an additional argument comment here what data is stored in that shaped vector.

src/ocean_emulators/models/modules/augment_input.py
```diff
@@ -0,0 +1,35 @@
+import torch
+from aurora.model.posencoding import lat_lon_meshgrid
+from jaxtyping import Float
+from torch import nn
+
+from ocean_emulators.constants import Lat, Lon
+
+
+class Add3dCoordinates(nn.Module):
```
jder: Maybe `Concat3dCoordinates`?

src/ocean_emulators/models/modules/augment_input.py
```diff
@@ -0,0 +1,35 @@
...
+        self.lat, self.lon = lat, lon
+
+    def forward(
+        self, fts: Float[torch.Tensor, "batch channel height width"]
+    ) -> Float[torch.Tensor, "batch channel+3 height width"]:
+        lat_lon_grid = lat_lon_meshgrid(self.lat, self.lon)  # [2, H, W]
+        lat_rad = torch.deg2rad(lat_lon_grid[0])  # [H, W]
+        lon_rad = torch.deg2rad(lat_lon_grid[1])  # [H, W]
+
+        # Compute Cartesian coordinates (naturally bounded in [-1, 1])
```
jder:
```suggestion
        # Compute 3D coordinates; earth center is (0, 0, 0)
        # (1, 0, 0) is at lat, lon = (0, 0)
        # (0, 1, 0) is at lat, lon = (0, 90)
        # (0, 0, 1) is at the north pole
```

src/ocean_emulators/models/modules/augment_input.py
```diff
@@ -0,0 +1,35 @@
...
+        x = torch.cos(lat_rad) * torch.cos(lon_rad)
+        y = torch.cos(lat_rad) * torch.sin(lon_rad)
+        z = torch.sin(lat_rad)
+
+        grid = torch.stack([x, y, z], dim=0)  # [3, H, W]
+        grid = grid.float().to(fts.device).unsqueeze(0).expand(fts.shape[0], -1, -1, -1)
```
jder: Should all of this except the last `to`/`expand` get precomputed in `__init__` or else cached on the device? Otherwise I think this is probably all happening on the CPU every forward pass.

src/ocean_emulators/models/modules/augment_input.py
```diff
@@ -0,0 +1,35 @@
...
+from jaxtyping import Float
+from torch import nn
+
+from ocean_emulators.constants import Lat, Lon
+
+
+class Add3dCoordinates(nn.Module):
+    """Add 3d Cartesian Coordinates on a unit sphere to the channel dimension.
+
+    This provides better pole handling than raw lat/lon coordinates.
```
jder: Is there a reference we can include for this? Should I be surprised these are not fourier encoded somehow?

jder (approved):
Thanks; all optional stuff except for I think we should avoid recomputing these every forward pass.

src/ocean_emulators/models/modules/augment_input.py
alxmrs: Great idea!

src/ocean_emulators/models/modules/augment_input.py
alxmrs: I see what you mean -- the placement of the comment indicates a statement of data rather than type. I do still think this is a type-level semantic. `Lat` is a vector of length `n_lat`, same for `Lon`. These types will remain the same while the length of either lat or lon or both (via `Grid`) get bigger or smaller. 

How about I move the comment to this module and other places where the center issue matters? And, while I don't want to rename them everywhere, I'm happy to rename them in a class like this one.

jder: Moving the comment sounds great, thanks.

alxmrs: Thanks for raising this. Now that I've documented it, it does indeed make more sense.

PR #461, opened 2025-11-13
Relax formatting checks for notebooks

jder:
If we're treating this directory as "these are things we ran" repository then it seems unnecessary to require formatting checks pass, so disabled them here. See https://github.com/Open-Athena/Ocean_Emulator/pull/440 for an example where this is happening.


alxmrs (approved):
LGTM, thanks for fixing the secrets whitespace, too.

PR #454, opened 2025-11-07
Minor doc fixes needed from running one degree eval and viz

jder:
These were changes needed to work with 1-degree data. If we're going to be hopping around between datasets a bunch we should probably make this better.

PR #462, opened 2025-11-14
Ensure zero bias is white-colored

jder:
Previously this plot would produce images like this:
SST_map_snapshot_t_597 (2)

Where zero bias was quite red-colored. We now ensure the colorbar is centered at zero:

SST_map_snapshot_t_597 (3)

The OHC plots look like they have a similar bug but I can't actually make it happen so I'm guessing it's not actually a problem due to the way the data is processed before plotting.

### DIFF
--- a/src/ocean_emulators/viz/core.py
+++ b/src/ocean_emulators/viz/core.py
@@ -2507,6 +2507,16 @@ class Viz:
         colormap = cm.cm.balance
         colormap.set_bad(color=(0.7, 0.7, 0.7, 0))
         sst_bias = sst_data - gt_sst_data
+        if vmin is None:
+            vmin = sst_bias.min().compute().item()
+        if vmax is None:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -265,7 +265,7 @@ To run a remote viz job with SkyPilot, please use the following command:
 
 ```shell
-uv run sky launch skypilot/eval.sky.yaml \
+uv run sky launch skypilot/viz.sky.yaml \
```
alxmrs: Nice, great correction.

alxmrs (approved):
Really nice to have sane plots. Thanks for this fix, LGTM.

jder:
oh, sorry, I didn't mean to include all those other fixes from #454 in this branch. I'll back those out and pull them over separately.

PR #463, opened 2025-11-14
Fix space leak due to local vars existing beyond a single loop iteration

jder:
Previously `single_step_data`, `input` and `label` all continued to exist throughout the next train step and kept that GPU-side data alive.

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -612,15 +612,7 @@ class Trainer:
             metric_logger.update(loss=loss_value_reduce.item())
             metric_logger.update(lr=lr)
 
-            if update := getattr(self.loss_fn, "update", None):
-                with torch.no_grad():
-                    # TODO(jder): could avoid a second forward pass here
-                    single_step_data = TrainData(data.num_prognostic_channels)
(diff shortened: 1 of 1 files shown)


jder:
Before: 
Before

After:
After

src/ocean_emulators/train.py
```diff
@@ -630,6 +622,17 @@ def train_one_epoch(self, epoch):
         logger.info(f"Aggregating train logs")
         return train_aggregator.get_logs()
 
+    def _call_loss_update(self, data: TrainData):
+        # This is a separate function to ensure locals are dropped immediately after use
```
alxmrs, line 626: Nice trick! I like this better than inserting `del` calls.

alxmrs (approved):
Very clever idea for the fix -- and thanks for finding the memory leak. I wouldn't be able to see this extra memory just looking at the code!

src/ocean_emulators/train.py
```diff
@@ -630,6 +622,17 @@ def train_one_epoch(self, epoch):
...
+        if update := getattr(self.loss_fn, "update", None):
```
ihodes, line 627: TIL the "walrus operator", lovely

PR #464, opened 2025-11-14
Refactoring dynamic loss to compose any loss method, not just MSE.

alxmrs:
We now use a builder function to create the loss function. Loss can be defined as a simple metric, or a more sophisticated loss mixed in with a metric, such as dynamic loss. This should allow us to combine MAE with dynamic loss, for example.

My hope with this change is to follow up with experimenting with this idea from the Cambdige/NVIDIA paper: https://ar5iv.labs.arxiv.org/html/2410.07472v1#S4.SS10. 

(TL;DR: I have reason to believe that dynamic MAE will perform better than dynamic MSE).

### DIFF
--- a/configs/samudra_om4/train.yaml
+++ b/configs/samudra_om4/train.yaml
@@ -9,8 +9,8 @@ batch_size: 4
 learning_rate: 0.0006
 scheduler: { type: cosine }
 loss:
-  metric: mse
   type: dynamic
+  metric: mse
 finetune: false
 resume_ckpt_path: null
 inference_epochs: []
--- a/configs/samudra_om4/train.yaml
+++ b/configs/samudra_om4/train.yaml
@@ -10,7 +10,7 @@ learning_rate: 0.0006
 scheduler: { type: cosine }
 loss:
   metric: mse
-  limit: false
+  type: dynamic
 finetune: false
 resume_ckpt_path: null
 inference_epochs: []
--- a/configs/samudra_om4/train.yaml
+++ b/configs/samudra_om4/train.yaml
@@ -8,7 +8,9 @@ epochs: 70
 batch_size: 4
 learning_rate: 0.0006
 scheduler: { type: cosine }
-loss: mse_dynamic_no_limit
+loss:
+  metric: mse
+  limit: false
 finetune: false
 resume_ckpt_path: null
 inference_epochs: []
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -1,13 +1,21 @@
 import abc
-from functools import cached_property
+from functools import cached_property, partial
 from pathlib import Path
-from typing import Annotated, Literal, Self, assert_never
+from typing import Annotated, Any, Literal, Self, assert_never, get_args
 
 import cftime
+import numpy as np
+import pydantic
 import torch
 import xarray as xr
 from perceiver_pytorch import Perceiver as NaivePerceiver
-from pydantic import Field, PlainSerializer, PlainValidator, WithJsonSchema
+from pydantic import (
+    BeforeValidator,
+    Field,
+    PlainSerializer,
+    PlainValidator,
+    WithJsonSchema,
+)
 from torch import nn
 from torch.nn import GELU
 
@@ -33,6 +41,16 @@ from ocean_emulators.models.modules.augment_input import Concat3dCoordinates
 from ocean_emulators.models.modules.blocks import ZonallyPeriodicBilinearUpsample
 from ocean_emulators.utils.data import DataContainer, DataSource, validate_data
 from ocean_emulators.utils.location import LocalLocation, Location, ResolvedLocation
+from ocean_emulators.utils.loss import (
+    DynamicLoss,
+    LossFn,
+    decomposed_mae,
+    decomposed_mse,
+    decomposed_mse_cos_weighted,
+    decomposed_mse_diff_weighted,
+    decomposed_mse_mae,
+    decomposed_mse_scaled,
+)
 from ocean_emulators.utils.profiler import Profiler
 from ocean_emulators.utils.schedule import SchedulerConfig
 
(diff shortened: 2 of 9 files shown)


src/ocean_emulators/config.py
```diff
@@ -670,6 +673,7 @@ class TrainConfig(TopLevelConfig):
     learning_rate: float = 2e-4
     scheduler: SchedulerConfig | None = None
     loss: LossType = "mse"
+    loss_extension: LossExtensionType = "off"
```
jder: It seems confusing to me that this makes 2 ways to ask for the dynamic loss now. I think we should either totally separate this from loss (ie remove `mse_dynamic` and break backwards compatibility) or remove `loss_extension` (and just use `loss: mae_dynamic`). 

🐑 If you keep it, don't like the name "extension" -- it seems pretty generic. If you want to separate the metric from the scaling, how about `loss_scaling`? 

Alternatively we could go all out and have a LossConfig class so we can support eg:

```yaml
loss:
  type: dynamic
  limit: false
  metric: mae
```

and do something like `Location`'s `string_to_unresolved` so that `loss: mae` still works. This might be worthwhile if @amogh-gulati's PR about ensemble losses will want a more configurable set of loss functions here but not sure if that's needed or not.

jder (changes requested):

src/ocean_emulators/config.py
amogh-gulati: It will be helpful to have dynamic loss, at least for CRPS!

src/ocean_emulators/config.py
```diff
@@ -648,14 +666,110 @@ def build(self, output_dir: Path, device: torch.device) -> Profiler:
...
-    "mse_mae",
-    "mse_dynamic",
-    "mse_dynamic_no_limit",
+]
+LossType = Literal["standard", "dynamic"]
+
+
+class LossConfig(pydantic.BaseModel):
+    metric: LossMetric = "mse"
+    type: LossType = "standard"
```
jder: My goal in suggesting a separate class and "type" member was to permit different types with different members. For example, the "dynamic" type would have a `limit` field, the "standard" type would not. (Like S3Location and LocalLocation)

This would then let us have eg a "gradient" type which includes how much to weigh between the gradients and the underlying metric loss without having a "gradient_weight" member here which does nothing most of the time.

src/ocean_emulators/config.py
```diff
@@ -648,14 +666,110 @@ def build(self, output_dir: Path, device: torch.device) -> Profiler:
...
+            f"  loss:\n"
+            f"    metric: mse\n"
+            f"    type: dynamic\n"
+            f"    limit: false"
+        )
+    return candidate
+
+
+Loss = Annotated[
+    Annotated[LossConfig, WithJsonSchema({"type": "string"})],
```
jder: I believe this will cause the JSON schema to reject the structured form (ie you will get red squiggles in the editor). Depending on how you resolve the above comment, you probably want either this to be `|`'d with another type (as it is in Location) or else you'll need to use a [fancier method](https://docs.pydantic.dev/latest/concepts/json_schema/#implementing-__get_pydantic_json_schema__) to extend rather than totally replacing the JSON schema.

jder (commented):
I think maybe I wasn't very clear about my suggestion. Happy to discuss live or else if this is turning into too much of a diversion I think it's fine to go back to the previous strategy where we have `loss` of `mse` or `mae` or whatever and a `loss_scaling` (or whatever name) of `none` or `dynamic`.

src/ocean_emulators/train.py
```diff
@@ -232,59 +222,21 @@ def __init__(self, cfg: TrainConfig) -> None:
...
+        stds = self.src.stds[self.prognostic_var_names]
+        scaling_residuals = None
+        if self.data_container.scaling_residuals is not None:
+            scaling_residuals = self.data_container.scaling_residuals[
+                self.prognostic_var_names
+            ]
+        self.loss_fn: LossFn = cfg.loss.build(
+            hist=cfg.data.hist,
+            wet=self.wet,
+            y_coord=self.data.y,
```
jder: (You might want to just delete these losses we have never tried rather than trying to fix whatever is going wrong here)

src/ocean_emulators/config.py
```diff
@@ -648,17 +659,69 @@ def build(self, output_dir: Path, device: torch.device) -> Profiler:
...
+        case "mse_diff_weighted":
+            loss_fn = partial(decomposed_mse_diff_weighted, wet=wet)
+        case "mse_cos_weighted":
+            area_weights = np.sqrt(np.cos(np.deg2rad(y_coord))).to_numpy()
+            area_weights = torch.from_numpy(area_weights).to(device=device)
+            loss_fn = partial(decomposed_mse_cos_weighted, wet=wet, cos=area_weights)
+        case _:
+            assert_never(metric)
+
+    if isinstance(loss_cfg, DynamicLossConfig):
```
alxmrs: I am on the fence if this should delegate a `build` method on top of `DynamicLossConfig`.

src/ocean_emulators/utils/loss.py
```diff
@@ -82,23 +93,25 @@ class MseDynamic:
     See: https://openathena.slack.com/archives/C08CYM42DT3/p1752275713570969
     """
 
-    N_WINDOW = 25
     """Rolling window size to average over. (~number of steps)"""
+    N_WINDOW = 25
```
jder: As I said [last time you asked about this](https://github.com/Open-Athena/Ocean_Emulator/pull/310#discussion_r2252375402), I believe the original order is the standard. Do you have a reference for this order?

src/ocean_emulators/utils/loss.py
```diff
@@ -82,23 +93,25 @@ class MseDynamic:
...
 
     def __init__(
         self,
-        wet: Grid,
+        loss_fn: LossFn,
```
jder, line 135: Nice, I like the cleanup of using device rather than passing wet around here.

src/ocean_emulators/config.py
```diff
@@ -648,17 +659,69 @@ def build(self, output_dir: Path, device: torch.device) -> Profiler:
...
+def build_loss_fn(
+    loss_cfg: Loss,
+    wet: Grid,
+    y_coord: xr.DataArray,
+    stds: xr.Dataset,
+    device: torch.device,
+) -> LossFn:
+    if isinstance(loss_cfg, str):
+        metric = loss_cfg
+    elif hasattr(loss_cfg, "metric"):
```
jder: 🐑 this is maybe just too python for me but given we know the loss is either a LossMetric or a DyanmicLossConfig I would probably do something like
```python
match loss_cfg:
  case DynamicLossConfig(metric=metric, limit=limit)
     metric = fn_for_metric(metric)
     return DynamicLoss(…)
  case metric:
     return fn_for_metric(metric)
    ...
```

and then move the later match into fn_for_metric

src/ocean_emulators/train.py
```diff
@@ -232,59 +222,14 @@ def __init__(self, cfg: TrainConfig) -> None:
...
-            case "mse_residual_scaled":
-                logger.info("Using decomposed mse loss with scaled residuals")
-                assert self.data_container.scaling_residuals is not None, (
-                    "With loss of 'mse_residual_scaled' you"
-                    " must supply a scaling_residuals_file"
-                )
-                scale = torch.from_numpy(
-                    (
-                        self.src.stds[self.prognostic_var_names]
-                        / self.data_container.scaling_residuals[
```
jder, line 261: Does this mean we can get rid of `scaling_residuals`?

src/ocean_emulators/config.py
jder: Made a suggestion up above. I think a build would also be fine.

src/ocean_emulators/config.py
```diff
@@ -648,17 +659,69 @@ def build(self, output_dir: Path, device: torch.device) -> Profiler:
...
+    limit: bool = Field(
+        description="For dynamic loss, should we limit the range of weighted loss by the variance. Default: off.",
+        default=False,
+    )
+
+
+Loss = LossMetric | DynamicLossConfig
+
+
+def build_loss_fn(
```
jder, line 659: 🐑 could move this into utils/loss

configs/samudra_om4/train.yaml
```diff
@@ -8,7 +8,9 @@ epochs: 70
 batch_size: 4
 learning_rate: 0.0002
 scheduler: { type: cosine }
-loss: mse_dynamic_no_limit
+loss:
+  metric: mse
+  type: dynamic
```
jder, line 12: 🐑 I might put type first since you need to know what type it is to understand the rest of the members

jder (approved):
Thanks, all optional stuff from me. Looks great!

src/ocean_emulators/utils/loss.py
alxmrs: Sorry, I think I missed your comment the last time. Thanks for pointing that out (so weird, right?).

src/ocean_emulators/config.py
alxmrs: super slick! I didn't realize that case matching unpacked like that.

src/ocean_emulators/train.py
alxmrs: hmm, maybe? I'll double check.

src/ocean_emulators/config.py
alxmrs: I switched back and forth on this, but I kind of like that the build methods are all in the config file.

configs/samudra_om4/train.yaml
alxmrs: Very reasonable.

src/ocean_emulators/train.py
alxmrs: Yes! we can! 🎉

src/ocean_emulators/utils/loss.py
jder: Yes, very weird!

PR #468, opened 2025-11-17
Prevent dimension mismatch by squeezing tensors

mkeutgen:
Squeeze target and gen tensors to prevent dimension mismatch.

### DIFF
--- a/src/ocean_emulators/aggregator/inference/reduced.py
+++ b/src/ocean_emulators/aggregator/inference/reduced.py
@@ -140,9 +140,14 @@ class AreaWeightedReducedMetric:
                 "target and gen must have the same shape, got "
                 f"{target.shape} and {gen.shape}"
             )
+        # Note(mkeutgen):  this prevents a dimension mismatch error
+        target = target.squeeze()
+        gen = gen.squeeze()
+        
         new_value = (
             self._compute_metric(target=target, gen=gen).mean(dim=0).to(self._device)
         )
+        
         if self._total is None:
             self._total = torch.zeros(
                 [self._n_timesteps], dtype=new_value.dtype, device=self._device
--- a/src/ocean_emulators/aggregator/inference/reduced.py
+++ b/src/ocean_emulators/aggregator/inference/reduced.py
@@ -143,11 +143,11 @@ class AreaWeightedReducedMetric:
         # Note(mkeutgen):  this prevents a dimension mismatch error
         target = target.squeeze()
         gen = gen.squeeze()
-        
+
         new_value = (
             self._compute_metric(target=target, gen=gen).mean(dim=0).to(self._device)
         )
-        
+
         if self._total is None:
             self._total = torch.zeros(
                 [self._n_timesteps], dtype=new_value.dtype, device=self._device


src/ocean_emulators/aggregator/inference/reduced.py
```diff
@@ -140,9 +140,14 @@ def record(self, target: torch.Tensor, gen: torch.Tensor, i_time_start: int):
                 "target and gen must have the same shape, got "
                 f"{target.shape} and {gen.shape}"
             )
+        # Note(mkeutgen):  this prevents a dimension mismatch error
+        target = target.squeeze()
+        gen = gen.squeeze()
```
jder, line 145: Thanks for the PR! Can you share more details about what error this fixes?

The comment on this method says that both of these tensors should have shape `[batch, time, height, width]` and then the code below assumes that; are we getting an extra dimension here somehow?

jder: (cc @alxmrs since I know you were looking at this too)

mkeutgen:
Hi
This was the error traceback I was getting:

  File
"/scratch/cimes/maximek/INMOS/Ocean_Emulator/src/ocean_emulators/aggregator/inference/main.py",
line 131, in record_batch
    aggregator.record_batch(
  File
"/scratch/cimes/maximek/envs/ocean-emulator/lib/python3.11/site-packages/torch/utils/_contextlib.py",
line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File
"/scratch/cimes/maximek/INMOS/Ocean_Emulator/src/ocean_emulators/aggregator/inference/reduced.py",
line 270, in record_batch
    variable_metrics[metric][name].record(
  File
"/scratch/cimes/maximek/INMOS/Ocean_Emulator/src/ocean_emulators/aggregator/inference/reduced.py",
line 151, in record
    self._total[time_slice] += new_value
RuntimeError: The size of tensor a (0) must match the size of tensor b (6)
at non-singleton dimension 0

Alex figured this was a dimension mismatch error and suggested this fix.
Happy to provide more information if needed!

src/ocean_emulators/aggregator/inference/reduced.py
alxmrs: From our pair debugging session: Maxime is running the train and eval pipeline on different source data than we are (I'm not sure the name of the dataset). Training ran fine after changing some constants! But, there was an error around here in the eval script -- namely, that the `mean()` method below summed along dim 0, which was a batch dimension of 1. 

After digging around a bit, we found a setting up the stack called `long_rollouts` that are hard-coded as True. It seems that for this dataset, we'd like long rollouts to be false. The difference, more or less, is between two einsum operations, one of which unpacks the batch dimension in a way that is very suspect.

jder:
Ah, got it. Yes, the long_rollout thing is a bit surprising for sure (basically, BaseModel.inference returns a set of results which have a first "batch" dimension which is actually a time dimension, which we then pass along in various places that has to handle it differently). I don't think I understand how/why this would depend on the dataset, and I'm a bit worried that a squeeze here might be papering over a deeper issue (eg the time range is unexpectedly empty or out of range). @mkeutgen would you be willing to share your code with me so I can take a quick look and see if there might be either a more robust fix we can make here or else suggest a tweak to how this code is being used?

alxmrs:
Thanks, Jesse. You're more familiar with the eval side of the codebase than I am right now.

mkeutgen:
Of course, I just invited you to my fork of Ocean_Emulator, the changes
I've made to Ocean_Emulator are only related to constants.py and config.py
and the .yaml config (I had to change the calendar class from JulianDate to
NoLeap), I am also on M2Lines slack if you would like to discuss there

jder:
I am confused by this error; gonna grab time with Maxime to try and debug live.

PR #469, opened 2025-11-19
Updated `add_3d_coordinates` feature to preserve backwards compatibility with previous checkpoints.

alxmrs:
When the `nn.Module` is `nn.Identity`, we now omit adding it to the modules list. This allows for backwards compatibility in the Samudra model. Since we have so few checkpoints in FOMO, I'm not applying this pattern there.


amogh-gulati (approved):
looks good!

jder (approved):
Thanks!

PR #480, opened 2025-11-24
Memory snapshots are Pickle files, not JSON

jder:
I don't know how I managed to get this wrong and it doesn't seem to confuse the profiling tool, but this makes me feel better.


alxmrs (approved):
haha, thanks for fixing this.

PR #482, opened 2025-11-25
Minor changes to our current strong baseline config.

alxmrs:
- Renamed the config to `v2`.
- Turned on 3d coordinates -- these should lead to loss improvement.


alxmrs:
Let me run an experiment before I turn this on; closing.

PR #483, opened 2025-11-26
Clean up our config a bit

jder:
There are a number of small changes in here to clear up what folks should try to run and how. 

* I deleted all configs that (AFAIK) no one has used in > 6 months.
* I grouped train, eval, viz, and model configs for various efforts into folders (samudra_om4_v0, samudra_om4_vnext, fomo_om4, and test for our test configs). Note these are not named after resolutions anymore since they are effectively independent of resolution. 
  * If we end up wanting per-resolution stuff here like batch size or checkpointing strategy differences, we can make eg `samudra_om4_highres_vnext/*.yaml`.
* I made all the naming of wandb groups, experiments, etc, consistent across these various configs. 
* I pulled out common model definitions from the above.
* I fixed a small bug which prevented all YAML schemas from being written out.
* I added an error if you try to eval to the same output zarr file more than once.
* Updated various docs to reflect all of the above and added more hints about experiment names and file locations.


src/ocean_emulators/config_schema.py
```diff
@@ -37,6 +38,10 @@ def get_pydantic_models(
         field_type = field.annotation
         if isinstance(field_type, type) and issubclass(field_type, pydantic.BaseModel):
             get_pydantic_models(field_type, seen)
+        elif isinstance(field_type, UnionType):
+            for type_ in field_type.__args__:
```
alxmrs: 🐑 typing has a `get_args` that is useful for checking membership in literals.

alxmrs (approved):
Thanks for deleting old configs and I agree, this new folder and import system for configs is very clean. My main point of feedback is I am not so sure about calling our new strong baseline model "vnext". What do you think about calling the original Samudra v1 and the new model v2? I am also happy with v0 and v1, but I could see how that would be confusing to many audiences. 

This is somewhat of a marketing question, certainly one we can kick down the road.

jder:
> My main point of feedback is I am not so sure about calling our new strong baseline model "vnext".

What do you not like about it? I was searching for a name which indicated "in progress" as opposed to "frozen" (v0).

src/ocean_emulators/config_schema.py
jder: Ah, thank you! I was so confused why there was nothing on UnionType itself.

alxmrs:
That's an interesting naming problem. I don't think of v# as frozen, but I appreciate it being taken that way. 

"vnext" reminds me of the "new bridge" -- maybe I've just seen too many really old products call "next generation".

Another concern: will vnext be the final release name? Any idea what we might call the second model release in the paper? Maybe we should ask the lab chat.

jder:
> Another concern: will vnext be the final release name? Any idea what we might call the second model release in the paper? Maybe we should ask the lab chat.

I would assume we'd decide it's v1 when we're ready to "release" it. I could also just drop the suffix altogether and we can decide when to freeze "samudra_om4" into "samudra_om4_v1" or whatever. Does that seem better to you?

jder:
(I'm just going to go with that)

alxmrs:
Sorry I was late to respond! Yeah, the naming decision is very reversible. Thanks for merging for now.

PR #484, opened 2025-11-26
Log data attributes with wandb run

### LINKED ISSUE #473: Update training launch script to ensure that we log the physical location of where data came from in our experiment tracker (i.e. wandb).
I've hard a hard time tracking in the logs or config system where data came from. Since we rclone it and pass that into our config, the data source location is lost from the perspective of the training command run.

jder:
Logging the original S3 path is a bit tricky since we copy the data down (this will also probably get better when/if we stop using skypilot) but logging these attributes seems like a reasonable proxy. 

Closes #473

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -311,7 +311,10 @@ class Trainer:
 
         # Set up wandb run
         self.wandb_id, self.wandb_name = self.wandb_logger.setup_run(
-            cfg.resume_ckpt_path, cfg, finetune=cfg.finetune
+            cfg.resume_ckpt_path,
+            cfg,
+            data_container=self.data_container,
+            finetune=cfg.finetune,
         )
 
         self.num_batches_seen = 0
--- a/src/ocean_emulators/utils/wandb.py
+++ b/src/ocean_emulators/utils/wandb.py
@@ -20,6 +20,7 @@ MetricsDict = dict[str, float | torch.Tensor | WBValue]
 
 if TYPE_CHECKING:
     from ocean_emulators.config import AnyTopLevelConfig
+    from ocean_emulators.utils.data import DataContainer
 
 
 class WandBLogger(Multiton):
@@ -32,10 +33,17 @@ class WandBLogger(Multiton):
     def enabled(self):
         return self._enabled
 
+    def _make_config(self, cfg: "AnyTopLevelConfig", data_container: "DataContainer"):
+        return {
+            "config": cfg.model_dump(),
+            "data/attrs": data_container.source.data.attrs,
+        }
+
     def setup_run(
         self,
         checkpoint_path: str | None,
         cfg: "AnyTopLevelConfig",
+        data_container: "DataContainer",
         finetune: bool = False,
     ):
         """Set up a wandb run, either resuming from checkpoint or creating new run.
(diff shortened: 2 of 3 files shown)


src/ocean_emulators/utils/wandb.py
```diff
@@ -72,31 +81,27 @@ def setup_run(
...
-            cfg.experiment.name
-            if hasattr(cfg.experiment, "name")
-            else ".LOCAL" + "//" + cfg.experiment.name
-        )
-
+        wandb_name = cfg.experiment.name
         if self._enabled:
             self.init(
-                config=cfg.model_dump(),
+                config=self._make_config(cfg, data_container),
```
jder, line 104: This is arguably not "config" but it's also not "summary metrics/outputs" which are the two options. I think wandb wants you to use artifacts but then we need a wandb run for each data prep run (which isn't a bad idea, but not sure we want to go back and do that now)

src/ocean_emulators/utils/wandb.py
```diff
@@ -72,31 +81,27 @@ def setup_run(
...
             cfg: Configuration object
+            data_container: Data container to log attributes of
         Returns:
             tuple: (None, generated_name) for new run
         """
-        wandb_name = (
-            cfg.experiment.name
-            if hasattr(cfg.experiment, "name")
-            else ".LOCAL" + "//" + cfg.experiment.name
-        )
```
jder, line 95: This code made no sense to me, I think it must have gotten mangled by an earlier refactor

src/ocean_emulators/utils/wandb.py
```diff
@@ -32,27 +33,35 @@ def _initialize(self):
     def enabled(self):
         return self._enabled
 
+    def _make_config(self, cfg: "AnyTopLevelConfig", data_container: "DataContainer"):
+        return {
+            "config": cfg.model_dump(),
+            "data/attrs": data_container.source.data.attrs,
```
alxmrs, line 39: Oh, nice.

src/ocean_emulators/utils/wandb.py
alxmrs: Yeah, happy with this KISS change.

src/ocean_emulators/utils/wandb.py
alxmrs: I like the tradeoff you took here.

src/ocean_emulators/train.py
```diff
@@ -311,7 +311,10 @@ def __init__(self, cfg: TrainConfig) -> None:
 
         # Set up wandb run
         self.wandb_id, self.wandb_name = self.wandb_logger.setup_run(
-            cfg.resume_ckpt_path, cfg, finetune=cfg.finetune
+            cfg.resume_ckpt_path,
+            cfg,
+            data_container=self.data_container,
+            finetune=cfg.finetune,
```
alxmrs, line 317: 🐑 nit: if we're already passing in `cfg` and modifying `setup_run`, how about we parse `cfg.finetune` within the method instead of via argument?

alxmrs: And, does it make sense to do this for `resume_ckpt_path` too? I could go either way.

alxmrs (approved):
One main point of feedback. In general, this is definitely an imporvement and will make parsing the wandb logs a lot better. Thanks for doing this.

Shall we close #473 for now?

src/ocean_emulators/train.py
jder: I don't think this is the right direction, for two reasons:

* This is used in eval as well, and we'd have to check the type or presence of these attributes in setup_run in a way which kinda seems sad
* I'd actually rather move the other way and do the model_dumping out here so that the wandb logger doesn't need to know about our config structure (ie moving closer towards a `build()`-like structure for it)

PR #485, opened 2025-11-26
Remove unused method with a bad bug

jder:
As per https://openathena.slack.com/archives/C08CYM42DT3/p1764188280159869, we should be masking out land somehow before computing these statistics. But this method is unused (and already kinda unlikely to be what we want in the future) so I figured it makes sense to delete it rather than fix.


alxmrs (approved):
LGTM.

PR #487, opened 2025-11-27
Turn checkpointing=all by default in Samudra v1.

alxmrs:


jder:
So this is turning it on for the (frozen, non-wide) original samudra aka `samudra_om4_v1`. Is that what you're running? I was intending to turn it on for the (wide, dynamic-mse, etc) current WIP samudra aka `samudra_om4`.

alxmrs:
I... did not totally understand the naming conventions. Thanks for pointing this out. That the OG samudra runs on the halfdegree data with checkpointing turned on is cool but not what I intended. Putting on my bob ross hat for a second, I think this is a happy accident in that we can more fairly compare samudra v1 with the current samudra model.

In any case, I kicked off a parallel training run just now, and I'll close this issue. Thanks for pointing this out.

jder:
Sorry, my bad for not explaining where I landed after our chat about naming. I will add some docs too.

PR #489, opened 2025-11-29
Add Weighted Gradient-Enhanced MAE Loss Function

mkeutgen:
**Summary**
Implements decomposed_mae_gradient_weighted() - a configurable loss function that combines standard MAE with spatial gradient penalties to preserve mesoscale ocean features (eddies, fronts) during training.

**Problem**
Pure MAE loss produces smooth predictions that lose sharp oceanographic features. Models achieve good R² scores but generate unrealistic, overly-diffused fields that lack physical structure

**Solution**
Add tunable gradient penalty term:
`Loss = MAE(pred, target) + α × gradient_penalty(pred, target)
`
Where α (gradient_weight) controls the accuracy vs. sharpness tradeoff:

α = 0.1: Conservative balance (recommended starting point)
α = 0.25: More aggressive feature preservation
α = 0.5: Better sharpness

**Implementation Details**

Per-channel loss computation for multi-variable training
Spatial gradients computed via finite differences (∂/∂x, ∂/∂y)
Gradient penalties averaged and weighted before combination

**Benefits**

Preserves mesoscale eddy coherence during autoregressive rollouts
Maintains fronts and boundaries
Tunable via a single hyperparameter
Computationally efficient (simple finite differences)

### DIFF
--- a/src/ocean_emulators/aggregator/inference/reduced.py
+++ b/src/ocean_emulators/aggregator/inference/reduced.py
@@ -140,9 +140,14 @@ class AreaWeightedReducedMetric:
                 "target and gen must have the same shape, got "
                 f"{target.shape} and {gen.shape}"
             )
+        # Note(mkeutgen):  this prevents a dimension mismatch error
+        target = target.squeeze()
+        gen = gen.squeeze()
+        
         new_value = (
             self._compute_metric(target=target, gen=gen).mean(dim=0).to(self._device)
         )
+        
         if self._total is None:
             self._total = torch.zeros(
                 [self._n_timesteps], dtype=new_value.dtype, device=self._device
--- a/src/ocean_emulators/aggregator/inference/reduced.py
+++ b/src/ocean_emulators/aggregator/inference/reduced.py
@@ -143,11 +143,11 @@ class AreaWeightedReducedMetric:
         # Note(mkeutgen):  this prevents a dimension mismatch error
         target = target.squeeze()
         gen = gen.squeeze()
-        
+
         new_value = (
             self._compute_metric(target=target, gen=gen).mean(dim=0).to(self._device)
         )
-        
+
         if self._total is None:
             self._total = torch.zeros(
                 [self._n_timesteps], dtype=new_value.dtype, device=self._device
--- a/src/ocean_emulators/utils/loss.py
+++ b/src/ocean_emulators/utils/loss.py
@@ -72,6 +72,68 @@ def decomposed_mse_mae(
     combined = (mse + mae) / 2
     return combined.mean(dim=(0, 2, 3))
 
+def decomposed_mae_gradient_weighted(
+    pred: torch.Tensor, 
+    target: torch.Tensor, 
+    wet: torch.Tensor,
+    gradient_weight: float = 0.1
+) -> torch.Tensor:
+    """
+    MAE loss with WEIGHTED spatial gradient matching penalty.
+    
+    By controlling gradient_weight,
+    we can balance accuracy (MAE term) vs sharpness (gradient term).
+    
+    Loss = MAE(pred, target) + α * gradient_penalty(pred, target)
+    
+    where α = gradient_weight is a tunable hyperparameter.
+    
+    Recommended starting values:
+    - α = 0.05: Very conservative, prioritize accuracy
+    - α = 0.1:  Conservative, good balance 
+    - α = 0.25: Moderate, more sharpness 
+    - α = 0.5:  Aggressive sharpening
+    - α = 1.0:  Equal weighting 
+    
+    Args:
+        pred: Predicted tensor [batch, channels, height, width]
+        target: Target tensor [batch, channels, height, width]
+        wet: Wet mask [batch, channels, height, width]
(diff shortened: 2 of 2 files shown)


src/ocean_emulators/aggregator/inference/reduced.py
```diff
@@ -140,9 +140,14 @@ def record(self, target: torch.Tensor, gen: torch.Tensor, i_time_start: int):
                 "target and gen must have the same shape, got "
                 f"{target.shape} and {gen.shape}"
             )
+        # Note(mkeutgen):  this prevents a dimension mismatch error
+        target = target.squeeze()
```
alxmrs, line 144: Would you mind removing this commit from this branch/PR? It seems to be left over from a change unrelated to the primary purpose of this PR. 

The easiest way may be to make a new branch from main and then to cherry pick the latest commit in this branch. 

Another option could be to do an interactive rebase (there are some instructions in our contributing guide) and to drop the first two commits. Then, when the change and commit history looks good, I recommend doing a force push with lease (see the contrib guide).

alxmrs (commented):
A quick note! A full review will come soon as I find time.

src/ocean_emulators/utils/loss.py
```diff
@@ -72,6 +72,68 @@ def decomposed_mse_mae(
     combined = (mse + mae) / 2
     return combined.mean(dim=(0, 2, 3))
 
+def decomposed_mae_gradient_weighted(
+    pred: torch.Tensor, 
+    target: torch.Tensor, 
+    wet: torch.Tensor,
+    gradient_weight: float = 0.1
```
alxmrs, line 79: Let's omit adding a default kwarg, (except in test code). In my experience, it helps prevent subtle configuration bugs.

src/ocean_emulators/utils/loss.py
```diff
@@ -72,6 +72,68 @@ def decomposed_mse_mae(
...
+) -> torch.Tensor:
+    """
+    MAE loss with WEIGHTED spatial gradient matching penalty.
```
alxmrs, line 82: This docstring is excellent. Thanks for explaining what this loss method does. 

Is there a paper associated with this method? It would be nice to add a citation (even just a URL).

src/ocean_emulators/utils/loss.py
```diff
@@ -72,6 +72,68 @@ def decomposed_mse_mae(
     combined = (mse + mae) / 2
     return combined.mean(dim=(0, 2, 3))
 
+def decomposed_mae_gradient_weighted(
+    pred: torch.Tensor, 
+    target: torch.Tensor, 
+    wet: torch.Tensor,
```
alxmrs, line 78: Let's use the jaxtyping value here called Grid from our constants file.

src/ocean_emulators/utils/loss.py
```diff
@@ -72,6 +72,68 @@ def decomposed_mse_mae(
...
+    
+    Recommended starting values:
+    - α = 0.05: Very conservative, prioritize accuracy
+    - α = 0.1:  Conservative, good balance 
+    - α = 0.25: Moderate, more sharpness 
+    - α = 0.5:  Aggressive sharpening
+    - α = 1.0:  Equal weighting 
+    
+    Args:
+        pred: Predicted tensor [batch, channels, height, width]
```
alxmrs, line 99: Optional nit: let's end each line of the docstring with a full stop ".".

src/ocean_emulators/utils/loss.py
```diff
@@ -72,6 +72,68 @@ def decomposed_mse_mae(
...
+        target: Target tensor [batch, channels, height, width]
```
alxmrs, line 100: Optional: consider incorporating this into the type via jaxtyping.

src/ocean_emulators/utils/loss.py
```diff
@@ -72,6 +72,68 @@ def decomposed_mse_mae(
...
+        Loss per channel [channels]
+    """
+    pred = pred * wet
+    target = target * wet
+    
+    # MAE term (main accuracy objective)
+    mae_loss = F.l1_loss(pred, target, reduction="none")
+    mae_per_channel = mae_loss.mean(dim=(0, 2, 3))
+    
+    # Gradient penalty: Match spatial gradients
```
alxmrs, line 114: Can you help me understand, why do we do this differencing across height and width?

alxmrs: To me, this reminds me of some kind of simple edge detector, see

https://cseweb.ucsd.edu/classes/sp16/cse152-a/lec7.pdf

alxmrs (commented):
More notes and a question.

src/ocean_emulators/utils/loss.py
alxmrs: Sorry, I read your README just now. This makes more sense!

mkeutgen: Hi Alex thanks for your very useful feedback, I've created a new clean PR I hope this new PR is better :) I still need to find a paper about gradient loss, I've not found one yet

PR #490, opened 2025-11-29
Update train.py

mkeutgen:
Updated train.py for compatibility with gradient_loss_weighted


alxmrs:
Hey @mkeutgen, would you mind TAL at the build? It seems like there is an important error.

alxmrs:
Can we integrate this change with #489? They seem part of the same concern.

PR #491, opened 2025-12-01
Spatial grad loss

mkeutgen:
**Summary**
Implements decomposed_mae_gradient_weighted() - a configurable loss function that combines standard MAE with spatial gradient penalties to preserve mesoscale ocean features (eddies, fronts) during training.

**Problem**
Pure MAE loss produces smooth predictions that lose sharp oceanographic features. Models achieve good R² scores but generate unrealistic, overly-diffused fields that lack physical structure

**Solution**
Add tunable gradient penalty term:
Loss = MAE(pred, target) + α × gradient_penalty(pred, target) 
Where α (gradient_weight) controls the accuracy vs. sharpness tradeoff:

α = 0.1: Conservative balance (recommended starting point)
α = 0.25: More aggressive feature preservation
α = 0.5: Better sharpness

**Implementation Details**

Per-channel loss computation for multi-variable training
Spatial gradients computed via finite differences (∂/∂x, ∂/∂y)
Gradient penalties averaged and weighted before combination

**Benefits**

Preserves mesoscale eddy coherence during autoregressive rollouts
Maintains fronts and boundaries
Tunable via a single hyperparameter
Computationally efficient (simple finite differences)

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -659,7 +659,7 @@ class GradientLossConfig(pydantic.BaseModel):
     # at the moment this metric is only used for the non-gradient loss
     # (and would take a bit of refactoring to make it work for the gradient loss too)
     # so we fix it to MAE for now until it's clear we what flexibility is needed here.
-    # TODO: support other metrics for the gradient loss
+    # TODO(#497): support other metrics for the gradient loss
     metric: Literal["mae"] = "mae"
     alpha: float = Field(
         description="Scaling factor for the gradient penalty term (alpha in the gradient-weighted loss).",
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -656,7 +656,11 @@ class DynamicLossConfig(pydantic.BaseModel):
 
 class GradientLossConfig(pydantic.BaseModel):
     type: Literal["gradient"] = "gradient"
-    metric: LossMetric = "mae"
+    # at the moment this metric is only used for the non-gradient loss
+    # (and would take a bit of refactoring to make it work for the gradient loss too)
+    # so we fix it to MAE for now until it's clear we what flexibility is needed here.
+    # TODO: support other metrics for the gradient loss
+    metric: Literal["mae"] = "mae"
     alpha: float = Field(
         description="Scaling factor for the gradient penalty term (alpha in the gradient-weighted loss).",
         default=0.1,
--- a/src/ocean_emulators/utils/loss.py
+++ b/src/ocean_emulators/utils/loss.py
@@ -72,6 +72,64 @@ def decomposed_mse_mae(
     combined = (mse + mae) / 2
     return combined.mean(dim=(0, 2, 3))
 
+def decomposed_mae_gradient_weighted(
+    pred: torch.Tensor, 
+    target: torch.Tensor, 
+    wet: torch.Tensor,
+    gradient_weight: float
+) -> torch.Tensor:
+    """MAE loss with WEIGHTED spatial gradient matching penalty.
+    
+    By controlling gradient_weight, we can balance accuracy (MAE term) vs sharpness (gradient term).
+    
+    Loss = MAE(pred, target) + α * gradient_penalty(pred, target).
+    
+    where α = gradient_weight is a tunable hyperparameter.
+    
+    Recommended starting values:
+    - α = 0.05: Very conservative, prioritize accuracy
+    - α = 0.1:  Conservative, good balance 
+    - α = 0.25: Moderate, more sharpness 
+    - α = 0.5:  Aggressive sharpening
+    - α = 1.0:  Equal weighting 
+    
+    Args:
+        pred: Predicted tensor [batch, channels, height, width].
+        target: Target tensor [batch, channels, height, width]
+        wet: Wet mask [batch, channels, height, width]
+        gradient_weight: Scaling factor α for gradient penalty
+    
+    Returns:
+        Loss per channel [channels]
+    """
+    pred = pred * wet
+    target = target * wet
+
+    # MAE term (main accuracy objective)
+    mae_loss = F.l1_loss(pred, target, reduction="none")
+    mae_per_channel = mae_loss.mean(dim=(0, 2, 3))
+
+    # Gradient penalty: Match spatial gradients
+    pred_grad_y = pred[:, :, 1:, :] - pred[:, :, :-1, :]
+    pred_grad_x = pred[:, :, :, 1:] - pred[:, :, :, :-1]
+
+    target_grad_y = target[:, :, 1:, :] - target[:, :, :-1, :]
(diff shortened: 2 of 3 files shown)


src/ocean_emulators/utils/loss.py
```diff
@@ -72,6 +72,64 @@ def decomposed_mse_mae(
...
+    target_grad_y = target[:, :, 1:, :] - target[:, :, :-1, :]
+    target_grad_x = target[:, :, :, 1:] - target[:, :, :, :-1]
+
+    grad_loss_y = F.l1_loss(pred_grad_y, target_grad_y, reduction="none")
+    grad_loss_x = F.l1_loss(pred_grad_x, target_grad_x, reduction="none")
+
+    # Average gradient losses
+    grad_loss = (
+        F.pad(grad_loss_y, (0, 0, 0, 1), value=0).mean(dim=(0, 2, 3)) +
+        F.pad(grad_loss_x, (0, 1, 0, 0), value=0).mean(dim=(0, 2, 3))
```
jder: What's the purpose of padding before meaning?

src/ocean_emulators/utils/loss.py
```diff
@@ -72,6 +72,64 @@ def decomposed_mse_mae(
...
+    # MAE term (main accuracy objective)
+    mae_loss = F.l1_loss(pred, target, reduction="none")
+    mae_per_channel = mae_loss.mean(dim=(0, 2, 3))
+
+    # Gradient penalty: Match spatial gradients
+    pred_grad_y = pred[:, :, 1:, :] - pred[:, :, :-1, :]
+    pred_grad_x = pred[:, :, :, 1:] - pred[:, :, :, :-1]
+
+    target_grad_y = target[:, :, 1:, :] - target[:, :, :-1, :]
+    target_grad_x = target[:, :, :, 1:] - target[:, :, :, :-1]
```
jder: Not sure it matters too much, but should we theoretically be using circular padding for the x grads here rather than truncating? (For whole-globe case at least.)

jder (commented):
Thanks @mkeutgen! Just a couple questions from me. I think @alxmrs and I need to resolve some open questions in #464 before landing this. (Whatever changes come out of that discussion will be easy to add here.) Looking forward to trying this out!

src/ocean_emulators/utils/loss.py
mkeutgen: The padding restores the gradients to the full (lat, lon) grid so that both directions are averaged on the same spatial support. Without padding, grad_loss_y and grad_loss_x have shapes (H-1, W) and (H, W-1), which would weight the two directions differently when averaged.

alxmrs:
Hey @mkeutgen, how does this method compare with https://github.com/Open-Athena/Ocean_Emulator/blob/dc94a80ecdea6748dcd532a2912f95f33003552c/src/ocean_emulators/utils/loss.py#L18? 

Is there a greater generalization we could make, similar to #464?

src/ocean_emulators/utils/loss.py
alxmrs: would be cool to make use of a `roll` operator here. Good shout, Jesse. 

https://docs.pytorch.org/docs/stable/generated/torch.roll.html

mkeutgen: Yes absolutely! Sorry for the oversight, I'm training on a model with rectangular boundaries hence I didn't think of periodicity but for global case a roll operator would be convenient

mkeutgen:
> Hey @mkeutgen, how does this method compare with
> 
> https://github.com/Open-Athena/Ocean_Emulator/blob/dc94a80ecdea6748dcd532a2912f95f33003552c/src/ocean_emulators/utils/loss.py#L18
> ?
> 
> Is there a greater generalization we could make, similar to #464?

Hi Alex, am I right that this is basically a cross-channel MSE?

pred[:, 1:] - pred[:, :-1]
target[:, 1:] - target[:, :-1]


So it’s penalizing differences between adjacent channels. If the channels are vertical levels (temp_0, temp_1, …, temp_19), that makes sense because they’re implicitly ordered, it’s basically a vertical gradient loss. But when channels are different variables (temp, salt, u, v, …), that ordering isn’t meaningful anymore, so interpreting the penalty gets trickier.

The gradient loss I’m proposing is doing something different: it penalizes loss of spatial structure (in 2D) inside each channel. decomposed_mse_diff_weighted doesn’t touch that part at all. So in that sense the two losses aren’t redundant.

That said, there is a conceptual link: restricted to vertical channels, decomposed_mse_diff_weighted already behaves like a vertical derivative term, so the two approaches could probably be unified in a cleaner way.

Also: in practice, the first-order spatial gradient term seems necessary but not always sufficient to keep fine-scale eddies/fronts from diffusing away, which is why I’m experimenting with adding a second-order (Laplacian/curvature) penalty. I’ll test that this week.

alxmrs:
> The gradient loss I’m proposing is doing something different: it penalizes loss of spatial structure (in 2D) inside each channel. decomposed_mse_diff_weighted doesn’t touch that part at all. So in that sense the two losses aren’t redundant.

Thanks for your explanation -- that's clear. It didn't occur to me that the existing method diff'd across channels; you're reasoning makes a lot of sense. 

I may want to delete `decomposed_mse_diff_weighted` soon, since we haven't used it in quite some time.

jder:
@codex can you update this with the latest changes to loss functions from main? We'll probably make a new class analogous to the dynamic loss class but for gradient losses which takes a metric, and the alpha factor described above. Please also fix the pre-commit issue and update the x-padding according to the specified pad mode which is available on BaseModel.

jder:
@alxmrs I made some updates to accommodate your other PR and fix the couple little things above, can you review and ✅ if it looks good to you?

src/ocean_emulators/config.py
```diff
@@ -653,7 +654,21 @@ class DynamicLossConfig(pydantic.BaseModel):
     )
 
 
-Loss = LossMetric | DynamicLossConfig
+class GradientLossConfig(pydantic.BaseModel):
+    type: Literal["gradient"] = "gradient"
+    # at the moment this metric is only used for the non-gradient loss
+    # (and would take a bit of refactoring to make it work for the gradient loss too)
+    # so we fix it to MAE for now until it's clear we what flexibility is needed here.
+    # TODO: support other metrics for the gradient loss
```
```suggestion
    # TODO(#497): support other metrics for the gradient loss
```

src/ocean_emulators/config.py
```diff
@@ -653,7 +654,21 @@ class DynamicLossConfig(pydantic.BaseModel):
...
+    metric: Literal["mae"] = "mae"
+    alpha: float = Field(
+        description="Scaling factor for the gradient penalty term (alpha in the gradient-weighted loss).",
+        default=0.1,
+        ge=0.0,
```
alxmrs, line 667: TIL!

alxmrs (approved):
One minor nit. Otherwise, this LGTM; ship it!

jder:
Thanks again @mkeutgen; merging this and excited to give it a try!

PR #493, opened 2025-12-01
Readmes for the various model configs, add checkpointing for v1

jder:
Hopefully this (a) clarifies what is what and (b) means we can train both samudra models on 1/2- and 1-degree datasets the same way.


alxmrs (approved):
Thanks so much. These are really nice to have.

PR #498, opened 2025-12-03
Clean up the TrainData internals.

alxmrs:
A small refactoring PR. Made this along the way of thinking about how we might load multiple datasets.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -363,7 +363,7 @@ class TrainData:
         return len(self.example_by_step)
 
     def __iter__(self):
-        return range(len(self))
+        return iter(range(len(self)))
 
     def to(self, device: torch.device) -> None:
         for step in self:
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -321,58 +321,65 @@ class RawTrainData:
 
 
 class TrainData:
+    """A single batch of training data.
+
+    For the Input (i.e. Example[0]), the top `num_prognostic_channels` must be prognostic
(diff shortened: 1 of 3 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -321,59 +321,65 @@ def pin_memory(self):
...
 
     def to(self, device: torch.device) -> None:
-        for step in self.td_dict:
-            self.td_dict[step] = (
-                self.td_dict[step][0].to(device, non_blocking=True),
-                self.td_dict[step][1].to(device, non_blocking=True),
+        for step in self:
+            self.example_by_step[step] = (
+                self[step][0].to(device, non_blocking=True),
+                self[step][1].to(device, non_blocking=True),
```
alxmrs, line 375: I think this and the below fn could be made even simpler (probably just a list comprehension), but I'm wary of changing too much all at once.

jder:
@codex review

src/ocean_emulators/datasets.py
```diff
@@ -321,59 +321,65 @@ def pin_memory(self):
 
 
 class TrainData:
+    """A single batch of training data.
+
+    Constraint: The `Input` tensor is a combination of (flattened) prognostic variables (at all depth levels) and
+    boundary forcings. The top `num_prognostic_channels` number of channels must be prognostic variables whereas the
+    remaining bottom channels are boundary forcings.
```
jder, line 331: Maybe worth mentioning that step means autoregressive rollout step?

jder (approved):
Thanks, looks great!

src/ocean_emulators/datasets.py
alxmrs: Yeah, that's a good idea.

PR #499, opened 2025-12-04
Masking is same concern as the DataSource.

alxmrs:
Introducing a `Masks` dataclass the in `DataSource` which has extracted wet masks. It turns out, near everywhere that `DataSource` is used, we have to separately plumb the surface and default wetmask. Thinking about what it would take to load multiple datasets of different resolutions, it seems that ocean/land masks are actually part of the same concern in as the data source. My hope is that after this PR, we can introduce a separate, but related masking concept -- namely, a dataloader-time "nodata" mask that is only necessary when combining disparate datasets.

Along the way, this refactor also brings in data validation into the creation of the `DataSource`.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -157,7 +157,6 @@ class DataConfig(BaseConfig):
         data_root: ResolvedLocation,
         boundary_var_names: BoundaryVarNames,
         prognostic_var_names: PrognosticVarNames,
-        hist: int,
     ) -> DataContainer:
         loader_version = LoaderVersion(self.loader_version)
         use_dask = loader_version != LoaderVersion.OM4_TORCH
@@ -173,7 +172,7 @@ class DataConfig(BaseConfig):
             use_dask=use_dask,
         )
         source = validate_data(source, boundary_var_names, self.static_data_vars)
-        source = source.mask(prognostic_var_names, hist)
+        source = source.mask(prognostic_var_names, self.hist)
 
         if use_dask:
             # If we're already using dask, we don't need a second source
@@ -189,7 +188,7 @@ class DataConfig(BaseConfig):
             source_using_dask = validate_data(
                 source_using_dask, boundary_var_names, self.static_data_vars
             )
-            source_using_dask = source_using_dask.mask(prognostic_var_names, hist)
+            source_using_dask = source_using_dask.mask(prognostic_var_names, self.hist)
 
         static_data = (
             source.data[self.static_data_vars]
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -13,7 +13,14 @@ from torch import nn
 from torch.nn import GELU
 
 from ocean_emulators.config_base import BaseConfig, TopLevelConfig
-from ocean_emulators.constants import Grid, Lat, LoaderVersion, Lon, PrognosticVarNames
+from ocean_emulators.constants import (
+    BoundaryVarNames,
+    Grid,
+    Lat,
+    LoaderVersion,
+    Lon,
+    PrognosticVarNames,
+)
 from ocean_emulators.models import FOMO, Samudra
 from ocean_emulators.models.base import BaseModel
 from ocean_emulators.models.modules import (
@@ -149,6 +156,7 @@ class DataConfig(BaseConfig):
         self,
         data_root: ResolvedLocation,
         prognostic_var_names: PrognosticVarNames,
+        boundary_var_names: BoundaryVarNames,
     ) -> DataContainer:
         loader_version = LoaderVersion(self.loader_version)
         use_dask = loader_version != LoaderVersion.OM4_TORCH
@@ -162,6 +170,7 @@ class DataConfig(BaseConfig):
             means_location=means_location,
             stds_location=stds_location,
             prognostic_var_names=prognostic_var_names,
+            boundary_var_names=boundary_var_names,
             static_data_vars=self.static_data_vars,
             use_dask=use_dask,
         )
@@ -176,6 +185,7 @@ class DataConfig(BaseConfig):
                 means_location=means_location,
                 stds_location=stds_location,
                 prognostic_var_names=prognostic_var_names,
+                boundary_var_names=boundary_var_names,
                 static_data_vars=self.static_data_vars,
                 use_dask=True,
             )
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -13,14 +13,7 @@ from torch import nn
 from torch.nn import GELU
 
 from ocean_emulators.config_base import BaseConfig, TopLevelConfig
-from ocean_emulators.constants import (
-    BoundaryVarNames,
-    Grid,
-    Lat,
-    LoaderVersion,
-    Lon,
-    PrognosticVarNames,
-)
+from ocean_emulators.constants import Grid, Lat, LoaderVersion, Lon, PrognosticVarNames
 from ocean_emulators.models import FOMO, Samudra
 from ocean_emulators.models.base import BaseModel
 from ocean_emulators.models.modules import (
@@ -39,7 +32,7 @@ from ocean_emulators.models.modules import (
 )
 from ocean_emulators.models.modules.augment_input import Concat3dCoordinates
 from ocean_emulators.models.modules.blocks import ZonallyPeriodicBilinearUpsample
-from ocean_emulators.utils.data import DataContainer, DataSource, validate_data
+from ocean_emulators.utils.data import DataContainer, DataSource
 from ocean_emulators.utils.location import LocalLocation, Location, ResolvedLocation
 from ocean_emulators.utils.loss import (
     DynamicLoss,
@@ -155,7 +148,6 @@ class DataConfig(BaseConfig):
     def build(
         self,
         data_root: ResolvedLocation,
-        boundary_var_names: BoundaryVarNames,
         prognostic_var_names: PrognosticVarNames,
     ) -> DataContainer:
         loader_version = LoaderVersion(self.loader_version)
@@ -165,15 +157,13 @@ class DataConfig(BaseConfig):
         means_location = data_root.resolve(self.data_means_location)
         stds_location = data_root.resolve(self.data_stds_location)
 
-        source = (
-            DataSource.from_locations(
-                data_location=data_location,
-                means_location=means_location,
-                stds_location=stds_location,
-                use_dask=use_dask,
-            )
-            .pipe(validate_data, boundary_var_names, self.static_data_vars)
-            .mask(prognostic_var_names, self.hist)
+        source = DataSource.from_locations(
+            data_location=data_location,
+            means_location=means_location,
+            stds_location=stds_location,
+            prognostic_var_names=prognostic_var_names,
+            static_data_vars=self.static_data_vars,
+            use_dask=use_dask,
         )
 
         if use_dask:
@@ -181,15 +171,13 @@ class DataConfig(BaseConfig):
             source_using_dask = source
         else:
             # If we're not using dask for the main source, create a separate one
-            source_using_dask = (
-                DataSource.from_locations(
-                    data_location=data_location,
-                    means_location=means_location,
-                    stds_location=stds_location,
-                    use_dask=True,
-                )
-                .pipe(validate_data, boundary_var_names, self.static_data_vars)
-                .mask(prognostic_var_names, self.hist)
+            source_using_dask = DataSource.from_locations(
+                data_location=data_location,
+                means_location=means_location,
+                stds_location=stds_location,
+                prognostic_var_names=prognostic_var_names,
+                static_data_vars=self.static_data_vars,
+                use_dask=True,
             )
 
         static_data = (
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -13,7 +13,14 @@ from torch import nn
 from torch.nn import GELU
 
 from ocean_emulators.config_base import BaseConfig, TopLevelConfig
-from ocean_emulators.constants import BoundaryVarNames, Grid, Lat, LoaderVersion, Lon
+from ocean_emulators.constants import (
+    BoundaryVarNames,
+    Grid,
+    Lat,
+    LoaderVersion,
+    Lon,
+    PrognosticVarNames,
+)
 from ocean_emulators.models import FOMO, Samudra
 from ocean_emulators.models.base import BaseModel
 from ocean_emulators.models.modules import (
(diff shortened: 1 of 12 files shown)


src/ocean_emulators/config.py
```diff
@@ -149,6 +156,8 @@ def build(
         self,
         data_root: ResolvedLocation,
         boundary_var_names: BoundaryVarNames,
+        prognostic_var_names: PrognosticVarNames,
+        hist: int,
```
alxmrs: Don't need to pass this in. It's a class member.

src/ocean_emulators/config.py
```diff
@@ -164,6 +172,7 @@ def build(
             use_dask=use_dask,
         )
         source = validate_data(source, boundary_var_names, self.static_data_vars)
+        source = source.mask(prognostic_var_names, self.hist)
```
alxmrs: I'm not sure if I should roll both of these into the static factory method `from_locations`. It would be simpler to use, but more complicated in that factory method.

src/ocean_emulators/train.py
```diff
@@ -188,13 +188,8 @@ def __init__(self, cfg: TrainConfig) -> None:
...
         self.metadata = construct_metadata(self.data)
-        self.wet, self.wet_surface = extract_wet_mask(
-            self.data, self.prognostic_var_names, cfg.data.hist
-        )
-        self.wet_without_hist_cpu, _ = extract_wet_mask(
-            self.data, self.prognostic_var_names, 0
-        )
-        self.wet = self.wet.to(self.device)
+        self.wet = self.src.masks.wet.to(self.device)
+        self.wet_without_hist = self.src.masks.wet_without_hist_cpu
```
alxmrs: Ironically, in the old implementation on the left, this "cpu" version was moved to `self.device`. I've removed that in this change; though, I can add it back.

src/ocean_emulators/config.py
```diff
@@ -157,27 +165,31 @@ def build(
...
-            use_dask=use_dask,
+        source = (
+            DataSource.from_locations(
+                data_location=data_location,
+                means_location=means_location,
+                stds_location=stds_location,
+                use_dask=use_dask,
+            )
+            .pipe(validate_data, boundary_var_names, self.static_data_vars)
+            .mask(prognostic_var_names, self.hist)
```
alxmrs: It might be less confusion to call this method `extract_mask`. I'm open to naming suggestions.

src/ocean_emulators/datasets.py
```diff
@@ -22,7 +22,11 @@
     PrognosticMask,
     PrognosticVarNames,
 )
-from ocean_emulators.utils.data import DataSource, LoadStats, conditional_rearrange
+from ocean_emulators.utils.data import (
+    LoadStats,
+    MaskedDataSource,
```
alxmrs: Similarly, it might be less confusing to call this `DataSourceWithMasks`. I'm open to naming suggestions.

src/ocean_emulators/utils/data.py
```diff
@@ -239,11 +244,45 @@ def from_locations(
             stds=stds,
         )
 
+    def mask(
```
alxmrs: I'm all ears for names!

jder:
@codex review

src/ocean_emulators/utils/data.py
```diff
@@ -239,11 +244,45 @@ def from_locations(
...
+    wet: PrognosticMask
+    wet_surface: GridMask
+    wet_without_hist_cpu: PrognosticMask
+
+
+@dataclasses.dataclass
+class MaskedDataSource(DataSource):
+    """A `DataSource` with a `masks` field; created via `DataSource.mask`."""
+
+    masks: Masks
```
jder: I don't understand why we want to distinguish between data sources with masks and data sources without masks. As far as I can tell, outside of test code we never make a DataSource which doesn't have masks -- every place we create a data source we immediately call validate_data and then mask. So I think we should not have two classes here. And then `mask` should happen as part of `validate_data`. 

I also [still think](https://github.com/Open-Athena/Ocean_Emulator/pull/211#pullrequestreview-2779019674) that `validate_data` (and now also `mask`) should be in `__init__` or `from_locations` or _something_ so that (a) we can de-duplicate the 4 places we init, then call validate data, then call mask and (b) we don't need a "data source without masks" type which is effectively unused.

src/ocean_emulators/utils/data.py
```diff
@@ -112,6 +113,10 @@ def filter(
 
         return dataclasses.replace(self, name=name, data=data, means=means, stds=stds)
 
+    def pipe(self, func: Callable[..., Self], *args, **kwargs) -> Self:
+        """Apply `func(self, *args, **kwargs)`, assuming `func` returns `Self`."""
+        return func(self, *args, **kwargs)
```
jder: 🐑 this gives up type safety and lintability of the given callable and adds indirection in exchange for a small amount of noise reduction in the caller which could also be achieved by making validate_data a member function. I don't think it's a good tradeoff.

src/ocean_emulators/utils/data.py
```diff
@@ -239,11 +244,45 @@ def from_locations(
...
+        )
+
+
+@dataclasses.dataclass
+class Masks:
+    """A collection of masks to expose the ocean and mask land."""
+
+    wet: PrognosticMask
+    wet_surface: GridMask
+    wet_without_hist_cpu: PrognosticMask
```
jder: 🐑 The fact that `wet` and `wet_without_hist` have the same type is a smell. Either the hist is part of the prognostic vars or it's not. I think we generally treat `prognostic_vars` as including history, so this should be `wet_without_hist_cpu: Bool[GridMask, "prognostic_vars_no_hist"]` or something.

jder (changes requested):
Thanks for doing this! I am very much in favor of having masks get passed around in the data source. I think it's pretty clear from all the places we now do things like `self.wet = src.wet_no_hist_cpu.to(device)` that the previous approach was hiding various mild bugs which we can now easily see & fix in the future. 

My one big piece of feedback is that I don't think this should be a separate class from DataSource; some comments below. (I also agree with Codex's two comments.)

src/ocean_emulators/utils/data.py
alxmrs: I'm mindway through a refactor where I combine masks into datasource along with validation, and it seems pretty hairy. A lot of tests are failing because masks are not validating or properly set up, even though masks are not used in tests. Maybe if you're online later today or next Monday, we can review what I have in progress to see if it's still worth it to reorganize the codebase. 

Especially hairy: our `conftest` data setup often doesn't have the right metadata (prog vars, static vars) to construct a DataSource since it was originally designed to be a light-weight holder of the primary datasets.

src/ocean_emulators/utils/data.py
alxmrs: I agree with this: I need to tease it apart now from a larger refactor, but I think we can get away with `wet` (singe history), `wet_surface` and a member function that extends the prognostic mask when we need it.

src/ocean_emulators/utils/data.py
alxmrs: hmmm, maybe I got lost in the big change, I think I'll be able to pull through.

src/ocean_emulators/config.py
```diff
@@ -161,9 +161,10 @@ def build(
             data_location=data_location,
             means_location=means_location,
             stds_location=stds_location,
+            prognostic_var_names=prognostic_var_names,
```
alxmrs, line 172: TODO(alxmrs): Can we infer this automatically?

src/ocean_emulators/datasets.py
```diff
@@ -93,8 +91,8 @@ def __init__(
                 f" output at {data.time.values[self.hist + 1]}"
             )
 
-        self.wet: torch.Tensor = wet.bool()
-        self.wet_surface: torch.Tensor = wet_surface.bool()
+        self.wet: PrognosticMask = src.masks.wet.bool()
```
alxmrs: TODO(alxrms) let's rename these two masks as "prognostic" and "boundary".

src/ocean_emulators/eval.py
```diff
@@ -90,20 +89,18 @@ def __init__(self, cfg: EvalConfig) -> None:
...
-        self.wet, self.wet_surface = extract_wet_mask(
-            self.data, self.prognostic_var_names, cfg.data.hist
-        )
-        self.wet_without_hist_cpu, _ = extract_wet_mask(
-            self.data, self.prognostic_var_names, 0
+        self.wet, self.wet_surface = (
+            self.src.masks.repeat_prognostic(cfg.data.hist),
+            self.src.masks.wet_surface,
         )
+        self.wet_without_hist_cpu = self.src.masks.wet
```
alxmrs: TODO(alxmrs) rename this.

src/ocean_emulators/train.py
```diff
@@ -185,13 +184,8 @@ def __init__(self, cfg: TrainConfig) -> None:
...
         self.metadata = construct_metadata(self.data)
-        self.wet, self.wet_surface = extract_wet_mask(
-            self.data, self.prognostic_var_names, cfg.data.hist
-        )
-        self.wet_without_hist_cpu, _ = extract_wet_mask(
-            self.data, self.prognostic_var_names, 0
-        )
-        self.wet = self.wet.to(self.device)
+        self.wet = self.src.masks.repeat_prognostic(cfg.data.hist).to(self.device)
+        self.wet_without_hist = self.src.masks.wet.to(self.device)
```
alxmrs: TODO(alxmrs): rename this.

src/ocean_emulators/utils/data.py
```diff
@@ -43,6 +44,25 @@ def _var_name_encode_level(var_name: str) -> bool:
...
+
+
+@dataclasses.dataclass
+class Masks:
+    """A collection of masks to expose the ocean and mask land."""
+
+    wet: PrognosticMask
+    wet_surface: GridMask
+
+    def repeat_prognostic(self, hist: int):
```
alxmrs: I'm not sure if this name is clear.

src/ocean_emulators/utils/data.py
```diff
@@ -225,18 +242,48 @@ def from_locations(
...
+        data: xr.Dataset,
+        means: xr.Dataset,
+        stds: xr.Dataset,
+        *,
+        prognostic_var_names: PrognosticVarNames,
+        static_data_vars: list[str] | None = None,
+        name: str = "DataSource",
+    ) -> Self:
+        data, means, stds = validate_data(
+            data, means, stds, static_data_vars=static_data_vars
```
alxmrs: I don't know if we'd rather handle the static data processing differently; then we wouldn't need to plumb the static data names around.

src/ocean_emulators/utils/data.py
```diff
@@ -225,18 +242,48 @@ def from_locations(
...
+        )
+        wet, wet_surface = extract_wet_mask(data, prognostic_var_names)
```
alxmrs: Maybe this could just return `Masks`.

src/ocean_emulators/utils/data.py
```diff
@@ -545,47 +590,44 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
...
-        def validated(data, means, stds):
-            data = (
-                data.pipe(flatten_masks)
-                .pipe(with_level_index_vars)
-                .pipe(with_lat_lon_coords)
+        for var in static_data_vars:
+            assert var in data.variables, (
+                f"Static data variable {var} not found in data"
             )
+            if "time" in data[var].dims:
```
alxmrs, line 618: If we did this processing in advance, or validated it elsewhere, we could probably get rid of a lot of parameter plumbing. Though, it's not the biggest deal.

alxmrs (commented):
Initial thoughts.

src/ocean_emulators/config.py
alxmrs: No, I think we cannot.

alxmrs:
@codex review, please.

alxmrs:
darn

alxmrs:
@claude will you review this PR, please?

jder:
@codex review

src/ocean_emulators/utils/data.py
```diff
@@ -545,47 +588,44 @@ def with_lat_lon_coords(data: xr.Dataset) -> xr.Dataset:
...
+        # Check if data variables are in the right format
+        # This check is to ensure we convert data to the correct format
+        means = with_level_index_vars(means)
+        stds = with_level_index_vars(stds)
+        boundary_vars = [str(var) for var in data.data_vars.keys() if "_" not in var]
 
     # Check if any anomalies are needed to be computed
     anomalies_vars = get_anomalies_vars(boundary_vars)
-    out = compute_anomalies(src_, anomalies_vars) if anomalies_vars else src_
+    out = (
```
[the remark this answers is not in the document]
alxmrs: Thanks for catching this. I'll fix this now.

src/ocean_emulators/utils/data.py
jder: How about `prognostic_with_hist(self, hist: int) -> Bool[GridMask, "prognostic_vars*(hist+1)"]`

jder: (Really, think the problem is that what we should really have here is masks per level and then allow extracting these like `mask_for_surface` and `masks_for_depth` or whatever and then the caller can reshape to broadcast along the number of prognostic vars or surface vars or whatever. I don't think we should make this change now, though.)

src/ocean_emulators/datasets.py
```diff
@@ -93,8 +91,8 @@ def __init__(
                 f" output at {data.time.values[self.hist + 1]}"
             )
 
-        self.wet: torch.Tensor = wet.bool()
-        self.wet_surface: torch.Tensor = wet_surface.bool()
+        self.wet: PrognosticMask = src.masks.prognostic.bool()
+        self.wet_surface: GridMask = src.masks.boundary.bool()
```
```suggestion
        self.wet: PrognosticMask = src.masks.prognostic
        self.wet_surface: GridMask = src.masks.boundary
```

src/ocean_emulators/datasets.py
```diff
@@ -458,8 +454,8 @@ def __init__(
             indices_da + stride * window_dim
         )
 
-        self.wet = wet.bool().to(self.device)
-        self.wet_surface = wet_surface.bool().to(self.device)
+        self.wet: PrognosticMask = src.masks.prognostic.bool().to(self.device)
+        self.wet_surface: GridMask = src.masks.boundary.bool().to(self.device)
```
```suggestion
        self.wet: PrognosticMask = src.masks.prognostic.to(self.device)
        self.wet_surface: GridMask = src.masks.boundary.to(self.device)
```

jder (approved):
Thank you, looks great! Sorry this was a lot of thrashing to get tests working. 

One day, now that we can generate datasets easily, we should consider getting rid of much of this preprocessing. In the meantime I think this is a great improvement and the mask thing is definitely better.

PR #502, opened 2025-12-04
Various minor viz doc & config fixes

jder:
This is a bit awkward because the viz configs don't really have anything to do with the folder they're in (other than trying to be helpful by having the source eval run name included, though arguably this is confusing more than helpful). And really they vary mostly by what resolution of data you need (and so data path + basin mask path). But I think this is at least improving the situation a bit by fixing typos and defaulting everything to 1-degree data.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -262,9 +262,9 @@ To run a remote viz job with SkyPilot, please use the following command:
 uv run sky launch skypilot/viz.sky.yaml \
   --env WANDB_API_KEY \
   --env-file <my-vars>.env \
-  --env NAME <my-experiment-name>-viz \
-  --env BASIN_PATH basin_masks_original.zarr \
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -262,9 +262,9 @@ To run a remote viz job with SkyPilot, please use the following command:
 uv run sky launch skypilot/viz.sky.yaml \
   --env WANDB_API_KEY \
   --env-file <my-vars>.env \
-  --env NAME <my-experiment-name>-viz \
-  --env BASIN_PATH basin_masks_original.zarr \
-  --env RUNS=[{"location": "/inputs/<my-experiment-name>-eval/predictions.zarr"}]
+  --env NAME=<my-experiment-name>-viz \
```
alxmrs, line 265: man, I can't tell you how many times I've been burnt before from mixing space vs =s in CLI args. Thanks for the fix.

alxmrs (commented):
Nice fixes, LGTM.

alxmrs (approved):

PR #504, opened 2025-12-04
Multi-scale data loading in two ways: "merged" and "multi" example loading.

alxmrs:
This PR provides two types of multi-scale data loading: "merge" and "multi". The former merges all representations into the highest resolution grid and pads missing data with zeros, the latter returns a collection of examples at different scales. 

This is the initial part of a multi-step change to get multi-scale data loading to work. After this change, the multi-scale data loader will function only on a single scale, since we have no configuration system to register multiple data sources at different scales (that will come soon). However, the multi-scale loader and masking system is fully "backwards compatible" -- it behaves like the standard torch loader when used at a single scale.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -647,28 +647,27 @@ def build_nodata_masks(
     )
     ref_lat, ref_lon = sorted_scales.pop(0)
 
-    all_prognostic_masks = [
-        torch.ones(1, num_prognostic_channels, ref_lat, ref_lon, dtype=torch.bool),
+    prognostic_masks = [
+        torch.ones(num_prognostic_channels, ref_lat, ref_lon, dtype=torch.bool),
     ]
-    all_boundary_masks = [
-        torch.ones(1, num_boundary_channels, ref_lat, ref_lon, dtype=torch.bool),
+    boundary_masks = [
+        torch.ones(num_boundary_channels, ref_lat, ref_lon, dtype=torch.bool),
     ]
 
     for lat, lon in sorted_scales:
-        lat_stride = ref_lat // lat
-        lon_stride = ref_lon // lon
+        lat_stride, lon_stride = ref_lat // lat, ref_lon // lon
 
-        prog_mask = torch.zeros_like(all_prognostic_masks[0], dtype=torch.bool)
-        bound_mask = torch.zeros_like(all_boundary_masks[0], dtype=torch.bool)
+        prog_mask_per_scale = torch.zeros_like(prognostic_masks[0], dtype=torch.bool)
+        bound_mask_per_scale = torch.zeros_like(boundary_masks[0], dtype=torch.bool)
 
-        prog_mask[:, :, ::lat_stride, ::lon_stride] = True
-        bound_mask[:, :, ::lat_stride, ::lon_stride] = True
+        prog_mask_per_scale[:, ::lat_stride, ::lon_stride] = True
+        bound_mask_per_scale[:, ::lat_stride, ::lon_stride] = True
 
-        all_prognostic_masks.append(prog_mask)
-        all_boundary_masks.append(bound_mask)
+        prognostic_masks.append(prog_mask_per_scale)
+        boundary_masks.append(bound_mask_per_scale)
 
-    input_mask = torch.cat(all_prognostic_masks + all_boundary_masks, dim=1)
-    label_mask = torch.cat(all_prognostic_masks, dim=1)
+    input_mask = torch.cat(prognostic_masks + boundary_masks, dim=0)
+    label_mask = torch.cat(prognostic_masks, dim=0)
 
     return input_mask, label_mask
 
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -16,9 +16,11 @@ from ocean_emulators.config_base import BaseConfig, TopLevelConfig
 from ocean_emulators.constants import (
     BoundaryVarNames,
     Grid,
+    InputMask,
     Lat,
     LoaderVersion,
     Lon,
+    PrognosticMask,
     PrognosticVarNames,
 )
 from ocean_emulators.models import FOMO, Samudra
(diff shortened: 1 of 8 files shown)


src/ocean_emulators/config.py
```diff
@@ -637,6 +639,39 @@ def resolved_data_root(self) -> ResolvedLocation:
         return default_root.resolve(self.data_root)
 
 
+def build_nodata_masks(
+    sources: list[DataSource], num_prognostic_channels: int, num_boundary_channels: int
+) -> tuple[InputMask, PrognosticMask]:
```
alxmrs, line 644: Since I only need the prognostic mask, I could make this simpler by not calculating the InputMask. However, we might want the input mask later. Should I keep it, or YAGNI?

src/ocean_emulators/datasets.py
```diff
@@ -421,7 +461,7 @@ def __init__(
         executor: ThreadPoolExecutor | None = None,
     ):
         super().__init__()
-        self.id = f"{self.__class__.__name__}_{str(id(self))}"
+        self.id = f"{self.__class__.__name__}({str(id(self))})"
```
alxmrs, line 517: Not essential, but this seems like a better way to encode strings (consistent with the new ID).

src/ocean_emulators/datasets.py
```diff
@@ -643,8 +683,118 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        self,
+        srcs: list[DataSource],
+        make_dataset: Callable[[DataSource], TorchTrainDataset],
+    ):
+        self.datasets = [make_dataset(src) for src in srcs]
+        self.id = f"{self.__class__.__name__}({str(id(self))}, {', '.join([ds.id for ds in self.datasets])})"
+
+    @elapsed(level=logging.DEBUG)
+    def __getitem__(self, idx: int) -> RawMultiscaleTrainData:
+        tds = [ds[idx] for ds in self.datasets]
```
alxmrs: Could be done in parallel, say, with our compute executor. I think, at least, the xarray stuff will be done in parallel anyway, so I don't think this optimization is needed for now.

src/ocean_emulators/datasets.py
```diff
@@ -643,8 +683,118 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        Args:
+            tds: TrainData objects at different resolutions. Spatial dimensions must be
+                evenly divisible (e.g., 360x180 and 180x90 with stride=2).
+
+        Returns:
+            TrainData with inputs/labels at the largest resolution, where num_prognostic_channels
+            equals the sum of prognostic channels across all scales.
+        """
+        tds_sorted = sorted(tds, key=lambda td: td[0][0].shape, reverse=True)
+        total_prog_channels = sum(td.num_prognostic_channels for td in tds_sorted)
```
alxmrs: Since we assume that the prognostic channels are the same for all data sources, this sum could probably be a mul. But, doing it this way would make it easier in the future to mix prog variables.

src/ocean_emulators/utils/data.py
```diff
@@ -69,6 +69,12 @@ def prognostic_with_hist(
     ) -> Bool[GridMask, " prognostic_vars*({hist}+1)"]:
         return torch.concat([self.prognostic] * (hist + 1), dim=0)
 
+    # TODO(alxmrs): It would probably be better to make a "repeat_prognostic" method, right?
```
alxmrs, line 72: On the other hand, this approach may have better static type info.

src/ocean_emulators/datasets.py
```diff
@@ -292,13 +294,20 @@ def __getitem__(self, idx):
...
 class RawTrainData:
-    def __init__(self, dataset_id: "TorchTrainDataset.Id"):
-        self.dataset_id: TorchTrainDataset.Id = dataset_id
+    def __init__(self, dataset_id: "TorchTrainDataset.Id | MultiscaleTrainDataset.Id"):
+        self.dataset_id = dataset_id
         self.raw_data: list[tuple[torch.Tensor, torch.Tensor]] = []
         self.load_stats: LoadStats | None = None
 
-    def insert(self, all_prognostic: torch.Tensor, all_boundary: torch.Tensor):
+    def append(self, all_prognostic: torch.Tensor, all_boundary: torch.Tensor):
```
alxmrs, line 311: Drive by refactor (more appropriate name).

tests/test_datasets.py
```diff
@@ -390,6 +420,64 @@ def make_config(src: DataSource):
...
+        original_samples = [extract_sample_arrays(sample) for sample in flat_loader]
+    with make_loader(
+        flat_config, version=LoaderVersion.OM4_MULTISCALE_TORCH
+    ) as multi_scale:
+        new_samples = [extract_sample_arrays(sample) for sample in multi_scale]
+
+    assert_equal_samples(original_samples, new_samples)
+
+
+def test_multiscale_merge():
```
alxmrs, line 464: I'm all ears for ideas to test this merge.

tests/test_datasets.py
```diff
@@ -390,6 +420,64 @@ def make_config(src: DataSource):
...
+
+    merged = merge(tds)
+
+    assert len(merged) == 4
+    merged_input, merged_label = merged[0]
+
+    assert merged_input.shape == (1, 13 * num_scales, scale, scale * 2)
+    assert merged_label.shape == (1, 10 * num_scales, scale, scale * 2)
+    assert merged.num_prognostic_channels == 10 * num_scales
+    assert set(merged_input[0, :, 0, 0].tolist()) == set(range(1, 6)) | {9, 17}
```
alxmrs, line 489: I think there is something more principled we could do here.

src/ocean_emulators/train.py
```diff
@@ -181,11 +182,23 @@ def __init__(self, cfg: TrainConfig) -> None:
         # TODO(jder): Could rewrite inference dataset like we did for TorchTrainDataset
         # see https://github.com/suryadheeshjith/Ocean_Emulator/issues/208
         self.inference_src = self.data_container.source_using_dask
-
         self.loader_version = self.data_container.loader_version
 
+        # This nodata mask is backwards compatible: when there is only one data source (i.e. one scale), the "mask"
+        # will be all ones (i.e., there is data everywhere).
+        # TODO(alxmrs): replace with multiple sources when we have a config system.
```
alxmrs, line 189: Since we punt on configuring multi-scales till later, I imagine we would only include this masking if we used the "merge" multi-scale loader and not otherwise.

jder:
@codex review

jder:
@codex review

jder:
@codex review

src/ocean_emulators/config.py
```diff
@@ -637,6 +639,39 @@ def resolved_data_root(self) -> ResolvedLocation:
...
+        bound_mask_per_scale = torch.zeros_like(boundary_masks[0], dtype=torch.bool)
+
+        prog_mask_per_scale[:, ::lat_stride, ::lon_stride] = True
+        bound_mask_per_scale[:, ::lat_stride, ::lon_stride] = True
+
+        prognostic_masks.append(prog_mask_per_scale)
+        boundary_masks.append(bound_mask_per_scale)
+
+    input_mask = torch.cat(prognostic_masks + boundary_masks, dim=0)
+    label_mask = torch.cat(prognostic_masks, dim=0)
```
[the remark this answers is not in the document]
alxmrs: Not the most useful because we mean to increase the channel dim in this way when "merge" is turned on.

tests/test_datasets.py
```diff
@@ -88,22 +95,55 @@ def make_loader(
...
+                srcs = [src]
 
-                raw_loader = DataLoader(
-                    data,
-                    batch_size=cfg.batch_size,
-                    drop_last=drop_last,
-                    collate_fn=collate_fn,
+                mode: MultiscaleMode = (
+                    "merge"
+                    if version == LoaderVersion.OM4_MULTISCALE_TORCH
```
alxmrs:
```suggestion
                    if version == LoaderVersion.OM4_MULTISCALE_MERGE
```

alxmrs:
Closing in favor of #523.

PR #512, opened 2025-12-08
Use new wandb config, remove infrequently-changed options

jder:


alxmrs (approved):
Nice, it's great to not have so much config noise. Thanks.

PR #513, opened 2025-12-09
Ohc plots rescale

amogh-gulati:
This PR fixes OHC scaling by real grid-cell areas.

Key changes

- Added spherical_area_weights_real to compute true cell areas using spherical geometry.
- Updated OHC computation and post-processing pipelines to propagate both areacello and areacello_real. 
- Ensures OHC anomaly now matches expected magnitude (±50 ZJ)

### DIFF
--- a/src/ocean_emulators/utils/data.py
+++ b/src/ocean_emulators/utils/data.py
@@ -473,35 +473,35 @@ def spherical_area_weights(data: xr.Dataset) -> Grid:
 def spherical_area_weights_real(data: xr.Dataset) -> Grid:
     """
     Compute real grid cell areas on a spherical Earth.
-    
+
     Uses the spherical geometry formula:
     A = R² × Δλ × (sin(φ₂) - sin(φ₁))
-    
+
     where:
     - R is Earth's radius (6371 km)
     - Δλ is the longitude spacing in radians
     - φ₁, φ₂ are the latitude bounds of the cell in radians
-    
+
     Args:
         data: Dataset containing lat/lon coordinates
-        
+
     Returns:
         Grid cell areas in m²
     """
     R = 6371000  # Earth radius in meters
-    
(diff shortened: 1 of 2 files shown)


src/ocean_emulators/utils/data.py
```diff
@@ -412,6 +412,49 @@ def spherical_area_weights(data: xr.Dataset) -> Grid:
...
+
+    # Compute cell areas: A = R² × dlon × (sin(lat + dlat/2) - sin(lat - dlat/2))
+    areas = np.zeros((len(lats), len(lons)))
+    for i, lat_rad in enumerate(lats_rad):
+        lat_north = lat_rad + dlat_rad / 2
+        lat_south = lat_rad - dlat_rad / 2
+        area = R**2 * dlon_rad * (np.sin(lat_north) - np.sin(lat_south))
+        areas[i, :] = area
+
+    return torch.from_numpy(areas)
```
jder, line 513: @adam-subel if you haven't already, would you mind verifying this looks right to you?

jder (changes requested):
Thanks for tracking this down! 

My only request is some naming changes to avoid future confusion. I think we should call this new function `spherical_area`. The old one is `spherical_area_weights` -- the `_weights` suffix already suggests to me that they're normalized. (Using `_real` seems dangerous given these are still a spherical approximation.)

Then in the viz code, I don't think we should use `areacello_real` for these new weights for the same reason. How about `areacello_spherical`? And perhaps you could rename the existing references to `areacello` in there to `areacello_weights` to avoid this confusion again? (I actually think most of the places which use `areacello` today want `areacello_spherical` rather than `areacello_weights` but we can review & fix in a follow-up if you don't want to change the behavior now.) 

WDYT?

jder (approved):

PR #517, opened 2025-12-15
Fix typo in dmse limit

jder:
Must have gotten broken after I ran a first training run with this turned on but before checking this code in.


alxmrs (approved):
LGTM

PR #518, opened 2025-12-18
Multi-scale data loading on a "match" or "mix" schedule.

alxmrs:
A simpler alternative to #504. In our multi-scale model, we need one of two things: a `TrainData` batch that varies by scale per the sample index, or a "mixed" `TrainData` batch, where each Input/Label pair may come from one of the input scales. This new data loader accomplishes this need by providing a "schedule" interface and by delegating the actual loading to the existing Torch loader. In addition, this new load accomplishes a necessary IO-bound optimization: we only ever load the `TrainData`(s) batch that will actually get used. So, instead of loading `TrainData` batches for every scale per index, we only load the data source that actually gets returned for that index. For this "match" case, this means we only look up one data source, for the "mix" case, this means we look up two.

Last, since data loading is no longer one-to-one (a single index in a batch could refer to multiple sources of data), this PR also includes a new batch sampler that samples across equivalence classes.

### DIFF
--- a/configs/test/train_default_2step.yaml
+++ b/configs/test/train_default_2step.yaml
@@ -7,7 +7,7 @@ disk_mode: true
 pin_mem: true
 save_freq: 5
 epochs: 1
-batch_size: 1
+batch_size: 2
 preemptible: false
 learning_rate: 0.0001
 scheduler: null
--- a/configs/test/train_default_2step.yaml
+++ b/configs/test/train_default_2step.yaml
@@ -7,7 +7,7 @@ disk_mode: true
 pin_mem: true
 save_freq: 5
 epochs: 1
-batch_size: 2
+batch_size: 1
 preemptible: false
 learning_rate: 0.0001
 scheduler: null
--- a/src/ocean_emulators/constants.py
+++ b/src/ocean_emulators/constants.py
@@ -234,7 +234,8 @@ def construct_metadata(data: xr.Dataset) -> dict[str, dict[str, str]]:
 
 class LoaderVersion(enum.Enum):
     OM4_TORCH = "om4-torch"
-    OM4_MULTI = "om4-multi"
+    OM4_MULTI_MATCH = "om4-multi-match"
+    OM4_MULTI_MIX = "om4-multi-mix"
 
 
 # TODO(#95): See if this can be removed and replaced.
--- a/src/ocean_emulators/constants.py
+++ b/src/ocean_emulators/constants.py
@@ -234,6 +234,7 @@ def construct_metadata(data: xr.Dataset) -> dict[str, dict[str, str]]:
 
 class LoaderVersion(enum.Enum):
     OM4_TORCH = "om4-torch"
+    OM4_MULTI = "om4-multi"
 
 
 # TODO(#95): See if this can be removed and replaced.
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -755,7 +755,7 @@ class MultiscaleTrainDataset(GpuResolvedDataset[RawMultiscaleTrainData]):
     def __getitem__(self, idx: int) -> RawMultiscaleTrainData:
         sub_idx, rem = self.index_by_schedule(idx)
         # In this optimization, we only look up the `TrainData`s that we might eventually use.
-        needed = list(self.multiplex[rem]) if self.schedule == "mix" else [rem]
+        needed = self.multiplex[rem] if self.schedule == "mix" else [rem]
         tds = {i: ds[sub_idx] for i, ds in enumerate(self.datasets) if i in needed}
         return RawMultiscaleTrainData(dataset_id=self.id, datasets=tds, index=idx)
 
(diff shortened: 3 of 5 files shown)


src/ocean_emulators/constants.py
```diff
@@ -234,6 +234,7 @@ def construct_metadata(data: xr.Dataset) -> dict[str, dict[str, str]]:
 
 class LoaderVersion(enum.Enum):
     OM4_TORCH = "om4-torch"
+    OM4_MULTI = "om4-multi"
```
alxmrs: Question for the reviewer: should I create two loaders for multi (one per schedule)?

src/ocean_emulators/datasets.py
```diff
@@ -421,7 +471,7 @@ def __init__(
         executor: ThreadPoolExecutor | None = None,
     ):
         super().__init__()
-        self.id = f"{self.__class__.__name__}_{str(id(self))}"
+        self.id = f"{self.__class__.__name__}({str(id(self))})"
```
alxmrs, line 476: Drive-by-refactor: I think this is nicer.

tests/test_datasets.py
```diff
@@ -41,6 +46,16 @@ def inference_loader_pair(trainer_pair: TrainPair) -> tuple[TrainConfig, DataLoa
     return cfg, trainer.inference_loader
 
 
+def coarsen(
```
alxmrs: TODO(alxmrs): Currently unused.

tests/test_datasets.py
```diff
@@ -88,22 +103,45 @@ def make_loader(
                     )
                     for stride in cfg.data_stride
                 ]
-
-                data: ConcatDataset = ConcatDataset(dataset_list)
                 collate_fn = collate_raw_train_data
+            case LoaderVersion.OM4_MULTI:
+                srcs = [src, src]
```
alxmrs: TODO(alxmrs): add coarsen here; it will mean tests need updating.

tests/test_datasets.py
```diff
@@ -88,22 +103,45 @@ def make_loader(
...
+                        stride=stride,
+                    )
 
-                loader = TrainDataLoader(raw_loader, dataset_list, torch.device("cpu"))
-                yield loader
+                dataset_list = [
+                    MultiscaleTrainDataset(
+                        srcs=srcs,
+                        make_dataset=partial(make_dataset, stride=stride),
+                        schedule="match",
```
alxmrs: TODO(alxmrs): we're only exercising one of two pathways.

src/ocean_emulators/utils/train.py
```diff
@@ -36,6 +40,23 @@ def collate_raw_train_data(data: Sequence[RawTrainData]) -> RawTrainData:
     return batched_data
 
 
+def collate_raw_multiscale_train_data(
+    data: Sequence[RawMultiscaleTrainData],
+) -> RawMultiscaleTrainData:
+    batched_data = RawMultiscaleTrainData(data[0].dataset_id, {}, data[0].index)
```
alxmrs, line 50: The index value here is suspect.

configs/test/train_default_2step.yaml
```diff
@@ -7,7 +7,7 @@ disk_mode: true
 pin_mem: true
 save_freq: 5
 epochs: 1
-batch_size: 1
+batch_size: 2
```
alxmrs: Needed to test if this PR actually worked with realistic batch sizes.

src/ocean_emulators/utils/train.py
```diff
@@ -36,6 +40,27 @@ def collate_raw_train_data(data: Sequence[RawTrainData]) -> RawTrainData:
     return batched_data
 
 
+def collate_raw_multiscale_train_data(
+    data: Sequence[RawMultiscaleTrainData],
+) -> RawMultiscaleTrainData:
+    # When using MultiscaleGroupedBatchSampler, items in a batch have the same multiplex
+    # position (idx % multiplex_size) but not necessarily the same exact index.
+    # For example, indices [0, 4, 8] all have multiplex position 0 when multiplex_size=4.
+    # We use data[0].index as the representative since all items map to the same position.
```
alxmrs:
```suggestion
    # When using EquivalenceGroupedBatchSampler, items in a batch have the same group
    # position (idx % group_size) but not necessarily the same exact index.
    # For example, indices [0, 4, 8] all have group position 0 when group_size=4.
    # We use data[0].index as the representative since all items map to the same position.
```

tests/test_datasets.py
```diff
@@ -251,32 +305,58 @@ def test_loader__data_shape(
...
+                }
+                observed_patterns = set(example_resolutions)
+                # All observed patterns must be valid
+                assert observed_patterns <= valid_patterns, (
+                    f"All resolutions must be valid members of the cartesian product for 'mix' schedule. "
+                    f"Valid patterns: {valid_patterns}, got {observed_patterns}, "
+                    f"invalid patterns: {observed_patterns - valid_patterns}"
+                )
+                # We should have at least 1 pattern
+                assert len(observed_patterns) > 0, "Should have at least one pattern"
```
alxmrs, line 359: AI generated; I think this could use some cleaning up.

alxmrs:
Closing in favor of #523.

PR #521, opened 2025-12-19
Allow using autocast for bfloat16

jder:
This seems to largely be a memory win and neutral for our metrics. I haven't asked for review of the viz results but it seems worth at least having this option checked in for running quarter-degree runs.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -508,6 +508,10 @@ class SamudraConfig(BaseModelConfig):
         default=0,
         description="""Number of channels used for a learned positional embedding""",
     )
+    use_bfloat16: bool = Field(
+        default=False,
+        description="Use bfloat16 for most layers rather than float32.",
+    )
 
     def build(
         self,
@@ -547,6 +551,7 @@ class SamudraConfig(BaseModelConfig):
             wet=wet,
             static_data=static_data,
             gradient_detach_interval=self.gradient_detach_interval,
+            use_bfloat16=self.use_bfloat16,
         )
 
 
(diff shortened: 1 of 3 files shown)


src/ocean_emulators/models/samudra.py
```diff
@@ -24,6 +25,7 @@ def __init__(
         wet: Grid,
         static_data: xr.Dataset | None,
         gradient_detach_interval: int,
+        use_bfloat16: bool,
     ):
         super().__init__(
             in_channels=in_channels,
```
alxmrs, line 31: Do we want this as a property of our base models?

src/ocean_emulators/utils/device.py
```diff
@@ -3,6 +3,13 @@
 _CHOSEN_DEVICE: torch.device | None = None
 
 
+def autocast(enabled: bool, dtype: torch.dtype) -> torch.autocast:
```
alxmrs, line 6: TIL autocast! Neat.

src/ocean_emulators/config.py
```diff
@@ -508,6 +508,10 @@ class SamudraConfig(BaseModelConfig):
         default=0,
         description="""Number of channels used for a learned positional embedding""",
     )
+    use_bfloat16: bool = Field(
```
alxmrs, line 511: 🐑 As commented below, I could see this as a feature used in all our models. Is it worth putting in the base model config?

alxmrs (approved):
Sweet. One minor reuse question. I'm excited to see this in the quarterdeg run.

src/ocean_emulators/config.py
jder: I think FOMO should probably always use bfloat16, so I suggest we make that change at some point instead? (Adding it now to FOMO but having it do nothing or doing the work as part of this PR to make it work in FOMO both seem worse to me than the current state.)

PR #522, opened 2025-12-19
Drive by: don't do an unnecessary copy

jder:
`astype` defaults to `copy=True` for mutation reasons but we don't mutate these so this is just wasted work.


alxmrs (approved):
Nice catch.

PR #523, opened 2025-12-19
TorchTrainDataset now contains a `src` _and_ `dst`.

alxmrs:
Our primary data loader has been modified to differentiate between loading input + boundary vs the label, where the input and label are prognostics. By separating these concepts into two different data sources (well, a data "source" and a data "destination"), we can easily create data loading schedules that allow for training across multiple scales. 

Having "match" or "mix" schedule training examples poses a problem when we have batch sizes greater than 1: we might get mixed resolution examples in a single batch, which will break during the collate step. To address this, this PR includes a batch sampler that we recommend to be used as the default sampler implementation.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -2,9 +2,10 @@ import itertools
 import logging
 import random
 import time
+from collections.abc import Callable
 from concurrent.futures import wait
 from concurrent.futures.thread import ThreadPoolExecutor
-from typing import TypeAlias, final
+from typing import Any, Self, TypeAlias, final
 
 import numpy as np
 import torch
@@ -718,8 +719,8 @@ class EquivalenceGroupBatchSampler(Sampler[list[int]]):
     batches are globally shuffled each epoch to avoid sequential group processing.
 
     Args:
-        dataset_sizes: Optional list of individual dataset sizes. If provided, groups are created
-            based on dataset boundaries.
+        dataset_sizes: List of individual dataset sizes. Groups are created based on dataset boundaries,
+            where each dataset forms its own equivalence group.
         batch_size: Number of samples per batch
         shuffle: Whether to shuffle indices within groups and shuffle batches globally
         drop_last: Whether to drop incomplete batches at the end of each group
@@ -744,6 +745,77 @@ class EquivalenceGroupBatchSampler(Sampler[list[int]]):
         for size in dataset_sizes:
             self.groups.append(list(range(cumsum, cumsum + size)))
             cumsum += size
+
+    @classmethod
+    def from_datasets(
+        cls,
+        datasets: list["TorchTrainDataset"],
+        group_key: Callable[["TorchTrainDataset"], Any],
+        batch_size: int,
+        shuffle: bool,
+        drop_last: bool,
+    ) -> Self:
+        """Create sampler by grouping datasets using a key function.
+
+        This factory method allows grouping datasets by arbitrary criteria (e.g., resolution,
+        regardless of other parameters like stride). Datasets with the same key are batched together.
+
+        Args:
+            datasets: List of TorchTrainDataset instances to group
+            group_key: Callable that extracts grouping key from a dataset.
+            batch_size: Number of samples per batch
+            shuffle: Whether to shuffle indices within groups and shuffle batches globally
+            drop_last: Whether to drop incomplete batches at the end of each group
+
+        Examples:
+                - lambda ds: (ds._input_src.data.sizes['lat'], ds._input_src.data.sizes['lon'])  # group by resolution
+                - lambda ds: ds._input_src.data.sizes['lat']  # group by latitude size only
+            batch_size: Number of samples per batch
+            shuffle: Whether to shuffle indices within groups and shuffle batches globally
+            drop_last: Whether to drop incomplete batches at the end of each group
+
+        Returns:
+            EquivalenceGroupBatchSampler configured to group by the provided key
+
+        Example:
+            >>> # Group datasets by resolution, allowing different strides to be batched together
+            >>> sampler = EquivalenceGroupBatchSampler.from_datasets(
+            ...     datasets=dataset_list,
+            ...     group_key=lambda ds: (ds._input_src.data.sizes['lat'], ds._input_src.data.sizes['lon']),
+            ...     batch_size=32,
+            ...     shuffle=True,
+            ...     drop_last=True,
+            ... )
+        """
+        from collections import defaultdict
+
+        # Group indices by their key
+        groups: dict[tuple, list[int]] = defaultdict(list)
+
+        cumsum = 0
+        for ds in datasets:
+            key = group_key(ds)
+            # Make key hashable if it isn't already
+            if not isinstance(key, (int, str, tuple)):
+                key = tuple(key) if hasattr(key, "__iter__") else (key,)
+            groups[key].extend(range(cumsum, cumsum + len(ds)))
+            cumsum += len(ds)
+
+        # Convert groups to dataset_sizes format
+        # Sort by key for deterministic ordering across runs
+        sorted_groups = sorted(groups.items(), key=lambda x: str(x[0]))
+
+        # Create instance with computed dataset_sizes
+        instance = cls.__new__(cls)
+        instance.batch_size = batch_size
+        instance.shuffle = shuffle
+        instance.drop_last = drop_last
+        instance.group_size = len(sorted_groups)
+
+        # Store the actual grouped indices (not just sizes)
+        instance.groups = [indices for _, indices in sorted_groups]
+
+        return instance
 
     def __iter__(self):
         # Choose sampler based on shuffle setting
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -470,8 +470,9 @@ class TorchTrainDataset(Dataset[RawTrainData]):
             indices_da + stride * window_dim
         )
 
-        self.wet: PrognosticMask = src.masks.prognostic.to(self.device)
+        self.wet_input: PrognosticMask = src.masks.prognostic.to(self.device)
         self.wet_surface: GridMask = src.masks.boundary.to(self.device)
+        self.wet_label: PrognosticMask = dst.masks.prognostic.to(self.device)
 
         def flatten_to_device(means_or_stds: xr.Dataset) -> torch.Tensor:
             if "lev" in means_or_stds.dims:
@@ -591,11 +592,15 @@ class TorchTrainDataset(Dataset[RawTrainData]):
             input_all[:, : self.hist + 1, :, :, :],
             self.input_means,
             self.input_stds,
+            self.wet_input,
             boundary_all[:, : self.hist + 1, :, :, :],
         )
         # grab future steps, repeat as we do for input
         label = self._prep_tensor_steps(
-            label_all[:, self.hist + 1 :, :, :, :], self.label_means, self.label_stds
+            label_all[:, self.hist + 1 :, :, :, :],
+            self.label_means,
+            self.label_stds,
+            self.wet_label,
         )
         return total_input, label
 
@@ -604,6 +609,7 @@ class TorchTrainDataset(Dataset[RawTrainData]):
         prognostic_steps: Float[torch.Tensor, "batch time variable lat lon"],
         prognostic_means: Float[torch.Tensor, " variable"],
         prognostic_stds: Float[torch.Tensor, " variable"],
+        prognostic_mask: Float[torch.Tensor, " variable"],
         boundary_steps: Float[torch.Tensor, "batch time variable lat lon"]
         | None = None,
     ) -> Input:
@@ -641,7 +647,7 @@ class TorchTrainDataset(Dataset[RawTrainData]):
             prognostic_steps,
             prognostic_means,
             prognostic_stds,
-            self.wet,
+            prognostic_mask,
         )
         if boundary_steps is not None:
             boundary_steps = normalize_and_mask(
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -1,18 +1,15 @@
-import itertools
 import logging
-import random
 import time
-from collections.abc import Callable
 from concurrent.futures import wait
 from concurrent.futures.thread import ThreadPoolExecutor
-from typing import Any, Self, TypeAlias, final
+from typing import TypeAlias, final
 
 import numpy as np
 import torch
 import xarray as xr
 from einops import rearrange
 from jaxtyping import Float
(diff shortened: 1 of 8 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -544,14 +586,15 @@ def _get_input_and_label(
...
+        label_all: Float[torch.Tensor, "batch time variable lat lon"],
     ) -> tuple[Input, Prognostic]:
         # grab past steps and prep for model
         total_input = self._prep_tensor_steps(
             prognostic_all[:, : self.hist + 1, :, :, :],
             boundary_all[:, : self.hist + 1, :, :, :],
         )
         # grab future steps, repeat as we do for input
-        label = self._prep_tensor_steps(prognostic_all[:, self.hist + 1 :, :, :, :])
+        label = self._prep_tensor_steps(label_all[:, self.hist + 1 :, :, :, :])
```
jder: This is normalizing the label steps using prognostic_mean/stds. I think probably this is not quite right -- we should use the same means and stds across the whole training run, regardless of whether we're using 1/2 degree or 1 degree as input or output, but probably worth confirming with the broader group. (Perhaps means/stds should be passed in to TorchTrainDataset in that case? Or we should have each datasource carry the training-run-wide means/stds?) Either way we should remove self.label_means/stds if they are unused.

src/ocean_emulators/datasets.py
```diff
@@ -295,25 +295,35 @@ def __getitem__(self, idx):
...
         self.dataset_id: TorchTrainDataset.Id = dataset_id
-        self.raw_data: list[tuple[torch.Tensor, torch.Tensor]] = []
+        self.raw_data: list[tuple[torch.Tensor, torch.Tensor, torch.Tensor]] = []
         self.load_stats: LoadStats | None = None
 
-    def insert(self, all_prognostic: torch.Tensor, all_boundary: torch.Tensor):
-        self.raw_data.append((all_prognostic, all_boundary))
+    def insert(
+        self,
+        all_prognostic: torch.Tensor,
```
jder: I believe the meaning of "prognostic" is "direct output" or maybe "input and output" when we're talking autoregressive. Given that this is no longer used for output, perhaps we should rename it here (and elsewhere in this PR where we've introduced label) to `input`?

src/ocean_emulators/datasets.py
```diff
@@ -382,6 +392,9 @@ def pin_memory(self):
         return self
 
 
+TrainSchedule = Literal["standard", "match", "mix"]
```
jder: I think this should probably be in config.py? It's unused in this file.

tests/test_datasets.py
```diff
@@ -73,11 +102,32 @@ def make_loader(
             cfg.experiment.prognostic_vars_key, cfg.experiment.boundary_vars_key
         )
 
+        match schedule:
+            case "standard":
+                srcs: Iterable[tuple[DataSource, DataSource]] = [(src, src)]
+            case "match":
+                coarsened_src = src.map_data(coarsen_data, suffix="half-size")
```
jder, line 110: :sheep: probably we should make `map` operate on the mask as well and use it here rather than this + the next line.

tests/test_datasets.py
```diff
@@ -73,11 +102,32 @@ def make_loader(
...
+                coarsened_src = dataclasses.replace(
+                    coarsened_src, masks=coarsen_masks(src.masks)
+                )
+                scales = [src, coarsened_src]
+                srcs = itertools.permutations(scales, 2)
```
jder: I think this returns `((src, coarsened_src), (coarsened_src, src))` but I think you want `((src, src), (coarsened_src, coarsened_src))`?

tests/test_datasets.py
```diff
@@ -279,6 +340,91 @@ def test_loader__data_shape(
...
+
+        assert len(samples) == n_samples, (
+            f"Current config {train_config} only supports {n_samples} examples; "
+            f"got {len(samples)}."
+        )
+
+        example_resolutions = []
+        # Only check the first 4 samples; this should be proof enough that everything is
+        # the right shape.
+        for sample in samples[:4]:
```
jder: Why 4? Won't this be drawn from the first dataset of the loader only? 

I assume this is not passing yet because we don't have the logic where a batch includes only data from one dataset?

jder (changes requested):
Thanks! This looks great; just a few suggestions/questions at this stage.

jder:
@codex review

tests/test_datasets.py
alxmrs: Now, I plan to sample from the whole list, striding by 3. This lets me use the strong assert for the mix schedule.

src/ocean_emulators/datasets.py
```diff
@@ -295,25 +298,36 @@ def __getitem__(self, idx):
...
+        self.raw_data: list[tuple[torch.Tensor, torch.Tensor, torch.Tensor]] = []
         self.load_stats: LoadStats | None = None
 
-    def insert(self, all_prognostic: torch.Tensor, all_boundary: torch.Tensor):
-        self.raw_data.append((all_prognostic, all_boundary))
+    def insert(
+        self,
+        all_input: torch.Tensor,
+        all_boundary: torch.Tensor,
+        all_label: torch.Tensor,
```
jder: :sheep: FWIW the "all_" prefix here was meant to indicate that it was both input and label. Now that we've separated them out I think we can drop it.

src/ocean_emulators/datasets.py
```diff
@@ -523,40 +566,53 @@ def __getitem__(self, idx: int):
...
         return TD
 
     def to_train_data(self, raw_train_data: RawTrainData) -> TrainData:
         train_data = TrainData(self.num_prognostic_channels)
-        for prognostic_all, boundary_all in raw_train_data.raw_data:
-            input, label = self._get_input_and_label(
-                prognostic_all.to(device=self.device, non_blocking=True),
+        for input_all, boundary_all, label_all in raw_train_data.raw_data:
+            input_, label = self._get_input_and_label(
+                input_all.to(device=self.device, non_blocking=True),
```
jder: :sheep: seems a little sad that "get_input_…" takes in something we're calling "input". The former means "the input to the model, including boundary" and the latter means "the input prognostic variables". Maybe it should be "_get_full_input_and_label" or something?

src/ocean_emulators/datasets.py
```diff
@@ -643,6 +699,160 @@ def load_variable_data(var: xr.Variable) -> None:
     wait(futures)
 
 
+class _SimpleSubsetSampler(torch.utils.data.Sampler):
```
```suggestion
class _SimpleSubsetSampler(Sampler):
```

src/ocean_emulators/datasets.py
```diff
@@ -643,6 +699,160 @@ def load_variable_data(var: xr.Variable) -> None:
...
+
+    def __len__(self):
+        """Calculate total number of batches across all groups."""
+        total_batches = 0
+        for group in self.groups:
+            if self.drop_last:
+                total_batches += len(group) // self.batch_size
+            else:
+                total_batches += (len(group) + self.batch_size - 1) // self.batch_size
+
```
jder: Could also extract the logic to make the sampler from `__iter__` and ask it for its length; not sure if that's better or worse.

src/ocean_emulators/datasets.py
```diff
@@ -643,6 +699,160 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        self.indices = indices
+
+    def __iter__(self):
+        return iter(self.indices)
+
+    def __len__(self):
+        return len(self.indices)
+
+
+class EquivalenceGroupBatchSampler(Sampler[list[int]]):
```
jder: Worth adding some tests directly for this class? Maybe breaking it out to its own file?

src/ocean_emulators/datasets.py
```diff
@@ -643,6 +699,160 @@ def load_variable_data(var: xr.Variable) -> None:
...
+    def __init__(
+        self,
+        dataset_sizes: list[int],
+        batch_size: int,
+        shuffle: bool = True,
+        drop_last: bool = False,
+    ):
+        super().__init__()
+        self.group_size = len(dataset_sizes)
+        self.batch_size = batch_size
```
jder: Might we want different batch sizes for different groups, too? (Don't need to do this now, just thinking out loud.)

src/ocean_emulators/datasets.py
```diff
@@ -643,6 +699,160 @@ def load_variable_data(var: xr.Variable) -> None:
...
+                key = tuple(key) if hasattr(key, "__iter__") else (key,)
+            groups[key].extend(range(cumsum, cumsum + len(ds)))
+            cumsum += len(ds)
+
+        # Convert groups to dataset_sizes format
+        # Sort by key for deterministic ordering across runs
+        sorted_groups = sorted(groups.items(), key=lambda x: str(x[0]))
+
+        # Create instance with computed dataset_sizes
+        instance = cls.__new__(cls)
```
jder: Rather than bypassing the initializer which seems like a smell to me, should `__init__` take a list of ranges instead? Or maybe this from_datasets method should just *be* `__init__`?

src/ocean_emulators/datasets.py
```diff
@@ -643,6 +699,160 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        from collections import defaultdict
+
+        # Group indices by their key
+        groups: dict[tuple, list[int]] = defaultdict(list)
+
+        cumsum = 0
+        for ds in datasets:
+            key = group_key(ds)
+            # Make key hashable if it isn't already
+            if not isinstance(key, (int, str, tuple)):
```
jder: I think we shuld instead enforce that the group key is hashable (eg have the group_key be TorchTrainDataset -> Hashable)

src/ocean_emulators/datasets.py
```diff
@@ -643,6 +699,160 @@ def load_variable_data(var: xr.Variable) -> None:
...
+        dataset_sizes: List of individual dataset sizes. Groups are created based on dataset boundaries,
+            where each dataset forms its own equivalence group.
+        batch_size: Number of samples per batch
+        shuffle: Whether to shuffle indices within groups and shuffle batches globally
+        drop_last: Whether to drop incomplete batches at the end of each group
+    """
+
+    def __init__(
+        self,
+        dataset_sizes: list[int],
```
jder: I suggest we remove the term "Dataset" from here and the description of this class and use "group" instead given these might not be actual individual datasets. (`from_datasets` can then map from actual datasets into abstract groups)

jder (changes requested):
Thanks! Looks great; mostly 🐑 from me and a few minor suggestions.

jder:
@codex review again please

src/ocean_emulators/datasets.py
```diff
@@ -643,6 +699,160 @@ def load_variable_data(var: xr.Variable) -> None:
...
+            # This is regenerated each epoch, giving different orderings
+            all_batches = list(batch_sampler)
+            random.shuffle(all_batches)
+            yield from all_batches
+
+    def __len__(self):
+        """Calculate total number of batches across all groups."""
+        total_batches = 0
+        for group in self.groups:
+            if self.drop_last:
```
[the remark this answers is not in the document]
jder: Indeed. Maybe we should turn on the warning about missing types which would have caught this.

src/ocean_emulators/datasets.py
alxmrs: What do you think about using the term "example"? How about `self._to_example` or `self._data_to_example` or `self._make_example`...

src/ocean_emulators/datasets.py
alxmrs: Very interesting! I like this idea, but will say YAGNI for now. I'll keep track of it in an issue + comment though.

src/ocean_emulators/datasets.py
alxmrs: Sorry, I'm a bit confused by this suggestion. I see that you mean the logic could be extracted into something simpler or more DRY, but I can't quite see how. It's probably because it's been a minute since I looked at this.

alxmrs: I think I get it! Sorry, needed to refresh myself. Making a patch.

src/ocean_emulators/datasets.py
alxmrs: I see now! Yes, I refactored the code to do this and it's much simpler.

src/ocean_emulators/datasets.py
alxmrs: I agree!

src/ocean_emulators/datasets.py
jder: Sounds good to me!

jder:
@alxmrs taking a look now. In the future for big PRs it would be easier for me to track what's going on if you didn't rebase all the commits when making changes. We're going to squash them all anyway so you can just merge main.

alxmrs:
Sure, maybe we can discuss and then document a different development workflow. I'm curious: for big PRs, why a merge over a rebase? Does it have to do with how github trackes comments with commits, for example?

jder:
Yeah, take a look at this page and see if you can tell which commits I should review as a delta from my last review.

alxmrs:
Yeah, github does a poor job at handling this for sure. I don't often check intra PR commit diffs in reviews; mainly I check the comments on the main "timeline" and then look at all files. I'm happy to adjust my workflow if this makes thinks easier to review next time. I may pick your brain later about git habits to adopt -- I think a demo over VC may make it stick for me.

src/ocean_emulators/utils/samplers.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        from collections import defaultdict
+
+        # Group indices by their key
+        groups: dict[Hashable, list[int]] = defaultdict(list)
+
+        cumsum = 0
+        for ds in datasets:
+            key = group_key(ds)
+            assert isinstance(key, (int, str, tuple, Hashable)), (
```
jder:
```suggestion
            assert isinstance(key, Hashable), (
```

I think `int`, `str` and `tuple` of Hashable are all Hashable? Though really we have a type hint here already so hopefully this shouldn't happen.

src/ocean_emulators/utils/samplers.py
```diff
@@ -0,0 +1,169 @@
...
+                "`group_key` must be hashable."
+            )
+            groups[key].extend(range(cumsum, cumsum + len(ds)))
+            cumsum += len(ds)
+
+        # Sort by key for deterministic ordering across runs
+        sorted_groups = sorted(groups.items(), key=lambda x: str(x[0]))
```
jder:
```suggestion
        sorted_groups = sorted(groups.items(), key=lambda x: x[0])
```
seems like a possible footgun to me

jder (approved):
Thanks, looks great! Just two minor (I think) Claudeisms that I don't like :)

alxmrs:
Thanks! Will merge soon.

PR #526, opened 2025-12-23
Fixes from running eval mistakenly on CPU

jder:
I was (because of skypilot config [mis-]features) mistakenly running the quarter-degree eval on CPU. This PR:

* Ensures that eval set_device's correctly so we use bfloat16 etc as expected
* Produces useful diagnostics for why we're not using CUDA when we're not
* Changes eval + train skypilot jobs to request CUDA backends, not auto, so they will fail if CUDA is not used.

### DIFF
--- a/src/ocean_emulators/backend.py
+++ b/src/ocean_emulators/backend.py
@@ -4,6 +4,8 @@ import torch
 
 logger = logging.getLogger(__name__)
 
+import os
+
 from ocean_emulators.config import (
     DistributedConfig,
     EvalBackendConfig,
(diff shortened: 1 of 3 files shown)


src/ocean_emulators/backend.py
```diff
@@ -13,6 +15,19 @@
 from ocean_emulators.utils.distributed import init_distributed_mode
 
 
+def _cuda_diagnostics() -> str:
+    """Return a short string that helps debug why CUDA is (not) available."""
+    visible = os.getenv("CUDA_VISIBLE_DEVICES", None)
+    return (
+        "CUDA diagnostics: "
+        f"torch={torch.__version__}, "
+        f"torch.version.cuda={torch.version.cuda}, "
```
alxmrs, line 24: 🐑 you may want `{torch.version.cuda=}`.

alxmrs (approved):
Damn, that seems annoying to debug. Good catch, and I look fwd to hearing the full story later.

src/ocean_emulators/backend.py
```diff
@@ -13,6 +15,19 @@
...
+        f"torch.cuda.is_available()={torch.cuda.is_available()}, "
+        f"torch.cuda.device_count()={torch.cuda.device_count()}, "
```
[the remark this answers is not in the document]
jder: This seems like a good point but at least in my setup, torch.cuda.device_count() returns 0 when torch is compiled without cuda.

src/ocean_emulators/backend.py
jder: Thanks, I always forget about this but I think I prefer the symmetry here with the other lines over the brevity.

PR #530, opened 2025-12-29
Adding the ability to configure multi-scale training runs.

alxmrs:
Changes our config system and skypilot scripts to support multi resolution training.

### DIFF
--- a/configs/fomo_om4/train_multiscale.yaml
+++ b/configs/fomo_om4/train_multiscale.yaml
@@ -1,0 +1,53 @@
+# yaml-language-server: $schema=../schemas/TrainConfig.json
+
+debug: false
+disk_mode: true
+pin_mem: true
+save_freq: 5
+epochs: 70
+batch_size: 4
+learning_rate: 0.0006
+scheduler: { type: cosine }
+loss:
+  type: dynamic
+  metric: mse
+finetune: false
+resume_ckpt_path: null
+inference_epochs: []
+data_percent: 1.0
+train_time:
+  start: "1975-01-03"
+  end: "2013-10-05"
+val_time:
+  start: "2013-10-05"
+  end: "2014-10-05"
+inference_times:
+  - start: "1975-01-03"
+    end: "1976-01-03" # "1985-01-03"
+  - start: "1985-01-03"
+    end: "1986-01-03" # "1995-01-03"
+  - start: "1995-01-03"
+    end: "1996-01-03" # "2005-01-03"
+  - start: "2004-10-05"
+    end: "2005-10-05" # "2014-10-05"
+data_stride: [1]
+steps: [4]
+step_transition: []
+preemptible: false
+
+backend: auto
+
+experiment:
+  name: fomo_om4
+  rand_seed: 15
+  base_output_dir: .LOCAL
+  wandb:
+    mode: disabled
+    project: default
+  prognostic_vars_key: thermo_dynamic_all
+  boundary_vars_key: tau_hfds_hfds_anom
+  train_schedule: match
+  prefixes_by_scale:
+    - "om4_onedeg_v3"
+data: !include ../data/om4.yaml
+model: !include model.yaml
--- a/configs/fomo_om4/train_multiscale.yaml
+++ b/configs/fomo_om4/train_multiscale.yaml
@@ -47,7 +47,14 @@ experiment:
   prognostic_vars_key: thermo_dynamic_all
   boundary_vars_key: tau_hfds_hfds_anom
   train_schedule: match
-  prefixes_by_scale:
-    - "om4_onedeg_v3"
-data: !include ../data/om4.yaml
+data:
+  static_data_vars: []
+  concurrent_compute: true
+  sources:
+    - data_location: OM4.zarr
+      data_means_location: OM4_means.zarr
+      data_stds_location: OM4_stds.zarr
+    - data_location: om4_onedeg_v3/OM4.zarr
+      data_means_location: om4_onedeg_v3/OM4_means.zarr
+      data_stds_location: om4_onedeg_v3/OM4_stds.zarr
 model: !include model.yaml
--- a/skypilot/train.sky.yaml
+++ b/skypilot/train.sky.yaml
@@ -70,20 +70,29 @@ setup: |
   rclone config create nyu-osn s3 provider=AWS endpoint=https://nyu1.osn.mghpcc.org/ access_key_id="$OSN_KEY_ID" secret_access_key="$OSN_KEY_SECRET"
 
   mkdir -p ~/data/
-  if [ -z "$OSN_DATA_PATHS" ] || [ "$OSN_DATA_PATHS" = "null" ]; then
-    OSN_DATA_PATHS="$OSN_DATA_PATH"
-  fi
 
-  IFS=',' read -r -a DATA_PATHS <<< "$OSN_DATA_PATHS"
-  for DATA_PATH in "${DATA_PATHS[@]}"; do
-    DATA_PATH=$(echo "$DATA_PATH" | xargs)
-    if [ -n "$DATA_PATH" ]; then
-      BASENAME=$(basename "$DATA_PATH")
-      echo "Copying data from $DATA_PATH to ~/data/$BASENAME"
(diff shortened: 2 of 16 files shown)


configs/fomo_om4/train_multiscale.yaml
```diff
@@ -0,0 +1,53 @@
...
+  rand_seed: 15
+  base_output_dir: .LOCAL
+  wandb:
+    mode: disabled
+    project: default
+  prognostic_vars_key: thermo_dynamic_all
+  boundary_vars_key: tau_hfds_hfds_anom
+  train_schedule: match
+  prefixes_by_scale:
+    - "om4_onedeg_v3"
```
alxmrs: These are the differences from the default training config.

skypilot/train.sky.yaml
```diff
@@ -42,6 +42,11 @@ envs:
   OSN_KEY_ID: null
   OSN_KEY_SECRET: null
   OSN_DATA_PATH: emulators/jr7309/data/om4_halfdeg  # Tip: Don't add a trailing `/`!
+  # Eventually, we'll want to move to a CSV list of paths; but for now, this will cover all the multiscale training
+  # needs we'll have for a while.
```
jder: Definitely seems like a reasonable strategy for now (though we could also do what I do for the viz inputs and accept some json). But again makes me want to have our config system outside skypilot, not inside. Someday maybe!

src/ocean_emulators/train.py
```diff
@@ -715,10 +720,23 @@ def init_data_loaders(self, cur_step: int) -> None:
...
+        scales = [self.src]
+        if self.data_container.additional_sources is not None:
+            scales.extend(self.data_container.additional_sources)
+        match self.train_schedule:
+            case "standard":
+                srcs: Iterable[tuple[DataSource, DataSource]] = [(self.src, self.src)]
+            case "match":
+                srcs = [(s, s) for s in scales]
+            case "mix":
+                srcs = itertools.product(scales, repeat=2)  # type: ignore
```
jder: Codex noticed that this will be exhausted by the first use of srcs below and so not have any values in val_datasets

src/ocean_emulators/config.py
```diff
@@ -160,59 +162,64 @@ def build(
...
-        if use_dask:
-            # If we're already using dask, we don't need a second source
-            source_using_dask = source
-        else:
-            # If we're not using dask for the main source, create a separate one
-            source_using_dask = DataSource.from_locations(
+        def make_source(
+            local_dir_prefix: str = "", turn_on_dask: bool = use_dask
+        ) -> tuple[DataSource, bool]:
+            prefix_root = data_root.resolve(UnresolvedLocation(path=local_dir_prefix))
```
jder: If I'm reading this correctly, this assumes a structure like:

```
data_root/
    OM4.zarr/ # <- 1/2 degree data
    OM4_means.zarr/
    OM4_stds.zarr/
    om4_onedeg_v3/
        OM4.zarr/ # <- 1 degree data
        OM4_means.zarr
        OM4_stds.zarr
```

Is that right?

jder (changes requested):
One bug from Codex and I think the main merge of your other PR has kinda broken the diff here. Happy to continue reviewing next week once we've rebased this on main. (I think we need to rebase in this case because of the squash merge?)

My high-level feedback is that it seems like we're treating these other sources as a new thing rather than a generalization of having a single source. My comment below about data layout is one result of this, but generally the "additional"ness seems like a smell to me; am I missing something? I asked codex to take a cut at what it would look like to make a list of equal sources instead of having two kinds of data source and I think it's kinda nice? https://github.com/Open-Athena/Ocean_Emulator/compare/main...u/jder/int-multi

Not perfect but I actually think all the places where it's accessing sources[0] are probably places we need to do some (future) cleanup so I don't mind them looking a bit suspicious for now. WDYT?

src/ocean_emulators/config.py
alxmrs: Yes, that file structure is correct.

src/ocean_emulators/train.py
alxmrs: Good bot, that explains the obtuse type error.

alxmrs:
I like "another take"! Incorporated.

jder:
@codex review

src/ocean_emulators/utils/wandb.py
```diff
@@ -36,7 +36,7 @@ def enabled(self):
     def _make_config(self, cfg: "AnyTopLevelConfig", data_container: "DataContainer"):
         return {
             "config": cfg.model_dump(),
-            "data/attrs": data_container.source.data.attrs,
+            "data/attrs": data_container.sources[0].data.attrs,
```
jder: I don't think we need to fix all these right now but this one seems worth unrolling to me (ie log attrs from all sources)

src/ocean_emulators/utils/data.py
```diff
@@ -301,11 +301,16 @@ def from_datasets(
 
 @dataclasses.dataclass
 class DataContainer:
-    source: DataSource
+    sources: list[DataSource]
     source_using_dask: DataSource
```
jder: Maybe worth renaming this to `inference_source` now since that's what it's (already) used for and also explains why there's only one of them (for now).

src/ocean_emulators/utils/data.py
```diff
@@ -301,11 +301,16 @@ def from_datasets(
...
     loader_version: LoaderVersion
     supports_fork: bool
     static_data: xr.Dataset | None = None
+    additional_sources: list[DataSource] | None = None
+
+    @property
+    def source(self) -> DataSource:
+        return self.sources[0]
```
jder, line 312: Can we at least rename this to `primary_source` or `first_source` or maybe even `deprecated_source` something that will stick out so we know we should review those places later? (Also lots of things seem to be using `.sources[0]` instead, should we standardize?)

src/ocean_emulators/train.py
```diff
@@ -173,8 +177,8 @@ def __init__(self, cfg: TrainConfig) -> None:
                 max_workers=None, thread_name_prefix="concurrent_compute"
             )
 
-        self.src = self.data_container.source
-        self.data = self.data_container.source.data
+        self.src = self.data_container.sources[0]
```
jder: I skimmed quickly and it seems like most (all?) of the places we use `self.src` or `self.data` below are now bugs if there are multiple sources. Is the plan to fix those in follow-ups rather than now?

jder (approved):
Thanks for the changes. These are all non-blocking 🐑 since they only affect multi-scale training.

skypilot/train.sky.yaml
```diff
@@ -68,7 +70,20 @@ setup: |
...
+  fi
+
+  IFS=',' read -r -a DATA_PATHS <<< "$OSN_DATA_PATHS"
+  for DATA_PATH in "${DATA_PATHS[@]}"; do
+    DATA_PATH=$(echo "$DATA_PATH" | xargs)
+    if [ -n "$DATA_PATH" ]; then
+      BASENAME=$(basename "$DATA_PATH")
+      echo "Copying data from $DATA_PATH to ~/data/$BASENAME"
+      mkdir -p ~/data/${BASENAME}
+      rclone copy --progress nyu-osn:${DATA_PATH}/ ~/data/${BASENAME} --ignore-existing --transfers=32
```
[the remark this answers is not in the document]
jder: Yeah, this is a good point, I think this is broken for `data_location: OM4.zarr` in both multi-scale and single-scale training. Probably easiest to just have skypilot handle `OSN_DATA_PATH` like it always has and `OSN_DATA_PATHS` with subdirectories like this?

src/ocean_emulators/train.py
alxmrs: In #532, I work towards getting rid of all uses of `src`. I will make sure not to break anything in this PR, but that's where we're headed.

alxmrs: Ok, I'm now raising a couple of errors if a user tries to turn on multi-scale training before it's ready.

skypilot/train.sky.yaml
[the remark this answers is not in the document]
alxmrs: How does the latest version look?

PR #531, opened 2025-12-29
Handle loss_fn masking in one place.

alxmrs:
It appears we apply a wet mast at the start of every loss function (ironically, very [WET](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself#WET) code). This PR handles this logic in one place. While this may add a small amount of object overhead, this refactor will make it easier to handle multi-scale data loading. In such a scenario, the scale of the wetmask will be different depending on the "training schedule".


alxmrs:
Closing in favor of #555.

PR #532, opened 2025-12-29
FOMO supports training at multiple scales of OM4 data.

alxmrs:
Refactors the logic around training to support multiple scales. Updates the FOMO model to support multi-scale training. Adds a proof-of-concept config for multi-scale training. This change introduces the `GridContext` class, which holds data necessary for training in the forward passes and loss function. Grid context is necessary since the assumptions we have about the grid is no longer fixed.

### TODO: 
- [x] Refactor the add_3d_coords module to use on-the-fly lat/lon values.
- [ ] Fix logging / wandb plots (Implemented in #569).


src/ocean_emulators/models/modules/encoder.py
```diff
@@ -78,26 +79,40 @@ def forward(self, x: Input) -> Float[torch.Tensor, "batch {self.embed_dim} h w"]
...
+        # Training run
+        # 1/2 + 1 degree
+        # - [x] (a) in a given batch, encoder sees either 1/2 or 1 degree data (not mixed)
+        # - [x]     in a given epoch, encoder see all 1/2 and 1 degree data
+        # either:
+        # - [x] forward call of encoder takes patch size or grid info with data
+        # - [ ] or we have two PerceiverEncoders (1 degree vs 1/2 degree) which share parameters
+        #       (ie the self.perceiver (+pos_embed and scale_embed?) module)
+        # encoder produces a grid of patches each with an embedding
+        # - [ ] (b) the number of patches & physical extent is constant across all batches
```
alxmrs:
```suggestion
        # - [x] (b) the number of patches & physical extent is constant across all batches
```

src/ocean_emulators/models/fomo.py
```diff
@@ -54,21 +53,25 @@ def __init__(
...
-        self.layers = nn.ModuleList(layers)
-        self.unpatch = nn.Linear(
-            out_channels, out_channels * self.patch_size[0] * self.patch_size[1]
+        self.decoder = nn.Conv2d(
+            processor.out_channels,
+            out_channels,
+            last_kernel_size,
+            padding=last_kernel_size // 2,
+        )
+        all_patches = [self.encoder.patch_from(*grid) for grid in all_grids]
```
alxmrs: make this simply work on all grids

src/ocean_emulators/config.py
```diff
@@ -587,29 +568,37 @@ class FOMOConfig(BaseModelConfig):
...
-        area_weights: Grid,
         static_data: xr.Dataset | None,
-        lat: Lat,
-        lon: Lon,
+        srcs: list[DataSource],
     ) -> FOMO:
         total_in_channels = in_channels + (3 if self.add_3d_coordinates else 0)
         add_3d_coordinates = (
-            Concat3dCoordinates(lat, lon) if self.add_3d_coordinates else nn.Identity()
+            Concat3dCoordinates() if self.add_3d_coordinates else nn.Identity()
```
alxmrs: I'm on the fence, but this could probably just be a function.

src/ocean_emulators/config.py
```diff
@@ -587,29 +568,37 @@ class FOMOConfig(BaseModelConfig):
...
+        )
+        all_grids = [(len(s.resolution[0]), len(s.resolution[1])) for s in srcs]
```
alxmrs: TODO(alxmrs): s.grid.

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -13,6 +13,17 @@
 from ocean_emulators.constants import Input, Lat, Lon
 
 
+def patch_from(extent: tuple[float, float], height: int, width: int) -> tuple[int, int]:
```
alxmrs: Would be good to have a few tests for this.

src/ocean_emulators/models/samudra.py
```diff
@@ -82,4 +81,6 @@ def forward_once(self, fts: torch.Tensor) -> torch.Tensor:
 
         if self.corrector is not None:
             fts = self.corrector(fts_input, fts)
-        return torch.where(self.wet, fts, 0.0)
+        # Ensure mask is on the same device as fts
```
alxmrs: TODO(alxmrs): do this in advance somewhere up the stack.

src/ocean_emulators/stepper.py
```diff
@@ -49,8 +52,14 @@ def validate_batch(
...
-        outs = model.forward_once(input)
-        loss_per_channel = loss_fn(outs, label, wet=batch.label_mask)
+        outs = model.forward_once(
+            input,
+            wet=batch.label_mask.to(device=input.device),
+            resolution=batch.input_res,
+        )
+        loss_per_channel = loss_fn(
+            outs, label, wet=batch.label_mask, lat=batch.input_res[0]
+        )
```
alxmrs: Maybe in the future, we will group the wet and resolution extras for the model and loss into a dataclass.

alxmrs: Will need help naming this.

src/ocean_emulators/train.py
```diff
@@ -179,8 +174,11 @@ def __init__(self, cfg: TrainConfig) -> None:
                 max_workers=None, thread_name_prefix="concurrent_compute"
             )
 
+        # TODO(alxmrs): This is only used in the aggregator. Remove?
         self.src = self.data_container.primary_source
+        # TODO(alxmrs): This is only used to get lats in the loss fn. Remove?
         self.data = self.src.data
+        # TODO(alxmrs): This is not used and incorrect. Refactor (as part of src)?
```
alxmrs: Replace with issue.

src/ocean_emulators/train.py
```diff
@@ -179,8 +174,11 @@ def __init__(self, cfg: TrainConfig) -> None:
                 max_workers=None, thread_name_prefix="concurrent_compute"
             )
 
+        # TODO(alxmrs): This is only used in the aggregator. Remove?
         self.src = self.data_container.primary_source
+        # TODO(alxmrs): This is only used to get lats in the loss fn. Remove?
```
alxmrs: Remove now; this was refactored.

src/ocean_emulators/train.py
```diff
@@ -190,12 +188,7 @@ def __init__(self, cfg: TrainConfig) -> None:
 
         self.loader_version = self.data_container.loader_version
 
-        self.metadata = construct_metadata(self.data)
-        self.wet = self.src.masks.prognostic_with_hist(cfg.data.hist).to(self.device)
-        self.area_weights: Grid = spherical_area_weights(self.data)
-
-        self.area_weights = self.area_weights.to(self.device)
-
+        # TODO(alxmrs): This is only used in the corrector IIUC, which is not used. Remove?
```
alxmrs: File an issue to delete the corrector.

alxmrs (commented):
Review part one. Taking off.

src/ocean_emulators/config.py
```diff
@@ -587,29 +568,38 @@ class FOMOConfig(BaseModelConfig):
...
-        wet: Grid,
-        area_weights: Grid,
-        static_data: xr.Dataset | None,
-        lat: Lat,
-        lon: Lon,
+        static_data_for_corrector: xr.Dataset | None,
+        srcs: list[DataSource],
     ) -> FOMO:
         total_in_channels = in_channels + (3 if self.add_3d_coordinates else 0)
+        # TODO(alxmrs): Is Identity OK, or should I use the `None` pattern?
```
alxmrs: Will Identity mess things up, like checkpointing?

alxmrs:
@codex review

alxmrs:
@codex please review again.

alxmrs:
@codex review

alxmrs:
@codex review

src/ocean_emulators/constants.py
```diff
@@ -236,6 +237,31 @@ class LoaderVersion(enum.Enum):
     OM4_TORCH = "om4-torch"
 
 
+@dataclasses.dataclass(frozen=True)
+class GridContext:
```
jder: 🐑 I think this should probably live in a different file?

src/ocean_emulators/constants.py
```diff
@@ -236,6 +237,31 @@ class LoaderVersion(enum.Enum):
...
+            variable. Shape: (num_prognostic_vars, lat, lon). Land cells are False.
+        input_resolution: Tuple of (latitude, longitude) coordinate tensors defining
+            the spatial grid. Used for cosine-latitude weighting in loss functions.
+    """
+
+    label_mask: PrognosticMask
+    input_resolution: tuple[Lat, Lon]
+
+    def to(self, device: torch.device) -> Self:
+        """Move the label mask to the specified device."""
```
jder: Should it also move `input_resolution`?

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -23,14 +34,16 @@ class PerceiverEncoder(nn.Module):
     > Note: We assume that data along the lat/lon coordinates are positioned at the center of each grid point! Please
     > ensure this is the case at the data processing time.
 
+    This encoder is designed to make the same number of patches out of the same spatial extents across different scales
+    (each scale's patch will have a different lat/lon grid). To accomplish this with a single perceiver model, our
```
jder: I don't understand what it means that "each scale's patch will have a different lat/lon grid". This means the *input* that populates a patch will have a different resolution grid depending on scale?

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -23,14 +34,16 @@ class PerceiverEncoder(nn.Module):
...
+    `forward` call requires supplementary information: the resolution (a pair of Lat/Lon tensors), which is used to make
+     consistent positional encodings for patches across different scales.
+
     Args:
         in_channels (int): the number of input channels (roughly:  time x variable x (surface + depths)).
         out_channels (int): size of the latent dimension (aka, the embedding dimension).
-        patch_size (int | tuple[int, int]): the size of the patches to embed. Patches must evenly divide the input grid.
-          If a tuple is supplied, then it represents the (height, width) of the patches to embed.
+        spatial_extent (tuple[float, float]): spatial extent of the lat/lon coordinates.
```
jder: I also don't understand what the "spatial extent of the lat/lon coordinates" means. This indicates the the spatial extent of _each patch_ measured in degrees of lat and lon? I thought patch_size was more clear. How about patch_extent? (I guess this applies to `spatial_extent` in the encoder config too)

src/ocean_emulators/models/modules/encoder.py
jder: I think this could be more clear if these things were labeled as input vs output sizes, whether these are extends/sizes of the whole grid or just a patch and what units they are in.

src/ocean_emulators/models/fomo.py
```diff
@@ -86,22 +89,32 @@ def __init__(
...
+
+        with autocast(enabled=self.use_bfloat16, dtype=torch.bfloat16):
+            if self.maybe_add_3d_coordinates is not None:
+                fts = self.maybe_add_3d_coordinates(fts, ctx.input_resolution)
+            fts = self.encoder(fts, ctx.input_resolution)
+            fts = self.processor(fts)
+
+        # Convert back to float32 for decoder and unpatchify operations
+        fts = fts.to(torch.float32)
+        fts = self.decoder(fts)
```
jder, line 107: Nice, I find this much nicer to read than the layer list.

src/ocean_emulators/models/fomo.py
```diff
@@ -86,22 +89,32 @@ def __init__(
...
 
         # Unpatchify: project to patch area, then reshape back to original spatial dimensions
+        patch_size = patch_from(self.encoder.extent, H, W)
```
jder: Should this be something like `self.encoder.patch_size(grid_h=H, grid_w=W)`?

src/ocean_emulators/models/fomo.py
```diff
@@ -32,43 +33,45 @@ def __init__(
...
+            processor.out_channels,
+            out_channels,
+            last_kernel_size,
+            padding=last_kernel_size // 2,
+        )
+        all_patches = [patch_from(self.encoder.extent, *grid) for grid in all_grids]
+
+        self.unpatch = nn.ModuleDict(
+            {
+                str(patch_size): nn.Linear(
```
jder, line 74: why str(...)?

src/ocean_emulators/utils/data.py
```diff
@@ -85,6 +88,26 @@ def is_compact(self) -> bool:
...
+            torch.from_numpy(self.data.lon.values),
+        )
+
+    @cached_property
+    def grid(self) -> tuple[int, int]:
+        res = self.resolution
+        return res[0].shape[0], res[1].shape[0]
+
+    @cached_property
+    def area_weights(self) -> Grid:
```
```suggestion
    def spherical_area_weights(self) -> Grid:
```
spherical is an important hint that these are not 'real' weights.

src/ocean_emulators/utils/loss.py
```diff
@@ -28,34 +29,40 @@ def __call__(
...
+        target: torch.Tensor,
+        ctx: GridContext,
     ) -> torch.Tensor:
-        wet = wet.to(device=pred.device)
+        wet = ctx.label_mask.to(device=pred.device)
         pred = pred * wet
         target = target * wet
-        return loss_fn(pred, target)
+        if (
+            metric == "mse_cos_weighted"
```
jder: 🐑 This seems kinda sad. Just my 2 cents but I would either just delete this (unused) loss function or change decomposed_mse etc to all take ctx themselves and so be something like:

```
def decomposed_mse(pred: torch.Tensor, target: torch.Tensor, ctx: GridContext) -> torch.Tensor:
    """Standard MSE loss (l2) computed per channel."""
    pred, target = ctx.masked(pred, target)
    return F.mse_loss(pred, target, reduction="none").mean(dim=(0, 2, 3))
```
Which would also (IMO) nicely remove this extra indirection we have around two different kinds of function types here.

src/ocean_emulators/config.py
```diff
@@ -587,29 +568,35 @@ class FOMOConfig(BaseModelConfig):
     processor: UNetBackboneConfig = UNetBackboneConfig()
     # decoder will go here.
     embedding_dim: int = 128
+    use_bfloat16: bool = Field(
+        default=True,
+        description="Use bfloat16 for most layers rather than float32. Required for flash attention.",
```
jder, line 577: Can we get a nice error if we don't set this and request flash attention?

src/ocean_emulators/utils/data.py
```diff
@@ -85,6 +88,26 @@ def is_compact(self) -> bool:
...
 
+    @cached_property
+    def resolution(self) -> tuple[Lat, Lon]:
+        return (
+            torch.from_numpy(self.data.lat.values),
+            torch.from_numpy(self.data.lon.values),
+        )
+
+    @cached_property
+    def grid(self) -> tuple[int, int]:
```
jder: I would have expected something called grid to return a grid :). How about grid_size? And can we propagate that to the various other places we're calling this tuple a "grid"? Or is this just some terminology I'm not familiar with?

src/ocean_emulators/config.py
```diff
@@ -394,36 +393,24 @@ def build(
...
+        max_patch_size = patch_from(extent, max_lat, max_lon)
         return PerceiverEncoder(
             in_channels=in_channels,
             out_channels=out_channels,
-            patch_size=patch_size,
-            perceiver=self.perceiver.build(in_channels, out_channels, patch_size),
-            lat=lat,
-            lon=lon,
+            spatial_extent=extent,
+            perceiver=self.perceiver.build(in_channels, out_channels, max_patch_size),
```
jder, line 413: We're passing `max_patch_size` into an argument named `patch_size` which smells to me; is the argument actually `max_patch_size`?

src/ocean_emulators/config.py
```diff
@@ -394,36 +393,24 @@ def build(
...
-        ):
-            patch_size: tuple[int, int] = self.patch_size[0], self.patch_size[1]
-        elif isinstance(self.patch_size, int):
-            patch_size = self.patch_size, self.patch_size
-        else:
-            raise ValueError(
-                "`patch_size` must be either a scalar integer or a two-tuple of integers (height: int, width: int)."
-            )
-
+        assert len(self.spatial_extent) == 2, "spatial extent must be a pair of floats."
```
jder: :sheep: Could get an earlier error (+in editor) using the pydantic `min_length` / `max_length` validators in the Field above.

src/ocean_emulators/config.py
```diff
@@ -394,36 +393,24 @@ def build(
...
+    spatial_extent: list[float] = Field(
+        default=[6.0, 10.0],
+        description="Target physical extent of each patch in degrees [height_deg, width_deg]. "
+        "Patch sizes will be calculated to match this extent for each grid resolution.",
     )
     perceiver: PerceiverConfig = PerceiverConfig()
 
     def build(
-        self, in_channels: int, out_channels: int, lat: Lat, lon: Lon
+        self, in_channels: int, out_channels: int, max_lat: int, max_lon: int
```
jder: 🐑 "max lat" sounds to me like "largest lat" not "largest number of lats". How about max_lats? or max_lat_resolution?

src/ocean_emulators/config.py
```diff
@@ -545,21 +529,19 @@ def build(
...
-        wet: Grid,
-        area_weights: Grid,
-        static_data: xr.Dataset | None,
-        lat: Lat,
-        lon: Lon,
+        static_data_for_corrector: xr.Dataset | None,
+        srcs: list[DataSource],
     ) -> Samudra:
         corrector = None
+        src = srcs[0]
```
jder, line 540: Should we be asserting there's exactly 1 src?

src/ocean_emulators/utils/loss.py
```diff
@@ -28,34 +29,40 @@ def __call__(
...
+            and (lat := ctx.input_resolution[0]) is not None
```
jder: I think this probably should be label_resolution here? (Or else just delete this loss fn)

src/ocean_emulators/train.py
```diff
@@ -206,20 +194,17 @@ def __init__(self, cfg: TrainConfig) -> None:
             in_channels=self.num_in,
             out_channels=self.num_out,
             hist=cfg.data.hist,
-            wet=self.wet,
-            area_weights=self.area_weights,
-            static_data=self.static_data,
-            lat=torch.from_numpy(self.data.lat.values),
-            lon=torch.from_numpy(self.data.lon.values),
+            # TODO(559): This won't work at multiple scales. Refactor as part of src.
```
jder, line 197: Do we have an assert for corrector use while doing multi-scale somewhere?

jder (commented):
Thanks! Generally makes sense to me, basically all my comments are about naming and safety. Didn't look closely at the tests yet.

src/ocean_emulators/models/modules/encoder.py
alxmrs: Thanks, these are great ideas.

src/ocean_emulators/models/modules/encoder.py
alxmrs: Thanks, this could be worded clearer.

alxmrs: Changed this sentence and added another one to clarify the intent.

src/ocean_emulators/models/modules/encoder.py
alxmrs: Yes, this was obvious to me and me alone. Thanks, `patch_extent` is much clearer.

src/ocean_emulators/models/fomo.py
alxmrs: I don't like it either, but ModuleDict only supports keys that are strings: https://docs.pytorch.org/docs/stable/generated/torch.nn.ModuleDict.html

src/ocean_emulators/models/fomo.py
alxmrs: Yeah, agreed. I liked that in the Samudra model.

src/ocean_emulators/utils/data.py
alxmrs: No, not a new terminology. I was having a hard time finding succinct names for things. Yes, `grid_size` is more accurate.

src/ocean_emulators/utils/loss.py
alxmrs: Thanks, I've been keen to delete this loss fn.

src/ocean_emulators/config.py
alxmrs: Good shout.

src/ocean_emulators/constants.py
alxmrs: I was torn about this. Ideally, this would go to `util.train`, but that causes hairy circular imports. Since this is such a central thing, we might as well create a new utility file.

src/ocean_emulators/constants.py
alxmrs: We don't need to move `input_resolution`, so I haven't. IIUC, doing so might cause device errors in the _other_ direction, where it's expected that the resolution should be on CPU and it's not. I'll hold off on moving this to device until it's needed.

src/ocean_emulators/train.py
alxmrs: Good idea.

alxmrs: I added an assert in Samudra to protect against training it at multiple scales. So, this should covert the Corrector case.

src/ocean_emulators/train.py
```diff
@@ -610,23 +599,22 @@ def _call_loss_update(self, data: TrainData):
...
 
+        # TODO(alxmrs): Aggregator only supports a single scale.
         val_aggregator = Aggregator.get_validation_aggregator(
-            self.metadata,
+            self.primary_src.metadata,
             self.hist,
-            self.area_weights,
-            self.src.masks.prognostic.to(self.device),
+            self.primary_src.area_weights.to(self.device),
+            self.primary_src.masks.prognostic.to(self.device),
```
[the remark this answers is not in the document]
alxmrs: This is fixed in #569.

src/ocean_emulators/constants.py
jder: ok, but I think this is a minor footgun to have a `to` method which does not move the whole object to the specified device.

jder:
@alxmrs LMK when I should take another look

alxmrs:
I'm sorry! I thought I clicked the "ask for another view" button. These sources are ready for another review.

alxmrs:
Wait! Let me address the GPU issue, just noticed that.

.github/workflows/test-gpu.yml
```diff
@@ -50,8 +50,10 @@ jobs:
       - name: Install dependencies
         run: uv sync --locked --dev --extra cuda
 
+      # Each test runs with a 5-minute timeout. The `-x` flag will stop on the first failure or hang.
       - name: Run pytest
-        run: uv run --locked pytest -n 2 -v -m "not manual and cuda"
+        run: uv run --with pytest-timeout --locked pytest -v -m "not manual and cuda" --timeout=300 --timeout-method=signal -x
+        timeout-minutes: 30
```
alxmrs, line 56: I'm happy to revert this change, but it might be nice to have in the future?

src/ocean_emulators/datasets.py
```diff
@@ -60,12 +61,15 @@ def __init__(
         long_rollout,
     ):
         super().__init__()
-        self.device = get_device()
+        # NOTE: Keep tensors on CPU during initialization. This allows the dataset
```
alxmrs, line 64: Note: this is a recent change to solve a device related issue.

src/ocean_emulators/stepper.py
```diff
@@ -116,6 +116,7 @@ def inference(
                 f"Inference [epoch {epoch}]: loop {loop} of {num_loops - 1}. "
                 f"Stepping {num_steps} steps forward."
             )
+            dataset.ctx.to(get_device())
```
alxmrs: Note: this is a recent change that solved a device related issue.

src/ocean_emulators/datasets.py
```diff
@@ -98,15 +102,29 @@ def __init__(
...
 
+    def to(self, device: torch.device) -> "InferenceDataset":
+        """Move the dataset's context tensors to the specified device.
+
+        Call this before using the dataset for inference to ensure tensors
+        are on the correct device (GPU).
+        """
+        self.ctx = self.ctx.to(device)
+        self.wet_label = self.wet_label.to(device, non_blocking=True)
+        return self
```
alxmrs, line 126: This is a recent change to pass GPU tests.

jder:
@codex review

.github/workflows/test-gpu.yml
jder: Let's keep it, sounds great. Do you know how much headroom 5 minutes gives us for the trainer tests?

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -23,14 +37,17 @@ class PerceiverEncoder(nn.Module):
...
+    our `forward` call requires supplementary information: the resolution (a pair of Lat/Lon tensors), which is used to
+    make consistent positional encodings for patches across different scales. While higher resolution scales will
+    contain more data per patch, the patch will refer to the same physical area on Earth as all other scales.
+
     Args:
         in_channels (int): the number of input channels (roughly:  time x variable x (surface + depths)).
         out_channels (int): size of the latent dimension (aka, the embedding dimension).
-        patch_size (int | tuple[int, int]): the size of the patches to embed. Patches must evenly divide the input grid.
-          If a tuple is supplied, then it represents the (height, width) of the patches to embed.
+        patch_extent (tuple[float, float]): spatial extent of each patch.
```
```suggestion
        patch_extent (tuple[float, float]): spatial extent of each patch measured in degrees of lat/lon.
```

src/ocean_emulators/utils/ctx.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Attributes:
+        label_mask: Boolean mask indicating valid ocean cells for each prognostic
+            variable. Shape: (num_prognostic_vars, lat, lon). Land cells are False.
+        input_resolution: Tuple of (latitude, longitude) coordinate tensors defining
+            the spatial grid. Used for cosine-latitude weighting in loss functions.
+    """
+
+    label_mask: PrognosticMask
+    input_resolution: tuple[Lat, Lon]
```
jder, line 25:
```suggestion
    input_resolution_cpu: tuple[Lat, Lon]
```

jder (approved):
Thank you! I'm gonna grab this code and kick off a "real" training run on lambda with samudra just to verify that things look OK. But otherwise this looks great to me now.

jder:
Smoke test is running here https://wandb.ai/ocean_emulators/default?nw=cbyjiox4nu9 -- the _smoke run should be comparable to the other one (same settings, data, hardware)

.github/workflows/test-gpu.yml
alxmrs: Let me check the logs and see if I can find out!

alxmrs:
Thanks for running the smoke test! I think it should fail in a predictable way at the end of the first epoch, but my hope is that #569 fixes this.

alxmrs:
Thanks for your thorough feedback on all the related PRs!

jder:
To be clear, the smoke test is a Samudra run, not FOMO. It seems to be going fine in epoch 14. Comparable memory usage, CPU usage, and accuracy. So I'd say it's good.

jder:
(btw the smoke test code includes merging in the memory fix from #576)

alxmrs:
Oh, that makes sense. Glad the smoke test is going well then! I'm glad this is backwards compatible.

.github/workflows/test-gpu.yml
alxmrs: I didn't measure, but reading the logs in real time, I can provide some anecdata: the two step trainer runs really fast. I think we have a lot of headroom, given that we allow 5 minutes per test.

PR #533, opened 2025-12-30
Make cached benchmark datasets match our training data: uncompressed.

alxmrs:
From my local cprofile traces, this should cut our benchmark times in half. A significant amount of time spent in our local profiles is on decompression, which we don't do in our real training runs.

After this PR is merged, we will need to delete our caches once to regenerate them: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches#deleting-cache-entries


jder (approved):

jder:
I also deleted all the test data caches. Will keep an eye on CI.

alxmrs:
Thank you!

Issue #410, opened 2025-10-03
Train on quarter-degree data


jder:
We should be able to replicate our half-degree training run on quarter-degree data.

Might need to tweak batch sizes (and correspondingly learning rate) to fit this in memory. 

- [x] Compute basins for quarter-degree, 
- [x] Run eval/viz for initial train run

runs: [train](https://wandb.ai/ocean_emulators/default/runs/9ei4iwxr), [eval](https://wandb.ai/ocean_emulators/default/runs/tkn5rd98), [viz](https://drive.google.com/drive/folders/1uUo7T7kBJMZjjEYgOW8J0aUYy3msvONF)

jder:
@amogh-gulati I believe you were planning to skip this one for now, right? If so I think I will give it a shot myself. (LMK if I misunderstood)

amogh-gulati:
@jder yes, I will be working on CRPS loss tomorrow

jder:
ok, so making use of a variety of new features that have recently landed & seem to produce reasonable outputs (bfloat16, gradient accumulation, instance norm), I have successfully kicked off a training run:

https://wandb.ai/ocean_emulators/default/runs/9ei4iwxr

```
AWS_PROFILE=skypilot uv run sky launch skypilot/train.sky.yaml --env WANDB_API_KEY --env CONFIG=configs/samudra_om4/train.yaml --env NAME=2025-12-19-samudra_om4_quarterdeg --env AWS_PROFILE --env OSN_DATA_PATH=emulators/am16581/data/2025-11/om4_quarterdeg_v2/ --env OSN_KEY_ID --env OSN_KEY_SECRET --env ARGS='--model.use_bfloat16=true  --gradient_accumulation_steps=4  --model.unet.core_block.norm=instance --batch_size=1'
```

It appears to be quite slow, with a fairly high variance in train step. I'm guessing data loading is the problem given that variance was due to individual GPUs not keeping up with the rest previously.

[...]

PR #537, opened 2026-01-08
Fixes for quarterdev eval and fix

jder:
This will be somewhat wasteful for half-/one-degree viz but seems better to work consistently and not worry too much about slight cost increases. See context in #410 for figuring these things out.


alxmrs (approved):
LGTM

PR #539, opened 2026-01-08
Make installs work on ARM/GH nodes (PyTorch 2.9 and CUDA 13)

jder:
Updates our pyproject to pull correct pytorch packages for arm machines using cuda 13. Unfortunately GPUs with such machines are not really available for CI, so we set up a scrub skill for fomo bot to run tests locally each night based on the image. We do run CPU tests at least in CI as normal. 

The flash attention packages we're using are not compatible with CUDA 13, so we filed #710 to track this but I would like to at least be able to start publishing & testing Samudra builds.


jder:
ok current state is that this kinda seems like it would work except we kill the instance when building flash-attention and the runner dies along with it. Feels like we could probably solve this somehow but building FA on each CI run is also probably not the right call. Looking into using [this docker image](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/physicsnemo/containers/physicsnemo?version=25.11) which is probably what we'll be running on in production anyway and includes FA.

jder:
ok, pulling the [PhysicsNEMO container](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/physicsnemo/containers/physicsnemo?version=25.11) which has FA and seems to work well unfortunately takes like 25 minutes to download. I guess we have to make our own AMI?

jder:
let's try:

* -> [Deep Learning ARM64 Base OSS Nvidia Driver GPU AMI \(Ubuntu 24.04\) 20260123 - AWS Deep Learning AMIs](https://docs.aws.amazon.com/dlami/latest/devguide/aws-deep-learning-ami-gpubaseossarm64-ul2404-2026-01-27.html)
  * -> X86 here: [Deep Learning Base AMI with Single CUDA \(Ubuntu 22.04\) 20260120 - AWS Deep Learning AMIs](https://docs.aws.amazon.com/dlami/latest/devguide/aws-deep-learning-ami-gpubasesinglecuda-ul2204-2026-01-21.html)

And send feedback to NVIDIA folks if they don't meet our needs somehow

jder:
@fomo-bot can you pull this branch and verify it works locally with your GPU before updating to the latest main?

jder:
@fomo-bot can you rebase this on the latest main and verify it lets you use CUDA locally? Then let's change the strategy here so arm tests are just run in a container like the current container tests (post-main-merge)

jder:
@fomo-bot great. Looks like ci is failing. Please fix and then verify the container build & both arch tests pass

jder:
@fomo-bot is there a different AWS instance we could use which would be compatible? What is the exact incompatibility between?

jder:
Let's look at modal or others.

jder:
Also it seems like FA might not actually be working with cuda 13 after updating to main.

jder:
@fomo-bot ok, new plan. It just does not seem possible to get cloud ARM machines with modern NVIDIA GPUs. So let's do this:

* Verify locally on your arm machine that the GPU-based tests do in fact work with this branch.
* Spin up an ARM non-GPU runner to build and push containers at the same time as we do for x86. Run CPU-based tests only (otherwise following the x86 logic of when to run tests in-container)
* Set up a nightly scrub to pull the latest container from main and run GPU-based tests.

jder:
@fomo-bot I meant that you should pull & run a nightly test with a scrub skill, not using GH CI.

jder:
@fomo-bot you should now be able to use docker locally, please confirm you can pull an arm image down and run the gpu tests

.github/workflows/container-physicsnemo.yml
```diff
@@ -90,21 +92,66 @@ jobs:
...
+            --filters \
+              "Name=name,Values=${ARM_UBUNTU_NAME_FILTER}" \
+              "Name=architecture,Values=arm64" \
+              "Name=virtualization-type,Values=hvm" \
+              "Name=root-device-type,Values=ebs" \
+              "Name=state,Values=available" \
+            --query 'reverse(sort_by(Images, &CreationDate))[:1].ImageId' \
+            --output text)"
+          echo "ami_id=${ami_id}" >> "${GITHUB_OUTPUT}"
+          echo "aws_region=${AWS_REGION}" >> "${GITHUB_OUTPUT}"
```
jder: This seems a bit fragile to me, I would rather run this once and hard-code the resulting AMI for future runs. Maybe leave a comment about how to get a new one?

containers/Dockerfile.physicsnemo-25.11
```diff
@@ -7,7 +7,9 @@ ARG VCS_REF=unknown
 ARG VCS_URL=unknown
 ARG BUILD_DATE=unknown
 ARG NSYS_CLI_VERSION=2026.1.1.204-3717666
-ARG NSYS_CLI_DEB_URL="https://developer.download.nvidia.com/devtools/nsight-systems/NsightSystems-linux-cli-public-${NSYS_CLI_VERSION}.deb"
+ARG NSYS_CLI_AMD64_DEB_URL="https://developer.download.nvidia.com/devtools/nsight-systems/NsightSystems-linux-cli-public-${NSYS_CLI_VERSION}.deb"
```
jder: Let's inline the version since otherwise it is deceptive (since aarch doesn't use it)

containers/Dockerfile.physicsnemo-25.11
```diff
@@ -7,7 +7,9 @@ ARG VCS_REF=unknown
...
+ARG NSYS_CLI_ARM64_DEB_URL="https://developer.download.nvidia.com/devtools/nsight-systems/nsight-systems-cli-2026.1.1_2026.1.1.204-1_arm64.deb"
```
jder, line 10: Looks like `[nsight-systems-cli-2026.1.1-2026.1.1.204_3717666-0.aarch64.rpm](https://developer.download.nvidia.com/devtools/nsight-systems/nsight-systems-cli-2026.1.1-2026.1.1.204_3717666-0.aarch64.rpm)` might be better since it matches the x86 version?

scripts/container/check_cuda_compat.py
jder, line 1: I think we can delete this file? Is it used?

scripts/container/scrub_arm64_gpu_container_tests.sh
```diff
@@ -0,0 +1,68 @@
...
+    GHCR_OWNER="${BASH_REMATCH[1]}"
+  else
+    GHCR_OWNER="Open-Athena"
+  fi
+fi
+
+owner_lower="$(echo "${GHCR_OWNER}" | tr '[:upper:]' '[:lower:]')"
+if [[ -z "${IMAGE_TAG}" ]]; then
+  IMAGE_TAG="ghcr.io/${owner_lower}/ocean-emulator-physicsnemo:25.11-arm64-latest"
+fi
```
jder: Let's declare `DOCKER_REPO=ghcr.io/Open-Athena` rather than trying to detect it.

scripts/container/scrub_arm64_gpu_container_tests.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
```
jder, line 1:
```suggestion
#!/usr/bin/env bash
# This script is run on our local arm64 GPU box nightly in order to smoke test our latest docker image.
```

src/ocean_emulators/config.py
```diff
@@ -823,7 +834,7 @@ class FOMiniConfig(BaseModelConfig):
     perceiver_implementation: PerceiverImpl = Field(
         default="auto",
         description="Perceiver attention implementation for the single PerceiverIO model. "
-        "'auto' selects flash attention when CUDA is available, otherwise naive.",
+        "'auto' selects flash attention when CUDA and the flash dependencies are usable, otherwise naive.",
```
jder: I really don't want to make this change to silently default to a slow implementation if the dependency is broken. Can we instead avoid running these flash-attention-requiring tests in the arm image and make an issue to fix FA-on-arm and re-enable them?

jder (commented):
@fomo-bot a few comments. please fix, push, and then verify the e2e publish/pull/test flow works locally

containers/Dockerfile.physicsnemo-25.11
```diff
@@ -6,8 +6,9 @@ WORKDIR /workspace
 ARG VCS_REF=unknown
 ARG VCS_URL=unknown
 ARG BUILD_DATE=unknown
-ARG NSYS_CLI_VERSION=2026.1.1.204-3717666
-ARG NSYS_CLI_DEB_URL="https://developer.download.nvidia.com/devtools/nsight-systems/NsightSystems-linux-cli-public-${NSYS_CLI_VERSION}.deb"
+ARG NSYS_CLI_AMD64_DEB_URL="https://developer.download.nvidia.com/devtools/nsight-systems/NsightSystems-linux-cli-public-2026.1.1.204-3717666.deb"
+ARG NSYS_CLI_ARM64_RPM_URL="https://developer.download.nvidia.com/devtools/nsight-systems/nsight-systems-cli-2026.1.1-2026.1.1.204_3717666-0.aarch64.rpm"
```
jder: @fomo-bot sorry, my bad. I meant to link to the appropriate DEB, not an RPM. You had it right the first time with https://developer.download.nvidia.com/devtools/nsight-systems/nsight-systems-cli-2026.1.1_2026.1.1.204-1_arm64.deb. Please return to that.

jder: @fomo-bot yeah please merge with the remote branch (I just merged in main)

containers/README.md
```diff
@@ -19,8 +19,48 @@ On CI, GitHub Actions builds and validates this image via:
...
+ARM64 EC2 runner to build, publish, pull, and CPU-test an ARM64 image. ARM64 GPU
+testing is intentionally not attempted on cloud runners because the available small
+ARM GPU instances expose older NVIDIA T4G GPUs that are unsupported by the current
+PyTorch/CUDA stack.
+
+The scheduled ARM GPU scrub is defined as a repo-local harness scrub skill, not a
+GitHub Actions workflow:
+
+```text
+.agents/skills/scrub-physicsnemo-arm64-gpu-container/SKILL.md
```
alxmrs, line 34: It's still kinda crazy to me that agents skills are dependable enough to be this load bearing in our infra. Magic.

src/ocean_emulators/config.py
```diff
@@ -398,7 +398,7 @@ def build(
         if _use_flash(implementation):
             try:
                 from flash_perceiver import Perceiver as FlashPerceiver  # type: ignore
-            except ModuleNotFoundError as e:
+            except ImportError as e:
```
alxmrs, line 401: Nice, this is a better error.

pyproject.toml
```diff
@@ -133,3 +133,16 @@ overrides = [
...
+[tool.uv.sources]
+torch = [
+  { index = "pytorch-cu130", marker = "sys_platform == 'linux' and platform_machine == 'aarch64'" },
+]
+torchvision = [
+  { index = "pytorch-cu130", marker = "sys_platform == 'linux' and platform_machine == 'aarch64'" },
+]
+
+[[tool.uv.index]]
+name = "pytorch-cu130"
```
alxmrs, line 146: Slick uv config.

alxmrs (approved):
As far as I got today, got side tracked. Happy to stamp for now, if I had more time to review, I'd try to understand the bash scripts at another level of depth. At a high level, this LGTM.

containers/README.md
jder: I mean I think we'll find out!

.github/workflows/container-physicsnemo.yml
```diff
@@ -181,9 +211,115 @@ jobs:
...
+          sudo systemctl enable --now docker || sudo service docker start
+          sudo chmod 666 /var/run/docker.sock
+          docker version
+          docker buildx version
+
+      - name: Cache test data
+        uses: actions/cache@v3
+        with:
+          path: .data_cache/
+          key: ${{ runner.os }}-${{ runner.arch }}-test-data-${{ hashFiles('tests/conftest.py', 'configs/test/*.yaml') }}
```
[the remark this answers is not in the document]
jder: @fomo-bot can you confirm this happens by making a dummy .data_cache and seeing if the git status line in fact fails? If so please move. Then go ahead and merge this PR once green.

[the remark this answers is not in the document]
jder: (by which I mean, move the line in this file, not move the data cache itself)

PR #540, opened 2026-01-09
Introduce a cap on max ratio of DMSE weights 

jder:
"Unlimited" DMSE seems to produce unwanted artifacts in rollouts due to increased variance. This lets us trade off between the positive effects of DMSE (reducing imprinting) and these negative ones. More context: https://openathena.slack.com/archives/C08CYM42DT3/p1766175789740149?thread_ts=1765824627.046769&cid=C08CYM42DT3

Note that the previous limit by stddev was in fact broken (we intended to limit by temporal stddev, not spatial as was written; again see details in the above thread).

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -658,9 +658,9 @@ TrainBackendConfig = Literal["cpu", "cuda", "nccl", "auto"]
 class DynamicLossConfig(pydantic.BaseModel):
     type: Literal["dynamic"] = "dynamic"
     metric: LossMetric = "mse"
-    limit: bool = Field(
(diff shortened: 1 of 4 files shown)


adam-subel (approved):
Under these changes, would it make sense to drop stds as an input and use something more general to get the structure of the weights? If not, is it worth maintaining the option to limit either with a ratio, max val or both? I think that the ratio is plenty, but feels like maybe worth maintaining some of that functionality. 

Otherwise looks good as is.

jder:
Yeah, fair point we no longer need to pass stds here, will fix. I think we can leave the ratio -- we can pretty easily add in other options later if they turn out to be helpful.

PR #542, opened 2026-01-10
skypilot train must use distributed backend

jder:
When I tried to force a non-CPU backend here I forgot that we separate "cuda" from "nccl". As a result our runs were training individually on each GPU (cc @alxmrs I think this also affected your recent multi-scale runs)

also fix a typo that's been bugging me :)


alxmrs (approved):
LGTM

PR #549, opened 2026-01-16
Fix indexing bug in dynamic mse

### LINKED ISSUE #548: Figure out why DMSE down-weights thetao
Context here: https://openathena.slack.com/archives/C08CYM42DT3/p1768491216543889?thread_ts=1766414069.316049&cid=C08CYM42DT3:

> Fundamentally I was confused why scaling thetao loss up would reduce the quality of thetao predictions as it appears to here. So I looked into it a bit more an in fact we are reducing how much we pay attention to the upper ocean heat in the dynamic MSE models. In the unlimited version the upper ocean heat is scaled to 1.7% of the total loss (which is about 8x less than it is represented in the unscaled wide-only model). In the 100-limit it is scaled to 2.8% of the total loss. And if you look at the actual loss estimates this checks out, too. The actual loss for upper ocean heat is significantly worse in the dynamic loss model and somewhat improved by the 100-limit version.
> My best guess is that the reason we’re seeing a reduction in imprinting is that uo is scaled even less (0.09% of total loss). vo and salinity are both scaled way up.

@adam-subel and I discussed. He noticed that basically all runs since we implemented dmse had this same phenomena, except for the the 2-in-1-out run from #519. Could be related to the reduction in the even-odd artifact there or could just be there's a bug in history handling this removes. We are also curious if this is ~nondeterministic since if you happen to start increasing the loss scale for some channel, that could cause a feedback loop which weights that channel more and more. Or it could a bug. So we're going to:

- [x] Try running the 2-in-1-out code but with 2-in-2-out (@jder)
- [x] Try initializing a dmse run with "expected" weights (ie deep thetao upweighted) and see if they stick or if they decay back to the scalings we see (@jder)
- [x] Look more into the 2-in-1-out behavior to see if it's clear why the thetao losses would be so different. (@adam-subel)

jder:
Fix bug found when investigating artifact in DMSE runs

closes #548

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -557,11 +557,11 @@ class Trainer:
                     self.loss_fn, "loss_scale_per_channel", None
                 ):
                     loss_scale_per_channel = loss_scale_per_channel_fn()
-                    # Reshape from channels * history to channels
-                    # by averaging along the `hist` dimension
+                    # Reshape from time-major channels to [hist, var] and
+                    # average along the history dimension.
                     loss_per_channel = TO.loss_per_channel.reshape(
-                        loss_scale_per_channel.shape[0], -1
-                    ).mean(dim=1)
+                        -1, loss_scale_per_channel.shape[0]
+                    ).mean(dim=0)
 
                     unscaled_loss_per_channel = (
                         loss_per_channel / loss_scale_per_channel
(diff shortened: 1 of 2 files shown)


adam-subel (commented):
Looks good! Hopefully this fix helps things

alxmrs (approved):
Wow, this is a really subtle bug. I keep wondering if some sort of type information on the arrays could have made this easier to detect by humans, but I think that might just kick the can down the road. I'm glad we have these text processors who can detect these detailed issues.

jder:
> I keep wondering if some sort of type information on the arrays could have made this easier to detect by humans

Yeah, this is the same kind of thing I've been mulling over -- you want something like jaxtyping or haliax, but that knows about dimensions you've flattened out for training purposes.

PR #551, opened 2026-01-21
Fix colorbars 

jder:
These runs are the same data but with the new color bars: [before](https://drive.google.com/drive/u/0/folders/1Kh32QlzLVNOkcpBm9YsOoInscQwselak), [after](https://drive.google.com/drive/u/0/folders/1inz0VAHbQNXg6NIoyD0jM3CwymXyWc8i). 

For example, the SST_Global_map:

before

after

Note the snapshots are no longer blown out over most of the domain and the bias map is better scaled in the other direction. cc @adam-subel

### DIFF
--- a/src/ocean_emulators/viz/core.py
+++ b/src/ocean_emulators/viz/core.py
@@ -23,6 +23,7 @@ from dask.array.core import Array as DaskArray
 from dask.base import compute
 from dask.delayed import delayed
 from dask.diagnostics.progress import ProgressBar
+from matplotlib import colors
 from matplotlib.ticker import FixedLocator, MaxNLocator, ScalarFormatter
 from tqdm.auto import tqdm
 from xarrayutils.plotting import box_plot, linear_piecewise_scale  # type: ignore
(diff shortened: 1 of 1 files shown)


jder:
@codex review please

jder:
@codex review

jder:
@codex please review again

src/ocean_emulators/viz/core.py
```diff
@@ -37,6 +38,62 @@
...
+        return np.array([], dtype=float)
+    else:
+        arr = data
+
+    if isinstance(arr, DaskArray):
+        arr = arr.compute()
+    return np.asarray(arr).ravel()
+
+
+def symmetric_percentile_norm(
```
alxmrs, line 61: 🐑  Could we make this method call `percentile_norm`? Would be less to maintain.

src/ocean_emulators/viz/core.py
```diff
@@ -37,6 +38,62 @@
...
+        vmax = np.percentile(flat, upper)
+        if not np.isfinite(vmin) or not np.isfinite(vmax):
+            vmin = 0.0
+            vmax = fallback
+        elif vmin == vmax:
+            vmin = vmin - fallback
+            vmax = vmax + fallback
+        elif vmin > vmax:
+            vmin, vmax = vmax, vmin
+    return colors.Normalize(vmin=vmin, vmax=vmax)
```
alxmrs, line 94: TIL about norms in matplotlib. Neat.

src/ocean_emulators/viz/core.py
```diff
@@ -2520,25 +2563,19 @@ def plot_sst(self, ax, sst_data, title, i):
             gl.left_labels = False
         return im
 
-    def plot_bias(
-        self, ax, sst_data, gt_sst_data, title, range: tuple[float, float] | None
-    ):
+    def plot_bias(self, ax, data, gt_data, title):
```
alxmrs, line 2566: Nice that these are now generic (not specific to one variable).

alxmrs (approved):
One minor suggestion. Otherwise, this PR LGTM. Thanks for improving the plots!

src/ocean_emulators/viz/core.py
jder: The names are maybe unclear, but the behavior is somewhat different -- symmetric_percentile_norm is 98% on the upper side and the negative of that on the lower (ie symmetric around zero). You can't express that will a call to percentile_norm AFAICT.

Issue #506, opened 2025-12-08
Verify we need periodic upscaling and enable if so


jder:
Ai2 did not see a need for this in their models and made a different change with circular padding: https://openathena.slack.com/archives/C09FXFDR2BF/p1762200012903659

We should verify whether we need the fancy periodic upscaling and if so enable it, or remove it.

jder:
Ok, so we have the padding fix already which James did in the linked thread. In our 1-degree outputs I'm pretty sure there's still a mild artifact. In 1/2-degree outputs I don't see anything; perhaps it's just too small or gets learned away with more data. Trying a run with a simpler version of the periodic upscaling:

```
AWS_PROFILE=skypilot uv run sky launch skypilot/train.sky.yaml --env WANDB_API_KEY --env CONFIG=configs/samudra_om4/train.yaml --env NAME=2026-01-21-samudra_om4_onedeg_periodic --env AWS_PROFILE --env OSN_DATA_PATH=emulators/am16581/data/2025-11/om4_onedeg_v3/ --env OSN_KEY_ID --env OSN_KEY_SECRET
```

jder:
ok, that one was actually not using the periodic upscaling and has the artifact, trying again. 

```
AWS_PROFILE=skypilot uv run sky launch skypilot/train.sky.yaml --env WANDB_API_KEY --env CONFIG=configs/samudra_om4/train.yaml --env NAME=2026-01-22-samudra_om4_onedeg_periodic --env AWS_PROFILE --env OSN_DATA_PATH=emulators/am16581/data/2025-11/om4_onedeg_v3/ --env OSN_KEY_ID --env OSN_KEY_SECRET
```

jder:
ok, this one in fact has no artifact whereas the identical run just before from yesterday with the existing upscaling does. I think we should enable this!

PR #554, opened 2026-01-22
Enable zonally-periodic upsampling for non-v1 runs

jder:
I re-ran some comparisons in #506 and can confirm that at least at 1-degree, the bilinear upsample has an artifact at lon=0 but this one does not. Here is an example of a (doctored to increase contrast) run with the artifact in a zos snapshot:
Screenshot 2026-01-22 at 5 10 33 PM

And the next identical one with the new upsampling and no artifact:
Screenshot 2026-01-22 at 5 11 32 PM

I also replaced the upsampling with a simpler approach that does less funny stuff and is hopefully more obviously correct.

### DIFF
--- a/src/ocean_emulators/models/modules/blocks.py
+++ b/src/ocean_emulators/models/modules/blocks.py
@@ -60,28 +60,20 @@ class ZonallyPeriodicBilinearUpsample(torch.nn.Module):
         self.scale_h, self.scale_w = upsampling
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
-        # First upsample along latitude/height using standard "bilinear" interpolation.
-        # (This is just linear interpolation but pytorch requires us to use "bilinear" for
-        # 2d data.)
-        up_height = torch.nn.functional.interpolate(
-            x,
-            scale_factor=(self.scale_h, 1),
+        # Upsample with periodic padding along longitude to avoid seams and
+        # keep interpolation aligned with PyTorch's bilinear sampling grid.
+        width = x.shape[-1]
+        padded = torch.nn.functional.pad(x, (1, 1, 0, 0), mode="circular")
+        upsampled = torch.nn.functional.interpolate(
+            padded,
+            scale_factor=(self.scale_h, self.scale_w),
             mode="bilinear",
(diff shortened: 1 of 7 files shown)


src/ocean_emulators/models/modules/blocks.py
```diff
@@ -60,28 +60,20 @@ def __init__(self, upsampling: int | tuple[int, int] = 2):
...
-        # (This is just linear interpolation but pytorch requires us to use "bilinear" for
-        # 2d data.)
-        up_height = torch.nn.functional.interpolate(
-            x,
-            scale_factor=(self.scale_h, 1),
+        # Upsample with periodic padding along longitude to avoid seams and
+        # keep interpolation aligned with PyTorch's bilinear sampling grid.
+        width = x.shape[-1]
+        padded = torch.nn.functional.pad(x, (1, 1, 0, 0), mode="circular")
+        upsampled = torch.nn.functional.interpolate(
```
alxmrs, line 67: While I like the striding trick on the left, I'm happy that we're delegating the sampling to this interpolate fn. It's more trustworthy for sure.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -60,28 +60,20 @@ def __init__(self, upsampling: int | tuple[int, int] = 2):
...
-        # First upsample along latitude/height using standard "bilinear" interpolation.
-        # (This is just linear interpolation but pytorch requires us to use "bilinear" for
-        # 2d data.)
-        up_height = torch.nn.functional.interpolate(
-            x,
-            scale_factor=(self.scale_h, 1),
+        # Upsample with periodic padding along longitude to avoid seams and
+        # keep interpolation aligned with PyTorch's bilinear sampling grid.
+        width = x.shape[-1]
+        padded = torch.nn.functional.pad(x, (1, 1, 0, 0), mode="circular")
```
alxmrs, line 66: A future research issue -- could be promising to look at reflective padding for the top/bottom (last two values).

alxmrs (approved):
LGTM! Just a few random thoughts.

PR #555, opened 2026-01-23
Apply wet mask to loss function late.

alxmrs:
To prepare for multi-scale training, we need to properly handle masking. In our loss function, making was previously done eagerly. This won't work when training on multiple scales, because the output resolution will differ based on the training schedule. Thus, this change applies masking late. To accomplish this, we make the label mask a primary property of TrainData and RawTrainData. 

This supersedes #531.

### DIFF
--- a/src/ocean_emulators/stepper.py
+++ b/src/ocean_emulators/stepper.py
@@ -1,5 +1,6 @@
 import logging
 from collections.abc import Callable
+from functools import partial
 from os import PathLike
 
 logger = logging.getLogger(__name__)
@@ -27,7 +28,7 @@ class Stepper:
     def train_batch(
         model: torch.nn.Module, batch: TrainData, loss_fn: Callable
     ) -> TrainBatchOutput:
-        loss_per_channel = model(batch, loss_fn=loss_fn)
+        loss_per_channel = model(batch, loss_fn=partial(loss_fn, wet=batch.label_mask))
         loss = torch.mean(loss_per_channel)
         return TrainBatchOutput(loss, loss_per_channel)
 
@@ -49,7 +50,7 @@ class Stepper:
             else model
         )
         outs = model.forward_once(input)
-        loss_per_channel = loss_fn(outs, label)
+        loss_per_channel = loss_fn(outs, label, wet=batch.label_mask)
         loss = torch.mean(loss_per_channel)
         return ValBatchOutput(loss, loss_per_channel, input, label, outs)
 
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -74,7 +74,7 @@ from ocean_emulators.utils.logging import (
     handle_logging,
     handle_warnings,
 )
-from ocean_emulators.utils.loss import LossFn
+from ocean_emulators.utils.loss import LossFnWithMask
 from ocean_emulators.utils.train import (
     CheckpointPaths,
     collate_inference_data,
@@ -217,10 +217,9 @@ class Trainer:
         self.network = self.model.__class__.__name__
 
         # Loss function
-        self.loss_fn: LossFn = build_loss_fn(
+        self.loss_fn: LossFnWithMask = build_loss_fn(
             cfg.loss,
-            wet=self.wet,
-            y_coord=self.data.lat,
+            y_coord=self.data.lat,  # TODO(alxmrs): Needs to be label lat vals. Remove?
             device=self.device,
             num_channels=self.N_prog,
             pad_mode=cfg.model.pad,
@@ -507,7 +506,8 @@ class Trainer:
                 r = self.gradient_accumulation_steps
 
             if self.num_batches_seen == 0:
-                get_model_summary(self.model, data, self.debug)
+                # Pass None since TrainData is not compatible with torchinfo's traversal
+                get_model_summary(self.model, None, self.debug)
 
             TO: TrainBatchOutput = Stepper.train_batch(self.model, data, self.loss_fn)
 
@@ -611,12 +611,14 @@ class Trainer:
         # This is a separate function to ensure locals are dropped immediately after use
         if update := getattr(self.loss_fn, "update", None):
             with torch.no_grad():
-                single_step_data = TrainData(data.num_prognostic_channels)
+                single_step_data = TrainData(
+                    data.num_prognostic_channels, data.label_mask
+                )
                 # Each entry in data is one step in a rollout.
                 input, label = data[0]
                 single_step_data.append(input, label)
                 pred = self.model(single_step_data)
-                update(pred[0], label)
+                update(pred[0], label, wet=data.label_mask)
 
     def validate_one_epoch(self, epoch):
         self.model.eval()
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -506,8 +506,7 @@ class Trainer:
                 r = self.gradient_accumulation_steps
 
             if self.num_batches_seen == 0:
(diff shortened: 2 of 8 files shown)


jder:
@codex review

src/ocean_emulators/stepper.py
```diff
@@ -27,7 +28,7 @@ def __init__(self):
     def train_batch(
         model: torch.nn.Module, batch: TrainData, loss_fn: Callable
     ) -> TrainBatchOutput:
-        loss_per_channel = model(batch, loss_fn=loss_fn)
+        loss_per_channel = model(batch, loss_fn=partial(loss_fn, wet=batch.label_mask))
```
jder, line 31: 🐑 `loss_fn` is a `LossFnWithMask` here (could be more tightly typed in the arg), right? Are we sure we want to assume the arg name is wet? ie would it be better to do `model(batch, loss_fn=lambda pred, target: loss_fn(pred, target, batch.label_mask)`? (As you know I would almost always prefer a lambda to a partial because the latter prevents any kind of typing analysis like this)

src/ocean_emulators/stepper.py
```diff
@@ -49,7 +50,7 @@ def validate_batch(
             else model
         )
         outs = model.forward_once(input)
-        loss_per_channel = loss_fn(outs, label)
+        loss_per_channel = loss_fn(outs, label, wet=batch.label_mask)
```
jder, line 53: 🐑 Same question here -- if LossFnWithMask always has a `wet` arg let's at least document that if we can't enforce it.

src/ocean_emulators/train.py
```diff
@@ -501,7 +500,8 @@ def train_one_epoch(self, epoch):
                 r = self.gradient_accumulation_steps
 
             if self.num_batches_seen == 0:
-                get_model_summary(self.model, data, self.debug)
+                # Pass None since TrainData is not compatible with torchinfo's traversal
```
jder: Why did this change? I thought we got some useful info from this before. (ie lat/lon sizes of each stage of the model)

jder (commented):
Thanks, generally LGTM. Wish we could have better typing here without having to get into classes etc. Happy to approve once we've resolved the torchinfo thing.

src/ocean_emulators/stepper.py
alxmrs: That's interesting! I usually prefer partials to lambdas because I typically am wary of including variables in the closure of the lambda vs explicitly controlling what is passed in via a partial. Reading this [SO post](https://stackoverflow.com/questions/11828410/differences-between-functools-partial-and-a-similar-lambda) (feels like a ghost town, taking a big grain of salt), it seems imprudent to worry about the closure wrt performance impact. 

I think it might be cool to bikeshed on this with the team, I feel like I always learn something from discussions around PL features like this.

alxmrs: I think protocols provide a good option here to document/(maybe) enforce the last argument being called "wet".

src/ocean_emulators/train.py
alxmrs: This changed because I added an argument to the TorchData constructor. See this stack trace, for example. 

<details>
```
test_trainer.py:24: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../.venv/lib/python3.11/site-packages/pytest_benchmark/fixture.py:156: in __call__
    return self._raw(function_to_benchmark, *args, **kwargs)
../.venv/lib/python3.11/site-packages/pytest_benchmark/fixture.py:180: in _raw
    duration, iterations, loops_range = self._calibrate_timer(runner)
../.venv/lib/python3.11/site-packages/pytest_benchmark/fixture.py:318: in _calibrate_timer
    duration = runner(loops_range)
../.venv/lib/python3.11/site-packages/pytest_benchmark/fixture.py:109: in runner
    function_to_benchmark(*args, **kwargs)
test_trainer.py:26: in run
    trainer.run()
../src/ocean_emulators/train.py:416: in run
    train_stats = self.train_one_epoch(epoch)
../src/ocean_emulators/train.py:503: in train_one_epoch
    get_model_summary(self.model, data, self.debug)
../src/ocean_emulators/utils/logging.py:225: in get_model_summary
    logger.info(summary(model, input_data=[data], depth=depth, verbose=0))
../.venv/lib/python3.11/site-packages/torchinfo/torchinfo.py:220: in summary
    x, correct_input_size = process_input(
../.venv/lib/python3.11/site-packages/torchinfo/torchinfo.py:246: in process_input
    correct_input_size = get_input_data_sizes(input_data)
../.venv/lib/python3.11/site-packages/torchinfo/torchinfo.py:496: in get_input_data_sizes
    return traverse_input_data(
../.venv/lib/python3.11/site-packages/torchinfo/torchinfo.py:448: in traverse_input_data
    [traverse_input_data(d, action_fn, aggregate_fn) for d in data]
../.venv/lib/python3.11/site-packages/torchinfo/torchinfo.py:448: in <listcomp>
    [traverse_input_data(d, action_fn, aggregate_fn) for d in data]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

data = <ocean_emulators.datasets.TrainData object at 0x1796ea1d0>
action_fn = <function get_input_data_sizes.<locals>.<lambda> at 0x31b508d60>
aggregate_fn = <class 'type'>

    def traverse_input_data(
        data: Any, action_fn: Callable[..., Any], aggregate_fn: Callable[..., Any]
    ) -> Any:
        """
        Traverses any type of nested input data. On a tensor, returns the action given by
        action_fn, and afterwards aggregates the results using aggregate_fn.
        """
        if isinstance(data, torch.Tensor):
            result = action_fn(data)
        elif isinstance(data, np.ndarray):
            result = action_fn(torch.from_numpy(data))
            # If the result of action_fn is a torch.Tensor, then action_fn was meant for
            #   torch.Tensors only (like calling .to(...)) -> Ignore.
            if isinstance(result, torch.Tensor):
                result = data
    
        # Recursively apply to collection items
        elif isinstance(data, Mapping):
            aggregate = aggregate_fn(data)
            result = aggregate(
                {
                    k: traverse_input_data(v, action_fn, aggregate_fn)
                    for k, v in data.items()
                }
            )
        elif isinstance(data, tuple) and hasattr(data, "_fields"):  # Named tuple
            aggregate = aggregate_fn(data)
            result = aggregate(
                *(traverse_input_data(d, action_fn, aggregate_fn) for d in data)
            )
        elif isinstance(data, Iterable) and not isinstance(data, str):
            aggregate = aggregate_fn(data)
>           result = aggregate(
                [traverse_input_data(d, action_fn, aggregate_fn) for d in data]
            )
E           TypeError: TrainData.__init__() missing 1 required positional argument: 'label_mask'

```
</details>

I bet there is a way to plumb the argument to torch info. Sonnet recommended that I just pass None, but maybe Opus knows how to address this issue.

alxmrs: Top of mind solution: I bet it would work if I removed `__iter__` on TD.

jder:
@codex review

src/ocean_emulators/train.py
jder: I am so confused why we didn't have this problem with the existing `num_prognostic_channels` argument. Can I just say again how much I miss real types?

jder (approved):
This seems complicated but I don't have a better suggestion! 🚢

jder:
@codex review

PR #556, opened 2026-01-26
Fix performance regression introduced by new multi modal data loader.

alxmrs:
This only fixes the performance regression when loading single datasets (i.e. "standard"). Loading multiple datasets is inherently slower. We will only know the final benchmark results when the are automatically run. Short of that, here is some evidence that the changes in this PR is an improvement: 

Start: 
```
------------------------------------------------------------------------------------------- benchmark: 1 tests ------------------------------------------------------------------------------------------
Name (time in ms)                                                                                               Min       Max      Mean   StdDev    Median      IQR  Outliers     OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_profile__loader__1gb[LoaderVersion.OM4_TORCH-cpu-extra_config_args0-mock-test/train_default.yaml]     766.8572  866.5532  806.3264  38.8650  806.3489  51.7769       2;0  1.2402       5           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
```

Finish:
```
------------------------------------------------------------------------------------------- benchmark: 1 tests ------------------------------------------------------------------------------------------
Name (time in ms)                                                                                               Min       Max      Mean   StdDev    Median      IQR  Outliers     OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_profile__loader__1gb[LoaderVersion.OM4_TORCH-cpu-extra_config_args0-mock-test/train_default.yaml]     414.9687  524.9424  458.3332  44.6805  448.1001  68.2806       1;0  2.1818       5           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
```

TL;DR: This cuts standard data loading time in half on average.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -436,7 +436,7 @@ class TorchTrainDataset(Dataset[RawTrainData]):
         self.id = f"{self.__class__.__name__}_{str(id(self))}"
         self.device = get_device()
 
-        # If the src and dst DataSource are the same, we can skip one set of normalizations.
+        # If the src and dst DataSource are the same, we can do a lot less work.
         self.independent_label = src != dst
 
         self.hist: int = hist
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -435,9 +435,8 @@ class TorchTrainDataset(Dataset[RawTrainData]):
         super().__init__()
         self.id = f"{self.__class__.__name__}_{str(id(self))}"
         self.device = get_device()
-
         # If the src and dst DataSource are the same, we can do a lot less work.
-        self.independent_label = src != dst
+        srcs = [src] if src is dst else [src, dst]
 
         self.hist: int = hist
         self.steps: int = steps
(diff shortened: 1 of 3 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -436,6 +436,9 @@ def __init__(
         self.id = f"{self.__class__.__name__}_{str(id(self))}"
         self.device = get_device()
 
+        # If the src and dst DataSource are the same, we can do a lot less work.
+        self.independent_label = src != dst
```
alxmrs: Question for the reader: Does this var have a clear name? Is `unique_label` better? Any other ideas?

jder:
@codex review

src/ocean_emulators/datasets.py
jder: Might be more clear and harder to mess up if we instead treat `self._label_src is not None` as the test for this?

jder (approved):
Thanks, minor naming suggestion and worth looking into codex's thing and maybe using `is` instead?

Totally up to you, but this might be cleaner if we instead had `self._prognostic_srcs` which was a list of either `[src]` or `[src, dst]` so all the below copy-pastes conditioned on `self.indepdent_label` like:

```
        self._input_src = src.filter(prognostic_var_names, prefix="input")
        self._boundary_src = src.filter(boundary_var_names, prefix="boundary")
        self._label_src = self._input_src
        if self.independent_label:
            self._label_src = dst.filter(prognostic_var_names, prefix="label")
```

were instead

```
        self.prognostic_srcs = [src.filter(prognostic_var_names, prefix="input") for src in self.prognostic_srcs]
        self._boundary_src = src.filter(boundary_var_names, prefix="boundary")
```

and then you do a final test for if there are 1 or 2 just before populating the final label.

PR #557, opened 2026-01-26
Adding distributed equivalence batch sampler.

alxmrs:
Adding a distributed version of the new equivalence batch sampler for use in multi-GPU training (with tests).

### DIFF
--- a/src/ocean_emulators/utils/samplers.py
+++ b/src/ocean_emulators/utils/samplers.py
@@ -203,6 +203,8 @@ class DistributedEquivalenceGroupBatchSampler(Sampler[list[int]]):
         seed: int = 0,
     ):
         super().__init__()
+        if num_replicas <= 0:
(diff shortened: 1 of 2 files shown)


jder:
@codex review

jder (changes requested):
Thanks, LGTM other than Codex's suggestion we should fix that behavior to match DistributedSampler:

> drop_last ([bool](https://docs.python.org/3/library/functions.html#bool), optional) – if True, then the sampler will drop the tail of the data to make it evenly divisible across the number of replicas. If False, the sampler will add extra indices to make the data evenly divisible across the replicas. Default: False.

alxmrs:
@claude will you review this PR?

alxmrs:
@codex may I have a review?

alxmrs:
@codex review

jder (approved):
LGTM! I think the codex comment is true but also P2 as it says. (Maybe worth a comment if you don't change the behavior)

PR #558, opened 2026-01-26
Extract similar concerns from the torch data loader into a `TorchDataSource`.

alxmrs:
A drive by refactor made during #532. I noticed that the means/stds/mask arrays really should be grouped with their associated tensor (either input, label, or boundary). The interface mirrors `DataSource`, except all operations are on `torch.Tensor`s rather than Xarray `Dataset`s. 

As a sanity check, I ran a local performance profile before and after this change. On my machine, both take basically the same median amount of time.

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -25,7 +25,7 @@ from ocean_emulators.constants import (
 from ocean_emulators.utils.data import (
     DataSource,
     LoadStats,
-    TorchDataSource,
+    OceanData,
     conditional_rearrange,
 )
 from ocean_emulators.utils.device import get_device, using_gpu
@@ -568,17 +568,17 @@ class TorchTrainDataset(Dataset[RawTrainData]):
 
         for input_, boundary, label in raw_train_data.raw_data:
             input_, label = self._to_example(
-                TorchDataSource.from_data_source(
+                OceanData.from_data_source(
                     input_,
                     self.wet_prognostic[0],
                     self._prognostic_srcs[0],
                 ).to(device=device, non_blocking=True),
-                TorchDataSource.from_data_source(
+                OceanData.from_data_source(
                     boundary,
                     self.wet_surface,
                     self._boundary_src,
                 ).to(device=device, non_blocking=True),
-                TorchDataSource.from_data_source(
+                OceanData.from_data_source(
                     label, self.wet_prognostic[-1], self._prognostic_srcs[-1]
                 ).to(device=device, non_blocking=True),
             )
@@ -589,26 +589,26 @@ class TorchTrainDataset(Dataset[RawTrainData]):
     def _to_example(
         self,
         # time includes (self.hist + 1) past steps and the (label) future steps
-        input_: TorchDataSource,
-        boundary: TorchDataSource,
-        label: TorchDataSource,
+        input_: OceanData,
+        boundary: OceanData,
+        label: OceanData,
     ) -> tuple[Input, Prognostic]:
         # Move normalization parameters to the same device as input data
         # grab past steps and prep for model
         total_input = self._prep_tensor_steps(
-            input_.map(lambda data: data[:, : self.hist + 1, :, :, :]),
-            boundary.map(lambda data: data[:, : self.hist + 1, :, :, :]),
+            input_.with_time(slice(0, self.hist + 1)),
+            boundary.with_time(slice(0, self.hist + 1)),
         )
         # grab future steps, repeat as we do for input
         label_tensor = self._prep_tensor_steps(
-            label.map(lambda data: data[:, self.hist + 1 :, :, :, :])
+            label.with_time(slice(self.hist + 1, None))
         )
         return total_input, label_tensor
 
     def _prep_tensor_steps(
         self,
-        prognostic: TorchDataSource,
-        boundary: TorchDataSource | None = None,
+        prognostic: OceanData,
+        boundary: OceanData | None = None,
     ) -> Input:
         """Prepare tensor steps by normalizing, masking and flattening dimensions."""
         prognostic_steps = prognostic.normalize_and_mask(
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -22,7 +22,12 @@ from ocean_emulators.constants import (
     PrognosticMask,
     PrognosticVarNames,
 )
-from ocean_emulators.utils.data import DataSource, LoadStats, conditional_rearrange
+from ocean_emulators.utils.data import (
+    DataSource,
+    LoadStats,
+    TorchDataSource,
+    conditional_rearrange,
+)
 from ocean_emulators.utils.device import get_device, using_gpu
 from ocean_emulators.utils.logging import elapsed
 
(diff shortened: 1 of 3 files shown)


src/ocean_emulators/utils/data.py
```diff
@@ -299,6 +299,59 @@ def from_datasets(
...
 
 
+@dataclasses.dataclass
+class TorchDataSource:
+    data: Float[torch.Tensor, "batch time variable lat lon"]
+    means: Float[torch.Tensor, " variable"]
+    stds: Float[torch.Tensor, " variable"]
+    mask: Bool[torch.Tensor, " variable"]
+
+    def map(self, data_func: Callable) -> Self:
```
alxmrs: Instead of `map`, this could be `slice` or `__getitiem__` where we would allow the user to pass the slicing information to get the data before or after `hist`.

src/ocean_emulators/datasets.py
```diff
@@ -470,11 +474,12 @@ def __init__(
             indices_da + stride * window_dim
         )
 
-        self.wet_input: PrognosticMask = src.masks.prognostic.to(self.device)
-        self.wet_surface: GridMask = src.masks.boundary.to(self.device)
-        self.wet_label: PrognosticMask = dst.masks.prognostic.to(self.device)
+        # NB(alxmrs): Keep masks on CPU - will be moved to GPU in to_train_data()
```
alxmrs, line 483: In my experiment branch, this was necessary to remove device related errors.

src/ocean_emulators/datasets.py
```diff
@@ -483,16 +488,16 @@ def flatten_to_device(means_or_stds: xr.Dataset) -> torch.Tensor:
...
 
-        self.boundary_means = flatten_to_device(self._boundary_src.means)
-        self.boundary_stds = flatten_to_device(self._boundary_src.stds)
+        self.boundary_means = flatten(self._boundary_src.means)
+        self.boundary_stds = flatten(self._boundary_src.stds)
 
-        self.label_means = flatten_to_device(self._label_src.means)
-        self.label_stds = flatten_to_device(self._label_src.stds)
+        self.label_means = flatten(self._label_src.means)
+        self.label_stds = flatten(self._label_src.stds)
```
alxmrs: Question for the reviewer: should this be moved into the new class? e.g. like on post init?

src/ocean_emulators/utils/data.py
```diff
@@ -299,6 +299,83 @@ def from_datasets(
...
+
+    @classmethod
+    def from_data_source(
+        cls,
+        data: Float[torch.Tensor, "batch time variable lat lon"],
+        mask: Float[torch.Tensor, " variable"],
+        src: DataSource,
+    ) -> Self:
+        means_torch = _flatten(src.means)
+        stds_torch = _flatten(src.stds)
```
alxmrs, line 349: I did a number of local profiles, and I haven't found a consistent result that says that we need to cache these numbers in the calling class. If anything, I have found that this "cleaner" approach is slightly faster. Hard to say until we run automated benchmarks.

src/ocean_emulators/datasets.py
jder: This is a little worrying from a performance perspective. We're now doing twice as many transfers to GPU. Probably doesn't matter too much but given it worked before I don't understand why it's a problem now.

jder:
@codex review

src/ocean_emulators/utils/data.py
```diff
@@ -299,6 +299,83 @@ def from_datasets(
...
+            "(variable lev)=var",
+            concat_dim="var",
+        ).rename({"var": "variable"})
+    else:
+        array = means_or_stds.to_dataarray()
+    return torch.from_numpy(array.to_numpy().flatten())
+
+
+@dataclasses.dataclass
+class TorchDataSource:
```
jder: I generally like the idea of trying to give semantic information around our torch tensors, thanks for doing this! I don't think "DataSource" is the right name, personally. It is an actual slice of ocean data, just with some metadata. How about "TorchDataSlice" or "OceanTensor" or "DecoratedData" or something?

Either way, please add a docstring here explaining what it represents.

src/ocean_emulators/utils/data.py
jder: Yeah, I don't think this is a good abstraction right now -- it's easy to call map with a lambda whose result no longer matches the means/stds/mask. So either the return type here should be the raw data Tensor, or we should restrict this to only do things to data which we can guarantee maintains the invariant of this class (ie that all the variable information lines up across the 4 members)

jder (changes requested):
Thanks! Generally in favor of this change. Just a few suggestions!

src/ocean_emulators/utils/data.py
jder: (Might also be nice to add a test or two)

src/ocean_emulators/utils/data.py
alxmrs: What do you think of the current `with_time` interface?

jder:
@codex review

src/ocean_emulators/utils/data.py
```diff
@@ -299,6 +299,104 @@ def from_datasets(
...
+        )
+
+    def pin_memory(self, device: torch.device | None = None) -> Self:
+        return dataclasses.replace(
+            self,
+            data=self.data.pin_memory(device=device),
+            means=self.means.pin_memory(device=device),
+            stds=self.stds.pin_memory(device=device),
+            mask=self.mask.pin_memory(device=device),
+        )
```
jder: This seems to be unused? Also I don't think pin_memory takes a device?

src/ocean_emulators/utils/data.py
```diff
@@ -299,6 +299,104 @@ def from_datasets(
...
+        """Normalize and mask tensors."""
+        tensor = self.data
+        if normalize_before_mask:
+            tensor = self._normalize(tensor)
+        tensor = torch.where(self.mask, tensor, masked_fill_value)
+        if not normalize_before_mask:
+            tensor = self._normalize(tensor)
+        return tensor
+
+    def to(self, device: torch.device, non_blocking: bool = False) -> Self:
```
jder: 🐑 we always seem to call this with non_blocking=True; maybe make the argument mandatory or default to True?

src/ocean_emulators/utils/data.py
jder: Looks great to me, thanks!

jder (approved):
Thanks! Just two little things.

src/ocean_emulators/utils/data.py
alxmrs: I remember you once mentioning that some valuable part of pytorch checks if a class has a `to` and `pin_memory` method in order to do something that we needed. Do you remember that context? Is that even relevant here?

src/ocean_emulators/utils/data.py
alxmrs:
```suggestion
    def to(self, device: torch.device, non_blocking: bool = True) -> Self:
```

src/ocean_emulators/utils/data.py
jder: Yes, the things that come out of the DataLoader (TrainData) have `pin_memory()` called on them

alxmrs: Ah, cool. So, since this never exits the loader, I'll delete this unused method. We can add it later if it's needed. Thanks.

PR #560, opened 2026-01-29
OSN_DATA_PATHS is intended to be optional

jder:
`null` in skypilot means "required", make these not-required.


alxmrs (approved):
LGTM, thank you!

PR #562, opened 2026-01-29
Attempt to fix memory leak during training run

### LINKED ISSUE #561: Memory leak (?) in recent commits
I think there might have been a recent memory usage regression; possibly a leak or just some inefficiency? [This view](https://wandb.ai/ocean_emulators/default?nw=of9pjio0cub&panelDisplayName=Process+Memory+Available+%28MB%29&panelSectionName=System&xAxisMax=5673.436901220998&xAxisMin=15.519669973) shows 3 runs. The teal run was run from [this commit](https://github.com/Open-Athena/Ocean_Emulator/commit/f11625140b03bcc9ef6fd7509061a51c27be0838) which is current main. You can see the machine ran out of memory after a few steps. The purple run is a run from a week ago, and the pink run is my restarting of teal (ie same config etc) at [the commit just before the src/dst change landed](https://github.com/Open-Athena/Ocean_Emulator/commit/09064250f63cfda8ecee366631fb1b5f6b45ecda). 

No rush @alxmrs (I can just stay pinned earlier for now) but can you take a look at some point?

alxmrs:
Fixes #561 caused by #555.

According to [this run](https://wandb.ai/ocean_emulators/default/runs/x9r2t1f9?nw=mz2c6oo6z8b), it appears that this PR fixes the memory leak. 

mem-leak-fix-memory-plots


src/ocean_emulators/datasets.py
```diff
@@ -568,7 +568,8 @@ def __getitem__(self, idx: int):
         return TD
 
     def to_train_data(self, raw_train_data: RawTrainData) -> TrainData:
-        train_data = TrainData(self.num_prognostic_channels, raw_train_data.label_mask)
+        label_mask = raw_train_data.label_mask.to(device=self.device, non_blocking=True)
```
alxmrs: This moves the mask to the GPU once instead of in every loss_fn call.

alxmrs:
@Claude Will you review this PR?

src/ocean_emulators/utils/data.py
```diff
@@ -53,17 +53,23 @@ def _is_compact(data: xr.Dataset, means: xr.Dataset, stds: xr.Dataset) -> bool:
...
     """A collection of masks to expose the ocean and mask land."""
 
     prognostic: PrognosticMask
     boundary: GridMask
 
-    def __post_init__(self):
-        self.prognostic = self.prognostic.bool()
-        self.boundary = self.boundary.bool()
+    @classmethod
+    def from_floating(cls, prognostic, boundary) -> Self:
```
jder: types, please

jder:
@codex review

src/ocean_emulators/utils/data.py
```diff
@@ -53,17 +53,23 @@ def _is_compact(data: xr.Dataset, means: xr.Dataset, stds: xr.Dataset) -> bool:
...
+        return cls(
+            prognostic=prognostic.bool(),
+            boundary=boundary.bool(),
+        )
 
+    # NB(alxmrs): Since this dataclass is long-lived, functools cache decorators shouldn't cause additional
+    #  memory issues.
+    @lru_cache(maxsize=5)
```
jder: I don't have strong feelings but this seems more likely to cause problems in the future than it helps. (eg if Masks becomes not-long-lived, or someone mutates it, or we keep 5 copies of the mask around, etc)

Would it be the same effect but less subtle if we had the TorchTrainDataset call `self._prognostic_srcs[-1].masks.prognostic_with_hist(self.hist)` once at init time and hold on to the result? Unless I'm missing something we could also avoid storing the label_mask in RawTrainData at all and keep the result of that call on GPU?

jder (commented):
Just one question from me

src/ocean_emulators/utils/data.py
alxmrs: I like the idea of avoiding adding this to RawTrainData -- this is something I think I'll add later (to make merging a bit easier). For now, I really like the idea of just caching this in the loader class. Adding that now.

alxmrs: No, wait -- on second though, I do want to do both now.

alxmrs:
@codex review

alxmrs:
@codex review please.

alxmrs:
I wasn't able to reproduce the memory leak on main. It appears that either #558 fixed it or the leak was cause by something else. Closing this PR for now.

PR #563, opened 2026-01-30
Warn on non-contiguous inference history stride and add failing regression test

jder:
### Motivation
- Make inference data stride issues visible during evaluation by logging a warning when inference history indices skip timesteps.
- Provide a minimal failing regression test that reproduces the non-contiguous history behavior for the inference dataset so the problem is obvious when running the test suite.

### Description
- Add a check in `InferenceDataset` initialization (`src/ocean_emulators/datasets.py`) that inspects the first rolling index window and logs a `logger.warning` when the inferred stride between history steps is not `1`.
- Add `test_inference_dataset_histories_are_contiguous` to `tests/test_datasets.py` which uses the `tiny_dataset_input` fixture and asserts that the first inference input contains contiguous time steps (for `hist=1`) by comparing normalized time values in the prognostic channels.
- The test is parametrized over `normalize_before_mask` and `masked_fill_value` to match existing dataset test coverage.

### Testing
- No automated tests were run as part of this change; the new regression test is expected to fail until the underlying stride issue is fixed.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_697ce98d7c348320acce1f3a47c14ff0)

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -90,6 +90,20 @@ class InferenceDataset(Dataset):
         )  # Skip indices based on history
         self.rolling_indices = self.rolling_indices.astype(int)
 
+        if self.hist > 0:
+            first_window = self.rolling_indices.isel(time=0).values
+            if len(first_window) > 1:
+                inferred_stride = int(first_window[1] - first_window[0])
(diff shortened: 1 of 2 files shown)


jder:
This was not actually a bug; bad robot.

PR #569, opened 2026-02-03
The Train and Validate Aggregators support multi-scale training.

alxmrs:
I need to test this e2e by running a training run, but all tests pass.


alxmrs:
@codex Please review.

alxmrs:
@codex may I have a review?

alxmrs:
@codex Please review

alxmrs:
@codex please review.

alxmrs:
May I have one last review, @codex?

alxmrs:
@codex Please review.

alxmrs:
@codex may I have a review?

alxmrs:

src/ocean_emulators/aggregator/validate/reduced.py
```diff
@@ -112,35 +119,38 @@ def record_batch(
...
         for metric in self._variable_metrics:
+            batch_key = metric.split("/")[-1]
             for key in self._variable_metrics[metric]:
                 data[f"{metric}/{key}"] = (
-                    self._variable_metrics[metric][key].get() / self._n_batches
+                    self._variable_metrics[metric][key].get()
+                    / self._n_batches[batch_key]
                 )
         meaned_data: dict[str, float] = {}
         for key in sorted(data.keys()):
```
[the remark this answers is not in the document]
alxmrs: @codex If we have a large number of batches, won't this regress to the mean anyway?

tests/test_aggregator_train_validate.py
```diff
@@ -0,0 +1,342 @@
...
+    def test_raises_when_no_batches_recorded(self, mock_data_source):
+        """Verify _get_data raises when no batches have been recorded."""
+        srcs = [mock_data_source]
+        agg = MeanAggregator(srcs, target_time=0)
+
+        with pytest.raises(ValueError, match="No batches have been recorded"):
+            agg._get_data()
+
+
+class TestMetricFunctions:
```
jder, line 240: 🐑 this feels like a pretty funny set of tests to me, and at least does not fit in this file.

tests/test_aggregator_train_validate.py
```diff
@@ -0,0 +1,342 @@
+"""Tests for train and validate aggregators.
+
+These tests verify the core functionality of:
+- TrainAggregator: loss accumulation across batches
+- ValidateAggregator: composition of sub-aggregators and multi-scale support
+- MeanAggregator (validate): area-weighted metric computation with multi-scale support
```
jder, line 6: This does not seem to test ValidateAggregator at all? And this comment doesn't mention AreaWeightedReducedMetric. Maybe just remove this comment? 

It would be nice to clarify in the tests below that they are about specific flavors of the aggregators. (eg `TestValidatorAreaWeightedReducedMetric`)

src/ocean_emulators/aggregator/validate/main.py
```diff
@@ -51,26 +50,34 @@ def record_validation_batch(self, batch: ValBatchOutput):
...
 
         assert batch.target_data.shape[1] == self.num_prognostic_channels
+
+        # NB(alxmrs): For non-standard schedules, the input/label grids could change on every batch! Thus, we need
+        # to check this every time.
+        label_grid_src = next(
+            s for s in self.srcs if s.grid_size == batch.target_data.shape[-2:]
+        )
+        input_grid_src = next(
+            s for s in self.srcs if s.grid_size == batch.input_data.shape[-2:]
```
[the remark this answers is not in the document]
jder: I agree this could eventually be a problem. Might be nice to change from `next` to something which asserts there's just one entry in the list? (A helper that does this named `single` or `only` or something is something I've appreciated in other python projects).

src/ocean_emulators/aggregator/validate/reduced.py
```diff
@@ -112,35 +119,38 @@ def record_batch(
         input_data_norm: dict[str, torch.Tensor] | None = None,
         i_time_start: int = 0,
     ):
-        if input_data is None:
-            input_data = {}
-        if input_data_norm is None:
-            input_data_norm = {}
+        # The label data shape will vary batch to batch during multi-scale training.
+        # thus, we look up the relevant are weights for this batch.
```
jder, line 123: > we look up the relevant are weights for this batch

Is this a typo? Not sure what it means.

src/ocean_emulators/aggregator/validate/main.py
```diff
@@ -15,27 +15,26 @@ class ValidateAggregator(TrainAggregator):
...
-        metadata: dict[str, dict[str, str]],
+        srcs: list[DataSource],
         hist: int,
-        area_weights: torch.Tensor,
-        wet: torch.Tensor,
         num_prognostic_channels: int,
     ):
         super().__init__()
+        self.srcs = srcs
+        primary_src = max(srcs, key=lambda s: s.grid_size)
```
jder, line 24: Why max grid size? Shouldn't this be srcs[0] or maybe the caller should just be passing one if validation only supports one src?

src/ocean_emulators/aggregator/validate/map.py
```diff
@@ -75,18 +79,20 @@ def get_logs(self, label: str):
...
+            # Get grid every time just in case it changes (but it shouldn't).
+            grid = name.split("/")[-1]
             # use first sample in batch
             gen = (
                 all_reduce_mean(
                     self._gen_data[name].select(dim=time_dim, index=target_time)
                 )
-                / self._n_batches
+                / self._n_batches[grid]
             )
```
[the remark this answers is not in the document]
jder: It also means we're all_reducing across potentially different sets of names in sorted_names?

src/ocean_emulators/aggregator/validate/map.py
jder, line 100: I think this name will now include the grid size which means the caption's metadata won't be found.

jder (commented):
Can you clarify what the goal of this PR is? ie what do we expect to happen for val metrics when using mixed training? For example, it looks like we're logging eg `val/mean_map/image-error/zos/180x360` which is the mean map across all batches which produce 180x360 output. Do we really want to take the mean across all input sizes? Same question for val snapshots; what is the desired behavior? Looks like losses for both train and val are meaned across all batches and not split out by resolution; is that what we want? 

One other high level comment is that we're doing a lot of string matching/splitting/building in various places across the code in a way that seems fragile to me. Could we instead pass more semantically meaningful values around, or consolidate this building/matching logic into a small piece of code? (I realize this is somewhat limited by the current structure so may not be feasible but eg seems like it would not be too hard to add more metadata into `ValBatchOutput` about what srcs it came from.)

jder:
Closed in favor of #591

PR #576, opened 2026-02-05
Memory usage fix

jder:
I did some more digging as to why my jobs were failing. I reduced the range of commits to find it was introduced in a534fc887cdeb10cc5179e3bccdcdfd188c3013e and Codex proposed the below fix. No idea if this is the right one (I haven't looked at it carefully) but it resolves the problem I was having. 

I believe the reason you were unable to reproduce it is that you were using 1-degree data and I was using 1/2-degree data. The bug causes a doubling of memory usage for the prognostic/boundary data so at 1-degree it was not fatal. Here's a screenshot of available memory for the 1-degree runs you were doing to diagnose compared with 1/2-degree runs. 

image

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -443,7 +443,7 @@ class TorchTrainDataset(Dataset[RawTrainData]):
         self.id = f"{self.__class__.__name__}_{str(id(self))}"
         self.device = get_device()
         # If the src and dst DataSource are the same, we can do a lot less work.
-        srcs = [src] if src is dst else [src, dst]
+        srcs = [src] if src.name == dst.name else [src, dst]
 
         self.hist: int = hist
         self.steps: int = steps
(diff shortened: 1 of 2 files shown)


jder:
@alxmrs can you take a look and decide if this is the correct fix or if there's another strategy to resolve that will work better with your plans?

src/ocean_emulators/datasets.py
```diff
@@ -310,7 +310,7 @@ def insert(
         boundary: torch.Tensor,
         label: torch.Tensor,
     ):
-        """Add a prognostic input, boundary, and prognostic label as the last step."""
+        """Add past input/boundary and future label for the last step."""
```
alxmrs, line 313: This is a useful comment.

alxmrs (commented):
I need to investigate this more, but at a high level, this makes sense. We should filter by time before moving to arrays, not after. That definitely makes sense why the status quo used more memory.

PR #577, opened 2026-02-06
Renaming ctx variable to make it clear that it always stays on the CPU.

alxmrs:
This was one of the remaining notes of feedback from the big multi-scale PR. I thought I pushed it before the auto-merge, but I had to click one additional "push" button on my IDE for it to have actually gone through! This tine PR addresses the suggestion and applies a rename.


jder (approved):
Thanks!

PR #579, opened 2026-02-13
Memory fix (with minimal diff).

alxmrs:
Based on #576. This PR attempts to reduce data loading memory usage. A memory regression was introduced recently when we refactored this area of code.


alxmrs:
@codex May I have a review, please?

jder (approved):
Thanks!

PR #591, opened 2026-02-18
NoOp the validate aggregator for the multiscale case.

alxmrs:
I was starting to go in a different direction in a refactor when I took a step back and realized that we could minimally modify the current aggregator interfaces to effectively "no op" in the multi-scale case in `train.py`. 

This PR introduces a clean way to do nothing (except log training loss) during the validate log aggregation. It also adds some tests for the val aggregator (interface). This required some refactoring.

### DIFF
--- a/src/ocean_emulators/aggregator/main.py
+++ b/src/ocean_emulators/aggregator/main.py
@@ -9,6 +9,10 @@ import torch
 from ocean_emulators.aggregator.inference import InferenceEvaluatorAggregator
 from ocean_emulators.aggregator.train import TrainAggregator
 from ocean_emulators.aggregator.validate import ValidateAggregator
+from ocean_emulators.aggregator.validate.map import MapAggregator
+from ocean_emulators.aggregator.validate.reduced import MeanAggregator
+from ocean_emulators.aggregator.validate.snapshot import SnapshotAggregator
+from ocean_emulators.aggregator.validate.sub_aggregator import ValidateSubAggregator
 
 
 class Aggregator:
@@ -21,14 +25,17 @@ class Aggregator:
         metadata: dict[str, dict[str, str]],
         hist: int,
         area_weights: torch.Tensor,
-        wet: torch.Tensor,
         num_prognostic_channels: int,
     ) -> ValidateAggregator:
+        val_aggregators: dict[str, ValidateSubAggregator] = {
+            "snapshot": SnapshotAggregator(metadata, hist),
+            "mean_map": MapAggregator(metadata, hist),
+            "reduced": MeanAggregator(area_weights, hist),
+        }
+
         return ValidateAggregator(
-            metadata=metadata,
+            val_aggregators,
             hist=hist,
-            area_weights=area_weights,
-            wet=wet,
             num_prognostic_channels=num_prognostic_channels,
         )
 
--- a/src/ocean_emulators/aggregator/main.py
+++ b/src/ocean_emulators/aggregator/main.py
@@ -13,6 +13,7 @@ from ocean_emulators.aggregator.validate.map import MapAggregator
 from ocean_emulators.aggregator.validate.reduced import MeanAggregator
 from ocean_emulators.aggregator.validate.snapshot import SnapshotAggregator
 from ocean_emulators.aggregator.validate.sub_aggregator import ValidateSubAggregator
+from ocean_emulators.utils.data import Normalize
 
 
 class Aggregator:
@@ -27,6 +28,8 @@ class Aggregator:
         area_weights: torch.Tensor,
         num_prognostic_channels: int,
     ) -> ValidateAggregator:
+        Normalize.get_instance()
+
         val_aggregators: dict[str, ValidateSubAggregator] = {
             "snapshot": SnapshotAggregator(metadata, hist),
             "mean_map": MapAggregator(metadata, hist),
--- a/src/ocean_emulators/aggregator/main.py
+++ b/src/ocean_emulators/aggregator/main.py
@@ -13,7 +13,6 @@ from ocean_emulators.aggregator.validate.map import MapAggregator
 from ocean_emulators.aggregator.validate.reduced import MeanAggregator
 from ocean_emulators.aggregator.validate.snapshot import SnapshotAggregator
 from ocean_emulators.aggregator.validate.sub_aggregator import ValidateSubAggregator
-from ocean_emulators.utils.data import Normalize
 
 
 class Aggregator:
@@ -28,9 +27,6 @@ class Aggregator:
         area_weights: torch.Tensor,
         num_prognostic_channels: int,
     ) -> ValidateAggregator:
-        # Probably not needed, but maintaining previous functionality.
-        Normalize.get_instance()
-
         val_aggregators: dict[str, ValidateSubAggregator] = {
             "snapshot": SnapshotAggregator(metadata, hist),
             "mean_map": MapAggregator(metadata, hist),
--- a/src/ocean_emulators/aggregator/main.py
+++ b/src/ocean_emulators/aggregator/main.py
@@ -28,6 +28,7 @@ class Aggregator:
         area_weights: torch.Tensor,
         num_prognostic_channels: int,
     ) -> ValidateAggregator:
+        # Probably not needed, but maintaining previous functionality.
         Normalize.get_instance()
 
         val_aggregators: dict[str, ValidateSubAggregator] = {
--- a/src/ocean_emulators/aggregator/main.py
+++ b/src/ocean_emulators/aggregator/main.py
@@ -45,7 +45,6 @@ class Aggregator:
         metadata: dict[str, dict[str, str]],
         hist: int,
         area_weights: torch.Tensor,
-        wet: torch.Tensor,
         num_prognostic_channels: int,
         channel_mean_names: list[str] | None = None,
     ) -> InferenceEvaluatorAggregator:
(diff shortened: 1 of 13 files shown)


alxmrs:
@codex May I have a review?

alxmrs:
@codex, may I have one more review?

tests/test_positional_channels.py
```diff
@@ -1,75 +1,48 @@
...
 from ocean_emulators.config import SamudraConfig, UNetBackboneConfig
-from ocean_emulators.constants import TensorMap
 from ocean_emulators.utils.ctx import GridContext
-from ocean_emulators.utils.data import DataSource, Masks, Normalize
-from ocean_emulators.utils.multiton import MultitonScope
+from ocean_emulators.utils.data import DataSource
 
 
-def test_positional_parameters_update():
+def test_positional_parameters_update(dummy_src: DataSource):
```
alxmrs, line 8: Extracting out this set up as a common test fixture.

tests/test_validate_aggregator.py
```diff
@@ -0,0 +1,103 @@
...
+from ocean_emulators.aggregator.train import TrainAggregator
+from ocean_emulators.aggregator.validate.sub_aggregator import ValidateSubAggregator
+from ocean_emulators.utils.ctx import GridContext
+from ocean_emulators.utils.data import DataSource
+from ocean_emulators.utils.output import TrainBatchOutput, ValBatchOutput
+from ocean_emulators.utils.wandb import Metrics
+
+
+@pytest.fixture
+def val_batch() -> ValBatchOutput:
```
alxmrs: Instead of a fixture, I may want to make this a function on top of DataSource, so the grid dimensions don't have to be hard coded.

src/ocean_emulators/constants.py
```diff
@@ -309,7 +309,7 @@ def _populate_dp_3d_idx(self):
                 self.DP_3D_IDX[self.DEPTH_SET[0]],
                 torch.tensor([self.VAR_3D_IDX[var_2D] for var_2D in self.VAR_SET_2D]),
             ]
-        )
+        ).to(torch.int32)
```
alxmrs, line 312: Now that we have real tests for the aggregators, I caught a bug.

src/ocean_emulators/aggregator/main.py
```diff
@@ -21,14 +25,17 @@ def get_validation_aggregator(
         metadata: dict[str, dict[str, str]],
         hist: int,
         area_weights: torch.Tensor,
-        wet: torch.Tensor,
         num_prognostic_channels: int,
     ) -> ValidateAggregator:
+        val_aggregators: dict[str, ValidateSubAggregator] = {
```
alxmrs, line 30: This factory class now does something more than call the constructor.

src/ocean_emulators/stepper.py
```diff
@@ -52,7 +52,7 @@ def validate_batch(
         outs = model.forward_once(input, batch.ctx)
         loss_per_channel = loss_fn(outs, label, batch.ctx)
         loss = torch.mean(loss_per_channel)
-        return ValBatchOutput(loss, loss_per_channel, input, label, outs)
+        return ValBatchOutput(loss, loss_per_channel, input, label, outs, batch.ctx)
```
alxmrs, line 55: Used to get masking information. This is helpful for dealing with multiple scales.

tests/conftest.py
```diff
@@ -559,3 +566,33 @@ def trainer_pair(
         trainer.init_data_loaders(cur_step=train_config.steps[0])
 
         yield train_config, trainer
+
+
+@pytest.fixture
+def dummy_src():
```
alxmrs, line 571: Drive by refactor -- this was helpful, so I put it in a common place.

tests/test_validate_aggregator.py
```diff
@@ -0,0 +1,113 @@
...
+from ocean_emulators.utils.output import TrainBatchOutput, ValBatchOutput
+from ocean_emulators.utils.wandb import Metrics
+
+
+def src_to_val_batch(src: DataSource) -> ValBatchOutput:
+    """Create a dummy Validation Batch loss / data from a DataSource."""
+    # If we can consume a DataSource, then TensorMap has to be initialized.
+    tm = TensorMap.get_instance()
+    n_prog = len(tm.prognostic_var_names)
+    n_boundary = len(tm.boundary_var_names)
```
alxmrs, line 18: In a future PR: this seems like a good reason why the var names should live inside the DataSource.

src/ocean_emulators/train.py
```diff
@@ -609,14 +609,21 @@ def _call_loss_update(self, data: TrainData):
...
-        )
+        if self.train_schedule == "standard":
+            # The standard val aggregator only supports a single scale.
+            val_aggregator = Aggregator.get_validation_aggregator(
+                self.primary_src.metadata,
+                self.hist,
+                self.primary_src.spherical_area_weights.to(self.device),
+                self.num_out,
+            )
+        else:
```
alxmrs, line 620: I plan on running a test e2e with this tomorrow.

alxmrs:
@codex may I have a review?

alxmrs:
Thanks for that catch. May I have another review, @codex?

src/ocean_emulators/aggregator/validate/main.py
```diff
@@ -45,6 +32,14 @@ def record_batch(self, batch):
     def record_validation_batch(self, batch: ValBatchOutput):
         super().record_batch(batch)  # Record losses
 
+        # If there are no log aggregators, omit doing any extra work.
+        if not self._aggregators:
+            return
+
+        # Translate the GridContext mask by removing history.
+        first_wetmask_chunk = batch.ctx.label_mask.shape[0] // (self.hist + 1)
+        wet = batch.ctx.label_mask[:first_wetmask_chunk]
```
jder: 🐑 seems like a bit of a smell this is happening here instead of in get_aggregator_dicts which is doing the same transform to the data itself. Maybe at least an assert here about the mask shape compared to the data shape would be nice to ensure this doesn't subtly break later?

src/ocean_emulators/utils/data.py
```diff
@@ -682,9 +682,12 @@ def get_aggregator_dicts(
         data_reshaped = rearrange(
             data, "n (hi c) h w -> (n hi) c h w", hi=hist + 1
         ).unsqueeze(0)  # add artificial batch dim
-    else:
+    elif data.ndim != 5:
         # Batches are independent rollouts during validation
         data_reshaped = rearrange(data, "n (hi c) h w -> n hi c h w", hi=hist + 1)
+    else:
+        # This case comes up in tests; typically, data is not in the desired shape automatically.
+        data_reshaped = data
```
jder, line 690: Wouldn't we want tests to be using the same shape as we use at non-test time? Otherwise it seems like we're not testing a critical thing here (that these shape transformations are correct).

src/ocean_emulators/utils/output.py
```diff
@@ -16,12 +18,14 @@ def __init__(
...
         target_data: torch.Tensor,
         gen_data: torch.Tensor,
+        ctx: GridContext,
     ):
         super().__init__(loss, loss_per_channel)
         assert target_data.shape == gen_data.shape
         self.input_data = input_data
         self.target_data = target_data
         self.gen_data = gen_data
+        self.ctx = ctx
```
jder, line 28: Thanks, having this info attached to the data itself seems like a good direction to move.

tests/test_validate_aggregator.py
```diff
@@ -0,0 +1,113 @@
...
+
+    h, w = src.grid_size
```
jder: This seems to be the only place we're using `src`. Should we instead accept a grid size? Or should we be using data form src for input_data/target_data/label_mask/input_resolution?

tests/test_validate_aggregator.py
```diff
@@ -0,0 +1,113 @@
...
+            input_resolution_cpu=(
+                torch.linspace(-90, 90, steps=h),
+                torch.linspace(-180, 180, steps=w),
+            ),
+        ),
+    )
+    return batch
+
+
+def val_to_train_batch(val_batch: ValBatchOutput) -> TrainBatchOutput:
```
jder: 🐑 Since ValBatchOutput is a TrainBatchOutput, I think you could instead do `train_agg.record_batch(val_batch)` below and delete this function.

jder (approved):
Thanks! Generally looks good to me, a few questions/suggestions below. 

I do think that snapshot tests (of the top-level aggregators) would on balance be better than testing train and val aggregators against each other. They share code, for example, and so eg if the train aggregator stops logging loss at all, this test will not catch it. But this is definitely fine for a targeted thing to unblock your training run.

tests/test_validate_aggregator.py
alxmrs: I wish I could give this the :brain: reaction. That's slick.

src/ocean_emulators/utils/data.py
alxmrs: I really did try to exercise one or both of the above blocks, but setting it to that shape caused assert areas in other areas. I think this is a side effect of how we reach this function through test code.

src/ocean_emulators/aggregator/validate/main.py
alxmrs: I like the idea of moving this into get_aggregator_dics, that's a good idea.

alxmrs: I think the assert might be a better, more minimal cut.

Issue #443, opened 2025-11-03
Remove extra forward pass from dynamic MSE loss


jder:
We don't need to do a second forward pass just for the loss, we can use a delayed estimate of the error, see: https://openathena.slack.com/archives/C08CYM42DT3/p1761925172329729?thread_ts=1761924209.500949&cid=C08CYM42DT3

jder:
We might also consider using the error from the end of the 4-step rollout rather than after a single step. This could help stabilize long-term behavior?

PR #592, opened 2026-02-18
Remove extra forward pass from dynamic MSE loss

amogh-gulati:
Resolved this issue #443

Eliminates the redundant second model forward pass that was used solely to update DynamicLoss channel scales. Instead, we reuse the per-channel loss already computed during the training rollout.

This should introduce a delayed estimate (if i interpret it correctly)

We are also using the error from the end of the 4-step rollout rather than after a single step.

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -599,7 +599,7 @@ class Trainer:
             metric_logger.update(loss=loss_value_reduce.item())
             metric_logger.update(lr=lr)
 
-            self._call_loss_update(data)
+            self._call_loss_update(TO.loss_per_channel)
 
             self.profiler.after_batch(self.num_batches_seen)
 
@@ -609,16 +609,24 @@ class Trainer:
         logger.info(f"Aggregating train logs")
         return train_aggregator.get_logs()
 
-    def _call_loss_update(self, data: TrainData):
-        # This is a separate function to ensure locals are dropped immediately after use
+    def _call_loss_update(self, loss_per_channel: torch.Tensor):
+        # Use the already-computed per-channel loss from the training rollout
+        # to update DynamicLoss scales, avoiding an second forward pass.
+        # This introduces delayed estimate but should be more efficient
         if update := getattr(self.loss_fn, "update", None):
-            with torch.no_grad():
-                single_step_data = TrainData(data.num_prognostic_channels, data.ctx)
-                # Each entry in data is one step in a rollout.
-                input, label = data[0]
-                single_step_data.append(input, label)
-                pred = self.model(single_step_data)
-                update(pred[0], label, ctx=data.ctx)
+            # Undo the dynamic scaling to recover the raw per-channel loss.
+            if scale_fn := getattr(self.loss_fn, "loss_scale_per_channel", None):
+                scale = scale_fn()
+                raw_loss = (
+                    loss_per_channel.detach().reshape(-1, scale.shape[0]).mean(dim=0)
+                    / scale
+                )
+                # Expand back to the full hist*var shape expected by update.
+                n_hist = self.hist + 1
+                raw_loss = raw_loss.repeat(n_hist)
+            else:
+                raw_loss = loss_per_channel.detach()
+            update(raw_loss)
 
     def validate_one_epoch(self, epoch):
         self.model.eval()
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -599,7 +599,7 @@ class Trainer:
             metric_logger.update(loss=loss_value_reduce.item())
             metric_logger.update(lr=lr)
 
-            self._call_loss_update(TO.loss_per_channel)
+            self._maybe_update_loss(TO)
 
             self.profiler.after_batch(self.num_batches_seen)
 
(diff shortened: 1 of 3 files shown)


src/ocean_emulators/train.py
```diff
@@ -595,16 +595,24 @@ def train_one_epoch(self, epoch):
         logger.info(f"Aggregating train logs")
         return train_aggregator.get_logs()
 
-    def _call_loss_update(self, data: TrainData):
-        # This is a separate function to ensure locals are dropped immediately after use
+    def _call_loss_update(self, loss_per_channel: torch.Tensor):
```
alxmrs: :sheep: consider making this accept the `TO` object instead of a raw tensor in case we need other data later / to add more context to the shape of the tensor.

src/ocean_emulators/train.py
```diff
@@ -595,16 +595,24 @@ def train_one_epoch(self, epoch):
...
+        # Use the already-computed per-channel loss from the training rollout
+        # to update DynamicLoss scales, avoiding an second forward pass.
+        # This introduces delayed estimate but should be more efficient
```
alxmrs: I appreciate you noting the tradeoffs involved here. By delayed estimate, do we mean a delay in wandb logs?

src/ocean_emulators/train.py
alxmrs: Can we change this name to `_maybe_call_loss_update` -- we don't take any action if there is no dynamic loss.

src/ocean_emulators/train.py
```diff
@@ -595,16 +595,20 @@ def train_one_epoch(self, epoch):
...
-                input, label = data[0]
-                single_step_data.append(input, label)
-                pred = self.model(single_step_data)
-                update(pred[0], label, ctx=data.ctx)
+            # Undo the dynamic scaling to recover the raw per-channel loss.
+            if scale_fn := getattr(self.loss_fn, "loss_scale_per_channel", None):
+                scale = scale_fn()
+                raw_loss = (
+                    loss_per_channel.detach().reshape(-1, scale.shape[0]) / scale
+                ).reshape(-1)
```
alxmrs: 🐑 A few naming nits, to be more consistent with the dynamic loss class: 

Totally optional.
```suggestion
            if get_scales := getattr(self.loss_fn, "loss_scale_per_channel", None):
                per_channel_scale = get_scales()
                raw_loss = (
                    loss_per_channel.detach().reshape(-1, scale.shape[0]) / per_channel_scale
                ).reshape(-1)
```

src/ocean_emulators/train.py
```diff
@@ -595,16 +595,20 @@ def train_one_epoch(self, epoch):
...
+            else:
+                raw_loss = loss_per_channel.detach()
```
alxmrs: I don't think we ever hit this case: the only function that has an `update` is the dynamic loss right now. Maybe, instead of a default behavior, we can throw an error -- this would alert us early that something is out of place (rather than silently getting strange results).

src/ocean_emulators/train.py
alxmrs: Thinking about it more, I like the name `_maybe_update_loss` better.

alxmrs (commented):
Thanks for doing this! Generally, this is looking good. A few notes.

src/ocean_emulators/train.py
amogh-gulati: This makes sense

src/ocean_emulators/train.py
amogh-gulati: by delayed estimate here we mean that the loss scales are one step older, we have already updated the model weights before we call this function. Earlier we were using the new weights to get the raw loss again which was more upto date with the weights ( but even that was only doing 1 step pass instead of 4 so that was not exactly precise either )

src/ocean_emulators/train.py
amogh-gulati: yup error makes sense sense

jder:
@amogh-gulati is this ready to merge? One thing that we might want to consider: this changes the semantics considerably (using the 4-step rollout which seems better, and delaying the update by 1 step which seems maybe worse but not enough to matter) but we might want to use the existing behavior for any other runs being done specifically for the paper. Could we add this as a new loss (or new configuration for the existing loss) until we ship the paper?

amogh-gulati:
> @amogh-gulati is this ready to merge? One thing that we might want to consider: this changes the semantics considerably (using the 4-step rollout which seems better, and delaying the update by 1 step which seems maybe worse but not enough to matter) but we might want to use the existing behavior for any other runs being done specifically for the paper. Could we add this as a new loss (or new configuration for the existing loss) until we ship the paper?

This should be ready to merge but yes I agree, lets keep it as a configuration (maybe a bool for delayed estimate )

alxmrs:
@amogh-gulati I see you marked this as a draft, is it ready for my review?

alxmrs (approved):
LGTM, thanks for the changes.

amogh-gulati:
> LGTM, thanks for the changes.

wanted to support delayed estimate as a bool first 😅

amogh-gulati:
> @amogh-gulati is this ready to merge? One thing that we might want to consider: this changes the semantics considerably (using the 4-step rollout which seems better, and delaying the update by 1 step which seems maybe worse but not enough to matter) but we might want to use the existing behavior for any other runs being done specifically for the paper. Could we add this as a new loss (or new configuration for the existing loss) until we ship the paper?

does the current change look good?

jder:
@codex review

jder:
@amogh-gulati thanks, this looks good to me now.

🐑 FWIW I think it would be fine to keep the old signature of `update` which takes pred/target and recomputes the (unscaled) loss rather than having to go through the unscaling and/or grabbing the underlying loss_fn from the dynamic loss to get unscaled loss externally. Would be slightly slower but I doubt meaningfully slower.

jder (approved):

amogh-gulati:
> @amogh-gulati thanks, this looks good to me now.
> 
> 🐑 FWIW I think it would be fine to keep the old signature of `update` which takes pred/target and recomputes the (unscaled) loss rather than having to go through the unscaling and/or grabbing the underlying loss_fn from the dynamic loss to get unscaled loss externally. Would be slightly slower but I doubt meaningfully slower.

that does sound cleaner but i dont have access to preds in the delayed estimate path

jder:
> > @amogh-gulati thanks, this looks good to me now.
> > 🐑 FWIW I think it would be fine to keep the old signature of `update` which takes pred/target and recomputes the (unscaled) loss rather than having to go through the unscaling and/or grabbing the underlying loss_fn from the dynamic loss to get unscaled loss externally. Would be slightly slower but I doubt meaningfully slower.
> 
> that does sound cleaner but i dont have access to preds in the delayed estimate path

Ah, right, because we still do loss calculation in the model. We really should fix that one of these days. Thanks for taking a look at it!

PR #597, opened 2026-02-23
Avoid getting OOM killed during container build with larger runner

jder:
I am pretty sure that [eg this](https://github.com/Open-Athena/Ocean_Emulator/actions/runs/22228379616/job/64301000021) is an OOM-kill. This instance has twice as much ram as the standard runners.


alxmrs (approved):
LGTM

PR #602, opened 2026-02-24
container: upgrade nsys CLI for nccl trace support

jder:
I wanted to do some profiling but the version of nsight systems in the NVIDIA image by default is a bit old, so we now install a new one. The image from https://github.com/Open-Athena/Ocean_Emulator/actions/runs/22362970435/job/64721289520 (this branch) successfully ran on torch with `nsys profile --trace=cuda,nvtx,osrt,nccl --sample=cpu` which is what we were using before (and did not work on the version installed by default).

### DIFF
--- a/docs/torch.md
+++ b/docs/torch.md
@@ -75,6 +75,10 @@ Key behavior:
 
 ### Example: 1 Node, 8x RTX6000 on the NYU Torch HPC
 
+For `rtx6000_lzanna` nodes, request CPUs at the node's full CPU:GPU ratio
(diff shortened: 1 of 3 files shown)


docs/torch.md
```diff
@@ -75,6 +75,25 @@ Key behavior:
 
 ### Example: 1 Node, 8x RTX6000 on the NYU Torch HPC
 
+For Torch RTX6000 nodes, size CPU and memory proportionally to GPUs. If you
```
alxmrs, line 78: Nice, this is what I'm looking for.

alxmrs (approved):
Nice! Good docs, and the nsys install in the docker image LGTM.

PR #603, opened 2026-02-25
Adding a perceiver-based decoder that complements the encoder.

### LINKED ISSUE #395: Add perceiver-based decoder

alxmrs:
Here is a PerceiverIO-based decoder. It consumes latent information and produces the physical output grid with a single perceiver-transformer. It accomplishes this by arranging tensors to take in the latent input tokens and "query" channels, which are points on a 3D unit sphere that correspond to the area on Earth. Because the output dimension is large and involves expensive cross attentions, we provide a `window_size` argument to the decoder that will only process a chunk of output information at a time. This should bound peak memory during decoding.

When the input patch extent is fine grained, attending over the whole input -- even when composed of latent information -- will likely explode memory during decoding. To address this, we provide the decoder with a `context_patches` that builds off the `window_size` argument: it adds a number of patches of `window_size` around the target patch as the only context (i.e. a local context) for the PerceiverIO call.

Fixes #395.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -495,7 +495,7 @@ class DecoderConfig(BaseConfig):
         description="Embedding dimension for pixel-position queries in the PerceiverIO decoder head.",
     )
     window_patches: int | None = Field(
-        default=None,
+        default=4096,
         description="Side length (in patches) of each spatial decode window. "
         "None = decode all patches at once (global attention). "
         "E.g. window_patches=8 means each PerceiverIO call covers an 8x8 block of patches.",
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -347,8 +347,6 @@ class PerceiverConfig(BaseConfig):
         # This is not really a "frequency" but a maximum of the width appears to be reasonable from looking at the code.
         max_freq = max(*max_patch_size)
 
-        # TODO(alxmrs,jder): Each implementation takes the mean of the num_latents dim to produce the final output_dim.
-        #  Why compute the mean? Is it better to directly project from the num_latents x latent_dim?
         if (
             self.implementation == "auto" and torch.cuda.is_available()
         ) or self.implementation == "flash":
@@ -358,7 +356,7 @@ class PerceiverConfig(BaseConfig):
                 raise ValueError(
                     "`implementation==flash` or flash was automatically chosen for `implementation==auto`, but the flash attention dependencies could not be imported. Please run `uv sync --extra cuda` or specify the `naive` attention implementation."
                 ) from e
-            perceiver = FlashPerceiver(
+            perceiver: nn.Module = FlashPerceiver(
                 latent_rotary_emb_dim=max_freq,
                 depth=self.depth,
                 input_dim=in_channels,
@@ -431,6 +429,7 @@ class DecoderConfig(BaseConfig):
             in_channels=in_channels,
             out_channels=out_channels,
             patch_extent=patch_extent,
+            latent_dim=self.perceiver.latent_dim,
             perceiver=self.perceiver.build(
                 perceiver_in_channels, out_channels, max_patch_size
             ),
@@ -663,7 +662,6 @@ class FOMOConfig(BaseModelConfig):
             hist=hist,
             checkpointing=self.checkpointing,
             gradient_detach_interval=self.gradient_detach_interval,
-            grid_sizes=all_grid_sizes,
             use_bfloat16=self.use_bfloat16,
         )
 
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -394,23 +394,21 @@ class PerceiverConfig(BaseConfig):
 
 
 class EncoderConfig(BaseConfig):
-    patch_extent: list[float] = Field(
-        default=[6.0, 10.0],
-        description="Target physical extent of each patch in degrees [height_deg, width_deg]. "
-        "Patch sizes will be calculated to match this extent for each grid resolution.",
-    )
     perceiver: PerceiverConfig = PerceiverConfig()
 
     def build(
-        self, in_channels: int, out_channels: int, max_lat_size: int, max_lon_size: int
+        self,
+        in_channels: int,
+        out_channels: int,
+        patch_extent: tuple[float, float],
+        max_lat_size: int,
+        max_lon_size: int,
     ) -> PerceiverEncoder:
-        assert len(self.patch_extent) == 2, "spatial extents must be a pair of floats."
-        extent = self.patch_extent[0], self.patch_extent[1]
-        max_patch_size = patch_from(extent, max_lat_size, max_lon_size)
+        max_patch_size = patch_from(patch_extent, max_lat_size, max_lon_size)
         return PerceiverEncoder(
             in_channels=in_channels,
             out_channels=out_channels,
-            patch_extent=extent,
+            patch_extent=patch_extent,
             perceiver=self.perceiver.build(in_channels, out_channels, max_patch_size),
         )
 
@@ -424,27 +422,21 @@ class DecoderConfig(BaseConfig):
     cross-attention from learned output queries.
     """
 
-    patch_extent: list[float] = Field(
-        default=[6.0, 10.0],
-        description="Target physical extent of each patch in degrees [height_deg, width_deg]. "
-        "Should match the encoder's patch_extent for consistent spatial semantics.",
-    )
     perceiver: PerceiverConfig = PerceiverConfig()
 
     def build(
         self,
         in_channels: int,
         out_channels: int,
+        patch_extent: tuple[float, float],
         max_lat_size: int,
         max_lon_size: int,
     ) -> PerceiverDecoder:
-        assert len(self.patch_extent) == 2, "patch_extent must be a pair of floats."
-        extent = self.patch_extent[0], self.patch_extent[1]
-        max_patch_size = patch_from(extent, max_lat_size, max_lon_size)
+        max_patch_size = patch_from(patch_extent, max_lat_size, max_lon_size)
         return PerceiverDecoder(
             in_channels=in_channels,
             out_channels=out_channels,
-            patch_extent=extent,
+            patch_extent=patch_extent,
             perceiver=self.perceiver.build(in_channels, out_channels, max_patch_size),
         )
 
@@ -606,6 +598,11 @@ class FOMOConfig(BaseModelConfig):
     encoder: EncoderConfig = EncoderConfig()
     processor: UNetBackboneConfig = UNetBackboneConfig()
     decoder: DecoderConfig = DecoderConfig()
+    patch_extent: list[float] = Field(
+        default=[6.0, 10.0],
+        description="Target physical extent of each patch in degrees [height_deg, width_deg]. "
+        "Shared by the encoder and decoder for consistent spatial semantics.",
+    )
     embedding_dim: int = 128
     use_bfloat16: bool = Field(
         default=True,
@@ -620,13 +617,16 @@ class FOMOConfig(BaseModelConfig):
         static_data_for_corrector: xr.Dataset | None,
         srcs: list[DataSource],
     ) -> FOMO:
+        assert len(self.patch_extent) == 2, "patch_extent must be a pair of floats."
+        extent = self.patch_extent[0], self.patch_extent[1]
+
         all_grid_sizes = [s.grid_size for s in srcs]
         max_lat_size, max_lon_size = (
             max(g[0] for g in all_grid_sizes),
             max(g[1] for g in all_grid_sizes),
         )
         encoder = self.encoder.build(
-            in_channels, self.embedding_dim, max_lat_size, max_lon_size
+            in_channels, self.embedding_dim, extent, max_lat_size, max_lon_size
         )
         if (
             hasattr(encoder.perceiver, "use_flash_attn")
@@ -645,6 +645,7 @@ class FOMOConfig(BaseModelConfig):
         decoder = self.decoder.build(
             processor.out_channels,
             out_channels,
+            extent,
             max_lat_size,
             max_lon_size,
         )
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -625,18 +625,10 @@ class FOMOConfig(BaseModelConfig):
             max(g[0] for g in all_grid_sizes),
             max(g[1] for g in all_grid_sizes),
         )
+
         encoder = self.encoder.build(
             in_channels, self.embedding_dim, extent, max_lat_size, max_lon_size
         )
-        if (
-            hasattr(encoder.perceiver, "use_flash_attn")
-            and encoder.perceiver.use_flash_attn
-            and not self.use_bfloat16
-        ):
-            raise ValueError(
-                "Encoder is configured to use flash attention. Please set `use_bfloat16=True`."
-            )
-
         processor = self.processor.build(
             self.embedding_dim,
             self.pad,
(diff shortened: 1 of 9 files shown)


alxmrs:
@codex, may I have your review?

src/ocean_emulators/config.py
```diff
@@ -570,7 +589,12 @@ def build(
 class FOMOConfig(BaseModelConfig):
     encoder: EncoderConfig = EncoderConfig()
     processor: UNetBackboneConfig = UNetBackboneConfig()
-    # decoder will go here.
+    decoder: DecoderConfig = DecoderConfig()
```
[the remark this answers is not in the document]
alxmrs: I have a draft of a commit that pulls up the perceiver config into the FOMO config. The idea is that we would configure the perceive once and build it in the encoder/decoder (with different input and output channels). Is that a better alternative? It would lead to a less brittle config, but it gives up having separate settings for encoder vs decoder. @jder, what do you think we should do here? I'm torn.

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,107 @@
...
+
+class PerceiverDecoder(nn.Module):
+    """A perceiver-based decoder that mirrors the PerceiverEncoder.
+
+    The encoder compresses each patch of physical data into a single latent
+    token via perceiver cross-attention. This decoder inverts that: for each
+    latent-grid position it runs a perceiver that maps one token back to
+    ``out_channels``, then FOMO's ``unpatch`` layer expands each position to
+    full patch resolution.
+
```
jder: Before I look too closely at the code, I want to make sure I understand the goal here. Considering a single patch for a moment, it looks like we're mapping the latent space of size `embedding_dim` values (L=128 right now) -- and turning it into a different latent vector of size `out_channels` (`C=(hist+1)*num_prog=154` or whatever right now) and then the unpatch layer goes from that C-sized vector to a matrix of size C*PW*PH? So a single patch goes from a shape of (L=128) via a perceiver to a shape of (C=154) and then via a linear layer to a shape of (C, PW, PH) = (154, 6, 10) for one degree or (154, 24, 40) for 1/4 degree?

I don't understand why (a) we want to use a perceiver to map from one latent space to another before unpatching rather than going directly from the L-sized latent space to (C, PW, PH) and (b) if we do want that two-step decode why the output of the decoder would produce a vector of size C -- there's nothing that tells those values to be specific channels and the unpatchify step then needs to upscale that by a factor of PW*PH=60x for 1 degree or 960x for 1/4 degree which seems like an extreme bottleneck.

src/ocean_emulators/config.py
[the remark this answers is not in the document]
jder: I think pulling up things that we almost certainly want to share (like `implementation`) is a good idea and can be passed to each build function. I agree pulling up everything is probably too limiting.

src/ocean_emulators/models/modules/decoder.py
alxmrs: Thanks for asking this. I don't think I have a clear mental model of how the unpatchify process reckons with the decoder, which I imagine to be the reverse of the encoder. Maybe we could have a quick call to chat about this; I'd like to have a better understanding of how this all fits together.

alxmrs:
Hey @codex, may I please have a review?

alxmrs:
@codex may I have one more review?

src/ocean_emulators/models/fomo.py
```diff
@@ -87,38 +65,22 @@ def __init__(
...
-
         with autocast(enabled=self.use_bfloat16, dtype=torch.bfloat16):
             if self.maybe_add_3d_coordinates is not None:
                 fts = self.maybe_add_3d_coordinates(fts, ctx.input_resolution_cpu)
             fts = self.encoder(fts, ctx.input_resolution_cpu)
             fts = self.processor(fts)
+            fts = self.decoder(fts, ctx.input_resolution_cpu)
 
-        # Convert back to float32 for decoder and unpatchify operations
+        # Convert back to float32
```
jder, line 83: Maybe worth at least a TODO that we're kinda tricking ourself here by pretending we have float32 outputs given the decoder is producing float16 outputs. (Previously it was at least possible that the decoder could use a larger latent space to get better-than-float16 precision in the outputs)

alxmrs:

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        patch_extent: tuple[float, float],
+        queries_dim: int,
+        perceiver_io: nn.Module,
+        window_size: int | None = None,
```
alxmrs: no defaults

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # --- Reshape to spatial ---
+        out = rearrange(out, "b (h w) c -> b c h w", h=H, w=W)
+
+        return out
+
+    def _decode(
+        self,
+        data: torch.Tensor,
+        queries: torch.Tensor,
```
alxmrs: Let's use jaxtyping here.

alxmrs (commented):
First pass.

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,163 @@
...
+from ocean_emulators.constants import Lat, Lon
+
+
+class PerceiverDecoder(nn.Module):
+    """A PerceiverIO-based decoder that maps a latent patch grid to full-resolution output.
+
+    All ``nh * nw`` pos/scale-encoded latent tokens are passed as **data** to
+    the PerceiverIO[3], and every output pixel position is a **query**.  Each
+    query cross-attends to the full latent representation, giving it global
+    spatial context — pixels near patch boundaries can attend to neighboring
```
jder, line 22: I was expecting this to all still be per-patch, but we can try both ways. I suspect this will be much more expensive and not sure it will be a big win in accuracy.

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,163 @@
...
+    patches, and the model can learn smooth inter-patch transitions.
+
+    Concretely:
+
+    1. Add Aurora-style pos/scale encoding to the ``nh * nw`` latent tokens
+       (telling the model *where on the globe* each patch is).
+    2. Pass all encoded latents as **data** to the PerceiverIO:
+       ``(B, nh * nw, C)``.
+    3. Build normalized 2D **queries** for every output pixel ``(i/H, j/W)``
+       in ``[0, 1)``, embed them via a learned linear layer, and feed them
```
jder: FWIW this could also be lat/lon or spherical coordinates if we're doing a whole-earth decode.

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,163 @@
...
+       to the PerceiverIO decoder head.
+    4. The PerceiverIO's decoder cross-attends from queries to the internal
+       latents (which themselves cross-attended to all patch tokens),
```
jder: This means the Perceiver did some initial cross-attention on the (nh * nw, C) tokens? And then maybe some self-attention on those internal latents? And then finally we do a single cross-attention to the queries? Might be more clear to just write this in that order.

src/ocean_emulators/models/modules/decoder.py
jder: (If we do it per patch we're going to have to make up some way to go from C patch latents to NxM latents to hand to the perceiver, either just reshaping or a linear layer?)

alxmrs:
Claude wrote an analysis about why a non-patched PerceiverIO should scale for the decoder: 

**The latent grid is small**                                                                                                                                                      
                                                                                                                                                                                
  The decoder's data isn't the raw 720 x 1440 grid — it's the **processor output** after the encoder has compressed it. With patch_extent = [6, 10]:                                
 ```                                                                                                                                                                               
  ┌───┬─────┬────┬───────────┬────┐
  │ Resolution │  Raw grid  │ Patch size │ Latent grid (nh x nw) │ Tokens │
  ├───┼─────┼────┼───────┼───┤
  │ 1°         │ 180 x 360  │ 3 x 10     │ 60 x 36               │ 2,160  │
  ├────┼─────┼───┼────────┼───┤
  │ 0.5°       │ 360 x 720  │ 6 x 20     │ 60 x 36               │ 2,160  │
  ├────┼─────┼───┼────────┼───┤
  │ 0.25°      │ 720 x 1440 │ 12 x 40    │ 60 x 36               │ 2,160  │
  └────┴─────┴────┴───────┴───┘
```
  The latent grid is the **same size** across all resolutions — that's the whole point of patch_extent. So the PerceiverIO encoding step (learned latents cross-attend to ~2,160
  data tokens, then self-attend) is cheap and resolution-independent.

  **The queries are the bottleneck**

  The expensive part is the **decode cross-attention**: each pixel query attends to num_latents internal latents. The query count scales with resolution:
```
  ┌────┬─────┬─────┐
  │ Resolution │   H x W    │ Pixel queries │
  ├───┼───────────┤
  │ 1°         │ 180 x 360  │ 64,800        │
  ├───┼─────┼─────┤
  │ 0.5°       │ 360 x 720  │ 259,200    │
  ├───┼─────┼─────┤
  │ 0.25°      │ 720 x 1440 │ 1,036,800   │
  └───┴──────┴─────┘
```
  At quarter degree, that's ~1M queries. The decode cross-attention is O(num_queries x num_latents). With num_latents=512, that's ~500M attention elements per call. This is
  large but manageable because:

  1. The decode cross-attention is a **single layer** (no depth), and each query independently attends to the same num_latents latent vectors — it's a matmul, not a quadratic
  self-attention.
  2. **Windowing** directly addresses this: with window_size=4096, we'd make ~253 calls at quarter degree, each doing a cheap 4096 x 512 attention. The encoding is repeated per
  window (wasteful), but it's only ~2,160 tokens — fast.

  **Where it could hurt**

  The real concern is memory, not compute. All 1M query embeddings (H*W, queries_dim) are built at once before windowing (line 131-133 of the decoder). At queries_dim=64 in
  float32, that's ~250MB — fine. But the PerceiverIO's backward pass through windowed calls accumulates gradients across all windows, which could strain GPU memory.

  **Is the per-patch approach better?**

  The previous per-patch approach had B*nh*nw independent PerceiverIO calls, each with a tiny data/query set. The total work was similar, but it had no inter-patch context (the
   core problem we fixed). The current approach is strictly better architecturally — windowing gives the same computational chunking but every query still sees the full latent
  grid.

  Bottom line: The latent grid is small (~2K tokens) and fixed across resolutions. The query cost scales linearly and is handled by windowing. This approach should scale fine
  to quarter degree. I'd recommend setting window_size to something like 4096-8192 as a default rather than None, so it's safe by default at high resolutions.

**Comment by jder** [comment]
yeah, this broadly makes sense to me except that I have a suspicion we'll end up waiting to scale up the size of the latent grid (which changes the calculus significantly). At the moment each cell is 6x10 degrees? I strongly suspect it will be hard for it to retain fine details from the 1/4 degree data at that granularity.

File: src/ocean_emulators/models/modules/decoder.py (hunk 7 above)
Comment (alxmrs) [reply src/ocean_emulators/models/modules/decoder.py]: I like this idea, implementing it now.

**Comment by alxmrs** [comment]
@codex please review.

File: src/ocean_emulators/config.py [hunk 9]
```diff
@@ -366,51 +367,162 @@ def build(
...
+    When ``window_patches`` is set, the decoder tiles the output grid into
+    spatial blocks of that many patches per side.  Each block's PerceiverIO
+    call receives only the overlapping latent tokens plus ``context_patches``
+    extra rings of neighbors, keeping cost bounded even when the latent grid
+    is large (i.e. fine ``patch_extent``).
+    """
+
+    perceiver: PerceiverConfig = PerceiverConfig()
+    queries_dim: int = Field(
+        default=64,
```
jder, line 494: Is this a lot to encode what is ultimately a 2d surface or is it normal to expand this much?

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,236 @@
...
+        for pi in range(0, nh, wp):
+            for pj in range(0, nw, wp):
+                pi_end = min(pi + wp, nh)
+                pj_end = min(pj + wp, nw)
+
+                if cp is None:
+                    # Full context: every window sees all latent tokens.
+                    local_data = full_data
+                else:
+                    # Expand data region by context_patches, clamped to grid bounds.
```
jder: We probably want this to use some kind of circular padding to avoid lon=0 artifacts.

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        patch_extent: tuple[float, float],
+        queries_dim: int,
+        perceiver_io: nn.Module,
+        window_patches: int | None = None,
+        context_patches: int | None = 1,
```
jder: 🐑 suggest removing default values here and leaving them in config classes instead

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,236 @@
...
+        # nh, nw: number of patches along height and width (the latent grid dims).
+        B, C, nh, nw = x.shape
+        lat, lon = resolution
+
+        H, W = len(lat), len(lon)
+
+        # For pos/scale encoding we need a patch size that produces exactly
+        # nh*nw tokens.  In the full pipeline patch_from gives the same result,
+        # but in unit tests the input grid may not match, so derive from tensor.
+        pos_patch_h, pos_patch_w = H // nh, W // nw
```
jder: I don't understand this comment. The problem is that in unit tests, the grids are not exact multiples of the patch size? Can we fix the unit tests and assert this is true here? Otherwise, at the moment self.patch_extent is totally unused.

src/ocean_emulators/models/modules/decoder.py
jder: Also we should probably assert somewhere that if `window_patches` is None, `context_patches` is also None? (Since any other value is meaningless when window_patches is None, right?)

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,236 @@
...
+        perceiver_io: nn.Module,
+        window_patches: int | None = None,
+        context_patches: int | None = 1,
+    ) -> None:
+        super().__init__()
+        self.in_channels = in_channels
+        self.out_channels = out_channels
+        self.patch_extent = patch_extent
+        self.window_patches = window_patches
+        self.context_patches = context_patches
```
jder, line 99: I think there are no tests for context_patches at the moment; probably worth adding one (especially if we add some special padding logic)

jder (approved):
Thanks, looks great! One small ~correctness issue about padding but the rest are just suggestions for hardening or testing. 

I am still uncertain if we need any of this windowing/contexting/etc logic so probably worth revisiting after running some experiments (eg do they actually reduce memory usage etc)

src/ocean_emulators/config.py
alxmrs: The Perceiver paper uses a value in the 100s. According to Claude, this value is already on the small side.

src/ocean_emulators/models/modules/decoder.py
```diff
@@ -0,0 +1,248 @@
...
+            data = rearrange(data_grid, "b nh nw c -> b c nh nw")
+            data_windows = data.unfold(2, wp, wp).unfold(3, wp, wp)
+        else:
+            # Pad: circular along longitude (last dim), zero along latitude.
+            data = rearrange(data_grid, "b nh nw c -> b c nh nw")
+            data = F.pad(data, (cp, cp, 0, 0), mode="circular")
+            data = F.pad(data, (0, 0, cp, cp), mode="constant", value=0)
+            win_size_h = wp + 2 * cp
+            win_size_w = wp + 2 * cp
+            data_windows = data.unfold(2, win_size_h, wp).unfold(3, win_size_w, wp)
```
alxmrs, line 213: I'm going to merge now, but I'm happy to do a follow up on this new implementation of the context windows, @jder.

PR #606, opened 2026-02-25
Multi-scale models need to allow unused parameters in DDP training.

alxmrs:
In Multi-scale FOMO, we now include a ModuleDict of linear projections after decoding. These are treated as unused parameters during DDP training. It's expected that they should be unused since we only want to project for the current scale the batch is training on. Thus, to get multi-scale training to work, we need to conditionally pass a flag to turn this check off.

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -279,7 +279,15 @@ class Trainer:
             # per-resolution unpatch layers.  Under a "match" or "mix" schedule
             # only one resolution's unpatch layer participates in each batch,
             # so the others have no gradients.  DDP must be told about this.
-            has_unused = self.train_schedule != "standard"
+            match self.train_schedule:
+                case "standard":
+                    has_unused = False
+                case "mix":
+                    has_unused = True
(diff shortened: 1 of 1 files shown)


alxmrs:
@codex please review.

src/ocean_emulators/train.py
```diff
@@ -275,9 +275,15 @@ def __init__(self, cfg: TrainConfig) -> None:
 
         # Modify DDP setup based on device
         if self.distributed is not None:
+            # Multiscale models (e.g. FOMO with multiple grid sizes) have
+            # per-resolution unpatch layers.  Under a "match" or "mix" schedule
+            # only one resolution's unpatch layer participates in each batch,
+            # so the others have no gradients.  DDP must be told about this.
+            has_unused = self.train_schedule != "standard"
```
jder: 🐑 I might slightly prefer a match spelling out the right values for each case (with assert_never) here so that later changes to the possible values of train_schedule doesn't silently break this.

jder (approved):

Issue #600, opened 2026-02-23
Axial attention


jder:

jder:
Updates here: https://openathena.slack.com/archives/C08CYM42DT3/p1773181264684409

alxmrs:
- Adding 2d embeddings to attention as well. 
- Talking to Adam to get land masking code.

PR #611, opened 2026-03-02
axial attention blocks

amogh-gulati:
#600 

This branch tracks the development of axial and full attention in unet blocks for samudra v1 

training run with axial attention at bottleneck : https://wandb.ai/ocean_emulators/default/runs/ps6jz798

training run vanilla samudra v1: https://wandb.ai/ocean_emulators/default/runs/c7of0p5t 

 Full attention at bottleneck with bigger attention block : https://wandb.ai/ocean_emulators/default/runs/xku4fh3o
 
----

fix made: 
Changed unetbackbone to capture encoder skip tensors at the end of each encoder stage instead of coreblock. Earlier skip paths were not using the output from the attention layer instead they were just being forwarded after the convnext layer. 

This branch enables optional U-Net attention in two forms: axial attention and full attention. Axial attention applies self-attention separately along the two spatial axes, first treating each longitude column as a sequence over height and then each latitude row as a sequence over width. Two axis-wise passes are applied sequentially with residual connections. 

Full attention instead flattens the whole  H×W feature map into one sequence of spatial tokens and attends over all token pairs directly. 

In both cases, the channel dimension is treated as the embedding dimension for each spatial token, and multi-head attention splits those channels before the outputs are projected back and added residually to the feature map.

The branch also leaves an space for models at the bottleneck for future transformer-like block families, so the bottleneck can later host attention, transformer, or MaxViT-style blocks without another config redesign.

-------


alxmrs:
Lmk when this is no longer a draft PR and I'll be happy to review it!

amogh-gulati:
> Lmk when this is no longer a draft PR and I'll be happy to review it!

should be good now!

configs/samudra_om4_v1_attention/model.yaml
```diff
@@ -0,0 +1,42 @@
+# yaml-language-server: $schema=../schemas/SamudraConfig.json
```
alxmrs, line 1: Nit: We may have _just_ renamed our Samudra version to v2.

configs/samudra_om4_v1_attention/model.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+pred_residuals: false
+last_kernel_size: 3
+pad: "circular"
+checkpointing: "all"
+
+unet:
```
alxmrs, line 8: It occurs to me, reading the dropout config that you have in the new attention blocks, that we should consider adding drop path dropout: we should have a small chance of turning off skip connections. We can address this in a future issue.

configs/samudra_om4_v1_attention/README.md
```diff
@@ -0,0 +1,3 @@
+# Samudra OM4 V1 — Attention
+
+Use [model.yaml](/scratch/ag11542/ocean/Ocean_Emulator/Ocean_Emulator/configs/samudra_om4_v1_attention/model.yaml) as the base example and [train.yaml](/scratch/ag11542/ocean/Ocean_Emulator/Ocean_Emulator/configs/samudra_om4_v1_attention/train.yaml) to launch training.
```
alxmrs: Would love to have a high level explanation for the experiments in the README. In fact, what you wrote in the PR comment should be fine to re-purpose for this.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    - ``axis='height'``: treats each column position independently, attends along H
+    - ``axis='width'``: treats each row position independently, attends along W
+
+    This decomposes full 2D attention :math:`O((HW)^2)` into two sequential 1D
+    operations :math:`O(H^2 + W^2)`, making it tractable for large spatial grids.
+
+    References:
+        Axial Attention in Multidimensional Transformers (Ho et al., 2019)
+        https://arxiv.org/abs/1912.12180
+    """
```
alxmrs, line 337: This is a good docstring.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+    def __init__(
+        self,
+        dim: int,
+        num_heads: int = 8,
+        qkv_bias: bool = True,
+        attn_drop: float = 0.0,
+        proj_drop: float = 0.0,
+        axis: Literal["height", "width"] = "height",
```
alxmrs: Let's omit default arguments, if possible.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        self.qkv = nn.Linear(dim, dim * 3, bias=qkv_bias)
+        self.proj = nn.Linear(dim, dim)
+        self.proj_drop = nn.Dropout(proj_drop)
+
+        # When True, the next forward pass stores attention weights in
+        # ``self.last_attn_weights`` for visualization / debugging.
+        self.capture_weights = False
+        self.last_attn_weights: torch.Tensor | None = None
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
```
alxmrs: Let's use jaxtyping here to check (and document) the input/output shapes.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            # (B, C, H, W) -> (B*H, W, C): attend along width for each row
+            x = x.permute(0, 2, 3, 1).reshape(B * H, W, C)
+
+        batch_size, seq_len, _ = x.shape
+
+        qkv = self.qkv(x).reshape(batch_size, seq_len, 3, self.num_heads, self.head_dim)
+        qkv = qkv.permute(2, 0, 3, 1, 4)  # (3, batch, heads, seq, head_dim)
+        q, k, v = qkv.unbind(0)
+
+        # Use scaled_dot_product_attention (supports flash / memory-efficient kernels)
```
alxmrs, line 384: Awesome! TIL that we only need to add a context manager to turn on the FlashAttention implementation.

alxmrs: We don't need to do this now, but I could see an update to this module that would configure the implementations via a flag.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        self.attn_drop_p = attn_drop
+
+        self.qkv = nn.Linear(dim, dim * 3, bias=qkv_bias)
+        self.proj = nn.Linear(dim, dim)
+        self.proj_drop = nn.Dropout(proj_drop)
+
+        # When True, the next forward pass stores attention weights in
+        # ``self.last_attn_weights`` for visualization / debugging.
+        self.capture_weights = False
+        self.last_attn_weights: torch.Tensor | None = None
```
alxmrs, line 363: It's nice to have this debugging system. I liked seeing these plots in wandb.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        attn_drop: Dropout rate for attention weights.
+        proj_drop: Dropout rate for output projection.
+    """
+
+    def __init__(
+        self,
+        channels: int,
+        num_heads: int = 8,
+        attn_drop: float = 0.0,
+        proj_drop: float = 0.0,
```
alxmrs, line 440: Ideally, these should be managing in one place (centrally in our config system).

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+        if self.axis == "height":
+            out = out.reshape(B, W, H, C).permute(0, 3, 2, 1)
+        else:
+            out = out.reshape(B, H, W, C).permute(0, 3, 1, 2)
+
+        return out
+
+
+class AxialAttentionBlock(nn.Module):
```
alxmrs, line 419: This is a nice, clean implementation. Well documented, too. Good work.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        self.capture_weights = False
+        self.last_attn_weights: torch.Tensor | None = None
+        self.last_spatial_shape: tuple[int, int] | None = None
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        B, C, H, W = x.shape
+        seq_len = H * W
+
+        # (B, C, H, W) -> (B, H*W, C)
+        x = x.permute(0, 2, 3, 1).reshape(B, seq_len, C)
```
alxmrs: 🐑 consider using einsum notation to compute this, since you could more or less use the comment as the actual code itself.

alxmrs: If you do like that style better, consider using it in the axial blocks above.

alxmrs: I do find the style a bit more readable, since you can use names in their order instead of having to track the indexes in a permute call.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+        qkv = self.qkv(x).reshape(B, seq_len, 3, self.num_heads, self.head_dim)
+        qkv = qkv.permute(2, 0, 3, 1, 4)  # (3, B, heads, seq, head_dim)
+        q, k, v = qkv.unbind(0)
```
alxmrs, line 527: This is a neat way to use a single linear to set up all three sets of weights before applying attention. Out of curiosity, do you know if there is a performance advantage implementing attention this way?

src/ocean_emulators/models/modules/unet_backbone.py
```diff
@@ -53,6 +53,9 @@ def __init__(
         downsampling_block: nn.Module,
         create_upsampling_block: UpsamplingBlockBuilder,
         checkpointing: "Checkpointing | None",
+        encoder_attention_blocks: list[nn.Module | None] | None = None,
+        bottleneck_attention_block: nn.Module | None = None,
+        decoder_attention_blocks: list[nn.Module | None] | None = None,
```
alxmrs, line 58: Same default arg note.

configs/samudra_om4_v1_attention/model.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+  attention:
+    encoder:
+      - null
+      - null
+      - null
+      - attention_type: "axial"
+        num_heads: 8
+        attn_drop: 0.0
+        proj_drop: 0.0
```
alxmrs, line 31: I kind of thing that the naming and implementation here could be made a bit more generic. Like, I see no reason that we couldn't use other types of CNN blocks here, depending on the block type. I'm loath to think of a name. What if we called this "mixins" or "extra_blocks" or "custom_blocks"? Like, what is a term for core blocks vs non-core blocks? The default and the specific. In any case, this means that "axial" could be treated like any other block. When other blocks are added, ideally, it would only be a matter of configuration to register them in this section.

configs/samudra_om4_v1_attention/model.yaml
```diff
@@ -0,0 +1,42 @@
...
+    bottleneck:
```
alxmrs, line 32: I do like "bottleneck" as a part of this new section of config. I think it is nice that we can explicitly define what the bottleneck will consist of.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    Args:
+        channels: Number of input/output channels.
+        num_heads: Number of attention heads.  Must divide *channels* evenly.
+        attn_drop: Dropout rate for attention weights.
+        proj_drop: Dropout rate for output projection.
+    """
+
+    def __init__(
+        self,
+        channels: int,
```
alxmrs, line 567: A pattern that is pretty consistent throughout the codebase is to list "in_channels" and "out_channels". This mirrors our CNN neighbor blocks. I think it could make this easier to mix in attention in our UNet later on.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
```
alxmrs, line 437: Same as a comment below, but we should probably keep the pattern of specifying "in_channels" and "out_channels", even if these end up being the same value. I.e. we should at least set these two variables as properties of the class.

src/ocean_emulators/models/modules/blocks.py
alxmrs, line 597: Could the attention blocks also be core blocks?

alxmrs: Reflecting on what I said about the config, maybe we'd have a new factory protocol for the 
"mixin" blocks.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -284,6 +284,257 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    used interchangeably in the UNet backbone.
+
+    Args:
+        channels: Number of input/output channels.
+        num_heads: Number of attention heads.  Must divide *channels* evenly.
+        attn_drop: Dropout rate for attention weights.
+        proj_drop: Dropout rate for output projection.
+    """
+
+    def __init__(
```
alxmrs, line 565: Ideally, these new attention blocks should implement a gradient checkpointing scheme (simple or all, likely all).

alxmrs: Feel free to implement this in a follow up PR.

src/ocean_emulators/models/modules/unet_backbone.py
```diff
@@ -106,6 +113,9 @@ def __init__(
             )
         )
 
+        if bottleneck_attention_block is not None:
+            layers.append(bottleneck_attention_block)
+
```
alxmrs: Do we want both the middle block and the bottleneck block? Maybe, when we specify a bottleneck, we don't add this middle block. WDYT?

src/ocean_emulators/models/modules/unet_backbone.py
```diff
@@ -126,6 +136,10 @@ def __init__(
                     checkpoint_simple=checkpoint_simple,
                 )
             )
+            if decoder_attention_blocks is not None:
+                attention_block = decoder_attention_blocks[i]
+                if attention_block is not None:
+                    layers.append(attention_block)
```
alxmrs, line 143: I think this should come before the `create_block` step to properly mirror the creation. Is that right? Please double check me on that.

src/ocean_emulators/models/modules/unet_backbone.py
```diff
@@ -139,6 +153,10 @@ def __init__(
                 checkpoint_simple=checkpoint_simple,
             )
         )
+        if decoder_attention_blocks is not None:
+            attention_block = decoder_attention_blocks[-1]
+            if attention_block is not None:
+                layers.append(attention_block)
```
alxmrs, line 160: Similarly, wouldn't it make sense for this to come before the final conv block?

src/ocean_emulators/config.py
```diff
@@ -243,6 +245,112 @@ def make_source(
 BlockType = Literal["conv_next_block", "conv_block"]
 ActivationType = Literal["relu", "gelu", "capped_gelu"]
 NormType = Literal["batch", "instance", "layer"]
+AttentionType = Literal["axial", "full"]
+BottleneckBlockType = Literal["attention", "transformer", "maxvit"]
```
alxmrs: Nice! Good to have these options.

src/ocean_emulators/config.py
```diff
@@ -243,6 +245,112 @@ def make_source(
...
+            )
+        assert_never(self.attention_type)
+
+
+class TransformerBottleneckConfig(BaseConfig):
+    """Reserved config surface for a future transformer bottleneck block."""
+
+
+class MaxViTBottleneckConfig(BaseConfig):
+    """Reserved config surface for a future MaxViT bottleneck block."""
```
alxmrs: Let's put a TODO here instead. I don't like the idea of adding a configuration that is not implemented yet.

src/ocean_emulators/config.py
```diff
@@ -243,6 +245,112 @@ def make_source(
...
+
+    attention_type: AttentionType = Field(
+        default="axial",
+        description="Attention implementation to use at this stage.",
+    )
+
+    num_heads: int = Field(
+        default=8,
+        description="Number of attention heads. Must divide the stage channel width evenly.",
+    )
```
alxmrs, line 263: Can we add an assert in the `build()` method below that checks for this property (that it divides the channel dim)?

src/ocean_emulators/config.py
```diff
@@ -243,6 +245,112 @@ def make_source(
...
 NormType = Literal["batch", "instance", "layer"]
+AttentionType = Literal["axial", "full"]
+BottleneckBlockType = Literal["attention", "transformer", "maxvit"]
+
+
+class AttentionBlockConfig(BaseConfig):
+    """Configuration for a single attention block in the U-Net."""
+
+    attention_type: AttentionType = Field(
+        default="axial",
```
alxmrs: 🐑 Let's make "full" attention the default. WDYT?

src/ocean_emulators/config.py
alxmrs: Same with the class immediately above.

src/ocean_emulators/config.py
```diff
@@ -243,6 +245,112 @@ def make_source(
...
+        description="Attention bottleneck settings used when `block_type` is `attention`.",
+    )
+    transformer: TransformerBottleneckConfig | None = Field(
+        default=None,
+        description="Reserved transformer bottleneck settings for future implementation.",
+    )
+    maxvit: MaxViTBottleneckConfig | None = Field(
+        default=None,
+        description="Reserved MaxViT bottleneck settings for future implementation.",
+    )
```
alxmrs: Let's omit these for now. Happy to add notes or TODOs here that these will come later.

src/ocean_emulators/config.py
```diff
@@ -582,6 +713,35 @@ def create_upsampling_block(in_channels: int, out_channels: int):
...
+            if bottleneck_attention_config is not None
+            else None
+        )
+        decoder_attention_blocks: list[nn.Module | None] | None = None
+        if decoder_attention_configs is not None:
+            decoder_attention_blocks = []
+            for cfg, channels in zip(
+                decoder_attention_configs,
+                reversed(self.ch_width),
+                strict=True,
```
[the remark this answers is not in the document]
alxmrs: This may be related to the block creation order that I pointed out above.

src/ocean_emulators/config.py
```diff
@@ -243,6 +245,112 @@ def make_source(
 BlockType = Literal["conv_next_block", "conv_block"]
 ActivationType = Literal["relu", "gelu", "capped_gelu"]
 NormType = Literal["batch", "instance", "layer"]
+AttentionType = Literal["axial", "full"]
+BottleneckBlockType = Literal["attention", "transformer", "maxvit"]
+
+
+class AttentionBlockConfig(BaseConfig):
```
alxmrs, line 252: In an ideal world, this would be rolled into `BlockConfig`. Maybe, to accomplish this, we would update the `BlockConfig` to be a union type, so we could pass context specific arguments depending on if the block was a CNN or Attention.

alxmrs: I'm imagining something similar to how the ModelConfig is done today. This is a bit tricky, and I (or I bet, Jesse) would be happy to meet to try to work through the pattern with you on this. I admit, the implementation path for this is fuzzy in my head.

alxmrs (changes requested):
I think there is a mistake that Codex and I caught related to the order of creating blocks on the decode side of the unet. I also provide some design suggestions for this feature. In general, the Axial and full transformers implementation are excellent. Here are some notes to help improve the PR for the next iteration.

configs/samudra_om4_v1_attention/model.yaml
amogh-gulati: I'm using the base v1 configs to run experiments with attention to keep the changes isolated, should I use v2 instead?

src/ocean_emulators/models/modules/blocks.py
amogh-gulati: yes!

src/ocean_emulators/config.py
amogh-gulati: agreed

src/ocean_emulators/models/modules/blocks.py
amogh-gulati: done!

src/ocean_emulators/models/modules/blocks.py
amogh-gulati: My understanding is that using one linear for QKV might be better than three separate projections because it reduces kernel, I have not benchmarked it though.

src/ocean_emulators/models/modules/unet_backbone.py
amogh-gulati: I agree. I think the cleaner semantics are that the bottleneck config should define the bottleneck stage itself. So if a bottleneck block is specified, it should replace the default middle block rather than be appended after it.

src/ocean_emulators/models/modules/unet_backbone.py
amogh-gulati: decoder attention runs after the decoder conv block, just like for encoder

src/ocean_emulators/models/modules/unet_backbone.py
amogh-gulati: same for above

src/ocean_emulators/config.py
amogh-gulati: What if we model each stage as an ordered list of components plus an explicit skip-capture point, instead of fixing every stage to has the same internal ordering? 
In the current CNN-style U-Net, the natural order is coreblock then scale change but for a ViT-style stage with patch embedding is tokenize first (scale change), then transform.  

(maybe transform keeps the spatial size consistent and transition changes it)

encoder_stages:
  - skip_after_component: 1
    components:
      - kind: transform
        block_type: convnext
      - kind: transform
        attention_type: axial
      - kind: transition
        transition_type: avg_pool

  - skip_after_component: 0
    components:
      - kind: transition
        patch_size: 4
        embed_dim: 256
      - kind: transform
        block_type: transformer
 
but this makes the skip logic pretty messy

alternative, a shape-preserving StageBlock abstraction that can contain one or more stage-preserving operations, such as a core block, an attention block, or both, and then keep ViTStageBlock as a separate stage type for the cases where tokenization/downsampling is coupled with the stage computation to keep skips simplified, VIT is the only one where we change shape before the feature mixing instead of after. It also keeps the skip logic cleaner, since the backbone would still only need to reason about one declared stage output per level

amogh-gulati: Right now, the conv or ConvNeXt block is the main block for each U-Net level. The attention block is just an extra step added after that main block. For the experiments with each level as a conv-based stage with optional attention on top the current config seems to let us experiment with axial and full attention on different coarse resolutions and if we dont need attention it does not pollute the config. 

I think Convnext + attention blocks will be a more natural fit for the coreblock moving forward (as they define a stage and the current attention block do not, they are just a optional bolt on) and for patchify + transformer based design a new model like fomo might be the most apt way to experiment

@alxmrs let me know what you think here, happy to discuss this more

configs/samudra_om4_v1_attention/model.yaml
alxmrs: Yes, I think we should base these off of v2.

src/ocean_emulators/models/modules/blocks.py
alxmrs: Is it possible to have these values be set by our config system? I'd prefer not to use default kwargs.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -319,6 +320,272 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        x: Float[torch.Tensor, "batch channels height width"],
+    ) -> Float[torch.Tensor, "batch channels height width"]:
+        # Height-axis attention with residual
+        x = x + self.attn_h(self.norm_h(x))
+        # Width-axis attention with residual
+        x = x + self.attn_w(self.norm_w(x))
+        return x
+
+
+class FullAttention(nn.Module):
```
alxmrs, line 473: In the future, I might slightly update this module to support both self and cross attention; since this PR doesn't need it, this is good for now.

src/ocean_emulators/models/modules/blocks.py
alxmrs: Happy to punt on in/out channels; this is now made more complicated with "prog_channels" and "boundary_channels" in a WIP feature.

src/ocean_emulators/models/modules/unet_backbone.py
alxmrs: In my understanding, this will break the "U" shaped symmetry of the network. Mirroring means reversing the order. This is only coming up now because we only ever had one kind of block.

If there is a convention for not having a strict reflection in symmetry, then I'm ok with this; I do want to check first.

src/ocean_emulators/config.py
alxmrs: Mind if we discuss this over video chat after our weekly meeting? I'd like to understand what you're proposing a bit better (and ask some questions).

alxmrs (commented):
Another review pass. Sorry I missed your responses from earlier this week.

PR #617, opened 2026-03-03
Fixes to get the flash perceiver working.

alxmrs:
Tested with a really small model on an A10 GPU via AWS.


alxmrs:
Closing in favor of #619.

PR #619, opened 2026-03-03
Fixes to get flash perceiver working.

alxmrs:
Closing #617 in favor of this PR, which has a cleaner history. This was tested on a small Flash Perceiver model config on an A10 GPU on AWS.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,8 +123,11 @@ overrides = [
     "fsspec.*",
     "perceiver_pytorch.*",
     "aurora.*",
-    "flash_perceiver.*",
   ], follow_untyped_imports = true },
+  { module = [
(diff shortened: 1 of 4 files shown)


alxmrs:

jder (approved):
Looks good!

FWIW this kind of stuff makes me feel like we should get rid of the non-flash attn. The checkpoints aren't compatible IIUC and I'm not sure it's worth all the extra code and logic. Just something to think about if this becomes more of a pain!

alxmrs:
I do like being able to test the core of our project without a cuda enabled device! Maybe instead of getting rid of the naive implementation, we simply rolled our own and had it work on CPU. In general, I like the idea of removing cruft.

Issue #613, opened 2026-03-02
Fix DMSE weighting to be sqrt(error), try other Adam suggestions


jder:
Context: https://openathena.slack.com/archives/C08CYM42DT3/p1771957988594599?thread_ts=1771434901.275589&cid=C08CYM42DT3

Single-container plan:
- Build one image from the branch that contains the rebased `2-in-1-out` work, `pos-only`, inverse-sqrt dynamic loss weighting, and config-driven init.
- Use `CONFIG=configs/samudra_om4/train.yaml` as the shared base config.
- Positional runs below assume `--model.pos_channels=16`.
- For `2-in-1-out`, rerun training with `--steps=[8]` to match the stability of the default 4-step `2-in-2-out` setup.

Train args per experiment: (See [all here](https://wandb.ai/ocean_emulators/default/groups/2026-03-14-sweep/workspace))
* [ ] 100-cap + delayed loss estimate  `ARGS="--loss.limit=100 --delayed_loss_estimate=true"`
* [x] 02: non-dynamic MAE + new pos embeddings
  `ARGS="--loss=mae --model.pos_channels=16"` [train](https://wandb.ai/ocean_emulators/default/runs/cq5s3mjy), [eval](https://wandb.ai/ocean_emulators/default/runs/2tuok913)
* [x] 03: dynamic MSE with sqrt(err) scaling + delayed loss estimate + 100-cap
  `ARGS="--loss.type=dynamic --loss.metric=mse --loss.weighting=inverse_sqrt_loss --loss.limit=100 --delayed_loss_estimate=true"` [train](https://wandb.ai/ocean_emulators/default/runs/wgyokowc), [eval](https://wandb.ai/ocean_emulators/default/runs/felaz708)
* [x] 04: dynamic MSE with sqrt(err) scaling + delayed loss estimate + 100-cap + 2-in-1-out
  `ARGS="--loss.type=dynamic --loss.metric=mse --loss.weighting=inverse_sqrt_loss --loss.limit=100 --delayed_loss_estimate=true --data.num_in_states=2 --data.num_out_states=1 --steps=[8]"` [train](https://wandb.ai/ocean_emulators/default/runs/rzzz1p5q), [eval](https://wandb.ai/ocean_emulators/default/runs/93o5dsga)
* [x] 05: dynamic MSE with sqrt(err) scaling + delayed loss estimate + 100-cap + 2-in-1-out + new pos embeddings
[...]

PR #620, opened 2026-03-03
Mark current model v2

jder:
At the moment it's the same as the current samudra but I intend to iterate on the current one based on #613.

I wrote this whole PR myself!

### DIFF
--- a/configs/samudra_om4/README.md
+++ b/configs/samudra_om4/README.md
@@ -1,5 +1,4 @@
 # Samudra OM4
 
-These configurations are our current best Samudra-style (UNet) model,
-trained on OM4 data. Major changes from the v1 model are that it is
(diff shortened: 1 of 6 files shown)


configs/samudra_om4/README.md
```diff
@@ -1,5 +1,4 @@
 # Samudra OM4
```
alxmrs, line 1: 🐑 What do you think of having the version in the title? (Same with below).

alxmrs (approved):
One nit, otherwise, these configs LGTM.

configs/samudra_om4/README.md
jder: This is the unversioned (next/current) readme, but I added it to the v2 readme below.

Issue #594, opened 2026-02-20
Metrics suite


jder:
Laure expects a draft of a set of metrics we should be driving forward on by the week of Feb 32. This issue is to capture needed work to turn those drafts into automated metrics logged during training/rollout.

PR #622, opened 2026-03-04
Spectra logging in eval

jder:


src/ocean_emulators/aggregator/spectra.py
```diff
@@ -0,0 +1,344 @@
...
+    def _deg_lon_to_km(delta_deg: float, lat_deg: float) -> float:
+        return (
+            abs(delta_deg)
+            * _KM_PER_DEGREE
+            * max(abs(math.cos(math.radians(lat_deg))), 1e-6)
+        )
+
+    @staticmethod
+    def _deg_lat_to_km(delta_deg: float) -> float:
+        return abs(delta_deg) * _KM_PER_DEGREE
```
jder, line 114: @YuanYuan98 or maybe @Pperezhogin we're using an approximation here for the grid spacing ultimately passed to torch.fft.rfftfreq below (we use the median delta-degrees in the region of interest, converted to km with this code). Do you think this is sufficient for spectra logging (here is [an example](https://wandb.ai/ocean_emulators/default/runs/odck1yr1?nw=nwuseroa_jder)) or does this seem like it's going to cause confusion/trouble down the road?

jder:
Replacing this with the #594 work.

PR #623, opened 2026-03-04
Turn off in memory cache when opening Xarray dataset (default is having it on)

alxmrs:
Claude found a simple, but potentially high ROI CPU memory optimization. Since Xarray targets analytics workloads, it turns on an in memory cache by default. This flag turns this off, since we only ever access a `sel` of data once. 

This should help address the OOM memory issue I faced when trying to load three Zarr datasets (1, 1/2, 1/4 degree versions of OM4) in a multi-scale training run on our SLURM cluster.

### DIFF
--- a/src/ocean_emulators/utils/location.py
+++ b/src/ocean_emulators/utils/location.py
@@ -82,10 +82,8 @@ class S3Location(ResolvedLocation, BaseModel):
             backend_kwargs={"storage_options": {"endpoint_url": self.endpoint_url}},
             engine="zarr",
             chunks=chunks,
-            # Disable xarray's MemoryCachedArray when not using dask.
-            # Without this, the first .isel().to_numpy() on any variable
-            # caches the ENTIRE array in application memory (not just the
-            # requested slice), causing OOM with forked DataLoader workers.
+            # Disable Xarray's MemoryCachedArray regardless of whether
+            # we're using dask (normally it is enabled when dask is disabled).
             # The OS file cache handles repeated reads efficiently.
             cache=False,
         )
@@ -138,10 +136,8 @@ class LocalLocation(ResolvedLocation, BaseModel):
 
     def open(self, chunks: dict[str, int] | None = None) -> xr.Dataset:
         engine = "netcdf4" if self.path.suffix == ".nc" else "zarr"
-        # Disable xarray's MemoryCachedArray when not using dask.
-        # Without this, the first .isel().to_numpy() on any variable
-        # caches the ENTIRE array in application memory (not just the
-        # requested slice), causing OOM with forked DataLoader workers.
+        # Disable Xarray's MemoryCachedArray regardless of whether
+        # we're using dask (normally it is enabled when dask is disabled).
         # The OS file cache handles repeated reads efficiently.
         return xr.open_dataset(self.path, engine=engine, chunks=chunks, cache=False)
 
--- a/src/ocean_emulators/utils/location.py
+++ b/src/ocean_emulators/utils/location.py
@@ -82,6 +82,12 @@ class S3Location(ResolvedLocation, BaseModel):
             backend_kwargs={"storage_options": {"endpoint_url": self.endpoint_url}},
             engine="zarr",
             chunks=chunks,
+            # Disable xarray's MemoryCachedArray when not using dask.
+            # Without this, the first .isel().to_numpy() on any variable
+            # caches the ENTIRE array in application memory (not just the
+            # requested slice), causing OOM with forked DataLoader workers.
+            # The OS file cache handles repeated reads efficiently.
+            cache=False,
         )
 
     def url(self) -> str:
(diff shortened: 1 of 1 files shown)


alxmrs:

src/ocean_emulators/utils/location.py
```diff
@@ -82,6 +82,12 @@ def open(self, chunks: dict[str, int] | None = None) -> xr.Dataset:
             backend_kwargs={"storage_options": {"endpoint_url": self.endpoint_url}},
             engine="zarr",
             chunks=chunks,
+            # Disable xarray's MemoryCachedArray when not using dask.
+            # Without this, the first .isel().to_numpy() on any variable
+            # caches the ENTIRE array in application memory (not just the
```
jder: Surely not the "ENTIRE array", right? If we were caching the entire array that would be the whole 1.5 TB dataset in memory for each data loader worker when training on 1/4-degree.

src/ocean_emulators/utils/location.py
```diff
@@ -82,6 +82,12 @@ def open(self, chunks: dict[str, int] | None = None) -> xr.Dataset:
             backend_kwargs={"storage_options": {"endpoint_url": self.endpoint_url}},
             engine="zarr",
             chunks=chunks,
+            # Disable xarray's MemoryCachedArray when not using dask.
```
jder: How about "Disable xarray's MemoryCachedArray regardless of whether we're using dask (normally it is enabled when dask is disabled)".

src/ocean_emulators/utils/location.py
```diff
@@ -82,6 +82,12 @@ def open(self, chunks: dict[str, int] | None = None) -> xr.Dataset:
...
+            # Without this, the first .isel().to_numpy() on any variable
+            # caches the ENTIRE array in application memory (not just the
+            # requested slice), causing OOM with forked DataLoader workers.
+            # The OS file cache handles repeated reads efficiently.
```
jder, line 87: The OS cache might be efficient but xarray overhead is a major concern thing we're trying to avoid in this code generally. It seems quite plausible that we're (unintentionally) relying on this cache for performance right now.

jder (commented):
A few comments, but at a high level if you can get a training run that shows this reduces memory usage and does not regress data loader performance I'm all for it!

src/ocean_emulators/utils/location.py
alxmrs: Yes, on second look this is a bit extreme. I've looked at bit at the implementation (which is a bit inscrutable), but my best guess is that it caches chunks of data at a time, and specifically caches arrays per variable.

src/ocean_emulators/utils/location.py
alxmrs: It's plausible, but I don't think we are. Maybe I misunderstand how this relates, but in data loading, we need random access across the time dimension. We load the data we need for the example and move on to the next one before relying on the cached info. 

Maybe, this isn't needed because we move the data to torch ASAP.

jder: Yeah I agree we are probably not relying on it but I would feel a bit better with a training run showing that 🙂

alxmrs:
> if you can get a training run that shows this reduces memory usage 

I didn't think we had anything that scientific, but then I remembered that the system characteristics are recorded in wandb! Check out this plot in particular: In the first run, I quickly used all processor memory (before this change). In the v1 run, I was able to train for 500 steps (before being cut off from preemption) -- but, there was process memory available. 

https://wandb.ai/ocean_emulators/default/panel/9m3e3uivc?nw=nwuseralxmrs

> and does not regress data loader performance I'm all for it!

Unfortunately, these multi-scale runs don't seem to have log aggregations for data loading. That is definitely something that I should add back in! I'll look into adding these logs soon so we can say so for sure.

jder:
Am I comparing the correct two runs? It looks like the v1 run had less memory available at all times and the earlier run was nowhere near to being out of memory? 

Screenshot 2026-03-05 at 7 57 13 PM


W&B is inexact, of course, but seems suspicious?

jder:
`train/batch/iter_time` is probably the most reliable way to answer the question of whether things have slowed down (at least in training).

jder:
Sorry for the comment chain, but one last thought: it's usually easier to see wins in memory usage in smaller runs where neither runs out of memory. 

Looks like across those two runs `num_workers` also changes?

alxmrs:
> train/batch/iter_time is probably the most reliable way to answer the question of whether things have slowed down (at least in training).

Thanks, I thought that was in its own dedicated bucket! Here's a comparison of this new run and a standard samudra run -- not a fair comparison, but it does indeed seem much slower.

https://wandb.ai/ocean_emulators/default?nw=lubb5ki452&panelDisplayName=train%2Fbatch%2Fiter_time&panelSectionName=train

alxmrs:
In the next kr1 experiment, I'll revert this change and see if it's really the `num_workers` config that is addressing the memory issue.

jder:
> In the next kr1 experiment, I'll revert this change and see if it's really the num_workers config that is addressing the memory issue.

Sounds great. If that's true I am truly confused FWIW. 

Also if you want to avoid running all this stuff of a branch, maybe throw this boolean under a config flag and we can merge it?

alxmrs:
Closing this issue -- I have now identified the culprit for my CPU OOM issue with initializing data loader workers (left over anomaly calculations copied many times). We can avoid this issue just by not using the anomaly boundary data, or pre-computing it (future direction). Since this flag isn't necessarily broken, I'm not going to mess with it for now.

PR #624, opened 2026-03-05
Adding small gif to README, compressed with `gifsicle`

alxmrs:
I first used ffmpeg to convert the MP4 to a (large) gif. Then, I used gifsicle with the arguments `gifsicle -O3 --lossy=30 --resize-width 800` to produce this ~9 MB gif of the figure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,7 @@
 
 ## Sea Surface Temperature (SST) Emulation
 
-<p align="center">
-  <video width="800" controls autoplay loop muted playsinline>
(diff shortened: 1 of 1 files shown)


alxmrs:
@codex may I have a review?

alxmrs:
Sorry @YuanYuan98, bit it looks like the absolute path to the video didn't fix the README issue. I looked around, and I think I may have deluded both of us into thinking that GitHub supported embedding videos in READMEs. I'm sorry to have done that, but it looks like we do need to use a gif after all. This PR is the same as the original attempt, but includes a much more compressed gif.

YuanYuan98 (approved):
Let's give it a shot. (

PR #626, opened 2026-03-05
Updating training to be preemptable on torch.

alxmrs:
This updates the train script, our torch docs and the sbatch script to allow us to make use of preemptable computing resources on torch. This PR is more of a refresh than anything else.

This is currently a draft because I want to test this out to see if it works.

### DIFF
--- a/docs/torch.md
+++ b/docs/torch.md
@@ -62,10 +62,14 @@ It expects environment variables:
 - `WANDB_API_KEY` (optional): if set and `WANDB_MODE` unset, defaults to W&B online
 - `WANDB_MODE` (optional): `online` or `disabled` (if unset, defaults based on
   whether `WANDB_API_KEY` is present)
+- `PREEMPTIBLE` (optional): set to `1` to enable preemptible training (allows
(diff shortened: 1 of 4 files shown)


jder:
Awesome! I was just thinking about needing this after the torch HPC session I went to today. Happy to share some notes but if you look in the torch docs you can probably find the additional sbatch args you need to opt into preemptible partitions.

alxmrs:
I figured this would help me iterate a lot faster! Yeah, I fed in the NYU HPC Google Sites page into Claude, and it determined what flags needed to be modified in our sbatch script. It seems that there are useful OS signals on SLURM that we can incorporate to make our training more resumable. I'll see if this works out in practice.

jder:
I don't believe the PR currently includes the correct flags to opt into the preemptible partitions

alxmrs:
I checked the docs again, it looks like you're right. Before, this script was using --requeue, but it seems like we actually need to use --comment. Thanks, Jesse, good catch.

jder:
We're going to close this for now, can revive later if needed.

PR #628, opened 2026-03-05
Initial FOMO model hyperparamter configuration.

alxmrs:
Turns on flash perceiver implementation. Configures the decoder to match the encoder in terms of depth and width. In addition, this configures the decode strategy for the decoder (window patch size, query dim, number of context patches for the memory-bound decode). Last, but not least, this sets a slightly higher spatial extent for patches to produce a 60 x 72 latent grid.

This was the configuration used in this run: https://wandb.ai/ocean_emulators/default/runs/0zeuoacu

### DIFF
--- a/configs/fomo_om4/model.yaml
+++ b/configs/fomo_om4/model.yaml
@@ -5,13 +5,24 @@ last_kernel_size: 3
 pad: "circular"
 checkpointing: "all"
 embedding_dim: 128  # Reduced from 280
-perceiver_implementation: "naive"
+perceiver_implementation: "flash"
+# TODO(alxmrs): How fine of a resolution should each patch be? This produces a 60x72 latent grid.
+patch_extent: [3.0, 5.0]
 
 encoder:
   perceiver:
     depth: 2  # Reduced from 3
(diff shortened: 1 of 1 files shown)


configs/fomo_om4/model.yaml
```diff
@@ -5,14 +5,25 @@ last_kernel_size: 3
 pad: "circular"
 checkpointing: "all"
 embedding_dim: 128  # Reduced from 280
-perceiver_implementation: "naive"
+perceiver_implementation: "flash"
+# TODO(alxmrs): How fine of a resolution should each patch be? This produces a 60x72 latent grid.
+patch_extent: [3.0, 5.0]
```
jder, line 10: Just my 2 cents but I think this will probably need to end up being closer to 0.5 degrees to have any hope of replicating 1/4 degree behavior.

jder (approved):

configs/fomo_om4/model.yaml
alxmrs: I'll target that patch size for one of my next runs!

PR #635, opened 2026-03-06
FOMO now uses all parameters.

### LINKED ISSUE #630: DDP find_unused_parameters=True causes extra autograd traversal every iteration
🤖

## Problem

When training with the `match` or `mix` schedule, we set `find_unused_parameters=True` in the DDP constructor. However, all 8 ranks report that no unused parameters were found:

```
[rank0]:[W306 14:11:30.969179112 reducer.cpp:1431] Warning: find_unused_parameters=True was specified in DDP constructor, but did not find any unused parameters in the forward pass. This flag results in an extra traversal of the autograd graph every iteration, which can adversely affect performance.
```

This warning fires on every rank, every first batch. The extra autograd graph traversal has a non-trivial per-iteration cost.

## Context

The flag is set in `train.py` because multi-scale models (e.g. FOMO with multiple grid sizes) have per-resolution unpatch layers, and under a `match` or `mix` schedule only one resolution's unpatch layer participates in each batch. However, the warning suggests this isn't actually happening — possibly because the model architecture changed or because the unused parameters are handled differently than expected.

The warning itself notes it could be a false positive due to flow control, so this needs investigation rather than a blind change.

## Potential approaches

- Investigate whether `find_unused_parameters=True` is actually needed for the current FOMO architecture under `match`/`mix` schedules
- If it's truly unnecessary, remove it for a free per-iteration performance gain
- If it's needed only intermittently, consider using `torch.nn.parallel.DistributedDataParallel`'s `no_sync()` context or setting unused params statically

alxmrs:
Fixes #630.

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -275,23 +275,9 @@ class Trainer:
 
         # Modify DDP setup based on device
         if self.distributed is not None:
-            # Multiscale models (e.g. FOMO with multiple grid sizes) have
-            # per-resolution unpatch layers.  Under a "match" or "mix" schedule
-            # only one resolution's unpatch layer participates in each batch,
-            # so the others have no gradients.  DDP must be told about this.
-            match self.train_schedule:
-                case "standard":
-                    has_unused = False
-                case "mix":
-                    has_unused = True
-                case "match":
-                    has_unused = True
-                case _:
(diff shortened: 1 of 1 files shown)


alxmrs:

jder (approved):
:all-the-parameters:

PR #637, opened 2026-03-06
Adding pointwise linear block as an alternative to 1x1 Conv.

### LINKED ISSUE #631: Grad strides mismatch warning during DDP backward pass
🤖

## Problem

During the first backward pass, PyTorch's DDP reducer logs a warning about gradient stride mismatch:

```
Grad strides do not match bucket view strides. This may indicate grad was not created
according to the gradient layout contract, or that the param's strides changed since
DDP was constructed. This is not an error, but may impair performance.
grad.sizes() = [320, 256, 1, 1], strides() = [256, 1, 256, 256]
bucket_view.sizes() = [320, 256, 1, 1], strides() = [256, 1, 1, 1]
```

The gradient has strides `[256, 1, 256, 256]` while DDP expects `[256, 1, 1, 1]`. This forces DDP to copy gradients into contiguous buckets rather than using views, which adds overhead to every gradient synchronization.

## Context

This likely comes from a Conv2d layer with kernel size 1×1 where the gradient layout doesn't match what DDP expects. The parameter shape `[320, 256, 1, 1]` suggests a pointwise convolution, possibly in the FOMO encoder/decoder or ConvNeXt blocks.

## Potential approaches

- Identify which layer(s) produce non-contiguous gradients
- Consider calling `.contiguous()` on the parameter or restructuring the layer to avoid the stride mismatch
- Alternatively, this may be acceptable if the copy overhead is small relative to the gradient sync

alxmrs:
Fixes #631. Also, this more closely mimics the official implementation of ConvNext: https://github.com/facebookresearch/ConvNeXt/blob/main/models/convnext.py#L18.

### DIFF
--- a/src/ocean_emulators/models/modules/blocks.py
+++ b/src/ocean_emulators/models/modules/blocks.py
@@ -239,7 +239,7 @@ class ConvNeXtBlock(CoreBlock):
         upscale_factor: int = 4,
         norm="batch",
         checkpoint_simple: bool = False,
-        pointwise_linear: bool = True,
+        pointwise_linear: bool = False,
     ):
         super().__init__(in_channels, out_channels, kernel_size, dilation, pad)
         assert n_layers == 1, "Can only use a single layer here!"
--- a/src/ocean_emulators/models/modules/blocks.py
+++ b/src/ocean_emulators/models/modules/blocks.py
@@ -4,6 +4,7 @@ from typing import Protocol
 import torch
 import torch.nn as nn
 import torch.utils.checkpoint
+from jaxtyping import Float
 
 from ocean_emulators.models.modules.activations import CappedGELU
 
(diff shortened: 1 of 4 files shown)


alxmrs:

alxmrs:

alxmrs:
@codex, may I have a review?

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -4,10 +4,33 @@
...
 from ocean_emulators.models.modules.activations import CappedGELU
 
 
+class PointwiseLinear(torch.nn.Module):
+    """A 1×1 convolution implemented as nn.Linear.
+
+    Mathematically equivalent to Conv2d(kernel_size=1), but avoids the
+    non-contiguous gradient strides that 1×1 convs produce, which cause
+    DDP to copy gradients instead of using zero-copy views.
+
```
jder, line 18: @fomo-bot can you help me understand what "non-contiguous gradient strides" means in this context and what it has to do with DDP or copying/views? Do you think this will be a (time) performance improvement?

jder (approved):

PR #640, opened 2026-03-10
Remove partition specification from sbatch script

jder:
NYU HPC recommends account and GPU requirements and not specify partitions.


alxmrs (approved):
LGTM, thanks!

PR #641, opened 2026-03-10
Stack 1/2: Initial configuration of skip connection drop path drop out.

alxmrs:
This PR introduces a form of drop path dropout on skip connections. This is put behind a flag we call "shortcut dropout rate", to refer to this stochastic form of "pruning" the skip connections from the encoder to decoder of the U-Net. 

We've internally conducted a few experiments (I'll put a link to a doc about this later) that provides some evidence that the standard U-Net structure without drop-out under-exercised the deeper layers of the U-Net, including the bottleneck. Further experiments using a variant of this feature (see part 2 of the stack; [early results](https://docs.google.com/presentation/d/1Wq4BLDJ3ptUhnhgIPSTG1MtCa3LcOM1Q0kvV2IJ3-O8/edit?slide=id.g397559857dc_1_22#slide=id.g397559857dc_1_22), [more recent results](https://docs.google.com/presentation/d/1HClqaXEiKd7l8b4Eold6meJai4I7lCg7xX24D7mDd2A/edit?slide=id.g397a5bf5ead_2_7#slide=id.g397a5bf5ead_2_7)) show improvements to the base samudra model when dropping these connections. Drop out is even more effective than [pruning the shallow connections](https://wandb.ai/ocean_emulators/default/runs/9jafz191?nw=nwuseralxmrs) or [pruning  the shallow connections with axial attenion at the bottleneck](https://wandb.ai/ocean_emulators/default/runs/6apycyv8?nw=nwuseramoghgulati).

We believe this method works similar to the intuition for why wide residual connections improve ResNets: the residual identity path in ResNets provides a liability for deeper networks where more depth means there's a greater of information flowing through the identity connections, not added parameters. Similarly, we believe that penalizing/regularizing skip connections during the encoders of the U-Net will incentivize the network from making more use of the encoder/decoder pathways. 

As mentioned below in this PR's conversation, this is different from the literature around "stochastic depth", which uses drop out in the opposite connection path -- here, drop path is applied to the non-skip connection channels.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -590,6 +590,10 @@ class UNetBackboneConfig(BaseConfig):
     core_block: BlockConfig = BlockConfig()
     down_sampling_block: DownSamplingBlocks = "avg_pool"
     up_sampling_block: UpSamplingBlocks = "zonally_periodic_upsample"
+    drop_path_rate: float = Field(
+        default=0.0,
+        description="Stochastic depth dropout rate. The chance we turn off residual connections in the UNet. Reasonable values are 0.1-0.3. Use 0.0 to disable.",
+    )
 
     def build(
         self,
@@ -634,6 +638,7 @@ class UNetBackboneConfig(BaseConfig):
             downsampling_block=downsampling_block,
             create_upsampling_block=create_upsampling_block,
             checkpointing=checkpointing,
+            drop_path_rate=self.drop_path_rate,
         )
 
 
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -592,7 +592,7 @@ class UNetBackboneConfig(BaseConfig):
     up_sampling_block: UpSamplingBlocks = "zonally_periodic_upsample"
     drop_path_rate: float = Field(
         default=0.0,
-        description="Stochastic depth dropout rate. The chance we turn off residual connections in the UNet. Reasonable values are 0.1-0.3. Use 0.0 to disable.",
+        description="Shortcut dropout rate. The chance we turn off skip connections in the UNet. Reasonable values are 0.1-0.3. Use 0.0 to disable.",
     )
 
     def build(
--- a/tests/test_blocks.py
+++ b/tests/test_blocks.py
@@ -4,7 +4,11 @@ import torch.nn as nn
 from hypothesis import given, settings
 from hypothesis import strategies as st
 
-from ocean_emulators.models.modules.blocks import ConvNeXtBlock, PointwiseLinear
+from ocean_emulators.models.modules.blocks import (
+    ConvNeXtBlock,
+    DropPath,
+    PointwiseLinear,
+)
 
 
 @given(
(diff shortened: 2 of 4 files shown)


alxmrs:
@codex Please review this PR

alxmrs:
Hey @codex, please review this.

tests/test_blocks.py
```diff
@@ -48,3 +52,43 @@ def test_convnext_block_backward_pass(pointwise_linear: bool):
...
+        drop_path = DropPath(drop_prob=1.0)
+        drop_path.train()
+
+        trunk = torch.randn(4, 16, 8, 8)
+        skip = torch.randn(4, 16, 8, 8)
+
+        output = trunk + drop_path(skip)
+        torch.testing.assert_close(output, trunk)
+
+    def test_gradient_flows_through_trunk_when_skip_dropped(self):
```
alxmrs, line 77: I'm on the fence if we should keep this test -- the guts of it might be covered in the `nn.Dropout` implementation. I've decided to keep it because it is covering something orthogonal to the other two tests.

alxmrs:
In light of #661 (discussed [here in slack](https://openathena.slack.com/archives/C08CYM42DT3/p1775064816082459)), I think we should pursue merging this PR. @jder, may I have your review?

src/ocean_emulators/config.py
```diff
@@ -550,6 +550,10 @@ class UNetBackboneConfig(BaseConfig):
     core_block: BlockConfig = BlockConfig()
     down_sampling_block: DownSamplingBlocks = "avg_pool"
     up_sampling_block: UpSamplingBlocks = "zonally_periodic_upsample"
+    drop_path_rate: float = Field(
+        default=0.0,
+        description="Stochastic depth dropout rate. The chance we turn off residual connections in the UNet. Reasonable values are 0.1-0.3. Use 0.0 to disable.",
```
jder: We should also remove the term "Stochastic depth" here since AFAICT this is different.

jder (changes requested):
So, as far as I understand, this is not what the papers you're linking to describe.

"Deep Networks with Stochastic Depth" describes dropping layers of a deep network and forcing the network to *use a skip connection* instead to reduce effective depth during training. We're not not changing the depth of the network when we apply this to the skip path (it's still just as deep, or arguably deeper) so doesn't seem like the same strategy.

The "Dropout Reduces Underfitting" describes how this should be done early and [the implementation](https://github.com/facebookresearch/dropout/blob/30cc113983fd99d62d8e2b0275ef066857681210/models/convnext.py#L51) also turns off the main block and forces the model to use the skip connection instead. 

If you want the current behavior, I think it would be more clear to rewrite the comment to say it drops the entire path and could be used for stochastic depth or other dropout where you want to drop the entire sample. And then I suggest removing the word "residual" from the comment/variable here since it's not really residual specific.

alxmrs:

jder (approved):
Thanks. I think the description is out of date; can you update it with the new wording you've chosen here and remove the part about how "there was little to no evidence of training in the bottleneck layer over time" since as I've pointed out the original experiment did not provide good evidence here either way and AFAIK we have still not done any experiment showing this. Can you instead link to the results showing this improving performance or otherwise motivating this change?

amogh-gulati:
> Thanks. I think the description is out of date; can you update it with the new wording you've chosen here and remove the part about how "there was little to no evidence of training in the bottleneck layer over time" since as I've pointed out the original experiment did not provide good evidence here either way and AFAIK we have still not done any experiment showing this. Can you instead link to the results showing this improving performance or otherwise motivating this change?

Hey @jder , here is a v2 run for the same : https://wandb.ai/ocean_emulators/default/runs/z6k7ac4m?nw=nwuseramoghgulati 

Looking over rmse over fields, it lands somewhere in between baseline and axial attention for Vo and Uos, and generally slightly better or slightly worse in Thetao, in OHC is remains marginally worse. LMK what you think!

PR #642, opened 2026-03-10
Stack 2/2: Adding early dropout to drop path feature using linear decay.

alxmrs:
This paper[1] and corresponding repo[2] suggest that dropout is even more effective when it is only used early during training. The paper reports that early dropout in ConvNext UNets works better than both using (s.d.) dropout or not using dropout at all:

Screenshot 2026-03-10 at 3 48 51 PM

Thus, this PR adds an enhancement to #641 that lets users define a number of epochs to linearly decay the dropout rate until it hits zero. This provides a really simple means of specifying early dropout in our experiments. According to [2], this provides just as much benefit as more complex schemes of implementing early drop out while minimizing the number of things we need to configure.

[1]: https://ar5iv.labs.arxiv.org/html/2303.01500
[2]: https://github.com/facebookresearch/dropout

### DIFF
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -666,7 +666,7 @@ class UNetBackboneConfig(BaseConfig):
     )
     drop_path_epochs: int = Field(
         default=0,
-        description="Number of epochs over which to linearly decay the drop path rate to 0 (early stochastic depth). "
+        description="Number of epochs over which to linearly decay the drop path rate to 0 (early shortcut dropout). "
         "0 means drop path is applied at a constant rate for the entire training run.",
     )
 
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -664,6 +664,11 @@ class UNetBackboneConfig(BaseConfig):
         default=0.0,
         description="Shortcut dropout rate. The chance we turn off skip connections in the UNet. Reasonable values are 0.1-0.3. Use 0.0 to disable.",
     )
+    drop_path_epochs: int = Field(
+        default=0,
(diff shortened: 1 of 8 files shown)


src/ocean_emulators/models/base.py
```diff
@@ -34,6 +34,12 @@ def __init__(
         self.hist = hist
         self.gradient_detach_interval = gradient_detach_interval
 
+    def set_epoch(self, epoch: int) -> None:
+        """Per-epoch hook for scheduling (e.g. early stochastic depth).
```
alxmrs:
```suggestion
        """Per-epoch hook for scheduling.
```

alxmrs:
@codex May I have a review?

alxmrs:
@codex may I have another review from you?

alxmrs:

src/ocean_emulators/config.py
```diff
@@ -594,6 +594,11 @@ class UNetBackboneConfig(BaseConfig):
         default=0.0,
         description="Shortcut dropout rate. The chance we turn off skip connections in the UNet. Reasonable values are 0.1-0.3. Use 0.0 to disable.",
     )
+    drop_path_epochs: int = Field(
+        default=0,
+        description="Number of epochs over which to linearly decay the drop path rate to 0 (early stochastic depth). "
```
```suggestion
        description="Number of epochs over which to linearly decay the drop path rate to 0 (early shortcut dropout). "
```

jder:
@fomo-bot can you rebase this on the latest main given the changes that happened on the (previous) base PR please?

alxmrs:
@fomo-bot will you rebase this against main once more?

PR #647, opened 2026-03-13
Adding a docker caching layer to our HPC Container CI

alxmrs:
We pull and cache the base Physics Nemo image before pulling the published image. My hope is that this speeds up the time to iterate on HPC runs from branches (this is the slowest part of the build).

After pulling the base image, we convert it to a gzip and store it inside of the GitHub cache artifact. My hope is that this leads to omitting downloading redundant parts of the newly published docker image.

### DIFF
--- a/.github/workflows/container-physicsnemo.yml
+++ b/.github/workflows/container-physicsnemo.yml
@@ -1,6 +1,10 @@
 # Build (and for main/manual runs, publish and test) the PhysicsNeMo-based container on x86 Linux.
 name: Container PhysicsNeMo 25.11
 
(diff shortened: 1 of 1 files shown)


alxmrs:
@codex Please review this change.

alxmrs:

jder:
@alxmrs do we have an example run where we can see the improved CI time?

alxmrs:
It just occurred to me that I can use the same `gh` command to test this out as I use to make feature containers! Will verify the caching and report back to you.

jder:
Seems like this doesn't help :(

PR #652, opened 2026-03-16
eval.py script supports single scale evaluations of mulit-scale FOMO models.

### LINKED ISSUE #651: The Eval script works for multi-scale FOMO models
This is a pre-requisite for #615.

alxmrs:
Added multiscale eval config + fixed config.py

Fixes #651.

### DIFF
--- a/configs/fomo_om4/eval_multiscale.yaml
+++ b/configs/fomo_om4/eval_multiscale.yaml
@@ -1,0 +1,46 @@
+# yaml-language-server: $schema=../schemas/EvalConfig.json
+#
+# Eval config for the multi-scale FOMO model.
+#
+# The eval code runs inference on the *first* data source only.
+# To evaluate all three scales, run three separate eval jobs and
+# reorder the sources list so the target resolution is first.
+# All three sources must always be present so the model builds
+# with the correct max grid size for the encoder patch grid.
+# Setting multiple data sources lets us use checkpoints for
+# FOMO models that use multiple scales.
+
+debug: false
+save_zarr: true
+disk_mode: true
+inference_time:
+  start: "2014-10-05"
+  end: "2015-10-05"
+num_model_steps_forward: 25
+
+experiment:
+  name: fomo_om4_eval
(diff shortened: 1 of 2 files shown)


alxmrs:
@codex Please review.

configs/fomo_om4/eval_multiscale.yaml
```diff
@@ -0,0 +1,46 @@
+# yaml-language-server: $schema=../schemas/EvalConfig.json
+#
+# Eval config for the multi-scale FOMO model.
+#
+# The eval code runs inference on the *first* data source only.
+# To evaluate all three scales, run three separate eval jobs and
+# reorder the sources list so the target resolution is first.
+# All three sources must always be present so the model builds
+# with the correct max grid size for the encoder patch grid.
```
jder, line 9: I don't follow why this would be necessary. What happens if you don't do this? Does the checkpoint fail to load? What parameters do we have that are dependent on the grid size?

jder (approved):
Seems like a fine first step to me, thanks!

configs/fomo_om4/eval_multiscale.yaml
alxmrs: Re-reading the code right now, I don't think the model would immediately fail. In fact, I think that specifically, if you at least kept the biggest grid here, it would work the way it is written. If an error occurred, it would happen in the encoder, specifically in the definition of the Perceiver. The grid size determines the latent rotary embed dimension. It's possible we wouldn't get a runtime error, but we would get a weird model performance error when running inference.

jder: Hrm, ok. Definitely seems like a smell to me that this isn't resolution-independent somehow but definitely doesn't need to get fixed in this PR.

alxmrs: Now that you mention it, I wonder if we could set this value based on the patch extent rather than the grid size (🤔 ).

PR #656, opened 2026-03-19
Updating our BatchSamplers to use the same scale across all replicas.

alxmrs:
This updates both our equivalence batch samplers to make it that all n_workers / num_replicas have batches with the same scale of data. This is needed because in DDP training across n_workers, when we mix scales, we're liable to NCCL timeouts. These timeouts are stochastic; they're caused by the scenario where we load 1/4 scale on at least one worker, but 1 or 1/2 scale on many of the other workers. When batch norm is turned on, and thus we need to synchronize between all workers, we have a high likelihood of facing GPU level timeouts, since data loading can take so much longer for larger datasets. 

This also fixes a small, but significant, bug where we didn't actually set epochs properly!

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -434,9 +434,9 @@ class Trainer:
                 cur_step = self.get_current_step(epoch)
                 self.init_data_loaders(cur_step)
 
-            if isinstance(self.train_sampler, DistributedSampler):
+            if hasattr(self.train_sampler, "set_epoch"):
                 self.train_sampler.set_epoch(epoch)
-            if isinstance(self.val_sampler, DistributedSampler):
+            if hasattr(self.val_sampler, "set_epoch"):
                 self.val_sampler.set_epoch(epoch)
 
             start_epoch_train_time = time.perf_counter()
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -888,6 +888,7 @@ class Trainer:
                 datasets=train_datasets,
                 group_key=group_key,
                 batch_size=self.batch_size,
+                num_replicas=1,
                 shuffle=True,
                 drop_last=True,
             )
@@ -896,6 +897,7 @@ class Trainer:
                 datasets=val_datasets,
                 group_key=group_key,
                 batch_size=self.batch_size,
+                num_replicas=1,
                 shuffle=True,
                 drop_last=False,
             )
--- a/src/ocean_emulators/utils/samplers.py
+++ b/src/ocean_emulators/utils/samplers.py
@@ -112,7 +112,7 @@ class EquivalenceGroupBatchSampler(Sampler[list[int]]):
             group_key: Callable that extracts grouping key from a dataset.
             batch_size: Number of samples per batch
             num_replicas: Coordinate batches across workers to prevent loading asymmetry.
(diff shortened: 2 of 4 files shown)


alxmrs:
@codex may I have a review?

alxmrs:

alxmrs:
Thanks @codex. May I have another review?

src/ocean_emulators/train.py
```diff
@@ -415,9 +415,9 @@ def run(self) -> None:
                 cur_step = self.get_current_step(epoch)
                 self.init_data_loaders(cur_step)
 
-            if isinstance(self.train_sampler, DistributedSampler):
+            if hasattr(self.train_sampler, "set_epoch"):
                 self.train_sampler.set_epoch(epoch)
-            if isinstance(self.val_sampler, DistributedSampler):
+            if hasattr(self.val_sampler, "set_epoch"):
                 self.val_sampler.set_epoch(epoch)
```
alxmrs, line 440: Bugfix: we weren't setting the epochs in the distributed sampler!

alxmrs:
Thanks, @codex. May I have another review?

alxmrs:
Good shout, @codex. may I have another review?

alxmrs:
Thanks again, @codex. May I have one last review?

src/ocean_emulators/utils/samplers.py
```diff
@@ -234,29 +263,52 @@ def set_epoch(self, epoch: int) -> None:
...
+        # data in the same training step, which causes load-time asymmetry and
+        # Gloo/NCCL timeouts.
+        chunks: list[tuple] = []
+        for sampler in self._inner._samplers:
+            group_chunks = list(itertools.batched(sampler, self.num_replicas))
+            if group_chunks and len(group_chunks[-1]) < self.num_replicas:
+                # Always pad incomplete chunks so every chunk is homogeneous
+                # and full-sized. Without padding, an incomplete chunk would
+                # break the stride alignment for all subsequent chunks.
+                # Dropping here would erase entire groups that have fewer
```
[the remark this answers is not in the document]
alxmrs: This is a contradiction with the previous comment. Keeping things as they are for now.

src/ocean_emulators/utils/samplers.py
```diff
@@ -142,27 +150,47 @@ def from_datasets(
...
-        batch_sampler = itertools.chain(*self._samplers)
+        # Batch within each group and align to num_replicas boundaries so that
+        # consecutive num_replicas batches always belong to the same group.
+        # This prevents different DDP ranks from receiving different-resolution
+        # data in the same training step, which causes load-time asymmetry and
+        # Gloo/NCCL timeouts.
+        per_replica = []
+        move_to_end = []
+        for sampler in self._samplers:
+            pre_replica_per_sample = list(itertools.batched(sampler, self.num_replicas))
```
jder, line 164: Is this supposed to be `per_replica_per_sample` as opposed to `pre...`?

jder (commented):
I was finding it hard to understand the logic here and ended up feeling like it was strange to have num_replicas at all in the non-distributed sampler. I ended up with an alternative and I think simpler option in #717 -- WDYT?

jder:
Closing in favor of #717

PR #657, opened 2026-03-19
NCCL timeouts: Stick some Gloo on it

alxmrs:
We use Pytorch's build in library for distributed communication primitives ([Gloo](https://github.com/pytorch/gloo)) to synchronize example data loading across DDP replicas on the CPU side. If and when the networked filesystem on SLURM gets throttled and has a slow down, this change means we'll wait only using CPU resources. This prevents the scenario where the slowdown happens during either the forward or backward pass, causing NCCL timeouts which ultimately crash the training job.

### DIFF
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -693,9 +693,6 @@ class Trainer:
                 if self.debug and (data_iter_step + 1) % 5 == 0:
                     break
 
-                if self._cpu_group is not None:
-                    dist.barrier(group=self._cpu_group)
-
                 VO: ValBatchOutput = Stepper.validate_batch(
                     self.model, data, self.loss_fn
                 )
--- a/src/ocean_emulators/train.py
+++ b/src/ocean_emulators/train.py
@@ -16,6 +16,7 @@ from typing import Any, assert_never
 
 import dask
 import torch
+import torch.distributed as dist
 import torch.nn as nn
 from torch.utils.data import (
     ConcatDataset,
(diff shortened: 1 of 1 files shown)


alxmrs:
@codex may I have a review?

alxmrs:

alxmrs:

alxmrs:

alxmrs:
Thanks, @codex. May I have another review?

src/ocean_emulators/train.py
```diff
@@ -284,12 +285,21 @@ def __init__(self, cfg: TrainConfig) -> None:
...
                 device_ids=[self.distributed.gpu],
             )
+            # Create a CPU-side (Gloo) process group for data-loading barriers.
+            # This lets all ranks synchronize after loading a batch without
+            # consuming the NCCL timeout budget, preventing NCCL timeouts when
+            # the networked filesystem stalls on some ranks.
+            self._cpu_group = dist.new_group(
+                backend="gloo",
+                timeout=datetime.timedelta(minutes=30),
+            )
```
jder, line 305: So if we are setting a 30-minute timeout here, could we instead pass that to `torch.distributed.init_process_group` rather than the (default) 10-minute timeout we have now? I think that would be the ~same behavior without needing to introduce our own synchronization and the separate data loader timeout?

jder (changes requested):

src/ocean_emulators/train.py
alxmrs: Disclaimer: I'm new to Gloo, torch.distributed, and NCCL. That said, I think there is some nuance here.  The behavior would be slightly different in that the timeouts would happen during the backward pass during GPU sync. I like that this explicitly synchronizes after data load time and before data is put on to the GPU. The remedy directly responds to spikes in data load times. By keeping the NCCL timeout tighter than this, we can still expose legitimate NCCL/GPU sync issues separately. 

Though, I do think you're right that your alternative would also work and it would be less code to maintain.

jder: I agree with your analysis. I would lean towards either just closing this (leaving a 10 minute timeout everywhere, which really should be plenty) or increasing that timeout if you like. 

If you'd prefer to merge this to separate data loading timeouts from other kinds, I'm happy to review in detail, just LMK.

alxmrs: Very fair point! I'm about to start a fresh training run, after identifying the root cause, this PR is probably not needed. I'll leave this open for now until I can confirm that the other fixes address the main concerns.

jder:
Hoping we don't need this, will reopen if that's not true.

PR #659, opened 2026-03-30
Immediately move TensorMap to GPU to reduce high AllReduce times during multiscale training.

alxmrs:
This doesn't totally fix the underlying issue with deploying our multi-scale model on torch, but it does significantly improve performance. It does so by reducing the CPU/GPU transfer overhead in one of the "chattiest" parts of the training loop.

### DIFF
--- a/src/ocean_emulators/constants.py
+++ b/src/ocean_emulators/constants.py
@@ -1,5 +1,6 @@
 import enum
 import logging
+from typing import Self
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


alxmrs:

jder (approved):
Thanks! Is it worth moving other things as well? (ie not just the index tensors)?

alxmrs:
>  Is it worth moving other things as well? (ie not just the index tensors)?

That's an excellent question! I will file a follow up issue to investigate. My initial guess is that most other things should be kept on the CPU.

PR #660, opened 2026-03-30
Each worker should have its own ThreadPoolExecutor. 

alxmrs:
This PR creates the ThreadPoolExecutors late when enabled. We add a private, static `_get_executor()` method to the DataLoader class. This means that there will be one thread pool created per worker (we create multiple dataloaders per zarr dataset; this only makes 1 pool per scale of data).

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -453,7 +453,7 @@ class TorchTrainDataset(Dataset[RawTrainData]):
     def _get_executor(cls) -> ThreadPoolExecutor:
         if cls._shared_executor is None:
             cls._shared_executor = ThreadPoolExecutor(
-                thread_name_prefix="concurrent_compute"
+                max_workers=None, thread_name_prefix="concurrent_compute"
             )
         return cls._shared_executor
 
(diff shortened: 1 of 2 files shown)


src/ocean_emulators/datasets.py
```diff
@@ -664,7 +677,7 @@ def load_variable_data(var: xr.Variable) -> None:
 
     futures = []
     for ds in datasets:
-        for var in ds.variables.values():
+        for var in ds.data_vars.variables.values():
```
alxmrs, line 683: Drive by fix: This only gets data variables instead of data variables and coordinates. It's slightly more efficient (we try to do less work).

alxmrs:

src/ocean_emulators/datasets.py
```diff
@@ -454,7 +467,7 @@ def __init__(
         normalize_before_mask: bool,
         masked_fill_value: float,
         stride: int = 1,
-        executor: ThreadPoolExecutor | None = None,
+        concurrent_compute_: bool = False,
```
jder, line 472: Why the trailing `_`?

jder (approved):

src/ocean_emulators/datasets.py
alxmrs: I got a warning in my IDE that this variable shadows a variable in outer scope with the same name.

PR #665, opened 2026-04-06
Attention positional embeddings

amogh-gulati:
This PR adds optional sinusoidal positional embeddings to the attention blocks on top of the axial_attention branch. It extends AttentionBlockConfig with a positional_embedding option and adds helper functions for building 1D and 2D sinusoidal embeddings, which are applied before QKV projection. It also adds per-head q/k layer norm in full attention, which helps stabilize training by making the attention weights more evenly distributed across tokens.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -269,24 +269,42 @@ class AttentionBlockConfig(BaseConfig):
         default=0.0,
         description="Dropout rate applied to the output projection.",
     )
+    positional_embedding: Literal["sinusoidal_1d", "sinusoidal_2d"] | None = Field(
+        default=None,
+        description=(
+            "Optional positional encoding added inside the attention block before "
+            "QKV projection. Use `sinusoidal_1d` with axial attention and "
+            "`sinusoidal_2d` with full attention."
+        ),
+    )
 
     def build(self, channels: int) -> nn.Module:
         assert channels % self.num_heads == 0, (
             f"channels {channels} must be divisible by num_heads {self.num_heads}"
         )
         if self.attention_type == "axial":
+            if self.positional_embedding not in (None, "sinusoidal_1d"):
+                raise ValueError(
+                    "Axial attention only supports positional_embedding='sinusoidal_1d'."
+                )
             return AxialAttentionBlock(
                 channels=channels,
                 num_heads=self.num_heads,
                 attn_drop=self.attn_drop,
                 proj_drop=self.proj_drop,
+                positional_embedding=self.positional_embedding,
             )
         if self.attention_type == "full":
+            if self.positional_embedding not in (None, "sinusoidal_2d"):
+                raise ValueError(
+                    "Full attention only supports positional_embedding='sinusoidal_2d'."
+                )
             return FullAttentionBlock(
                 channels=channels,
                 num_heads=self.num_heads,
                 attn_drop=self.attn_drop,
                 proj_drop=self.proj_drop,
+                positional_embedding=self.positional_embedding,
             )
         assert_never(self.attention_type)
 
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -1,7 +1,7 @@
 import abc
 from functools import cached_property
 from pathlib import Path
-from typing import Annotated, Literal, Self, assert_never
+from typing import Annotated, Literal, Self, assert_never, cast
 
 import cftime
 import pydantic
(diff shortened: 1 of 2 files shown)


src/ocean_emulators/models/modules/blocks.py
```diff
@@ -10,6 +11,50 @@
 from ocean_emulators.models.modules.activations import CappedGELU
 
 
+def sinusoidal_1d_position_embedding(
+    length: int,
+    dim: int,
+    *,
+    device: torch.device,
+) -> torch.Tensor:
```
alxmrs, line 19: 🐑 Would you mind adding jaxtyping types for the output tensor?

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -10,6 +11,50 @@
...
+    if dim <= 0:
+        return torch.empty(length, 0, device=device, dtype=torch.float32)
```
alxmrs, line 21: Do we want to use this dtype? Should it be an argument? What if we use a higher or lower fp resolution for the channels, will that cause any problems?

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -10,6 +11,50 @@
...
+def sinusoidal_2d_position_embedding(
+    height: int,
+    width: int,
+    dim: int,
+    *,
+    device: torch.device,
+    dtype: torch.dtype,
+) -> torch.Tensor:
+    row_dim = dim // 2
+    col_dim = dim - row_dim
```
alxmrs, line 44: I like this way of capturing the remainder in the col_dim.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -10,6 +11,50 @@
 from ocean_emulators.models.modules.activations import CappedGELU
 
 
+def sinusoidal_1d_position_embedding(
+    length: int,
+    dim: int,
+    *,
+    device: torch.device,
+) -> torch.Tensor:
+    if dim <= 0:
```
alxmrs, line 20: Does the `dim` also need to be even?

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -10,6 +11,50 @@
...
+
+    position = torch.arange(length, device=device, dtype=torch.float32).unsqueeze(1)
+    div_term = torch.exp(
+        torch.arange(0, dim, 2, device=device, dtype=torch.float32)
+        * (-(math.log(10000.0) / max(dim, 1)))
+    )
+
+    embedding = torch.zeros(length, dim, device=device, dtype=torch.float32)
+    embedding[:, 0::2] = torch.sin(position * div_term)
+    embedding[:, 1::2] = torch.cos(position * div_term[: embedding[:, 1::2].shape[1]])
```
alxmrs, line 31: Why do we index/filter the `div_term` here?

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -10,6 +11,50 @@
...
+    row_embedding = sinusoidal_1d_position_embedding(height, row_dim, device=device)
+    col_embedding = sinusoidal_1d_position_embedding(width, col_dim, device=device)
+
+    embedding = torch.cat(
+        [
+            row_embedding.unsqueeze(1).expand(-1, width, -1),
+            col_embedding.unsqueeze(0).expand(height, -1, -1),
+        ],
+        dim=-1,
+    )
```
alxmrs, line 54: I like that we can reuse the 1d encoding for the 2d! Do you have a reference I could check to know that this is correct?

src/ocean_emulators/config.py
```diff
@@ -269,24 +269,50 @@ class AttentionBlockConfig(BaseConfig):
         default=0.0,
         description="Dropout rate applied to the output projection.",
     )
+    positional_embedding: Literal["sinusoidal_1d", "sinusoidal_2d"] | None = Field(
```
alxmrs, line 272: 🐑 maybe we could include an option called "auto" that would turn on a positional embedding, but would choose a good default depending on the type of attention. I think having this defaulted to be set "off" could make mis-configuration more easy.

src/ocean_emulators/config.py
```diff
@@ -269,24 +269,50 @@ class AttentionBlockConfig(BaseConfig):
...
     def build(self, channels: int) -> nn.Module:
         assert channels % self.num_heads == 0, (
             f"channels {channels} must be divisible by num_heads {self.num_heads}"
         )
         if self.attention_type == "axial":
+            if self.positional_embedding not in (None, "sinusoidal_1d"):
+                raise ValueError(
+                    "Axial attention only supports positional_embedding='sinusoidal_1d'."
+                )
+            axial_positional_embedding = cast(
```
alxmrs, line 290: 🐑  IIRC, I think the type checker will better be able to infer the right type if we use an `assert ==` here, it might save you from casting.

alxmrs (approved):
A few minor nits and then some questions. This looks great! I think this is ready for merge after addressing a few small issues.

src/ocean_emulators/models/modules/blocks.py
```diff
@@ -368,6 +415,27 @@ def forward(
...
+            else:
+                positional_embedding = sinusoidal_1d_position_embedding(
+                    W,
+                    C,
+                    device=x.device,
+                )
+                positional_embedding = rearrange(
+                    positional_embedding, "w c -> 1 c 1 w"
+                ).to(dtype=x.dtype)
+            x = x + positional_embedding
```
jder, line 437: Just my 2 cents, but I would guess that using a 2d embedding here would work better. Otherwise there's no way for, say, the horizontal the axial attention to know if it's looking at the equator or at the pole and presumably you'd want very different behavior across those two cases.

PR #668, opened 2026-04-08
Split data loading config from worker settings

jder:
## Summary
This is the first PR in the LLC stack.

It separates `data.loading` from generic worker and DataLoader settings so CPU vs GPU loading can be configured explicitly without overloading one config object.

## Why
Later LLC and GPU-decode changes need an explicit loading mode. Pulling that distinction forward keeps the rest of the stack smaller and should preserve OM4 CPU behavior.

## Notes
This is intended as a draft stacked base PR for the rest of the LLC stack.

### DIFF
--- a/configs/test/train_default_2step.yaml
+++ b/configs/test/train_default_2step.yaml
@@ -47,7 +47,9 @@ data:
     - data_location: data.zarr
       data_means_location: means.nc
       data_stds_location: stds.nc
-  num_workers: 4
+  loading:
+    type: cpu
+    num_workers: 4
   hist: 0
 
 model:
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -146,15 +146,26 @@ class DataSourceConfig(BaseConfig):
     )
 
 
-class CpuDataLoadingConfig(BaseConfig):
+class BaseDataLoadingConfig(BaseConfig):
+    def num_pytorch_workers(self) -> int:
+        raise NotImplementedError
+
+
+class CpuDataLoadingConfig(BaseDataLoadingConfig):
     type: Literal["cpu"] = "cpu"
     num_workers: int = Field(default=4, ge=0)
 
+    def num_pytorch_workers(self) -> int:
+        return self.num_workers
 
-class GpuDataLoadingConfig(BaseConfig):
+
+class GpuDataLoadingConfig(BaseDataLoadingConfig):
     type: Literal["gpu"] = "gpu"
     kvikio_task_size: int = Field(default=64 * 1024 * 1024, gt=0)
     kvikio_num_threads: int = Field(default=8, gt=0)
+
+    def num_pytorch_workers(self) -> int:
+        return 0
 
 
 DataLoadingConfig = Annotated[
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -146,6 +146,23 @@ class DataSourceConfig(BaseConfig):
     )
 
 
+class CpuDataLoadingConfig(BaseConfig):
+    type: Literal["cpu"] = "cpu"
+    num_workers: int = Field(default=4, ge=0)
+
+
+class GpuDataLoadingConfig(BaseConfig):
+    type: Literal["gpu"] = "gpu"
+    kvikio_task_size: int = Field(default=64 * 1024 * 1024, gt=0)
+    kvikio_num_threads: int = Field(default=8, gt=0)
+
+
+DataLoadingConfig = Annotated[
+    CpuDataLoadingConfig | GpuDataLoadingConfig,
+    Field(discriminator="type"),
+]
+
+
 class DataConfig(BaseConfig):
     sources: list[DataSourceConfig] = Field(
         description=(
@@ -155,7 +172,7 @@ class DataConfig(BaseConfig):
         min_length=1,
     )
     static_data_vars: list[str] | None = None
-    num_workers: int = 4
+    loading: DataLoadingConfig = Field(default_factory=CpuDataLoadingConfig)
     hist: int = 1
     loader_version: str = str(LoaderVersion.OM4_TORCH.value)
     normalize_before_mask: bool = True
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -165,6 +165,8 @@ class GpuDataLoadingConfig(BaseDataLoadingConfig):
     kvikio_num_threads: int = Field(default=8, gt=0)
 
     def num_pytorch_workers(self) -> int:
+        # When loading data direct to GPU, we don't want worker processes.
+        # 0 means "load in the main process"
         return 0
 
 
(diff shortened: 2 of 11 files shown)


src/ocean_emulators/config_schema.py
```diff
@@ -36,17 +35,27 @@ def get_pydantic_models(
...
+    if isinstance(annotation, type) and issubclass(annotation, pydantic.BaseModel):
+        get_pydantic_models(annotation, seen)
+        return
+
+    origin = get_origin(annotation)
+    if origin is None:
+        return
+
+    for arg in get_args(annotation):
+        _collect_pydantic_models(arg, seen)
```
jder: I think this could now produce an infinite loop How about renaming `seen` to `models` and add a new seen thing which is just a set of all values we've collected across to avoid infinite regress?

jder:
@codex review

src/ocean_emulators/eval.py
```diff
@@ -43,10 +43,16 @@ def __init__(self, cfg: EvalConfig) -> None:
         self.device = init_eval_backend(cfg.backend)
 
         # Adjust workers and memory pinning based on device
+        cpu_loading = (
+            cfg.data.loading
+            if isinstance(cfg.data.loading, CpuDataLoadingConfig)
+            else None
+        )
+        data_num_workers = cpu_loading.num_workers if cpu_loading is not None else 0
```
jder: How about here and below doing cfg.data.loading.num_pytorch_workers() and implement for both types with a tiny common interface?

jder (commented):
@fomo-bot please resolve these comments

jder:
@fomo-bot please see above

src/ocean_emulators/config_schema.py
jder: Addressed in `415a59d2`.

`get_pydantic_models()` now keeps the collected `models` separate from a `seen` set of visited annotation identities, and `_collect_pydantic_models()` checks that `seen` set before descending. That keeps the recursive walk from re-entering the same annotation structure indefinitely while still collecting the model classes we need.

src/ocean_emulators/eval.py
jder: Addressed in `415a59d2`.

I added `num_pytorch_workers()` to both loading config variants and switched both `train.py` and `eval.py` to use that shared interface. CPU loading returns its configured worker count; GPU loading returns `0`, so the call sites no longer need the type check.

jder:
Addressed the review comments in `415a59d2`.

- added a shared `num_pytorch_workers()` interface on the loading config variants and updated `train.py`/`eval.py` to use it
- split collected models from visited annotations in `config_schema.py` so recursive annotation traversal can short-circuit safely

Validation: `uv run pytest tests/test_config.py -q`

configs/test/train_default_2step.yaml
```diff
@@ -67,4 +69,4 @@ model:
       norm: "batch"
 
     down_sampling_block: "avg_pool"
-    up_sampling_block: "zonally_periodic_upsample"
\ No newline at end of file
+    up_sampling_block: "zonally_periodic_upsample"
```
alxmrs, line 72: 🐑 🐑 🐑  seems to be extra whitespace here

src/ocean_emulators/config.py
```diff
@@ -146,6 +146,34 @@ class DataSourceConfig(BaseConfig):
...
+        return self.num_workers
+
+
+class GpuDataLoadingConfig(BaseDataLoadingConfig):
+    type: Literal["gpu"] = "gpu"
+    kvikio_task_size: int = Field(default=64 * 1024 * 1024, gt=0)
+    kvikio_num_threads: int = Field(default=8, gt=0)
+
+    def num_pytorch_workers(self) -> int:
+        return 0
```
alxmrs, line 170: 🐑 This confuses me a little bit (seems like I'm not [alone](https://discuss.pytorch.org/t/a-clear-explanation-of-what-num-workers-0-means-for-a-dataloader/177614)) -- can we leave a comment that says this sets the GPU data loading on to the main thread?

src/ocean_emulators/config.py
```diff
@@ -155,7 +183,7 @@ class DataConfig(BaseConfig):
         min_length=1,
     )
     static_data_vars: list[str] | None = None
-    num_workers: int = 4
+    loading: DataLoadingConfig = Field(default_factory=CpuDataLoadingConfig)
```
alxmrs, line 188: This is a slick means of configuration!

tests/test_config.py
```diff
@@ -0,0 +1,98 @@
+from pathlib import Path
```
alxmrs, line 1: Happy to have this test file.

src/ocean_emulators/config.py
```diff
@@ -146,6 +146,34 @@ class DataSourceConfig(BaseConfig):
...
+    num_workers: int = Field(default=4, ge=0)
+
+    def num_pytorch_workers(self) -> int:
+        return self.num_workers
+
+
+class GpuDataLoadingConfig(BaseDataLoadingConfig):
+    type: Literal["gpu"] = "gpu"
+    kvikio_task_size: int = Field(default=64 * 1024 * 1024, gt=0)
+    kvikio_num_threads: int = Field(default=8, gt=0)
```
alxmrs, line 165: I don't see how these are used here, but I assume they are used later on in the stack. Is that right?

src/ocean_emulators/config_schema.py
```diff
@@ -15,36 +14,50 @@
...
+        return
+    seen.add(annotation_id)
+
+    if isinstance(annotation, type) and issubclass(annotation, pydantic.BaseModel):
+        models[annotation.__name__] = annotation
+        for field in annotation.model_fields.values():
+            _collect_pydantic_models(field.annotation, models, seen)
+        return
+
+    origin = get_origin(annotation)
```
alxmrs, line 57: What does `get_origin` do? I get the sense that we still do the union type processing with the remainder of this method, but I'm not sure how this function works like the previous `isinstance` check.

src/ocean_emulators/config_schema.py
```diff
@@ -15,36 +14,50 @@
...
+    if origin is None:
+        return
+
+    for arg in get_args(annotation):
+        _collect_pydantic_models(arg, models, seen)
```
alxmrs, line 62: I like that we recur to do the type/basemodel check instead of repeating that logic in the loop.

alxmrs (approved):
Mostly questions, one ultra nit about whitespace and another nit about a helpful comment. LGTM

src/ocean_emulators/config.py
jder: Yes, that's right.

configs/test/train_default_2step.yaml
jder: Are you objecting to the newline at the end of a file? Many folks insist on it. I find opinions either way pretty baffling! Feel free to update pre-commit rules if you want formatting things enforced.

jder:
Thanks! I added some comments and will merge once CI passes.

configs/test/train_default_2step.yaml
alxmrs: They are baffling indeed! I have no opinion one way or the other, I just like a small diff. 

If this was created by a pre-commit check, then happy to keep it.

PR #669, opened 2026-04-08
Introduce dataset specs and explicit tensor state

### LINKED ISSUE #475: Make variable set configurable
Currently the set of variables used (eg uo, thetao, etc) is hard-coded in constants.py. We should (a) move this into configuration and (b) audit for places where we're assuming a particular name and make those configurable (eg we hard-code thetao in add_derived_variables). Unclear if we should also try and make the viz code generic in this way (I think probably not for the most part?).

Why move this into configuration at all? This will force us to start threading variable information around where needed, which is better than reaching out to a singleton or constants.py as we move towards simultaneous training from data sources with different sets of variables available.

jder:
## Summary
This is PR 2 in the LLC stack, based on #668.

It introduces a first-class dataset-family/spec layer and removes hidden global tensor state from the train, eval, and data paths. `TensorMap` and normalization are carried explicitly instead of being read from global mutable state.

## Why
LLC support needs dataset-specific semantics without pushing LLC behavior into OM4 globals. This also makes `TensorMap` and related tensor metadata explicit at the call sites that actually depend on them.

Fixes #475


jder:
@fomo-bot can you rebase this and the rest of the stack on the latest main please

src/ocean_emulators/aggregator/loss.py
```diff
@@ -3,46 +3,71 @@
 from ocean_emulators.constants import TensorMap
 
 
+def _index_on_tensor_device(index: torch.Tensor, tensor: torch.Tensor) -> torch.Tensor:
+    if index.device == tensor.device:
+        return index
+    return index.to(tensor.device, non_blocking=tensor.device.type == "cuda")
```
jder: I think we no longer need this since TensorMap indexes are moved to the device ahead of time?

src/ocean_emulators/utils/data.py
```diff
@@ -515,29 +521,47 @@ def conditional_rearrange(
 
 
 def extract_wet_mask(
-    data: xr.Dataset, prognostic_var_names: PrognosticVarNames
+    data: xr.Dataset,
+    prognostic_var_names: PrognosticVarNames,
+    *,
+    dataset_spec: DatasetSpec = OM4_DATASET_SPEC,
```
jder: I think these default values are asking for bugs downstream, let's remove them.

src/ocean_emulators/utils/data.py
```diff
@@ -548,35 +572,44 @@ def _parse_lev_from_output_var(prognostic_var_names: PrognosticVarNames) -> list
     return depth_inds
 
 
-def flatten_masks(data: xr.Dataset) -> xr.Dataset:
-    """Adds data_vars "mask_0"..."mask_18" with dimensions (y, x)."""
+def flatten_masks(
+    data: xr.Dataset,
+    dataset_spec: DatasetSpec = OM4_DATASET_SPEC,
+) -> xr.Dataset:
+    """Adds level-wise mask variables with dimensions (y, x)."""
```
jder: I don't see anything about y or x in this code, can we not claim things unenforced here + below?

src/ocean_emulators/utils/data.py
```diff
@@ -808,9 +849,33 @@ def validate_data(
...
+        if Multiton._current_scope.get(cls) is not None:
+            raise ValueError(f"{cls} already initialized")
+
+        instance = cls(
+            src,
+            prognostic_var_names=prognostic_var_names,
+            boundary_var_names=boundary_var_names,
+        )
+        Multiton._current_scope[cls] = instance
+        return instance
```
jder: Can we get rid of this now? (ie do we even need multitons for this class now)? If we still need them I don't think inlining this is a good idea. Same question for TensorMap below.

src/ocean_emulators/constants.py
```diff
@@ -46,171 +47,348 @@
...
+
+DEPTH_LEVELS = list(OM4_DATASET_SPEC.depth_levels)
+DEPTH_THICKNESS = list(OM4_DATASET_SPEC.depth_thickness)
+DEPTH_I_LEVELS = list(OM4_DATASET_SPEC.depth_i_levels)
+MASK_VARS = list(OM4_DATASET_SPEC.mask_vars)
+MASK_ALL_LEVELS_VAR = OM4_DATASET_SPEC.mask_all_levels_var
+SECONDS_PER_TIME_STEP = OM4_DATASET_SPEC.seconds_per_time_step
+PROGNOSTIC_VARS = OM4_DATASET_SPEC.prognostic_vars
+BOUNDARY_VARS = OM4_DATASET_SPEC.boundary_vars
+DEFAULT_METADATA = OM4_DATASET_SPEC.default_metadata
```
jder: These also seem like bugs waiting to happen

jder (commented):
@fomo-bot can you help resolve these comment please?

src/ocean_emulators/viz/core.py
```diff
@@ -119,7 +119,7 @@ def __init__(
             }
 
         key1 = runs[0].name
-        levels = len(DEPTH_LEVELS)
+        levels = len(OM4_DATASET_SPEC.depth_levels)
```
jder: Let's set the spec on self here and use that below, with a TODO to support other specs.

src/ocean_emulators/utils/data.py
```diff
@@ -843,8 +881,12 @@ def normalize_tensor_prognostic(
...
     ) -> torch.Tensor:
         """Normalize prognostic tensor."""
-        tensor_mean = self._to_tensor(self._prognostic_mean_np, data.device)
-        tensor_std = self._to_tensor(self._prognostic_std_np, data.device)
+        tensor_mean = self._to_tensor(self._prognostic_mean_np, data.device).to(
+            data.dtype
+        )
+        tensor_std = self._to_tensor(self._prognostic_std_np, data.device).to(
+            data.dtype
+        )
```
jder, line 878: I'm not sure why this got added but seems harmless.

src/ocean_emulators/config.py
```diff
@@ -176,7 +179,33 @@ def num_pytorch_workers(self) -> int:
...
+class LlcDatasetConfig(BaseConfig):
+    type: Literal["llc"] = "llc"
+    face: int = 1
+    i_start: int = 0
+    i_end: int = 720
+    j_start: int = 0
+    j_end: int = 720
+
+    def build_spec(self) -> DatasetSpec:
+        return get_dataset_spec(self.type)
```
jder: For future reference: this is stub until the next PR in the stack.

tests/conftest.py
```diff
@@ -570,28 +570,21 @@ def trainer_pair(
 @pytest.fixture
 def dummy_src():
     h, w = 4, 8
-    with MultitonScope():
```
jder, line 573: Are there no multitons left at all?

tests/test_positional_channels.py
```diff
@@ -20,12 +21,21 @@ def test_positional_parameters_update(dummy_src: DataSource):
         ),
         pos_channels=1,
     )
+    tensor_map = c.TensorMap("thetao_1", "hfds", dataset_spec=src.dataset_spec)
```
jder: Let's just import TensorMap rather than c

jder (commented):
@fomo-bot one question and two suggested changes (and a couple notes for future reviewers). Please resolve and then request review from @alxmrs

src/ocean_emulators/constants.py
```diff
@@ -46,171 +45,337 @@
...
 MAX_TRAIN_MODEL_STEPS_FORWARD = 200
 
+DatasetType = Literal["om4", "llc"]
+
+PrognosticVarNames = list[str]
+BoundaryVarNames = list[str]
+
+
+@dataclasses.dataclass(frozen=True)
+class DatasetSpec:
```
alxmrs, line 57: Really happy to have this.

src/ocean_emulators/constants.py
```diff
@@ -46,171 +45,337 @@
...
+    type: DatasetType
+    depth_levels: tuple[float, ...]
+    depth_thickness: tuple[float, ...]
+    mask_vars: tuple[str, ...]
+    mask_all_levels_var: str
+    seconds_per_time_step: int
+    prognostic_vars: dict[str, PrognosticVarNames]
+    boundary_vars: dict[str, BoundaryVarNames]
```
alxmrs: 🐑 Maybe we could name this property so it's more obvious that these are options/configs of var names rather than a particular set of variables. When I was new to this codebase, I found this confusing.

alxmrs: A docstring could help clarify this, too.

src/ocean_emulators/constants.py
```diff
@@ -46,171 +45,337 @@
...
+        },
     },
-    "tauuo": {
-        "long_name": "Surface Downward X Stress",
-        "units": "N/m^2",
+    ocean_heat_temperature_var="thetao",
+    surface_heat_flux_var="hfds",
+)
+
+LLC_DATASET_SPEC = DatasetSpec(
```
alxmrs: Very sick to have in the codebase!

src/ocean_emulators/constants.py
```diff
@@ -237,23 +404,32 @@ class LoaderVersion(enum.Enum):
     OM4_TORCH = "om4-torch"
 
 
-# TODO(#95): See if this can be removed and replaced.
-class TensorMap(Multiton):
-    def _initialize(self, prognostic_vars_key: str, boundary_vars_key: str):
+class TensorMap:
```
alxmrs, line 459: is TensorMap and DatasetSpec the same concern?

alxmrs: Like, could it just be one big DataClass, maybe with custom properties?

configs/data/om4.yaml
```diff
@@ -1,5 +1,8 @@
 # yaml-language-server: $schema=../schemas/DataConfig.json
 
+dataset:
+  type: om4
```
alxmrs, line 4: Nice! A good specification to have.

src/ocean_emulators/utils/data.py
```diff
@@ -83,6 +80,7 @@ class DataSource:
     means: xr.Dataset
     stds: xr.Dataset
     masks: Masks
+    dataset_spec: DatasetSpec
```
alxmrs, line 83: I like that we have this metadata holder in the data source. I think a while ago I created a TODO or issue that asked if the prog/boundary var names should live on the data source. I think this is good coupling.

jder:
@fomo-bot can you please resolve the CI issues

src/ocean_emulators/utils/data.py
```diff
@@ -515,29 +519,47 @@ def conditional_rearrange(
 
 
 def extract_wet_mask(
-    data: xr.Dataset, prognostic_var_names: PrognosticVarNames
+    data: xr.Dataset,
+    prognostic_var_names: PrognosticVarNames,
```
alxmrs, line 523: We now can get the prognostic var names from the dataset spec IIUC.

src/ocean_emulators/utils/data.py
```diff
@@ -770,6 +808,7 @@ def validate_data(
     data: xr.Dataset,
     means: xr.Dataset,
     stds: xr.Dataset,
+    dataset_spec: DatasetSpec,
     boundary_var_names: BoundaryVarNames,
     static_data_vars: list[str] | None = None,
```
alxmrs, line 806: I know we can probably elimiate the boundary_var_names kwarg, but maybe also the static_data_vars given the dataset_spec?

src/ocean_emulators/utils/data.py
```diff
@@ -843,8 +881,12 @@ def normalize_tensor_prognostic(
         self, data: torch.Tensor, fill_nan=True, fill_value=0.0
     ) -> torch.Tensor:
         """Normalize prognostic tensor."""
-        tensor_mean = self._to_tensor(self._prognostic_mean_np, data.device)
-        tensor_std = self._to_tensor(self._prognostic_std_np, data.device)
+        tensor_mean = self._to_tensor(self._prognostic_mean_np, data.device).to(
+            data.dtype
```
alxmrs: Good catch (preserving the type).

alxmrs: 🐑 Should `_to_tensor` now accept a data type, since it currently collects `data.device`? Or: since we're making the `to()` call anyway, maybe we could pull in the move to device to that call instead of doing it within `_to_tensor`?

src/ocean_emulators/config.py
```diff
@@ -186,7 +189,33 @@ def persistent_pytorch_workers(self) -> bool:
...
+
+
+DatasetConfig = Annotated[
+    Om4DatasetConfig | LlcDatasetConfig,
+    Field(discriminator="type"),
+]
+
+
 class DataConfig(BaseConfig):
+    dataset: DatasetConfig = Field(default_factory=Om4DatasetConfig)
```
alxmrs, line 227: Nice, happy to see this in the config lang.

src/ocean_emulators/config.py
```diff
@@ -693,6 +745,9 @@ def build(
         hist: int,
         static_data_for_corrector: xr.Dataset | None,
         srcs: list[DataSource],
+        tensor_map: TensorMap,
+        normalize: Normalize,
+        dataset_spec: DatasetSpec,
```
alxmrs, line 764: I can't wait to fuse these all into one context obj.

src/ocean_emulators/constants.py
```diff
@@ -46,171 +45,337 @@
...
-        "long_name": "hfds anomalies",
-        "units": "W/m^2",
-    },
-}
+    ocean_heat_temperature_var="Theta",
+    surface_heat_flux_var="oceQnet",
+)
+
+
+def get_dataset_spec(dataset_type: DatasetType) -> DatasetSpec:
```
alxmrs: I'm so old fashioned, I would just use a global dictionary for this. I do like how type safe this is with the match expression.

src/ocean_emulators/constants.py
alxmrs: Discussed in chat, we'll implement this later.

src/ocean_emulators/eval.py
```diff
@@ -97,12 +97,12 @@ def __init__(self, cfg: EvalConfig) -> None:
         self.src = self.data_container.inference_source
         self.data = self.src.data
         self.static_data = self.data_container.static_data
-        self.metadata = construct_metadata(self.data)
+        self.metadata = self.src.metadata
```
alxmrs, line 95: Slick

src/ocean_emulators/eval.py
```diff
@@ -97,12 +97,12 @@ def __init__(self, cfg: EvalConfig) -> None:
...
         self.wet = self.src.masks.prognostic_with_hist(cfg.data.hist)
         self.area_weights: Grid = spherical_area_weights(self.data)
         self.area_weights = self.area_weights.to(self.device)
 
-        self.normalize = Normalize.init_instance(
+        self.normalize = Normalize(
```
alxmrs, line 100: A regular ol' constructor for this feels really good.

src/ocean_emulators/train.py
```diff
@@ -116,16 +114,17 @@ def __init__(self, cfg: TrainConfig) -> None:
         set_seed(cfg.experiment.rand_seed)
 
         # Getting prognostic and boundary variables
-        self.prognostic_var_names: PrognosticVarNames = PROGNOSTIC_VARS[
-            cfg.experiment.prognostic_vars_key
-        ]
-        self.boundary_var_names: BoundaryVarNames = BOUNDARY_VARS[
-            cfg.experiment.boundary_vars_key
-        ]
+        self.dataset_spec = cfg.data.dataset.build_spec()
```
alxmrs: 🐑 Somehow, this feels like a break in convention. I would expect the interface to be: 

```suggestion
        self.dataset_spec = cfg.data.dataset_spec.build()
```

Both ways are quite an improvement and I am happy with what you have.

src/ocean_emulators/utils/data.py
alxmrs: I see, I think the current dataset spec does something different now that I've scrolled down again. I allows one to look up any option of prog vars, not a particular set of vars. 

Happy to keep this as is, but I also see the argument that we could store the key on the dataset_spec and thus look this up.

tests/conftest.py
alxmrs: Still quite an improvement!

alxmrs (approved):
A big PR! Most of the notes of feedback are more design questions for discussion. Thanks so much for this change, it seems like we're so so close to getting rid of multitons, which is quite a refactoring milestone.

src/ocean_emulators/utils/data.py
alxmrs: Update: this is not the case after a review (this comment escaped me).

src/ocean_emulators/utils/data.py
jder: Yeah, after looking at this a bit more I think I agree. @fomo-bot I think we should try making the DatasetSpec not be the constant (ie unconfigured) dataset specification but the specification of the particular version of that dataset to use. So for om4, that would mean moving the prognostic_vars_key and boundary_vars_key into the `Om4DatasetConfig` and having its build_spec call eg build_om4_spec(self.prognostic_vars_key, ...) which would have approximately the content currently in constants.py for the om4 spec but the output would be specialized to the chosen var names. ie DatasetSpec would just store prognostic_var_names instead of the prognostic_vars dict. (For LLC this will be more important because then we can also pass along the spatial subset etc in a future PR)

We can build a global one of these in conftest.py so tests can just reach out to TEST_DATASET_SPEC or something (since there will no longer be a global OM4_DATASET_SPEC in constants.py). Does that make sense?

src/ocean_emulators/models/corrector.py
```diff
@@ -173,19 +174,25 @@ def __init__(
...
         tensor_map: TensorMap,
         normalize: Normalize,
+        dataset_spec: DatasetSpec,
         hfgeou_tensor: torch.Tensor,
         sea_surface_fraction_tensor: torch.Tensor,
     ):
         super().__init__(hist, tensor_map, normalize)
         # Area weights are not on the correct scale.
+        self.dataset_spec = dataset_spec
         self.area_weights = area_weights
```
jder:
```suggestion
        self.dataset_spec = dataset_spec
        # Area weights are not on the correct scale.
        self.area_weights = area_weights
```

src/ocean_emulators/utils/data.py
```diff
@@ -515,29 +519,47 @@ def conditional_rearrange(
...
+) -> list[int]:
     """Parse the `lev` dimension from the output var names. Default: 0 for surface."""
     depth_inds = []
     for var_depth_i in prognostic_var_names:
+        if "_lev_" in var_depth_i:
+            _, lev_depth = var_depth_i.split("_lev_", maxsplit=1)
+            depth_inds.append(
+                dataset_spec.depth_levels.index(float(lev_depth.replace("_", ".")))
+            )
+            continue
```
jder: Why did this show up?

src/ocean_emulators/utils/data.py
jder: Yeah, agreed, let's just put this all inline here ie `tensor_mean = torch.from_numpy(...).to(data.device, data.dtype)` or similar.

src/ocean_emulators/config.py
```diff
@@ -186,7 +188,48 @@ def persistent_pytorch_workers(self) -> bool:
 ]
 
 
+class Om4DatasetConfig(BaseConfig):
+    type: Literal["om4"] = "om4"
+    prognostic_vars_key: str = "thermo_dynamic_all"
+    boundary_vars_key: str = "tau_hfds"
+
+    def build_spec(self) -> DatasetSpec:
```
jder: Please name this `build` to match all other config build methods.

src/ocean_emulators/constants.py
```diff
@@ -46,171 +45,409 @@
...
+BoundaryVarNames = list[str]
+
+
+@dataclasses.dataclass(frozen=True)
+class DatasetSpatialSubset:
+    face: int
+    i_start: int
+    i_end: int
+    j_start: int
+    j_end: int
```
jder: This is LLC-specific and so shouldn't live on all dataset specs. For now let's remove this and reintroduce in a later PR where we can actually handle this in LLC-specific code.

jder (commented):
@fomo-bot a few more requests, but generally looking great!

src/ocean_emulators/constants.py
```diff
@@ -46,171 +45,387 @@
...
+    depth_levels: tuple[float, ...]
+    depth_thickness: tuple[float, ...]
+    mask_vars: tuple[str, ...]
+    mask_all_levels_var: str
+    seconds_per_time_step: int
+    prognostic_var_names: PrognosticVarNames
+    boundary_var_names: BoundaryVarNames
+    default_metadata: dict[str, dict[str, str]]
+    ocean_heat_temperature_var: str
+    surface_heat_flux_var: str
```
jder, line 68: These both really make me want to just uniformly rename the variables some how but I think this can be a future-us cleanup.

jder (commented):
ok, I think this looks much nicer now. @alxmrs can you take another (small) look?

jder:
@fomo-bot can you merge in the latest main and resolve conflicts, please?

@alxmrs this is waiting for your [re-] review when you have a sec, thanks!

configs/test/train_default.yaml
```diff
@@ -39,10 +39,12 @@ experiment:
...
     entity: m2lines
     group: samudra-train-om4
-  prognostic_vars_key: thermo_dynamic_5
-  boundary_vars_key: tau_hfds
   data_root: .data_cache/
 data:
+  dataset:
+    type: om4
+    prognostic_vars_key: thermo_dynamic_5
+    boundary_vars_key: tau_hfds
```
alxmrs, line 47: Nice, I like this coupling.

src/ocean_emulators/utils/data.py
alxmrs: Seems more clear to me, I like this refactor.

src/ocean_emulators/constants.py
alxmrs: Agreed, a good thing to do later building upon this change.

tests/conftest.py
```diff
@@ -18,21 +18,36 @@
...
 FOMO_CONFIG = "test/train_fomo.yaml"
 ALL_CONFIGS = [DEFAULT_CONFIG, "test/train_default_2step.yaml", FOMO_CONFIG]
+TEST_DATASET_SPEC = c.build_om4_spec(
+    prognostic_vars_key="thetao_1",
+    boundary_vars_key="hfds",
+)
+TEST_FULL_DATASET_SPEC = c.build_om4_spec(
+    prognostic_vars_key="thermo_dynamic_all",
+    boundary_vars_key="tau_hfds_hfds_anom",
+)
```
alxmrs, line 36: This is nice.

tests/test_validate_aggregator.py
```diff
@@ -52,6 +51,18 @@ def val_batch_of(
...
+def tensor_map_for(src: DataSource) -> TensorMap:
+    return TensorMap(dataset_spec=src.dataset_spec)
+
+
+def normalize_for(src: DataSource, tensor_map: TensorMap) -> Normalize:
+    return Normalize(
+        src,
+        tensor_map.prognostic_var_names,
+        tensor_map.boundary_var_names,
+    )
```
alxmrs, line 63: I like these.

src/ocean_emulators/utils/data.py
```diff
@@ -553,13 +553,6 @@ def _parse_lev_from_output_var(
     """Parse the `lev` dimension from the output var names. Default: 0 for surface."""
     depth_inds = []
     for var_depth_i in prognostic_var_names:
-        if "_lev_" in var_depth_i:
```
alxmrs: Was this dead code?

src/ocean_emulators/config.py
```diff
@@ -234,10 +243,8 @@ class DataConfig(BaseConfig):
     def build(
         self,
         data_root: ResolvedLocation,
-        prognostic_var_names: PrognosticVarNames,
-        boundary_var_names: BoundaryVarNames,
     ) -> DataContainer:
-        dataset_spec = self.dataset.build_spec()
+        dataset_spec = self.dataset.build()
```
alxmrs, line 247: Thanks for the name change.

src/ocean_emulators/constants.py
```diff
@@ -77,11 +79,14 @@ def __post_init__(self) -> None:
     def depth_i_levels(self) -> tuple[str, ...]:
         return tuple(str(i) for i in range(len(self.depth_levels)))
 
-    def prognostic_var_names(self, key: str) -> PrognosticVarNames:
-        return self.prognostic_vars[key]
-
-    def boundary_var_names(self, key: str) -> BoundaryVarNames:
-        return self.boundary_vars[key]
+    @property
+    def num_prognostic_depth_levels(self) -> int:
```
alxmrs, line 83: This is a good general utility.

src/ocean_emulators/constants.py
```diff
@@ -91,285 +96,332 @@ def boundary_var_names(self, key: str) -> BoundaryVarNames:
...
+        return list(vars_by_key[key])
+    except KeyError as exc:
+        choices = ", ".join(sorted(vars_by_key))
+        raise ValueError(
+            f"Unsupported {dataset_type} {var_kind} variable key {key!r}. "
+            f"Expected one of: {choices}"
+        ) from exc
+
+
+def build_om4_spec(
```
alxmrs, line 116: I agree it's probably better to defer creating this spec.

alxmrs: I now see better how this is wired together.

src/ocean_emulators/constants.py
```diff
@@ -418,17 +468,14 @@ def __init__(
...
 
         self.INPT_BOUNDARY_IDX: dict[str, torch.Tensor] = {}
         self.VAR_SET_2D = []
         self.VAR_SET_3D = []
-        self.prognostic_var_names = dataset_spec.prognostic_var_names(
-            prognostic_vars_key
-        )
-        self.boundary_var_names = dataset_spec.boundary_var_names(boundary_vars_key)
+        self.prognostic_var_names = dataset_spec.prognostic_var_names
+        self.boundary_var_names = dataset_spec.boundary_var_names
```
alxmrs, line 478: Yeah, it's great we can just access the var names -- better to set this early rather than late (at every access point).

tests/conftest.py
```diff
@@ -26,6 +26,28 @@
...
+
+
+def _om4_canonical_var_names(var_names: Iterable[str]) -> list[str]:
```
alxmrs, line 39: Useful to have for consistent test set up.

alxmrs (approved):
Took another look, sorry to miss your ping. Still LGTM.

jder:
@fomo-bot can you please run a tiny-model training run locally using the real 1-degree om4 data and `--debug` to make sure this successfully can load it and get through a train + validation epoch please? And if so then please go ahead and merge this PR.

PR #670, opened 2026-04-08
Add LLC CPU data canonicalization

jder:
Add LLC data support (preprocessing, etc). Brings time configuration down into the data source since LLC data has different time formats than OM4, and we likely want to use different time ranges even for different resolutions of OM4 data. 

It looks big but a lot of it is mechanical config changes and tests; happy to walk through live and/or break up further if desired!


jder:
@fomo-bot can you please rebase this PR into a single clean commit on top of the latest main?

jder:
@fomo-bot can we unify the two different mock LLC datasets in tests, please? 

I also don't really understand all the complexity around the time stuff. Why do we need to change `slice` if we are already remapping the times in the canonicalize function?

src/ocean_emulators/utils/data.py
```diff
@@ -44,6 +46,74 @@
...
+                [_to_julian_datetime(value) for value in flat_values],
+                dtype=object,
+            ).reshape(values.shape)
+
+    units = time_coord.attrs.get("units")
+    if units is None:
+        raise ValueError(
+            "LLC time coordinate is numeric but missing 'units'; unable to decode "
+            "raw time values."
+        )
```
jder: @fomo-bot do we really need all these cases? I would expect only one of them is in fact used?

src/ocean_emulators/utils/data.py
```diff
@@ -280,11 +299,18 @@ def from_locations(
         boundary_var_names: BoundaryVarNames,
         static_data_vars: list[str] | None,
         use_dask: bool,
+        canonicalize: Callable[
+            [xr.Dataset, xr.Dataset, xr.Dataset],
+            tuple[xr.Dataset, xr.Dataset, xr.Dataset],
+        ]
+        | None = None,
```
jder: This seems a bit silly compared to just passing in the canonicalized datasets and pulling the open logic out into DataConfig.build

jder: @fomo-bot can you try this ^ please? And update to the latest main?

jder:
@fomo-bot can you rebase this on the latest main please

configs/data/om4.yaml
```diff
@@ -4,13 +4,20 @@
...
-
 sources:
-  - data_location: OM4.zarr
+  - type: om4
+    prognostic_vars_key: thermo_dynamic_all
+    boundary_vars_key: tau_hfds_hfds_anom
+    train_time:
+      start: "1975-01-03"
+      end: "2013-10-05"
+    val_time:
```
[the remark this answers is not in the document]
jder: Fixed in `73674621`. The smoke path now replaces the nested source configuration through the supported `--data.sources` JSON argument, preserving the shortened train/validation windows and the source metadata. Validated with `bash -n`, an actual `TrainConfig.from_yaml_and_cli` parse of the exact payload, and the repository pre-commit hooks.

src/samudra/utils/data.py
```diff
@@ -327,9 +313,145 @@ def from_datasets(
...
+    requested_data_vars = {
+        var_name.rsplit("_", maxsplit=1)[0]
+        if var_name.rsplit("_", maxsplit=1)[-1].isdigit()
+        else var_name
+        for var_name in (
+            dataset_spec.prognostic_var_names + dataset_spec.boundary_var_names
+        )
+    }
+    requested_data_vars.update(
+        [dataset_spec.mask_all_levels_var, "mask_c", *dataset_spec.mask_vars]
```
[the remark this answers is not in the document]
jder: @fomo-bot can you take a look at https://github.com/m2lines/samudra/tree/llc_cpu_working and see if it has the same problem and if not, how it solves this?

src/samudra/utils/data.py
```diff
@@ -330,6 +300,139 @@ def from_datasets(
...
+    )
+    data_copy = data_copy[
+        [name for name in data_copy.data_vars if name in requested_data_vars]
+    ]
```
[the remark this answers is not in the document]
jder: @fomo-bot this makes sense to me, let's include the static vars in the requested_data_vars above?

src/samudra/utils/data.py
```diff
@@ -327,9 +313,145 @@ def from_datasets(
...
+        )
+
+    data_copy = data_copy.drop_vars(["i_g", "j_g"], errors="ignore")
+    rename_map = {
+        old: new
+        for old, new in {
+            "k": "lev",
+            "mask_c": "wetmask",
+            "i": "x",
+            "j": "y",
```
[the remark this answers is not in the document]
jder: Made #801 for this.

src/samudra/utils/data.py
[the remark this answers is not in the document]
jder: @fomo-bot ok that is too complicated, can you please undo that nonsense and just reject static vars for LLC until we have a need for them please?

src/samudra/utils/data.py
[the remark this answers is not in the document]
jder: @fomo-bot poke

jder:
@fomo-bot please fix precommit

configs/data/llc.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+# yaml-language-server: $schema=../schemas/DataConfig.json
+
+sources:
+  - type: llc
+    face: 1
+    i_start: 0
+    i_end: 720
+    j_start: 0
+    j_end: 720
```
alxmrs, line 13: This is 1 of (13?) faces and a subset of that face along 2d (e.g. lat or lon dims) -- is that correct? I want to try to grok the new dataset.

configs/samudra_multi_om4/eval.yaml
```diff
@@ -34,20 +31,41 @@ experiment:
...
-    boundary_vars_key: tau_hfds
   static_data_vars: []
   concurrent_compute: true
   sources:
-    - data_location: om4_halfdeg_v4/OM4.zarr
+    # First source is used for inference. Reorder to eval a different scale.
+    - type: om4
+      prognostic_vars_key: thermo_dynamic_all
+      boundary_vars_key: tau_hfds
+      train_time: &train_time
```
alxmrs, line 41: What does the `&` imply? (probably will answer as we read). Having PL features in YAML makes me a little concerned.

src/samudra/utils/data.py
```diff
@@ -17,7 +17,7 @@
 from jaxtyping import Bool, Float
 
 if TYPE_CHECKING:
-    from samudra.config import TimeConfig
+    from samudra.config import SourceTimeConfig
```
alxmrs: Is there also a non-source TimeConfig that we're keeping around? If not, then should we just keep the name "TimeConfig"? WDYT?

src/samudra/utils/data.py
```diff
@@ -330,6 +305,167 @@ def from_datasets(
         )
 
 
+def _rename_llc_level_index_vars(ds: xr.Dataset) -> xr.Dataset:
```
alxmrs: I hope there are roundtrip tests for these. Nice to see a function and its inverse.

src/samudra/utils/data.py
```diff
@@ -330,6 +305,167 @@ def from_datasets(
...
+    if {"U", "oceTAUX"} & requested_data_vars:
+        if mask_var := _preferred_available_var(data, ("mask_w", "hFacW")):
+            mask_vars.add(mask_var)
+    if {"V", "oceTAUY"} & requested_data_vars:
+        if mask_var := _preferred_available_var(data, ("mask_s", "hFacS")):
+            mask_vars.add(mask_var)
+    return mask_vars
+
+
+def canonicalize_llc_datasets(
```
alxmrs: Idea: should we start to create data source specific utility files? e.g. keep data.py generic, and then separate out OM4 vs llc specific utilities? This might help us create better abstractions later on.

alxmrs: We could also do this after one more dataset gets added (to channel my inner YAGNI).

alxmrs: It seems like we're fairly close to creating new overridable abstract methods in DatasetSpec for these kinds of functions, but not quite yet.

src/samudra/utils/data.py
```diff
@@ -407,19 +543,29 @@ def to(self, device: torch.device, non_blocking: bool = True) -> Self:
         )
 
 
+@dataclasses.dataclass
+class DataSourceSplits:
```
alxmrs, line 387: Nice

src/samudra/config.py
```diff
@@ -80,55 +88,61 @@ class JulianDate:
...
+# We reuse pydantic's AwareDatetime to match JSONSchema's expected RFC3339 format
+_aware_datetime_adapter = pydantic.TypeAdapter(pydantic.AwareDatetime)
 
-    Endpoints are Julian dates (not times) but cftime stores them in datetimes.
-    The final endpoint is exclusive.
-    """
 
-    start: DateConfig
-    end: DateConfig
+def _datetime64_validator(value: str | np.datetime64) -> np.datetime64:
```
alxmrs, line 111: Do we have to worry about datasets using cftime? Do we convert from cftime before we reach this method?

alxmrs: I think the JulianDateConfig may handle it before this validator, but I think it would be worth adding a test to be sure.

alxmrs: Ok, it looks like the LLC dataset may not use cftime at all, which would be ideal.

src/samudra/config.py
```diff
@@ -138,13 +152,38 @@ def __str__(self) -> str:
...
+
+
 LOCATION_DOCS = (
     "Use a string relative to the `data_root` or use a structured location "
     "see location.py for possible types."
 )
 
 
-class DataSourceConfig(BaseConfig):
+class BaseDataSourceConfig[SourceTimeConfigT: (Om4TimeConfig, LlcTimeConfig)](
```
alxmrs: Very nice generic typing 🧠

src/samudra/config.py
```diff
@@ -155,6 +194,93 @@ class DataSourceConfig(BaseConfig):
         description="Location of the data standard deviations; " + LOCATION_DOCS
     )
 
+    @property
+    @abc.abstractmethod
+    def dataset_spec(self) -> DatasetSpec:
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def canonicalize_datasets(
```
alxmrs, line 197: Looks like you already had the idea I mentioned above, just abstracting at config time. Nice.

src/samudra/utils/data.py
alxmrs: Looks like this somewhat exists in config.py. Nice.

src/samudra/config.py
```diff
@@ -155,6 +194,93 @@ class DataSourceConfig(BaseConfig):
...
+    ) -> DataSourceSplits:
+        source = self._build_source(
+            data_root,
+            turn_on_dask=use_dask,
+            static_data_vars=static_data_vars,
+        )
+        inference_source = None
+        if is_primary and self.inference_times:
+            if use_dask:
+                full_inference_source = source
```
alxmrs, line 223: I would love to revisit the idea that we don't really need dask with an agent sometime soon.

src/samudra/config.py
```diff
@@ -155,6 +194,93 @@ class DataSourceConfig(BaseConfig):
...
+            else:
+                full_inference_source = self._build_source(
+                    data_root,
+                    turn_on_dask=True,
+                    static_data_vars=static_data_vars,
+                )
+            # TODO: remove multiple inference time ranges altogether
```
alxmrs: 🐑 Can we track this with an issue?

src/samudra/config.py
```diff
@@ -155,6 +194,93 @@ class DataSourceConfig(BaseConfig):
...
+            data_root.resolve(location)
+            for location in (
+                self.data_location,
+                self.data_means_location,
+                self.data_stds_location,
+            )
+        ]
+        resolved_data_location, resolved_means_location, resolved_stds_location = (
+            resolved_locations
+        )
```
alxmrs: Is this simpler? WDYT?

```suggestion
    ) -> DataSource:
        resolved_data_location = data_root.resolve(self.data_location)
        resolved_means_location = data_root.resolve(self.data_means_location)
        resolved_stds_location = data_root.resolve(self.data_stds_location)
```

alxmrs: Warning: hand written

src/samudra/config.py
```diff
@@ -196,43 +322,96 @@ def persistent_pytorch_workers(self) -> bool:
...
+    @property
+    def dataset_spec(self) -> DatasetSpec:
         return build_om4_spec(
             self.prognostic_vars_key,
             self.boundary_vars_key,
         )
 
+    @pydantic.model_validator(mode="after")
+    def validate_time_splits(self) -> Self:
+        if self.train_time.overlaps(self.val_time):
```
alxmrs, line 334: Nice guard.

src/samudra/config.py
```diff
@@ -196,43 +322,96 @@ def persistent_pytorch_workers(self) -> bool:
...
+            )
+        return self
+
+    def canonicalize_datasets(
+        self,
+        data: xr.Dataset,
+        means: xr.Dataset,
+        stds: xr.Dataset,
+    ) -> tuple[xr.Dataset, xr.Dataset, xr.Dataset]:
+        return data, means, stds
```
alxmrs, line 350: Don't we technically canonicalize the dataset upon open? Could a lot of that logic be moved to here?

src/samudra/config.py
```diff
@@ -196,43 +322,96 @@ def persistent_pytorch_workers(self) -> bool:
...
+            dataset_spec=self.dataset_spec,
+        )
+
 
-DatasetConfig = Annotated[
-    Om4DatasetConfig | LlcDatasetConfig,
+DataSourceConfig = Annotated[
+    Om4DataSourceConfig | LlcDataSourceConfig,
     Field(discriminator="type"),
 ]
```
alxmrs, line 413: Nice

src/samudra/config.py
```diff
@@ -248,64 +427,37 @@ class DataConfig(BaseConfig):
     masked_fill_value: float = 0.0
     concurrent_compute: bool = False
 
+    @pydantic.model_validator(mode="after")
+    def reject_llc_static_data_vars(self) -> Self:
```
alxmrs: I'm not sure the place, but this doesn't seem like the right place to have the validation. Maybe it would better be located inside `LlcDataSourceConfig`? WDYT?

src/samudra/config.py
```diff
@@ -248,64 +427,37 @@ class DataConfig(BaseConfig):
...
+        ):
+            raise ValueError("LLC data sources do not support static_data_vars")
+        return self
+
     def build(
         self,
         data_root: ResolvedLocation,
     ) -> DataContainer:
-        dataset_spec = self.dataset.build()
-        if self.dataset.type != "om4":
```
alxmrs, line 259: Nice

src/samudra/config.py
```diff
@@ -248,64 +427,37 @@ class DataConfig(BaseConfig):
...
-            resolved_means_location = data_root.resolve(means_location)
-            resolved_stds_location = data_root.resolve(stds_location)
-            return DataSource.from_locations(
-                data_location=resolved_data_location,
-                means_location=resolved_means_location,
-                stds_location=resolved_stds_location,
-                dataset_spec=dataset_spec,
-                prognostic_var_names=dataset_spec.prognostic_var_names,
-                boundary_var_names=dataset_spec.boundary_var_names,
+        source_splits = [
```
alxmrs, line 443: Oh yeah, much cleaner.

src/samudra/config.py
alxmrs: Furthermore: do we use static vars in OM4 in the first place? Should we cull it? 

And, if we do for OM4 and want to use static vars in LLC, what is a better place to factor that information? I think it probably should be pushed down to the data source instead of the "DataConfig".

src/samudra/train.py
```diff
@@ -364,37 +354,23 @@ def __init__(self, cfg: TrainConfig) -> None:
...
-                boundary_var_names=self.boundary_var_names,
-                hist=self.hist,
-                normalize_before_mask=self.normalize_before_mask,
-                masked_fill_value=self.normalize_fill_value,
-                long_rollout=True,
-            )
-
-            inference_datasets.append(inference_dataset)
-            num_steps_inf_set.append(num_time_steps)
+        assert self.inference_src is not None
```
alxmrs, line 348: Much cleaner, nice.

tests/llc_fixtures.py
```diff
@@ -0,0 +1,128 @@
...
+
+from pathlib import Path
+
+import numpy as np
+import xarray as xr
+
+from samudra.constants import build_llc_spec
+
+
+def raw_llc_datasets(n_time: int = 3) -> tuple[xr.Dataset, xr.Dataset, xr.Dataset]:
```
alxmrs, line 13: Should this live here or conftest (as a fixture)?

tests/test_utils_data.py
```diff
@@ -225,6 +227,115 @@ def test_normalize_compact_mixed_depth_and_surface_stats(data_source):
     assert normalize.normalize_tensor_prognostic(prognostic).shape == prognostic.shape
 
 
+def test_canonicalize_llc_datasets_standardizes_layout():
```
alxmrs, line 366: Nice

alxmrs (commented):
Looks really great! I mostly have questions about factoring (where to put the new things), and then some conversations with myself to understand the PR. 

Thanks for this change, this is really valuable.

configs/data/llc.yaml
jder: Yep, that's right.

configs/samudra_multi_om4/eval.yaml
jder: Yeah, this is YAML's built-in "structure sharing" feature. This names the following dictionary and then the later `*train_time` refers to that.

src/samudra/utils/data.py
jder: Will add.

src/samudra/utils/data.py
jder: Yeah, makes sense.

src/samudra/utils/data.py
jder: Yeah, I can break out the LLC-specific stuff.

src/samudra/config.py
jder: Yeah this is specifically for the LLC dataset, which does not use cftime.

src/samudra/config.py
jder: I still don't believe it has much value to include every piece of tech debt as an issue we never review, but ok. #813

src/samudra/config.py
jder: Yes, a follow-up could move all the OM4 canonicalization stuff in here. I will add a TODO.

src/samudra/config.py
jder: Sure, I will push it down. I think we should just delete them since they were only used for correctors we're not using.

tests/llc_fixtures.py
jder: I don't see any need for it to be a fixture, and generally prefer less magic when possible.

jder:
@alxmrs I think I did all the things above, RFAL!

configs/samudra_multi_om4/eval.yaml
alxmrs: Ah, I forget that YAML is pretty programmatic. Thanks.

alxmrs (approved):
LGTM! Thanks!

src/samudra/utils/llc.py
```diff
@@ -0,0 +1,196 @@
...
+
+    requested_data_vars = {
+        _var_without_level(var_name)
+        for var_name in (
+            dataset_spec.prognostic_var_names + dataset_spec.boundary_var_names
+        )
+    }
+    requested_data_vars.update(
+        [dataset_spec.mask_all_levels_var, "mask_c", *dataset_spec.mask_vars]
+    )
```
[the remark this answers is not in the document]
jder: @fomo-bot can you please verify if our LLC mock data includes hFacC? If it does, please fix this comment. Either way let's then get this merged.

PR #671, opened 2026-04-08
Enable LLC training configs

jder:
## Summary
This is PR 4 in the LLC stack, based on #670.

It lands the minimal model and config changes needed to make the initial LLC training configs valid on the standard training path. The intent is to support the initial LLC experiment shape without mixing in GPU decode or cluster-local launcher logic.

## Why
Once the LLC CPU loader exists, the next step is making LLC runs representable in the training config and model code. Keeping that separate from GPU decode keeps the review surface smaller.

## Notes
This remains a draft stacked PR.

### DIFF
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -519,7 +519,7 @@ class DataConfig(BaseConfig):
 
 BlockType = Literal["conv_next_block", "conv_block"]
 ActivationType = Literal["relu", "gelu", "capped_gelu"]
-NormType = Literal["batch", "instance", "layer"]
+NormType = Literal["batch", "instance", "group", "nonorm", "layer"]
 
 
 class BlockConfig(BaseConfig):
@@ -528,6 +528,7 @@ class BlockConfig(BaseConfig):
     activation: ActivationType = "capped_gelu"
     upscale_factor: int = 4
     norm: NormType = "batch"
+    group_norm_groups: int = Field(default=32, ge=1)
     pointwise_linear: bool = False
 
     def build(self) -> CoreBlockBuilder:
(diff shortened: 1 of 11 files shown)


alxmrs:
Hey @fomo-bot will you resolve the merge conflicts in this change? This looks quite tricky, so let me know if I can help.

alxmrs:
@fomo-bot will you rebase from main and address merge conflicts?

src/samudra/models/modules/blocks.py
```diff
@@ -280,6 +280,7 @@ def __init__(
         pad="circular",
         upscale_factor: int = 4,
         norm="batch",
+        group_norm_groups: int = 32,
```
jder: Rather than a separate group_norm_groups which is ignored when norm != group, an we pull the same trick as elsewhere were we supply a dataclass for norm but support `norm = "batch"` as a short form of `norm.type="batch"`?

src/samudra/config.py
```diff
@@ -1286,6 +1288,7 @@ class TrainConfig(TopLevelConfig):
 
     # Data parameters at root level
     data_stride: list[int] = [1]
+    temporal_stride: int = Field(default=1, ge=1)
```
jder: What is the difference between data_stride and temporal_stride? If they're not equivalent we need some docs here.

src/samudra/models/modules/blocks.py
```diff
@@ -343,6 +338,29 @@ def __init__(
...
+        if norm == "batch":
+            return torch.nn.BatchNorm2d(channels)
+        if norm == "instance":
+            return torch.nn.InstanceNorm2d(channels)
+        if norm == "group":
+            if group_norm_groups < 1:
+                raise ValueError("group_norm_groups must be >= 1")
+            num_groups = min(group_norm_groups, channels)
+            while channels % num_groups != 0:
+                num_groups -= 1
```
jder: Let's make this an error for now instead of silently accommodating; too easy to test two different values of group_norm_groups and think you're changing something when you're not. If it gets annoying we can reintroduce this with a warning or something.

jder (commented):
@fomo-bot can you please resolve these comments (+ codex's) & any (non-flake) test failures?

PR #672, opened 2026-04-08
Add GPU zarr decode path and runtime packaging

jder:
## Summary
This is PR 5 in the LLC stack, based on #671.

It combines the guarded GPU zarr decode runtime path with the dependency, container, and smoke-test updates needed to make that path reproducible. The PR includes GPU-aware xarray/zarr opening, tensor materialization changes, the pinned zarr/xarray/kvikio/nvcomp stack, and container smoke coverage.

## Why
The runtime path and the packaging changes are tightly coupled: the GPU decode code is only useful if reviewers can see the exact dependency and container assumptions it requires. Keeping them together makes the rollout boundary clearer.

## Notes
This remains a draft stacked PR.


jder:
This is fragile and requires us to maintain forks of various libraries, so closing for now until we've shown that #800 is insufficient. (At which point we might add the GPU-side decode into that instead of via xarray)

PR #675, opened 2026-04-09
Attention viz

amogh-gulati:
Add attention weight visualization during validation. 
The AttentionAggregator captures transient attention maps from AxialAttention and FullAttention layers inside the UNet backbone by holding direct references to those modules (which is why it takes the model as a constructor arg). To avoid overhead, the capture_attention context manager only wraps the last validation batch, temporarily setting capture_weights=True on each attention module so their forward pass materializes and stores the full weight matrices alongside the normal scaled_dot_product_attention output. 
The aggregator then reads those weights and logs heatmaps (per-axis attention maps, outer-product receptive fields for axial attention, and reshaped query-row receptive fields for full attention) to W&B under keys derived from the backbone's layer names (e.g. val/bottleneck/height, val/encoder_0/receptive_field). For models without attention blocks, everything is a no-op.


src/ocean_emulators/aggregator/validate/attention.py
```diff
@@ -0,0 +1,222 @@
+from __future__ import annotations
```
alxmrs, line 1: Minor nit: would you mind adding a docstring to the top of this file that explains what it does?

src/ocean_emulators/aggregator/validate/attention.py
```diff
@@ -0,0 +1,222 @@
...
+
+def _collect_attention_blocks(
+    model: nn.Module,
+) -> list[tuple[str, AxialAttentionBlock | FullAttentionBlock]]:
+    backbone = _get_unet_backbone(model)
+    if backbone is None:
+        return []
+
+    blocks: list[tuple[str, AxialAttentionBlock | FullAttentionBlock]] = []
+    for layer_name, layer in zip(backbone.layer_names, backbone.layers, strict=True):
```
alxmrs, line 57: Good catch with `strict=True`, I've been burned by that before.

src/ocean_emulators/aggregator/validate/attention.py
```diff
@@ -0,0 +1,222 @@
...
+            return backbone
+
+    for module in model.modules():
+        if isinstance(module, UNetBackbone):
+            return module
+
+    return None
+
+
+def _collect_attention_blocks(
```
alxmrs, line 49: Reading the code below, I think you would be able to omit a lot of conditional logic if this method returned two lists of tuples: `tuple[list[tuple[str, AxialAttentionBlock]], list[tuple[str, FullAttentionBlock]]]`. While this is a complicated return type, it would save you from having to check which type of attention block you're working with down the line.

src/ocean_emulators/aggregator/validate/attention.py
```diff
@@ -0,0 +1,222 @@
...
+        self,
+        model: nn.Module,
+        query_lat: int | None = None,
+        query_lon: int | None = None,
+    ):
+        self._blocks = _collect_attention_blocks(model)
+        self._query_lat = query_lat
+        self._query_lon = query_lon
+        self._axial_captures: dict[str, tuple[torch.Tensor, torch.Tensor]] = {}
+        self._full_captures: dict[str, tuple[torch.Tensor, tuple[int, int]]] = {}
```
alxmrs, line 115: It might be helpful to leave a comment that explains that these weights get overwritten every call to record_batch. Is that the intended behavior?

src/ocean_emulators/aggregator/validate/attention.py
```diff
@@ -0,0 +1,222 @@
...
+                )
+
+    @torch.no_grad()
+    def get_logs(self, label: str) -> Metrics:
+        logs: MetricsDict = {}
+
+        for name, (height_weights, width_weights) in self._axial_captures.items():
+            height_np = height_weights.float().numpy()
+            width_np = width_weights.float().numpy()
+            query_lat = (
```
alxmrs, line 170: 🐑 for my own brain, it would be helpful to annotate these two variables with the type:
```suggestion
            query_lat: int = (
```

src/ocean_emulators/aggregator/validate/attention.py
```diff
@@ -0,0 +1,222 @@
...
+            query_lon = (
+                self._query_lon
+                if self._query_lon is not None
+                else width_np.shape[0] // 2
+            )
+
+            logs[f"{label}/{name}/height"] = plot_attention_map(
+                height_np,
+                axis="height",
+                caption="Height-axis attention (avg over heads, batch, width)",
```
alxmrs, line 184: Are these averaged? Where does that happen?

src/ocean_emulators/aggregator/main.py
```diff
@@ -28,6 +30,7 @@ def get_validation_aggregator(
         num_prognostic_channels: int,
         *,
         include_image_aggregators: bool = True,
+        model: nn.Module,
```
alxmrs, line 33: Maybe, let's make this module or None -- it would be a performance advantage to be able to turn off attention aggregation if it is not included in the model.

src/ocean_emulators/aggregator/plotting.py
```diff
@@ -133,3 +135,130 @@ def _stitch_data_panels(data: list[list[np.ndarray]], fill_value) -> np.ndarray:
             stitched_data[start_row:end_row, start_col:end_col] = arr
 
     return stitched_data
+
+
+def _downsample_for_display(data: np.ndarray, max_size: int = 256) -> np.ndarray:
```
alxmrs: 🐑 Can we use jaxtyping with np.ndarray as the base array type?

alxmrs: If it's too complicated, I'm happy to skip it.

alxmrs (commented):
First pass on the review, I ran out of time!

PR #677, opened 2026-04-10
Small improvements from profiling on NYU torch

jder:
3 small changes to help with GPU utilization:

* Don't compute validation images if w&b is off.
* Only compute images on rank 0 (this doesn't really change latency but it does mean other CPUs/GPUs are not competing for resources while we do this)
* Turn on persistent workers to reduce start-of-epoch delays.

### DIFF
--- a/src/ocean_emulators/aggregator/validate/map.py
+++ b/src/ocean_emulators/aggregator/validate/map.py
@@ -2,7 +2,7 @@ import torch
 
 from ocean_emulators.aggregator.plotting import plot_paneled_data
 from ocean_emulators.aggregator.validate.sub_aggregator import ValidateSubAggregator
-from ocean_emulators.utils.distributed import all_reduce_mean
+from ocean_emulators.utils.distributed import all_reduce_mean, is_main_process
 
 
 class MapAggregator(ValidateSubAggregator):
@@ -73,6 +73,7 @@ class MapAggregator(ValidateSubAggregator):
         time_dim = 0
         target_time = self.hist  # Use latest time step
         image_logs = {}
+        render_images = is_main_process()
         sorted_names = sorted(list(self._gen_data.keys()))
         for name in sorted_names:
             # use first sample in batch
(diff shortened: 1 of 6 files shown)


src/ocean_emulators/aggregator/validate/map.py
```diff
@@ -88,6 +89,8 @@ def get_logs(self, label: str):
                 )
                 / self._n_batches
             )
+            if not render_images:
+                continue
```
alxmrs, line 93: Very slick!

tests/test_validate_aggregator.py
```diff
@@ -148,3 +151,62 @@ def test_validation_aggregator__reduced_only__omits_image_logs(
...
+def test_snapshot_aggregator__non_main_rank__skips_plot_rendering(
+    dummy_src: DataSource, monkeypatch: pytest.MonkeyPatch
+):
+    val_batch = val_batch_of(*dummy_src.grid_size)
+    aggregator = SnapshotAggregator(dummy_src.metadata, hist=0)
+
+    monkeypatch.setattr(
+        "ocean_emulators.aggregator.validate.snapshot.is_main_process",
+        lambda: False,
+    )
```
alxmrs, line 165: TIL about this pytest feature.

src/ocean_emulators/train.py
```diff
@@ -1056,6 +1063,15 @@ def remove_module_prefix(state_dict, prefix="module."):
     def is_wandb_enabled(self):
         return self.wandb_logger.enabled and is_main_process()
 
+    def _sync_flag_from_main(self, flag: bool) -> bool:
```
alxmrs, line 1069: What is this for?

alxmrs (approved):
LGTM, thank you for these performance improvements. Just one question for my understanding.

src/ocean_emulators/train.py
jder: It transmits from the main process to other GPU processes whether the main process is set up to log to W&B. (The other processes always have disabled W&B loggers.)

jder:
@fomo-bot can you merge with the latest main please

PR #679, opened 2026-04-13
Make config to document how we trained 1/4-degree models

jder:
Basically taking the cli args from https://wandb.ai/ocean_emulators/default/runs/9t6la3qq/overview?nw=nwuseroa_jder and turning them into a saved config we can reuse.


alxmrs (approved):
Glad we have this persisted in the codebase.

Issue #615, opened 2026-03-02
Support mixed scale forcings for FOMO model


alxmrs:
Our second KR of our first objective is:

> Produce ¼ output after 10-year rollout with 1-degree inputs by May 1st.

---update after 2026-04-06 discussion---

After thinking this over with Amogh and Jesse, we realized that the core engineering required to accomplish this capability is to have inference predict at a high resolution, but use the forcings from a low resolution. We imagine our multiscale model performing inference as such (at least, initially): 

* We train on all of 1 degree OM4 and a time limited range of 1/4 degree OM4. Each batch uses it's scale's forcings.
* At inference time, we start the rollout at 1/4 degrees. This is just to simplify the engineering work; we can add the capability to start a the coarse resolution and then further generate high resolution rollouts later.
* Each rollout step will consume 1/4 degree prognostics and 1 degree boundary forcings. It will predict 1/4 degree prognostics.

One alternative considered in our conversation is that we could update the rollout to operate on a latent space. In this formulation, we take the input at either fine or coarse resolution, encoding it into a latent space, perform a rollout taking coarse forcings as input and latent information as prognostics, and predict latent prognostic information. If we adopting this approach, we'd still need to merge the emergent shape of the prognostic representation with physical coarse forcings. 

The long term goal we have is to use 1/4 degree OM4 as the low resolution and produce 1/48 degree LLC outputs. This creates constraints in our solution space -- for example, we can't coarsen a high resolution output back into a coarse input for an easier rollout formulation (at least, not without using the model somehow to do the upscaling). In addition, this means that we won't be able to formulate the problem to have the model cross train on different boundaries x prognostics. Instead, we may bank of tactics like adding dropout to the boundary information to incentivize the model to not depend too much on boundary inputs.

jder:
See #702

PR #681, opened 2026-04-13
U/alxmrs/experiments/kr2

alxmrs:
This PR makes four contributions towards implementing O1KR2 (#615): 
1. Updates the encoder to use dual perceivers to separately represent prognostic and boundary forcings. This is needed because at inference, we plan to use coarse forcings for high res prognostic prediction. Latents from both boundary and prognostic perceivers are linearly mixed, then positional encodings are added before the output is passed on (to the processor). 
2. To enable (1) in the model, this PR also updates the data plumbing across the whole training and inference process to separate the prognostic and boundary tensors. These are no longer concatenated as an "input" tensor.
3. Also to enable (1), we now update the training and inference AR process to update the `GridContext` input resolution after step 1. In this new scheme, step 0 does either downscaling or upscaling, and step 1+ updates the grid context to make the input resolution the output resolution. This is needed during "mix" schedule multiscale training, since step 1 and after the prognostic is fed back to the previous decoder step at the output resolution. This does not affect "match" schedule multiscale training, because both resolutions are already equal. 
4. In addition to unit tests being updated, we add a new `test_fomo_cross_resolution` set of integration tests. These vet that single step cross resolution (prog/boundary), single step mix schedule, and two step AR mixed schedule training of the FOMO model all work as expected.


src/ocean_emulators/models/modules/encoder.py
```diff
@@ -27,27 +31,84 @@ def patch_from(
     return patch_h, patch_w
 
 
+class MultiHeadAttention(nn.Module):
```
alxmrs: TODO: this belongs in its own module

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -28,26 +29,34 @@ def patch_from(
 
 
 class PerceiverEncoder(nn.Module):
```
alxmrs, line 31: The changes in this branch are in service of this encoder update.

alxmrs:

alxmrs:

alxmrs:

src/ocean_emulators/models/base.py
```diff
@@ -44,33 +48,29 @@ def forward(
...
-                input_tensor = train_data.merge_prognostic_and_boundary(
-                    prognostic=prev_output, step=step
+                _, boundary_tensor = train_data.get_input(step)
+                prog_tensor = prev_output
+                # From step 1 onward the prognostic is the previous decoder
+                # output, which sits on the output grid — update the context
+                # so the encoder uses the correct resolution.
+                ctx = dataclasses.replace(
+                    ctx, input_resolution_cpu=ctx.output_resolution_cpu
                 )
```
alxmrs, line 70: This addresses the TODO on the left.

src/ocean_emulators/stepper.py
```diff
@@ -20,112 +20,112 @@
 from ocean_emulators.utils.writer import ZarrWriter
 
 
-class Stepper:
-    def __init__(self):
-        pass
+def train_batch(
```
alxmrs, line 23: The `Stepper` class was really three functions in a trenchcoat. So, I extracted them.

alxmrs:
Let's keep the branch, but I'll close this prototype.

PR #701, opened 2026-04-15
Separate prognostic and boundary tensors end-to-end

alxmrs:
Thread prognostic and boundary inputs as separate tensors through the data pipeline, stepper, base model, and all model implementations. Previously they were concatenated along the channel dim before reaching the model; now each model's forward_once receives (prog, boundary, ctx).

FOMO concatenates before its single-stream encoder (the dual-perceiver encoder that enables cross-resolution fusion lands in a follow-up). Samudra and FOMini concatenate inside their forward_once as before.

This is #681 broken up into a stack of three small PRs. 1/3

### DIFF
--- a/src/ocean_emulators/datasets.py
+++ b/src/ocean_emulators/datasets.py
@@ -13,6 +13,7 @@ from torch.utils.data import Dataset
 from xarray_einstats.einops import rearrange as xr_rearrange  # noqa: F401
 
 from ocean_emulators.constants import (
+    Boundary,
     BoundaryVarNames,
     Example,
     GridMask,
@@ -138,9 +139,9 @@ class InferenceDataset(Dataset):
         label = self._get_label(x_index)
         return label
 
-    def get_initial_input(self):
-        data = self.__getitem__(0)[0]
-        return data
+    def get_initial_input(self) -> tuple[Prognostic, Boundary]:
+        prog, boundary, _ = self.__getitem__(0)
+        return prog, boundary
 
     def get_target_time(self, start_step: int, num_steps: int):
         x_index = self._get_x_index(start_step)
@@ -154,20 +155,28 @@ class InferenceDataset(Dataset):
             )
         )
 
-    def merge_prognostic_and_boundary(self, prognostic: torch.Tensor, step: int):
+    def get_boundary_for_prognostic(
+        self, prognostic: Prognostic, step: int
+    ) -> tuple[Prognostic, Boundary]:
+        """Return ``(prognostic, boundary)`` at the requested step.
+
+        Prognostic is passed through verbatim; boundary is fetched from the
+        boundary source and returned on the same device as ``prognostic``. The
+        two tensors may live on different spatial grids once the cross-
+        resolution boundary source lands — the encoder is responsible for
+        fusing them.
+        """
         x_index = self._get_x_index(step)
         boundary = self._get_boundary(x_index).to(prognostic.device)
-        data = torch.cat((prognostic, boundary), dim=1)
-        return data
+        return prognostic, boundary
 
     @elapsed(level=logging.DEBUG)
     def __getitem__(self, idx):
         x_index = self._get_x_index(idx)
-        data_in = self._get_prognostic(x_index)
+        data_in_prog = self._get_prognostic(x_index)
         data_in_boundary = self._get_boundary(x_index)
-        data_in = torch.cat((data_in, data_in_boundary), dim=1)
         label = self._get_label(x_index)
-        return (data_in, label)
+        return (data_in_prog, data_in_boundary, label)
 
     def _get_x_index(self, idx):
         if isinstance(idx, slice):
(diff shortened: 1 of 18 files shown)


alxmrs:
@codex may I have a review of this PR?

src/ocean_emulators/models/base.py
```diff
@@ -101,7 +93,8 @@ def inference(
         num_steps=None,
         epoch=None,
     ) -> ModelInferenceOutput:
-        out_shape = (num_steps, *dataset[0][1].shape[1:])
+        # `dataset[idx]` returns `(prog, boundary, label)`.
```
jder, line 96: <3

src/ocean_emulators/datasets.py
```diff
@@ -154,20 +155,28 @@ def get_target_time(self, start_step: int, num_steps: int):
             )
         )
 
-    def merge_prognostic_and_boundary(self, prognostic: torch.Tensor, step: int):
+    def get_boundary_for_prognostic(
+        self, prognostic: Prognostic, step: int
+    ) -> tuple[Prognostic, Boundary]:
```
jder: I think this is kind of a weird method now? Should it be `get_boundary(step: int)`? I don't really think we need the `to` in here but if we do I'd rather have callers do `get_boundary(step).to(prognostic.device)`, I think?

src/ocean_emulators/models/base.py
```diff
@@ -113,26 +106,18 @@ def inference(
...
-                pred = (
-                    input_tensor[
-                        0,
-                        : self.out_channels,
-                    ].to(  # Residuals on last state in input
-                        device=get_device()
-                    )
-                    + decodings
-                )
+                pred = prog_tensor.to(device=get_device()) + decodings
```
jder: We seem to have lost the `[0, ...` part from the old code. IIUC the prog_tensor has a batch dimension which we're trying to get rid of here before adding decodings. I guess we don't have inference tests that would catch this? Or am I missing something?

src/ocean_emulators/datasets.py
```diff
@@ -357,44 +366,40 @@ def pin_memory(self):
...
-    autoregressive rollout in the training and inference process.
-
-    Constraint: The `Input` tensor is a combination of (flattened) prognostic variables (at all depth levels) and
-    boundary forcings. The top `num_prognostic_channels` number of channels must be prognostic variables whereas the
-    remaining bottom channels are boundary forcings.
+    A single batch contains multiple steps worth of ``Example`` entries, each
+    of which is a ``(prognostic_input, boundary_input, label)`` triple. The
+    prognostic and boundary tensors are carried *separately* (they are no
+    longer concatenated along the channel dim) so that the encoder can fuse
+    them at the token level, potentially at different spatial resolutions.
```
```suggestion
    prognostic and boundary tensors are carried separately because the FOMO model
    encodes them separately (Samudra just concatenates them later).
```

src/ocean_emulators/datasets.py
```diff
@@ -622,43 +636,27 @@ def to_train_data(
...
-            boundary_steps = boundary.normalize_and_mask(
-                self.normalize_before_mask, self.masked_fill_value
-            )
-            boundary_steps = flatten_dims(boundary_steps)
-            return torch.cat((prognostic_steps, boundary_steps), dim=1)
-
-        return prognostic_steps
+        return rearrange(
+            steps, "batch time variable lat lon -> batch (time variable) lat lon"
+        )
```
jder, line 649: Nice cleanup!

src/ocean_emulators/stepper.py
```diff
@@ -20,112 +20,112 @@
...
-                coords=coords,
-                hist=inf_aggregator.hist,
-                model_path=model_path,
-                time_chunk_size=chunk_size,
+    outs = model(batch)[0]
+    loss_per_channel = loss_fn(outs, label, batch.ctx)
+    loss = torch.mean(loss_per_channel)
+    # `input_data` in ValBatchOutput is used for val visualization; pass the
+    # channel-concatenated tensor for continuity with existing consumers.
+    # We only ever mix scales across boundary and prognostic during inference.
```
jder: Does "inference" mean "forward passes" here?

src/ocean_emulators/stepper.py
```diff
@@ -20,112 +20,112 @@
...
+    input_data = torch.cat((prognostic, boundary), dim=1)
+    return ValBatchOutput(loss, loss_per_channel, input_data, label, outs, batch.ctx)
+
+
+@torch.no_grad()
+def inference(
```
jder: Maybe worth a slightly more descriptive name now? eg `run_rollout`?

jder (approved):
A few suggestions but generally LGTM! Love the getting rid of Stepper. I kinda wish Claude did not leave so many vague comments about matching "existing behavior" but 🤷

src/ocean_emulators/datasets.py
alxmrs: Thanks, I was happy about this too.

src/ocean_emulators/stepper.py
alxmrs: hmm, yes, you are right, I'm confusing which AR process is done here.

alxmrs: Wait, looking closer, I do think that it is during eval/inference. Maybe I have the wrong mental model here. To be safe, I'll invert the negative of the sentence and say what I know to be true.

src/ocean_emulators/models/base.py
alxmrs: I know this is covered by tests because when I removed the `unexpand(0)` a test failed. I tried two variants (the status quo and a adhering to the original version) and both versions worked. Merging this version.

jder: @fomo-bot can you help me understand why this version works given my above comments?

jder: @fomo-bot ok, can you please try again to help me understand why this version works given my above comments?

PR #702, opened 2026-04-15
Perceiver encoder updated for cross-resolution fusion

alxmrs:
Replace the single-stream PerceiverEncoder with a dual-perceiver architecture: separate Perceivers for prognostic and boundary streams, fused via concat --> Linear projection. Because patch_extent is in degrees, both streams produce the same latent grid regardless of their spatial resolution, enabling cross-resolution forward passes (e.g. ¼° prog with 1° boundary).

This is https://github.com/Open-Athena/Ocean_Emulator/pull/681 broken up into a stack of three small PRs. 2/3

### DIFF
--- a/src/ocean_emulators/models/modules/encoder.py
+++ b/src/ocean_emulators/models/modules/encoder.py
@@ -2,6 +2,7 @@
 # - https://github.com/microsoft/aurora/blob/main/aurora/model/patchembed.py
 # - https://github.com/microsoft/aurora/blob/main/aurora/model/encoder.py
 # - https://github.com/lucidrains/vit-pytorch
+# - https://github.com/lucidrains/perceiver-pytorch (intra-patch Fourier encoding)
 
 
 import torch
@@ -9,10 +10,15 @@ from aurora.model.fourier import pos_expansion, scale_expansion
 from aurora.model.posencoding import pos_scale_enc
 from einops import rearrange
 from jaxtyping import Float
+from perceiver_pytorch.perceiver_pytorch import fourier_encode
 from torch import nn
 
 from ocean_emulators.constants import Boundary, Prognostic
 from ocean_emulators.utils.ctx import GridContext
+
+# Stream-type indices for the learned stream embedding.
+_PROG_STREAM = 0
+_BOUNDARY_STREAM = 1
 
 
 def patch_from(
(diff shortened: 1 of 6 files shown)


alxmrs:

alxmrs:
Discussion with @jder - this is not quite the direction that we want. There is a way to accomplish this with a single perceiver: if we treat the prognostic and boundary as 1d vectors and contact then, we could use the perceiver on the 1d sequence. To preserve the patch encoding, we'd need to use our own Fourier positional encoding scheme, which should be easy to implement or reuse. We would also need to encode which part of the sequence was the prognostic and boundary in an encoding scheme. 

There are many modeling reasons (large context / latent mixing) why this is preferred over using two perceivers.

jder:
Sounds great to me!

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -27,27 +35,62 @@ def patch_from(
...
-    contain more data per patch, the patch will refer to the same physical area on Earth as all other scales.
+class PerceiverEncoder(nn.Module):
+    """Single-perceiver encoder over a concatenated prog + boundary sequence.
+
+    Each input stream is patchified in 2-D using the same ``patch_extent``
+    in degrees, so both produce the same patch grid regardless of native
+    resolution.  Within each patch:
+
+      * Prog and boundary pixels are flattened to 1-D token sequences and
+        linearly projected to a common ``token_dim``.
```
alxmrs, line 70: This is equivalent to a 1x1 Conv, see comment in code.

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -56,59 +99,142 @@ class PerceiverEncoder(nn.Module):
...
         )
 
-        # Calculate and add positional + scale encoding
+        # Concat prog + boundary tokens along the sequence dim.  The Perceiver
+        # cross-attends over the unified sequence, so prog and boundary mix
+        # inside the latent set rather than being fused after the fact.
+        seq = torch.cat([prog_tokens, boundary_tokens], dim=1)
+
+        pooled = self.perceiver(seq)  # (B*lat_h*lat_w, latent_dim)
+        x = self.fusion_proj(pooled)  # (B*lat_h*lat_w, out_channels)
```
alxmrs, line 228: Why include the linear layer here? Why not just set the perceiver to use the output dim? I believe the last layer of the perceiver is a proj.

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -56,59 +99,142 @@ class PerceiverEncoder(nn.Module):
...
+        )
+        boundary_tokens = self._tokenize(
+            boundary, self.boundary_proj, b_ph, b_pw, _BOUNDARY_STREAM
         )
 
-        # Calculate and add positional + scale encoding
+        # Concat prog + boundary tokens along the sequence dim.  The Perceiver
+        # cross-attends over the unified sequence, so prog and boundary mix
+        # inside the latent set rather than being fused after the fact.
+        seq = torch.cat([prog_tokens, boundary_tokens], dim=1)
```
alxmrs, line 225: I ran out of time to finish this PR today, but @jder, was this more of what you were imagining?

jder: Yes, this looks like the right direction to me!

src/ocean_emulators/models/modules/encoder.py
```diff
@@ -56,59 +99,142 @@ class PerceiverEncoder(nn.Module):
...
+            self.num_freq_bands,
+            self.max_intra_patch_freq,
+            tokens.device,
+            tokens.dtype,
+        )
+        # TODO(alxmrs): Should we make a pos_proj for each stream, or use a shared projection?
+        tokens = tokens + self.pos_proj(pos_enc).unsqueeze(0)
+
+        stream_idx = torch.tensor(stream_id, device=tokens.device)
+        tokens = tokens + self.stream_embed(stream_idx)
```
jder, line 193: FWIW given the per-stream projection above in `tokens = proj(tokens)` I don't think we also need this since there's effectively a learned embedding inside the linear projection.

jder:
Jesse to review after getting KR1 branch merged + running.

PR #703, opened 2026-04-15
Track prognostic resolution across autoregressive rollout steps

alxmrs:
In BaseModel.forward and BaseModel.inference, update ctx.input_resolution_cpu to output_resolution_cpu from step 1 onward. The prognostic tensor fed back is the previous decoder output, which sits on the output grid — the encoder's positional encoding must match. This is a no-op when input and output resolutions are equal (match schedule).

Without this fix, multi-step mix-schedule training would use incorrect positional features from step 1 onward, and would crash with 3D coordinates enabled (shape mismatch).

This is https://github.com/Open-Athena/Ocean_Emulator/pull/681 broken up into a stack of three small PRs. 3/3

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -387,12 +387,22 @@ class PerceiverConfig(BaseConfig):
     def build(
         self,
         in_channels: int,
-        out_channels: int,
         max_patch_size: tuple[int, int],
         implementation: PerceiverImpl,
     ) -> nn.Module:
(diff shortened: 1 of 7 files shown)


alxmrs:
@codex may I have a review?

jder:
I'm going to mark this as a draft until we get the previous PR in the stack landed.

jder:
@alxmrs can we close this in favor of #783, which removes "mix" training?

jder:
I think given we've removed mix mode we can close this.

PR #708, opened 2026-04-20
Created script to publicize raw & curated OM4 data.

### LINKED ISSUE #692: Move final version of the OM4 data to a public OSN pod location + document data access
The main README (and contributing guide) should have instructions for accessing / copying the data we use for experiments.

alxmrs:
Fixes #692.


alxmrs:

scripts/migrate_datasets_to_pubs.sh
```diff
@@ -0,0 +1,110 @@
...
+
+SOURCE_REMOTE="nyu-osn:emulators"
+DEST_REMOTE="nyu-osn:m2lines-pubs/FOMO"
+
+# Each entry is "<source-subpath>|<destination-subpath>", relative to the
+# remotes above. `rclone copy SRC DST` copies the *contents* of SRC into DST,
+# so point both sides at directories (or both at single objects).
+MIGRATIONS=(
+    "jbusecke/ocean_emulators/OM4/OM4_raw_test.zarr|raw/om4_5daily.zarr"
+    "am16581/ocean_static_no_mask_table.zarr|raw/ocean_static_no_mask_table.zarr"
```
alxmrs, line 16: We don't have a good name for this, but I don't want this to block the data transfer. We can always rename it later.

scripts/migrate_datasets_to_pubs.sh
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
+
+# Migrate FOMO datasets from nyu-osn:emulators (private) to nyu-osn:m2lines-pubs
+# (public read), rewriting the layout to the FOMO/ publication structure.
+
+set -euo pipefail
+
+SOURCE_REMOTE="nyu-osn:emulators"
+DEST_REMOTE="nyu-osn-public:m2lines-pubs/FOMO"
```
jder, line 9: I think this path probably depends on the naming decisions we're making right now, or if this is intended for the immediate publication, it should probably be under Samudra or Samudra2 or something like that.

scripts/migrate_datasets_to_pubs.sh
```diff
@@ -0,0 +1,110 @@
...
+
+echo "Source:      ${SOURCE_REMOTE}"
+echo "Destination: ${DEST_REMOTE}"
+echo "Transfers:   ${#MIGRATIONS[@]}"
+echo ""
+
+FAILED=()
+for entry in "${MIGRATIONS[@]}"; do
+    src_sub="${entry%%|*}"
+    dst_sub="${entry##*|}"
```
jder, line 87: This makes me want to use python instead of bash.

jder (commented):
Broadly this looks ok other than the naming thing; I can't review the data itself at the moment because I lost access to the bucket. I don't think we actually want to merge this PR, though, do you?

scripts/migrate_datasets_to_pubs.sh
alxmrs: I have commenced the move, but it should be easy to rename the parent directory once we decide on a durable name.

alxmrs:
> I don't think we actually want to merge this PR, though, do you?

Happy to omit merging the PR, but even having it is useful for review/tracing how did the transfer in the issue. Closing for now.

scripts/migrate_datasets_to_pubs.sh
jder: I don't think we can easily rename it -- I think that requires a full copy and delete.

alxmrs: I agree, sorry, when I say it's easy, I mean that the rclone command will be simple to write. I agree that it will take about a day to re-transfer.

PR #716, opened 2026-04-23
Asymmetric eval/inference support: coarse forcings can perform a high resolution rollout.

### LINKED ISSUE #615: Support mixed scale forcings for FOMO model

alxmrs:
Fixes #615, 4/4 in the stack.

### DIFF
--- a/configs/fomo_om4/eval_cross_resolution.yaml
+++ b/configs/fomo_om4/eval_cross_resolution.yaml
@@ -1,0 +1,56 @@
+# yaml-language-server: $schema=../schemas/EvalConfig.json
+#
+# Cross-resolution eval for the FOMO model (KR2, issue #615).
+#
+# The rollout starts at ¼° and consumes 1° boundary forcings at each step.
+# Prognostics stay at ¼° throughout the rollout — only the boundaries come
+# from a coarser source. Enabled by the dual-perceiver encoder (PR #702),
+# which produces the same latent grid for both streams when patch_extent is
+# in degrees.
+#
+# As in eval_multiscale.yaml, all three resolutions must appear in
+# `sources` so FOMOConfig.build computes `max_patch_size` correctly for the
+# Perceiver's Fourier frequency range. The first source is the one used for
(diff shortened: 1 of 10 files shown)


alxmrs:

alxmrs:
@jder, is this still helpful for our O1 goals?

jder:
@alxmrs I think yes, given the conversation today. @fomo-bot can you please generalize/extend this to training, too? We will probably need it there as well as at inference time.

PR #717, opened 2026-04-23
Alternative replica-balanced sampler

jder:
Alternative to #656: 

This updates our distributed equivalent batch samplers to make it that all n_workers / num_replicas have batches with the same scale of data. This is needed because in DDP training across n_workers, when we mix scales, GPUs which see smaller resolutions will be idle for long periods of time which can lead to NCCL timeouts and/or low GPU utilizations. In the future we might consider dynamically changing batch sizes or similar. 

This also fixes a small, but significant, bug where we didn't actually set epochs properly!


alxmrs (approved):
Thanks, this LGTM. I like the direction where we don't depend on the non-distributed sampler, since we reimplement much of the logic anyway. I also think the two types you introduce make what we do in the distributed sampler way more legible.

PR #722, opened 2026-04-28
Add PyPI-backed Samudra 2 Colab quickstart

YuanYuan98:
## Summary

Adds a browser-based Google Colab quickstart for **Samudra 2** that installs the latest stable `samudra` release from PyPI and trains from scratch on public 2° OM4 data using a free-tier CUDA runtime.

The notebook is an onboarding smoke test rather than a scientifically useful pretrained model. It requires no repository checkout, HPC access, cloud credentials, Weights & Biases account, or bulk data download.

## Changes

- Add generated `notebooks/quickstart.ipynb` plus its reviewable source in `scripts/build_quickstart_notebook.py`
- Use the config presets bundled in the PyPI wheel and stream the public anonymous 2° OM4 Zarr stores directly from S3
- Configure a five-level, one-step run with `batch_size=4`, explicit CUDA, and a bounded train/validation window
- Preserve structured S3 data, means, and standard-deviation locations across the Pydantic config customization round-trip
- Show concise Colab stages for preparation, training epoch, validation epoch, completion, and checkpoint output while retaining warnings and errors
- Compare normalized sea-surface-height ground truth, prediction, and `Bias (prediction − truth)`
- Add README and documentation entry points plus focused notebook-currentness and compilation tests

## Root cause addressed

The package's `Location` union currently serializes structured `S3Location` objects through the unresolved string-path variant during `model_dump`. Re-validating that dump loses the S3 endpoint, bucket, and `anon=True`, causing `UnresolvedLocation.open()` to fail. The notebook restores all three validated location objects before rebuilding the customized config.

## User impact

A new user can open the notebook in Colab, select a GPU, install Samudra from PyPI, verify the public data stream, train a small smoke-test model, inspect epoch stages, save a checkpoint, and compare one held-out prediction without local setup.

## Validation

- `.venv/bin/python -m pytest tests/test_quickstart_config.py -q` — **2 passed**
- `.venv/bin/python scripts/build_quickstart_notebook.py --check` — notebook current
- `git diff --check` — clean
- Executed the corrected configuration cell against merged `main`; verified `batch_size=4`, `backend="cuda"`, and intact anonymous S3 data/statistics locations
- Executed the data-check logic with a shape-accurate lazy 90×180, 23-timestep dataset stand-in
- Live Colab review confirmed anonymous data access and CUDA trainer initialization

A full production-quality model evaluation is intentionally out of scope; the notebook trains a smoke-test checkpoint from scratch.


YuanYuan98:
@codex, can you review?

notebooks/quickstart.ipynb
```diff
@@ -0,0 +1,347 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "install",
+   "metadata": {},
+   "source": [
+    "## 2. Slim install\n",
+    "\n",
+    "Ocean Emulator's full dependency set includes a few packages that are heavy or\n",
+    "brittle on Colab (`xesmf` needs the ESMF system library; `skypilot`,\n",
```
alxmrs: This is good for now, but in the future, I think we should clean up our dependencies in our pyproject.toml

notebooks/quickstart.ipynb
```diff
@@ -0,0 +1,347 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "download-data",
+   "metadata": {},
+   "source": [
+    "## 3. Download a data slice\n",
+    "\n",
+    "`scripts/clone_data.py` pulls from the public OSN bucket\n",
+    "`https://nyu1.osn.mghpcc.org/m2lines-pubs/Samudra/` — no credentials needed.\n",
+    "\n",
```
alxmrs, line 164: This is the wrong data source! I'm about to change, this. Let's make sure this is fixed before we land this PR!

notebooks/quickstart.ipynb
```diff
@@ -0,0 +1,347 @@
...
+    "        [\n",
+    "            \"python\",\n",
+    "            f\"{REPO_DIR}/scripts/clone_data.py\",\n",
+    "            str(DATA_DIR),\n",
+    "            \"--time_start\", \"0\",\n",
+    "            \"--time_end\",   \"290\",\n",
+    "            \"--write_time_chunks\", \"1\",\n",
+    "        ],\n",
+    "        env={**os.environ, \"PYTHONPATH\": f\"{REPO_DIR}/src\"},\n",
+    "        check=True,\n",
```
alxmrs: I would recommend against subprocess and instead use `!` at the front of the line of the colab notebook to run shell scripts.

notebooks/quickstart.ipynb
```diff
@@ -0,0 +1,347 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "train",
+   "metadata": {},
+   "source": [
+    "## 4. Train\n",
+    "\n",
+    "`configs/quickstart/train.yaml` is the same shape as the production v2 config\n",
```
alxmrs: Figuring out this config is the hard part; thanks for finding a good small model setup.

notebooks/quickstart.ipynb
```diff
@@ -0,0 +1,347 @@
...
+    "plt.tight_layout()\n",
+    "plt.show()\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "next-steps",
+   "metadata": {},
+   "source": [
+    "## Where to go next\n",
```
alxmrs: Very nice note.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,341 @@
...
+    print(f"VRAM: {torch.cuda.get_device_properties(0).total_memory / 1e9:.1f} GB")
+else:
+    raise RuntimeError(
+        "No GPU detected. Switch to a GPU runtime: "
+        "Runtime → Change runtime type → T4 GPU."
+    )
+
+# T4 (Turing, sm_75) does not support bfloat16; the quickstart config disables
+# it. A100/L4/H100 do support it — feel free to flip use_bfloat16 back on.
+print(subprocess.run(["nvidia-smi"], capture_output=True, text=True).stdout)
```
alxmrs: I recommend using the `!` commands in colab to run things on the shell.

alxmrs (commented):
I ran out of time yesterday to complete a full review. Here are my pending notes to minimize blocking you.

alxmrs:
I uploaded this into google colab and one error I hit so far is that we cannot clone this repo because it is private.

YuanYuan98:
@copilot , can i have your review?

docs/getting-started/quick-start.md
```diff
@@ -6,6 +6,15 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # Quick Start
 
+## Run in Google Colab
+
+The [Colab quickstart](https://colab.research.google.com/github/m2lines/Samudra/blob/main/notebooks/quickstart.ipynb)
+installs the latest stable Samudra release from PyPI and trains a Samudra 2
+model on public 2° OM4 data using a free-tier GPU runtime. It streams the data
+directly from S3, so it requires no local installation, bulk data download, HPC
+access, or cloud credentials. The notebook is an onboarding smoke test rather
```
alxmrs: Maybe this is naive of me, but I dream of external users being able to prototype real scientific ideas in a free tier colab -- then, they could be promoted to larger scale (1°, 1/2°, 1/4°) experiments if proven useful there.

Is that a far fetched vision? Is this only valuable as a smoke test?

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+OUT = REPO / "notebooks" / "quickstart.ipynb"
+
+
+INTRO_MD = """\
+# Samudra 2 — Colab Quickstart
+
+## Goal
+
+Train a Samudra 2 ocean emulator on public 2° OM4 data using a free-tier Google
+Colab GPU. This is an onboarding smoke test, not a production-quality or
```
alxmrs: See comment/discussion above.

alxmrs: Rich Turner claims that 1.5° resolution weather models are enough to rapidly test ideas before scaling them up to 0.25°.

alxmrs: Not sure if this extends to the ocean.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+scientifically useful model.
```
alxmrs: While it's useful to describe the limits of what we can know through a coarse resolution colab budget of experimenting, I think it's also wise to be open to the possibility that folks could iterate through useful ideas cheaply.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+
+No HPC system, local installation, cloud credentials, or Weights & Biases
+account is required. Samudra is installed from PyPI, and the data is streamed
+directly from the public S3 store instead of being copied into Colab.
+
+**Expected runtime:** depends on the Colab GPU and public S3 throughput. There
+is no multi-gigabyte download step; training reads only the requested samples.
```
alxmrs: As a reader, this is more revealing of the cluade conversation used to generate this text rather than informative to me about what the contents or purpose of the notebooks are. One idea: maybe instead of saying what this notebook isn't, we should state positively what it is for.

Claude has a habit of writing text like this by default.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+
+You will:
+
+1. Check the Colab runtime.
+2. Install the latest stable Samudra release from PyPI.
+3. Inspect the public 2° data stream and configure a compact run.
+4. Train a Samudra 2 model.
+5. Compare one prediction with ground truth.
```
alxmrs: Solid

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+)
+subprocess.run(
+    [sys.executable, "-m", "pip", "install", "--quiet", "--upgrade",
+     "--no-deps", "samudra"],
+    check=True,
+)
+subprocess.run(
+    [sys.executable, "-m", "pip", "install", "--quiet", *RUNTIME_DEPENDENCIES],
+    check=True,
+)
```
alxmrs, line 126: This is a bit weird for colab notebooks. I normally would expect something like: 

```
!pip install xarray zarr ...
```
at the top of a notebook. Further, aren't all of these deps available when we `pip install samudra`? What is the diff of deps?

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+import nbformat as nbf
+
+REPO = Path(__file__).resolve().parent.parent
+OUT = REPO / "notebooks" / "quickstart.ipynb"
+
+
+INTRO_MD = """\
+# Samudra 2 — Colab Quickstart
+
+## Goal
```
alxmrs, line 24: Instead of demoing a smoke test, here's an idea for what the colab could be about: What if we try to walk the user through answering a specific scientific question about the model, for example -- what dilation schedule is optimal for our UNet to make forecasts? -- and then use the training/eval/viz process in the package as tools to answer that question. I could see a lot of value structuring the notebook through inquiry or mystery. This might prime the user into thinking of other ways to improve the model. Further, in the preamble, we might want to link users to good first issues in our GH issue tracker to invite them to try to make further experiments.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+# the config so the public endpoint, bucket, and anonymous-access flag survive.
+for quickstart_source, base_source in zip(
+    quickstart["data"]["sources"], base_cfg.data.sources, strict=True
+):
+    for location_field in (
+        "data_location",
+        "data_means_location",
+        "data_stds_location",
+    ):
+        quickstart_source[location_field] = getattr(base_source, location_field)
```
alxmrs: I don't understand why this is necessary? And, why wouldn't we set this via the `from_yaml_and_cli()` arg of some kind?

alxmrs: Same for the arguments below, too.

alxmrs: Or via writing a specific YAML config locally and then reading it.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+
+# The Python model class keeps the historical name `Samudra`; the default
+# architecture selected here is Samudra 2.
+assert cfg.model.unet.ch_width == [280, 380, 480, 520]
+assert cfg.model.unet.core_block.upscale_factor == 2
+assert cfg.model.unet.up_sampling_block == "zonally_periodic_upsample"
+print("Default model: Samudra 2")
+"""
+
+DATA_CHECK_PY = """\
```
alxmrs, line 318: For this block, it might be easier to simply print the YAML config -- or write a new one with a lot of comments that explains what each statement means.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+logical_bytes = sum(window[name].nbytes for name in expected_variables)
+probe = window["zos"].isel(time=0, **{lat_dim: 0, lon_dim: 0}).load()
+
+print(f"Streaming from: {source.data_location}")
+print(f"Grid: {ds.sizes[lat_dim]} × {ds.sizes[lon_dim]} (2°)")
+print(f"Quickstart window: {window.time.values[0]} → {window.time.values[-1]}")
+print(f"Selected model fields: {logical_bytes / 2**20:.1f} MiB before compression")
+print(f"Connectivity probe (zos): {probe.item()}")
+print("No local data copy was created.")
+ds.close()
```
alxmrs: I don't understand that well about utility or explainability of these specific traces. I also don't understand practically how the block above helps be better configure experiments -- new hparams, data sources, etc.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+TRAIN_MD = """\
+### 4. Train a Samudra 2 model
+
+The bundled `samudra_om4` preset supplies the production Samudra 2 channel
+widths, ConvNeXt expansion factor, zonally periodic upsampling, and dynamic
+variance-weighted loss. Training and validation still go through the normal
+`Trainer`, including normalization, masking, EMA, metrics, and checkpointing.
+
+The cell keeps Colab output concise: it prints when training and validation
+start for each epoch, while warnings and errors remain visible.
```
alxmrs: Let's show, not tell.

scripts/build_quickstart_notebook.py
alxmrs: We primarily want users to configure our training program from YAML, and secondarily from CLI args. We should demo these and not the workaround to make the edits in Python (not intended behavior).

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+
+samudra_logger = logging.getLogger("samudra")
+for handler in tuple(samudra_logger.handlers):
+    if handler.get_name() == "samudra-colab-progress":
+        samudra_logger.removeHandler(handler)
+        handler.close()
+samudra_logger.setLevel(logging.WARNING)
+
+
+class ColabTrainer(Trainer):
```
alxmrs, line 384: I don't mind this logging trainer, it's nice to follow. Might be a good follow up issue to use https://github.com/tqdm/tqdm.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+
+### 5. Compare a prediction with ground truth
+
+The final cell uses the last-epoch model on one held-out validation batch and
+plots normalized sea surface height (`zos`). It is deliberately small; the
+regular evaluation CLI runs long autoregressive rollouts and writes full
+prediction datasets.
+"""
+
+PREDICTION_PY = """\
```
alxmrs, line 451: I really like these plots.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,413 @@
...
+prognostic_names = cfg.data.sources[0].dataset_spec.prognostic_var_names
+zos_index = prognostic_names.index("zos")
+predicted_zos = prediction[0, zos_index].cpu().numpy()
+target_zos = target[0, zos_index].cpu().numpy()
+
+fig, axes = plt.subplots(1, 3, figsize=(14, 3.5))
+for ax, field, title in zip(
+    axes,
+    [target_zos, predicted_zos, predicted_zos - target_zos],
+    ["Ground truth", "Prediction", "Bias (prediction − truth)"],
```
alxmrs: A lot of useful information is captured in these two lines -- maybe there is a more legible way to present this.

alxmrs (commented):
Review of the notebook itself. At the top of the notebook, I have a new idea of what it should be about rather than a smoke test.

notebooks/quickstart.ipynb
YuanYuan98: Solved

notebooks/quickstart.ipynb
YuanYuan98: Solved

scripts/build_quickstart_notebook.py
YuanYuan98: Solved

scripts/build_quickstart_notebook.py
YuanYuan98: Solved

scripts/build_quickstart_notebook.py
YuanYuan98: Solved

scripts/build_quickstart_notebook.py
YuanYuan98: Solved

scripts/build_quickstart_notebook.py
YuanYuan98: Solved

scripts/build_quickstart_notebook.py
YuanYuan98: Solved

scripts/build_quickstart_notebook.py
YuanYuan98: Solved

scripts/build_quickstart_notebook.py
YuanYuan98: Solved

scripts/build_quickstart_notebook.py
YuanYuan98: Solved

notebooks/quickstart.ipynb
YuanYuan98: Yeah, totally agree. Thanks for flagging this. I kept the workaround local to the notebook and added a note explaining why it’s needed. Separating the core runtime dependencies from the HPC and infrastructure extras in `pyproject.toml` would be a really useful follow-up, since it would improve installs beyond just this Colab.

docs/getting-started/quick-start.md
YuanYuan98: I really like that vision, and I don’t think it’s far-fetched at all. Thanks for encouraging me to think about the notebook that way. I updated the quickstart so it’s no longer framed as just a smoke test. Users get editable YAML, a checkpoint, and clear ways to extend the run. For this first notebook, I kept the one-epoch run focused on learning the full workflow without implying that it’s already enough for a scientific conclusion.

scripts/build_quickstart_notebook.py
YuanYuan98: Thanks, I really like this idea too. I kept the parts that make the notebook easy to extend. The YAML is visible and editable, and there’s a link to good first issues. For this quickstart, I decided not to make dilation optimization the main question because the current one-epoch, one-seed budget wouldn’t give it a very meaningful answer. I’d love to explore that as a separate experiment notebook with a few configurations or seeds and a clear evaluation metric.

scripts/build_quickstart_notebook.py
YuanYuan98: Yes, completely agree. Thanks for calling this out. I removed the wording that made useful experimentation sound out of reach, and now frame the notebook as a low-cost, editable starting point for trying ideas. I’m just being careful not to overstate what we can learn from one epoch and one seed without a more controlled comparison.

scripts/build_quickstart_notebook.py
alxmrs: Thanks, this experiment decision SGTM. I appreciate your adjustments.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+
+Samudra installs from PyPI and streams the selected samples directly from
+public S3 storage. The workflow runs entirely in Colab, with experiment
+controls visible in YAML and local checkpoint output ready for follow-up work.
+
+**Expected runtime:** depends on the Colab GPU and public S3 throughput.
+Training and validation use short, explicit date windows and stream only the
+requested chunks.
+
+You will:
```
alxmrs: 🐑 I like explaining what will be done in this colab. One idea to add after or before this section: What are the desired learning outcomes from walking through this notebook? It could be a nice list like this.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+INSTALL_PY = """\
+!pip install --quiet --upgrade --no-warn-conflicts --progress-bar off \\
+    "numpy>=1.26.4,<2" "matplotlib>=3.10.1" \\
+    "cftime>=1.6.4.post1" "dacite>=1.9.2" "dask>=2025.2,<2026" "einops>=0.7" \\
+    "jaxtyping>=0.3" "microsoft-aurora>=1.8" "perceiver-pytorch>=0.8.8" \\
+    "pydantic-settings>=2.8.1" "pyyaml>=6.0.2" "s3fs==2025.5.1" \\
+    "torchinfo>=1.8" "tqdm>=4.67.1" "typing-extensions>=4.15" \\
+    "wandb>=0.19.8" "xarray>=2025.1.2" \\
+    "xarray-einstats>=0.8" "zarr<3"
+!pip install --quiet --upgrade --no-deps --no-warn-conflicts \\
```
alxmrs, line 109: Do we want to install this with the CUDA extra?

alxmrs: If we did the cuda extra, would it solve the problem and allow us to use one command?

alxmrs: I see; we want to use the torch versions already optimized for colab. I'm happy with that explanation; I could go either way.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+
+data:
+  hist: 0
+  static_data_vars: []
+  concurrent_compute: true
+  loading:
+    type: cpu
+    num_workers: 0
+    persistent_workers: false
+  sources:
```
alxmrs, line 212: Maybe you could ask your agent (not sure if this is possible), but can we link to the `data/om4_demo` data source here via the config distributed in the package (maybe, via an ! or @ import)? It would be my preference if we could, since it would be a more succinct example. 

If we can't do this here, we probably could by passing the path via the CLI.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+test_using_ema: false
+loss:
+  type: dynamic
+  metric: mse
+finetune: false
+resume_ckpt_path: null
+inference_epochs: []
+data_stride: [1]
+steps: [1]
+step_transition: []
```
alxmrs, line 191: I think it might be helpful to have comments explaining what these three args are (for the training curriculum for the rollout. -- this could be a good experiment in its own right).

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+
+
+QUICKSTART_YAML = """\
+# A compact Samudra 2 training run for a free-tier Colab GPU.
+# Every training, data, and model control is visible and editable.
+debug: false
+disk_mode: true
+pin_mem: true
+save_freq: 1
+epochs: 1
```
alxmrs, line 171: Maybe invite users to change this number to 70 to perform experiments more typical to Samudra training; 1 epoch is for instructional purposes.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+    ch_width: [280, 380, 480, 520]
+    n_layers: [1, 1, 1, 1]
+    core_block:
+      block_type: conv_next_block
+      kernel_size: 3
+      activation: capped_gelu
+      upscale_factor: 2
+      norm: batch
+    down_sampling_block: avg_pool
+    up_sampling_block: zonally_periodic_upsample
```
alxmrs, line 270: It would be nice to add comments to explain how these knobs effect our UNet.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+  corrector:
+    non_negative_corrector_names: null
+    ocean_heat_corrector: false
```
alxmrs: PSA: this is going away or just got deleted (thankfully).

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+  name: samudra_quickstart
+  rand_seed: 15
+  base_output_dir: /content/samudra_outputs
+  wandb:
+    mode: disabled
+    project: samudra_quickstart
+
+data:
+  hist: 0
+  static_data_vars: []
```
alxmrs: This was just removed or is about to be removed.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+experiment:
+  name: samudra_quickstart
+  rand_seed: 15
+  base_output_dir: /content/samudra_outputs
+  wandb:
+    mode: disabled
+    project: samudra_quickstart
+
+data:
+  hist: 0
```
alxmrs, line 206: I think it's important to comment and explain what hist means.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
```
alxmrs, line 1: Pro-tip: marimo might provide a nice way to generate notebooks in code link this, https://marimo.io/

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+CONFIG_LOAD_PY = """\
+from pathlib import Path
+
+from IPython.display import Markdown, display
+from samudra.config import TrainConfig
+
+CONFIG_PATH = Path("/content/samudra_quickstart.yaml")
+cfg = TrainConfig.from_yaml_and_cli([str(CONFIG_PATH)])
+
+display(Markdown(
```
alxmrs, line 291: Nice.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+"""
+
+TRAIN_MD = """\
+### 5. Train and validate Samudra 2
+
+The normal `Trainer` handles normalization, masking, dynamic
+variance-weighted loss, validation, and checkpointing. The small wrapper below
+adds one `tqdm` bar for the training batches and another for the validation
+batches. Each completed bar shows its final loss and elapsed time while keeping
+the output focused on progress, warnings, and errors.
```
alxmrs, line 362: This is very helpful.

scripts/build_quickstart_notebook.py
```diff
@@ -0,0 +1,581 @@
...
+        progress = tqdm(
+            original_loader,
+            total=len(original_loader),
+            desc=f"Training epoch {epoch}/{self.epochs}",
+            unit="batch",
+            dynamic_ncols=True,
+            leave=True,
+        )
+        self.train_loader = progress
+        try:
```
alxmrs, line 387: I think the most typical way to use tqdm is with a `with` expression.

alxmrs (commented):
Looking really great so far! Another round of notes to really make it pop. I think you'll like marimo :)

scripts/build_quickstart_notebook.py
YuanYuan98: Good idea, thanks! I added brief comments for these rollout settings.

scripts/build_quickstart_notebook.py
YuanYuan98: Thanks! The bundled config is too large for this quickstart, so I linked to it and kept the shorter config explicit.

scripts/build_quickstart_notebook.py
YuanYuan98: Exactly, thanks! I kept Colab’s CUDA-enabled PyTorch so we do not replace its working stack.

scripts/build_quickstart_notebook.py
YuanYuan98: Thanks, Alex! I updated this section to focus on the learning outcomes.

scripts/build_quickstart_notebook.py
YuanYuan98: Thanks! I added brief comments for the main U-Net settings.

scripts/build_quickstart_notebook.py
YuanYuan98: Thanks! I clarified that one epoch is for the quickstart and suggested 70 epochs for longer experiments.

scripts/build_quickstart_notebook.py
YuanYuan98: Thanks for the heads-up! I removed it.

scripts/build_quickstart_notebook.py
YuanYuan98: Agreed, thanks! I added a short explanation of `hist`.

scripts/build_quickstart_notebook.py
YuanYuan98: Thanks for the heads-up! I removed it.

scripts/build_quickstart_notebook.py
YuanYuan98: Thanks for the tip! marimo looks great. I will keep it as a follow-up proof of concept for now.

scripts/build_quickstart_notebook.py
YuanYuan98: Thanks! I updated both progress bars to use `with tqdm(...)`.

README.md
```diff
@@ -6,13 +6,22 @@ SPDX-License-Identifier: CC-BY-4.0
...
 
 Samudra is a global ocean emulator described in ["Samudra: An AI Global Ocean Emulator for Climate"](https://arxiv.org/abs/2412.03795) and updated in ["Samudra 2: Scaling Ocean Emulators across Resolutions"](https://m2lines.github.io/Samudra/samudra2/). Samudra efficiently emulates the ocean component of a state-of-the-art climate model, accurately reproducing key ocean variables including sea surface height, horizontal velocities, temperature, and salinity, across their full depth.
 
+To configure, train, validate, and visualize Samudra 2 in a browser on public
+data with a Colab GPU, open the
+[free-tier Colab quickstart](https://colab.research.google.com/github/m2lines/Samudra/blob/main/notebooks/quickstart.ipynb).
+It runs from an editable YAML configuration, streams public 2° OM4 data, shows
+live batch progress, saves a checkpoint, and visualizes one held-out
+sea-surface-height prediction.
+
```
alxmrs: optional nit: I think these paragraphs might be better fit in the `how to run` section below, WDYT?

alxmrs: I'm also happy with having this here.

README.md
```diff
@@ -6,13 +6,22 @@ SPDX-License-Identifier: CC-BY-4.0
...
 [Benchmark Results](https://m2lines.github.io/Samudra/dev/bench/)
 
 Samudra is a global ocean emulator described in ["Samudra: An AI Global Ocean Emulator for Climate"](https://arxiv.org/abs/2412.03795) and updated in ["Samudra 2: Scaling Ocean Emulators across Resolutions"](https://m2lines.github.io/Samudra/samudra2/). Samudra efficiently emulates the ocean component of a state-of-the-art climate model, accurately reproducing key ocean variables including sea surface height, horizontal velocities, temperature, and salinity, across their full depth.
 
+To configure, train, validate, and visualize Samudra 2 in a browser on public
+data with a Colab GPU, open the
+[free-tier Colab quickstart](https://colab.research.google.com/github/m2lines/Samudra/blob/main/notebooks/quickstart.ipynb).
+It runs from an editable YAML configuration, streams public 2° OM4 data, shows
+live batch progress, saves a checkpoint, and visualizes one held-out
+sea-surface-height prediction.
```
alxmrs: I think we can omit this paragraph:
```suggestion
```

alxmrs: I could take it or leave it.

alxmrs (approved):
Great work, Yuan! Thank you.

README.md
YuanYuan98: Agreed!

README.md
YuanYuan98: Good idea, thanks! I moved the Colab link under `How to run`.

PR #705, opened 2026-04-15
Add MkDocs Material developer documentation (#695)

YuanYuan98:
Set up MkDocs with the Material theme for developer documentation. The docs source lives in `docs-src/`, builds to `docs/api/`, and is hosted alongside the existing Samudra 2 paper page on GitHub Pages.

  ## Changes
  - `mkdocs.yml` — config with Material theme, mkdocstrings for auto-generated API reference
  - `docs-src/` — Markdown source (Getting Started, Architecture, API Reference for all models including Samudra, Samudra 2, and FOMO)
  - `.github/workflows/pages.yml` — updated to build MkDocs on deploy
  - `docs/index.html` — added API Docs card on landing page
  - `pyproject.toml` — added `docs` dependency group (mkdocs, mkdocs-material, mkdocstrings)

  ## Preview locally
  ```bash
  uv sync --group docs
  uv run mkdocs serve

  Note

  Build artifacts (docs/api/) are temporarily committed so reviewers can preview without building locally. Once CI is stable, these can be moved to .gitignore and built only in CI.

  Built with the help of Claude Code.
  ```

PR #725, opened 2026-04-30
Restore benchmark dashboard publishing to GitHub Pages

### LINKED ISSUE #723: Fix Performance Benchmark Github Pages Deployment
See: https://openathena.ai/Ocean_Emulator/dev/bench/, which produces a 404 currently. 

Now that we use a different way to build our pages site, we need to update our automated benchmark publishing system.

See: 
- https://github.com/Open-Athena/Ocean_Emulator/blob/main/.github/workflows/benchmarks.yml
- https://github.com/Open-Athena/Ocean_Emulator/blob/main/.github/workflows/pages.yml

YuanYuan98:
## Summary
- The pytest-benchmark dashboard at `/dev/bench/` has been 404ing since the Pages site moved to the `actions/deploy-pages` flow (#705/#721). `benchmarks.yml` keeps pushing results to the `gh-pages` branch via `benchmark-action/github-action-benchmark`, but `pages.yml` now publishes a Zensical-built artifact and never reads from `gh-pages`.
- Bridge the gap in `pages.yml`: check out `gh-pages` into a sibling path and copy `dev/bench/` into the `site/` artifact before upload.
- Add a `workflow_run` trigger so docs redeploys after each successful `Benchmark` run, with a conclusion guard so failed runs don't kick off pointless rebuilds.
- `benchmarks.yml` is unchanged — `gh-pages` remains the historical data store, which `github-action-benchmark` requires for its append-history behavior.

Closes #723

## Test plan
`workflow_run` triggers only fire from the default branch, so full end-to-end verification has to happen post-merge. After merge:
- [ ] Manually `workflow_dispatch` the `docs` workflow on `main`. Confirm `https://openathena.ai/Ocean_Emulator/dev/bench/` resolves and renders the existing benchmark history (data already in `gh-pages` as of `69e1e688`).
- [ ] On the next `Benchmark` run completion, confirm `docs` auto-runs via `workflow_run` and the dashboard reflects the new commit's data.
- [ ] Verify a failed `Benchmark` run does NOT trigger a `docs` rebuild (the `if: conclusion == 'success'` guard).

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -13,6 +13,8 @@ on:
     workflows: ["Benchmark"]
     types:
       - completed
+    branches:
+      - main
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -9,6 +9,10 @@ on:
       - "mkdocs.yml"
       - "src/**"
       - ".github/workflows/pages.yml"
+  workflow_run:
+    workflows: ["Benchmark"]
+    types:
+      - completed
   workflow_dispatch:
 
 permissions:
@@ -21,6 +25,7 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success' }}
     permissions:
       contents: read
       pages: write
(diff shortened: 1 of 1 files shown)


YuanYuan98:
Addressed in 689ac5d9: added `branches: [main]` to the `workflow_run` trigger so only canonical Benchmark runs (push to main, or workflow_dispatch on main) re-publish `/dev/bench/`. Feature-branch benchmark dispatches no longer trigger a docs deploy.

Note: the underlying `benchmark-action/github-action-benchmark` step in `benchmarks.yml` still pushes to `gh-pages` from any source branch with `auto-push: true` — that's a separate concern out of scope for #723. This PR just stops the public dashboard from auto-republishing on those runs.

.github/workflows/pages.yml
```diff
@@ -38,6 +45,22 @@ jobs:
...
         run: cp -R docs/static/. site/
 
+      # Benchmark results are pushed to the gh-pages branch by benchmarks.yml
+      # via benchmark-action/github-action-benchmark. We bring those files into
+      # the artifact so they're served at /dev/bench/ alongside the docs site.
+      - name: Check out benchmark data from gh-pages
+        uses: actions/checkout@v5
+        with:
+          ref: gh-pages
+          path: gh-pages-data
```
alxmrs, line 55: Nice, good idea to check out the branch to a local dir.

.github/workflows/pages.yml
```diff
@@ -38,6 +45,22 @@ jobs:
...
+
+      - name: Copy benchmark dashboard into build output
+        run: |
+          if [ -d gh-pages-data/dev/bench ]; then
+            mkdir -p site/dev/bench
+            cp -R gh-pages-data/dev/bench/. site/dev/bench/
+          fi
```
alxmrs, line 62: Solid.

alxmrs (approved):
This is an elegant fix to keep the gh-pages branch but merge it in to our pages deploy workflow. The path to build the website is easy enough to understand. Thanks, LGTM.

PR #732, opened 2026-05-04
U/alxmrs/experiments/kr1 v2

alxmrs:


alxmrs:
Looking at wandb, it seems like there are serious bugs in either the rollout or the model: 
Screenshot 2026-05-04 at 11 52 56 AM

Screenshot 2026-05-04 at 12 05 37 PM

Screenshot 2026-05-04 at 12 07 05 PM

The second image indicates that it is a bug in the rollout specifically. Though, I don't know how `_gen` plot are calculated. If I understand the third image (the _target),  maybe the model is fine but the rollout is busted.

jder:
@alxmrs I believe "_target" graphs are just "the gold data we are expecting to predict". The "_gen" plots are the (autoregressive) model outputs

PR #733, opened 2026-05-06
PCGB

alxmrs:


alxmrs:
Closing this experiment, but keeping the branch and idea around.

Notes from talking to Zack at OA: maybe a different intervention than controlled dropout, like a structured (batch) norm, could be useful to "up regulate or down regulate" (my words) subnetworks in the ensemble.

PR #736, opened 2026-05-07
Fix broken link to gif

alxmrs:
We put things under a "static/" folder but didn't update the README.


YuanYuan98 (approved):
Thanks! It should work now.

PR #741, opened 2026-05-12
Consistent 2d Fourier features for each (encoder) Perceiver impl.

alxmrs:
During trial and error experiments with FOMO, Claude and I detected a possible source of error in the initial runs where the production Perceiver used in the encoder has sub-par positional encodings. Specifically, our Flash perceiver impl doesn't encode the 2d patch information in the same way as the naive Perceiver impl. This PR augments the input of that Perceiver, borrowing the Fourier encoding scheme from our "naive" impl library (a Lucid Rains dependency). 

CC: @jder - I wonder if this affects your recent Fomini experiments at all. I'm not sure if I should apply this to the PerceiverIO/decoder module, too.

### DIFF
--- a/src/ocean_emulators/config.py
+++ b/src/ocean_emulators/config.py
@@ -459,8 +459,6 @@ class PerceiverConfig(BaseConfig):
         # This is not really a "frequency" but a maximum of the width appears to be reasonable from looking at the code.
         max_freq = max(*max_patch_size)
 
-        # Match the naive perceiver's `num_freq_bands=4`; both paths therefore
-        # encode intra-patch position with the same Fourier-feature layout.
         num_freq_bands = 4
         if _use_flash(implementation):
             try:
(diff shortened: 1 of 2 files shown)


alxmrs:

src/ocean_emulators/config.py
```diff
@@ -455,22 +459,30 @@ def build(
         # This is not really a "frequency" but a maximum of the width appears to be reasonable from looking at the code.
         max_freq = max(*max_patch_size)
 
+        # Match the naive perceiver's `num_freq_bands=4`; both paths therefore
+        # encode intra-patch position with the same Fourier-feature layout.
```
```suggestion
```

jder (approved):
Thanks! Good catch. (I again wish we just had 1 implementation here…)

I do think this probably hurt fomini a lot… maybe I will try rerunning my latest experiment there and see if it's dramatically different.

PR #754, opened 2026-06-02
Updating the format of our rollout (`eval.py`) to make it easier to calculate metrics (true to the original data).

### LINKED ISSUE #508: Improve Usability of Eval Outputs
Currently the outputs generated from the eval script create unnecessary friction when performing analysis, and may lead to unintended errors. I am including an example notebook below, which highlights several ways in which the analysis process is hindered.

1) The data is separated by level for ease of ML training, however, given that the eval outputs are not used in this context, we introduce the step of concatenating variables back to their respective depths, while this is not a big lift, and mistakes in doing this could an additional place for errors. 

2) The horizontal coordinates of the eval outputs do not align with the true data (['lat','lon'] vs ['y','x']) this is unnecessary and requires manually renaming axes

3) We are missing most of the grid information necessary for performing calculations. At the moment, I address this by recomputing them from the same grid files used in processing the data; however, for future proofing, I think that this grid information should either live in the eval datasets, or we should create a grid zarr for each resolution and store a path to that grid in the attrs. These differences, have an impact on conservation as slight changes to depth, area, or lat/lon locations can make a difference

[Sample_Eval_Notebook_OM4.ipynb](https://github.com/user-attachments/files/24037380/Sample_Eval_Notebook_OM4.ipynb)

alxmrs:
Fixes #508. This makes the three changes in this issue by correcting the underlying data (adding essential coordinate information) and propagating that data during eval time. Since this is a data fix along with an eval output fix (IMO, this was the simplest approach), I provide two data engineering corrections: 
1. I updated the original data engineering script to no longer omit the metadata that we need for evaluation. This way, we won't make the same mistake for future datasets.
2. I've added a script to update the existing datasets in place (they should be safe since we only add new coordinate information; we don't have to re-write any of the existing chunks). 

My plan is that after this PR is reviewed (and merged), I'll run the coordinate update script on the public FOMO data. For good measure, I've run the backup script on this data so we can inspect it: `s3://emulators/am16581/data/2025-11/om4_onedeg_v3/OM4.zarr`.


data/ocean_preprocessing/areacello_xesmf_check.py
```diff
@@ -0,0 +1,111 @@
+# SPDX-FileCopyrightText: 2026 Ocean Emulator Authors
```
alxmrs: This is a one-off script to make an estimation for a decision. I'm open to deleting it before final merge.

alxmrs:

alxmrs:
Thanks, that was a genuine bug. I pushed a fix. @codex, may I have another round of review?

data/ocean_preprocessing/backfill_grid_coords.py
```diff
@@ -0,0 +1,368 @@
...
+        1050,
+        1400,
+        1850,
+        2400,
+        3100,
+        4000,
+        5000,
+        6000,
+    ],
+    dtype="float64",
```
jder: Is there somewhere we could read these from?

src/samudra/utils/data.py
```diff
@@ -787,15 +787,45 @@ def with_level_index_vars(
...
+    """
+    data_copy = data.copy()
+    n_depths = len(dataset_spec.depth_levels)
+    mask_vars = set(dataset_spec.mask_vars)
+    for var in data.variables:
+        var_str = str(var)
+        if var_str in mask_vars:
+            continue
+        base, _, idx = var_str.rpartition("_")
+        if base and idx.isdigit() and int(idx) < n_depths:
```
jder: Instead of testing this heuristically, can we read the known set of variable names off the DatasetSpec?

src/samudra/utils/writer.py
```diff
@@ -94,3 +101,41 @@ def write(self):
         # Reset
         self.buffer = None
         self.time_buffer = None
+
+    def _output_coords(self) -> dict:
+        """Build CF-aligned output coordinates matching the ground-truth layout.
```
jder: What is "CF-aligned"? What is "ground-truth layout"?

src/samudra/utils/writer.py
```diff
@@ -94,3 +101,41 @@ def write(self):
...
+        x_vals = np.asarray(src["lon"].values)  # longitudes
+        ny, nx = y_vals.size, x_vals.size
+
+        spec = self.tensor_map.dataset_spec
+        n_levels = spec.num_prognostic_depth_levels
+        coords: dict[str, tuple] = {
+            "y": ("y", y_vals),
+            "x": ("x", x_vals),
+            "lat": (("y", "x"), np.broadcast_to(y_vals[:, None], (ny, nx)).copy()),
+            "lon": (("y", "x"), np.broadcast_to(x_vals[None, :], (ny, nx)).copy()),
```
jder: How about keeping the old lat/lon around as lat_2d lon_2d or something so that we don't need to recompute them here and potentially mess something up?

tests/test_writer.py
```diff
@@ -0,0 +1,187 @@
...
+from ocean_emulators.utils.writer import ZarrWriter
+from tests.conftest import TEST_FULL_DATASET_SPEC
+
+# write() never touches normalization (record_batch does), so these tests drive
+# the writer directly with a buffer and omit a real Normalize.
+_NO_NORMALIZE = cast(Normalize, None)
+
+
+def _source_coords(ny, nx):
+    """Coords as `get_coords_dict` returns them after backfill: 1D lat/lon dims,
```
jder: What does this have to do with backfill?

jder (changes requested):
Would be great to ask @adam-subel to look at some outputs here since this is a lot of new code and I'm finding it a bit hard to follow exactly how data is flowing & being transformed now.

I would rather re-run the full pipeline rather than using/keeping the backfill script + tests to avoid any drift. What datasets would you want to backfill? Can we reprocess them instead? 

More fuzzily, I am somewhat worried about the proliferation of names, magic numbers, and name patterns like foo_i in our code and tests especially given we are starting to use other datasets which presumably will not match. WDYT?

src/samudra/utils/writer.py
alxmrs: CF is the "climate and forest" metadata standards. They are the conventions that are defaults in Xarray, but only relevant to some Xarray-openable datasets, like ours. 

Let me get back to you on the other question with a closer look. 

https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html

src/samudra/utils/data.py
```diff
@@ -792,16 +792,44 @@ def with_level_index_vars(
...
     if "lat" not in data_copy.dims:
-        # Drop unnecessary coordinates and rename dimensions
-        data_copy = data_copy.drop_vars(
-            ["lat", "lon", "lat_b", "lon_b", "dayofyear"], errors="ignore"
-        ).rename({"x": "lon", "y": "lat"})
+        # Preserve the 2-D geographic coords under a non-colliding name, then
+        # rename the x/y dims to the 1-D lat/lon we standardize on.
+        preserve = {n: f"{n}_2d" for n in ("lat", "lon") if n in data_copy.coords}
+        data_copy = (
+            data_copy.drop_vars(["dayofyear"], errors="ignore")
```
jder: 🐑 maybe remove the drop_vars here? seems like a bug it was here in the first place

src/samudra/utils/writer.py
```diff
@@ -94,3 +101,57 @@ def write(self):
...
+        # correct on any grid; only broadcast the 1-D axes for legacy regular-grid
+        # data that did not carry them. Broadcasting fabricates wrong geometry on a
+        # curvilinear (e.g. tripolar) grid, so it is a fallback, not the default.
+        if "lat_2d" in src and "lon_2d" in src:
+            lat2d = np.asarray(src["lat_2d"].values)
+            lon2d = np.asarray(src["lon_2d"].values)
+        else:
+            lat2d = np.broadcast_to(y_vals[:, None], (ny, nx)).copy()
+            lon2d = np.broadcast_to(x_vals[None, :], (ny, nx)).copy()
+
```
jder, line 143: Seems like it might be better to omit the (incorrect) values here rather than make them up?

jder (approved):
Looks great, thanks for this! Can you update the PR description, too?

src/samudra/utils/data.py
adam-subel: This would be addressing an earlier bug, that should not be a variable in any dataset we use

src/samudra/utils/writer.py
adam-subel: For the gaussian grid, you can actually broadcast without any real issue, but this would cause some larger problems for the tripolar data. Not sure if storing the grid type as a dataset attribute would be worth carrying around as a check for places like this in the code

data/ocean_preprocessing/preprocessing.py
```diff
@@ -388,7 +388,29 @@ def flatten_by_depth_level(ds: xr.Dataset) -> xr.Dataset:
...
-    ds = ds.drop_vars(vars_3d).drop_dims("lev").reset_coords(drop=True)
+    ds = ds.drop_vars(vars_3d)
+
+    # Keep the grid-metadata coordinates needed for downstream physical analysis
+    # (areacello, ocean_fraction, cell bounds, depth axis). Previously these were
+    # discarded here via `.drop_dims("lev").reset_coords(drop=True)`, which forced
+    # eval/analysis to recompute them from external grid files. The stacked
+    # `wetmask` coord has just been flattened into the per-level `mask_i` vars, so
+    # it (and native-grid-only coords like `angle`/`ilev`) are dropped instead.
+    keep_coords = {
```
adam-subel, line 399: One detail, I am not sure we have access to at the moment, is the addition of explicit dx and dy values. Especially if we move to tripolar, this should not be estimated as simply shifting the coordinates. For any cross section this would be relevant (including looking at AMOC). For the gaussian grid, we can compute this using lon_b, lat_b and the relationship dy = R * (latitude_N-latitude_S) abd dx \approx R * (lon_E-lon_w)*cos(latitude). But for future proofing with the tripolar, we want to retain the grid metrics

adam-subel (commented):
I think this looks good overall, you have enough tests that I trust we are retaining what you intend and this will make eval much easier. 

I left a few notes on places where behavior may change between the gaussian and the tripolar data in the case a grid metric is missing. I think it is worth thinking about tracking the grid type through some of this as we will have different behavior for different grids. I think there it is worth considering supporting a couple cases, especially if the goal is to work with more datasets down the line, which have some subtleties with geometry that you may want to treat differently down the line; however, best practice would be to compute all grid metrics exactly when processing the data and just pass them along from the start.

alxmrs:
@code may I have a final review?

jder:
@codex review please

PR #757, opened 2026-06-05
Retry the Zarr write to survive transient blosc -1 read errors

alxmrs:
Reading blosc-compressed source chunks over S3 under heavy concurrency occasionally returns a truncated buffer, which surfaces as an intermittent `RuntimeError: error during blosc decompression: -1` (https://github.com/zarr-developers/numcodecs/issues/810) and kills the whole job near completion. The failure is transient -- re-fetching the chunk almost always succeeds.

Drive the final to_zarr with `client.compute(..., retries=write_retries)` on a distributed cluster (default 5) so a failed chunk task is re-run instead of aborting the job. Falls back to a plain compute when running without a cluster. After the retry budget is exhausted it still fails loudly.

🤖


alxmrs:

jder (approved):

PR #763, opened 2026-06-17
Update readme and contributing for public release

jder:
To make a bit more of a landing page. I removed contributing.md in the root since clicking it in the github UI is unfriendly when it is a symlink, and moving it out of docs raises various questions about how to keep links working.


README.md
```diff
@@ -11,11 +11,12 @@ SPDX-License-Identifier: CC-BY-4.0
 [![Benchmark CPU](https://github.com/m2lines/Samudra/actions/workflows/benchmarks.yml/badge.svg)](https://github.com/m2lines/Samudra/actions/workflows/benchmarks.yml)
 [Benchmark Results](https://m2lines.github.io/Samudra/dev/bench/)
 
+Samudra is a global ocean emulator described in ["Samudra: An AI Global Ocean Emulator for Climate"](https://arxiv.org/abs/2412.03795) and updated in ["Samudra 2: Scaling Ocean Emulators across Resolutions"](https://m2lines.github.io/Samudra/samudra2/). Samudra efficiently emulates the ocean component of a state-of-the-art climate model, accurately reproducing key ocean variables including sea surface height, horizontal velocities, temperature, and salinity, across their full depth.
+
+We are actively and openly developing this emulator to support new tasks and data sources with the goal of building a broadly useful foundation model for ocean and climate. Please see [our docs](https://m2lines.github.io/Samudra/docs/) for more or [our contributing guide](https://m2lines.github.io/Samudra/docs/contributing/) to join in!
```
alxmrs, line 16: +1 to pointing folks to our contributing webpage.

alxmrs: I also like the call to action about open development.

README.md
```diff
@@ -11,11 +11,12 @@ SPDX-License-Identifier: CC-BY-4.0
 [![Benchmark CPU](https://github.com/m2lines/Samudra/actions/workflows/benchmarks.yml/badge.svg)](https://github.com/m2lines/Samudra/actions/workflows/benchmarks.yml)
 [Benchmark Results](https://m2lines.github.io/Samudra/dev/bench/)
 
+Samudra is a global ocean emulator described in ["Samudra: An AI Global Ocean Emulator for Climate"](https://arxiv.org/abs/2412.03795) and updated in ["Samudra 2: Scaling Ocean Emulators across Resolutions"](https://m2lines.github.io/Samudra/samudra2/). Samudra efficiently emulates the ocean component of a state-of-the-art climate model, accurately reproducing key ocean variables including sea surface height, horizontal velocities, temperature, and salinity, across their full depth.
```
alxmrs, line 14: I like this summary; this will be useful for folks who are looking for the Samudra v1 model.

alxmrs (approved):
LGTM, thanks for doing this.

CONTRIBUTING.md
```diff
@@ -1 +0,0 @@
-docs/contributing.md
```
alxmrs, line 1: Does GH do some sort of magic, like register that this project has a contrib file, based on a file with this name at the project root?

alxmrs:
> Now that this content is the real root CONTRIBUTING.md, these relative links resolve on GitHub to /run.md and /config.md, but those files only exist under docs/; the same problem applies to the data.md link later in the file. Readers using the repository-level contributing guide will hit 404s unless these are changed to docs/run.md, docs/config.md, and docs/data.md or to absolute docs URLs.

I believe this is a real concern that should be addressed before merge.

CONTRIBUTING.md
jder: The [docs](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) say that it should also work with docs/CONTRIBUTING.md. I'm hoping that's case-insensitive but not sure how to find out other than merging this and seeing.

jder:
> > Now that this content is the real root CONTRIBUTING.md, these relative links resolve on GitHub to /run.md and /config.md, but those files only exist under docs/; the same problem applies to the data.md link later in the file. Readers using the repository-level contributing guide will hit 404s unless these are changed to docs/run.md, docs/config.md, and docs/data.md or to absolute docs URLs.

> I believe this is a real concern that should be addressed before merge.

Thanks, I think I already addressed this by not moving the contributing file.

CONTRIBUTING.md
alxmrs: Perfect! Thanks for checking.

alxmrs:
>Thanks, I think I already addressed this by not moving the contributing file.

I clicked into the file in the branch and clicked the link and got a 404. Happy to merge and then verify the change if it's still broken. These things are hard to check in advance.

jder:
I think the links should all work once we move to the M²LInES org

PR #764, opened 2026-06-17
Update copyright to "Samudra"

jder:


alxmrs:
I think you need to update the copyright variable in `annotate_license.sh` too. 

BTW, what was the tool you used to format code and HTML files? Was that just an IDE formatting? Ideally, we would omit the fmt changes to make the diff small so we could revert this patch easier. Would you mind reverting those changes as well -- the HTML indentation is my main concern.

jder:
@alxmrs thanks, fixed both of those things

scripts/annotate_license.sh
```diff
@@ -22,7 +22,7 @@
 
 set -euo pipefail
 
-COPYRIGHT="Ocean Emulator Authors"
+COPYRIGHT="Samudra Authors"
```
alxmrs, line 25: LGTM!

alxmrs (approved):
Thanks so much for doing this. Approved.

PR #768, opened 2026-06-18
Cleaned up the README (nits)

alxmrs:
I think this improves the flow.


jder:
How about putting the images before any of the text after the badges? I would guess no one will ever find the last line of text with this image immediately before it.

alxmrs:
I find that the image placement has the opposite effect, that I'm more likely to skim over the gif and then see the call to action right below it. I like how the image breaks up the descriptions of Samudra vs the call-to-action section. WDYT? Here's a screenshot: 

Screenshot 2026-06-18 at 2 30 41 PM

jder:
I don't have the same intuition but I also don't think it matters very much.

jder (approved):

alxmrs:
Thanks! In general, I'd love to get input from the team / other OA folks / Sophie to discuss how we could make our README more presentable.

jder:
Yeah, this was just my attempt to have there be *some* content; definitely agreed we should improve it.

PR #770, opened 2026-06-22
validation rollout sliced

amogh-gulati:
This PR adds long-horizon autoregressive rollout validation to the training loop, complementing the existing single-step validation.

Previously, validation only checked one-step-ahead prediction error each epoch. This change allows training to periodically roll the model forward autoregressively over the validation period and log rollout RMSE at configured horizons, such as 90 days and 360 days.

The implementation supports multiple rollout horizons in a single rollout. For example, with `rollout_validation_days: [360, 90]`, the code rolls out once to the maximum horizon, records the 90-day metrics at the intermediate cutoff, and then continues to 360 days. This avoids launching a separate 90-day rollout.

Metrics are logged separately by horizon, variable, depth level, and depth band using raw, un-normalized fields. Rollout validation currently runs only on rank 0, with other ranks waiting at a barrier, since the validation window is not sharded across workers. The rollout is processed in bounded chunks to avoid materializing the full forecast horizon’s targets at once.

This is currently wired up for the standard single-scale training schedule and is a no-op for FOMO’s multi-scale schedule.

* Added rollout validation config options:

  * `rollout_validation_days`
  * `rollout_validation_steps`
  * `rollout_validation_freq`
  * `rollout_validation_steps_forward`

## Config

In `configs/samudra_om4_v2/train.yaml`, rollout validation is enabled with:

```yaml
rollout_validation_days: [360, 90]
rollout_validation_steps_forward: 4
```
test run here - https://wandb.ai/ocean_emulators/default/runs/e7r70ugc?nw=nwuseramoghgulati 
metrics present in rollout_val for 90 days and 360 days


jder:
@codex review

jder:
Thanks @amogh-gulati! One high-level question: why is this separate from the existing `inference_one_epoch` code? ie should those be the same thing, generalized to cover the new things you're adding here like supporting multiple lengths and removing the strange multi-GPU behavior the inference code currently has?

Also looks like you have some (real) CI failures now that I've finally fixed the spurious ones.

src/samudra/train.py
```diff
@@ -13,9 +13,10 @@
 from collections import OrderedDict
 from multiprocessing.context import BaseContext
 from pathlib import Path
-from typing import Any
+from typing import Any, NamedTuple
 
 import dask
+import numpy as np
```
alxmrs: I'm surprised we didn't have numpy before!

src/samudra/train.py
```diff
@@ -90,15 +93,80 @@
 logger = logging.getLogger(__name__)
 
 
-def should_log_validation_images(epoch: int, frequency: int) -> bool:
-    """Return whether to log validation images for a 1-based training epoch."""
+def should_run_on_epoch_freq(epoch: int, frequency: int) -> bool:
```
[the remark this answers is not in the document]
alxmrs: Is this still a valid concern? I agree with codex.

src/samudra/train.py
```diff
@@ -88,15 +91,80 @@
...
+def resolve_rollout_validation_steps(requested_steps: int, available_steps: int) -> int:
+    """Resolve requested rollout validation steps against available val steps."""
+    if available_steps <= 0:
+        return 0
+    if requested_steps < 0 or requested_steps > available_steps:
+        return available_steps
+    return requested_steps
+
+
+class RolloutValidationSpec(NamedTuple):
```
alxmrs: 🐑  I prefer using data classes over named tuples. Though, if tuples make this cleaner, that's ok too (I could lack context).

alxmrs: If this is a public contract, it would be nice to add a docstring here to explain what this is for and what it does.

src/samudra/train.py
```diff
@@ -88,15 +91,80 @@
...
+def _elapsed_days(start: Any, end: Any) -> float:
+    delta = end - start
+    if isinstance(delta, np.timedelta64):
+        return float(delta / np.timedelta64(1, "D"))
+    if isinstance(delta, datetime.timedelta) or hasattr(delta, "total_seconds"):
+        return float(delta.total_seconds() / (24 * 60 * 60))
+    return float(delta)
+
+
+def resolve_rollout_validation_day_spec(
```
alxmrs: I think these new methods and classes should live in util somewhere to not clutter train.

src/samudra/train.py
```diff
@@ -311,6 +379,11 @@ def __init__(self, cfg: TrainConfig) -> None:
         self.step_transition = cfg.step_transition
         self.save_freq = cfg.save_freq
         self.validation_image_log_freq = cfg.validation_image_log_freq
+        self.rollout_validation_steps = cfg.rollout_validation_steps
+        self.rollout_validation_days = cfg.rollout_validation_days
+        self.rollout_validation_steps_forward = cfg.rollout_validation_steps_forward
+        self.rollout_validation_freq = cfg.rollout_validation_freq
+        self._rollout_validation_pg: torch.distributed.ProcessGroup | None = None
```
alxmrs: I think we should pull these up into a single rollout validation config object, not as top lvl train settings.

alxmrs: feel free to pull in related, existing vars from the train config into here.

src/samudra/train.py
```diff
@@ -419,6 +492,13 @@ def run(self) -> None:
             val_stats = self.validate_one_epoch(epoch)
             end_epoch_val_time = time.perf_counter()
 
+            if self.should_run_rollout_validation(epoch):
+                rollout_val_stats = self.validate_rollout_one_epoch(epoch)
+                end_epoch_rollout_val_time = time.perf_counter()
+            else:
+                rollout_val_stats = {}
+                end_epoch_rollout_val_time = None
```
alxmrs: What is the train time cost of this operation (approximately)?

src/samudra/train.py
```diff
@@ -695,6 +786,162 @@ def validate_one_epoch(self, epoch):
         logger.info(f"Aggregating validation logs")
         return val_aggregator.get_logs(label="val")
 
+    def should_run_rollout_validation(self, epoch: int) -> bool:
```
alxmrs: If we have a rollout validation config object, then this could just be a null check (does the object exist?) at the call site -- this function would go away.

alxmrs: To be clear, the config file would have a config object that builds a new data class or ADT. The ADT could live in util and would be built at the top of train.py.

src/samudra/train.py
```diff
@@ -695,6 +786,162 @@ def validate_one_epoch(self, epoch):
...
+                    "enough timesteps for one autoregressive rollout step."
+                )
+                return {}
+
+            if self.rollout_validation_days:
+                target_times = rollout_dataset.get_target_time(
+                    0, available_steps
+                ).values
+                specs = [
+                    resolve_rollout_validation_day_spec(
```
alxmrs: Maybe this method could live as a static build method on the `RolloutValidationSpec` data class?

src/samudra/train.py
```diff
@@ -695,6 +786,162 @@ def validate_one_epoch(self, epoch):
...
+        if is_main_process():
+            try:
+                yield True
+            finally:
+                torch.distributed.barrier(group=group)
+        else:
+            torch.distributed.barrier(group=group)
+            yield False
+
+    def validate_rollout_one_epoch(self, epoch):
```
alxmrs, line 743: I'm torn if a lot of the contents of this method should live in the stepper (like `validate_one_epoch`) or not. That would follow convention, but I'm not sure what the right convention is. The way this is done has some advantages.

alxmrs (commented):
Review part 1/N. Will return to this when I can.

src/samudra/config.py
```diff
@@ -1165,6 +1165,40 @@ class TrainConfig(TopLevelConfig):
             "a value of 10 logs on epochs 1, 11, 21, ..."
         ),
     )
+    rollout_validation_steps: int = Field(
```
alxmrs: See comment in train.py about packaging these new options differently.

src/samudra/stepper.py
```diff
@@ -57,6 +58,107 @@ def validate_batch(
     return ValBatchOutput(loss, loss_per_channel, input_data, label, outs, batch.ctx)
 
 
+def _get_rollout_step_chunks(
```
alxmrs, line 62: I think this function could be simpler if we used batched: 

https://docs.python.org/3/library/itertools.html#itertools.batched

src/samudra/aggregator/validate/rollout.py
```diff
@@ -0,0 +1,178 @@
...
+    def __init__(
+        self,
+        *,
+        hist: int,
+        area_weights: torch.Tensor,
+        wet: torch.Tensor,
+        num_prognostic_channels: int,
+        normalize: Normalize,
+        tensor_map: TensorMap,
+        distributed_reduce: bool = True,
```
alxmrs, line 114: 🐑  I could be wrong, but could we use the DatasetSpec here instead? Many of the arguments we need are there. 

Do we need to include TensorMap and Normalize? We're generally trying to move away from including those multitions.

src/samudra/aggregator/validate/rollout.py
```diff
@@ -0,0 +1,178 @@
...
+            for name in self._raw_field_names
+        }
+
+    def record_batch(self, data: ModelInferenceOutput):
+        if len(data.prediction) == 0:
+            raise ValueError("No prediction values in data")
+        if len(data.target) == 0:
+            raise ValueError("No target values in data")
+
+        _, target_unnorm = get_aggregator_dicts(
```
alxmrs: 🐑  optional: We may be able to create a replacement to this function and the one below to use "modern" normalization code that manages state better so we could avoid adding the multitons.

alxmrs (commented):
2/2. Looks pretty good so far! My main note of feedback is on how we package this new process.

alxmrs:
Hey @fomo-bot, will you address the concerns that I've raised in my review of Amogh's code? Where there is ambiguity, please let me know your thinking and I'll help resolve it. Feel free to push back on my suggestions where that is appropriate, given your reading of the code in the branch. Thank you!

src/samudra/train.py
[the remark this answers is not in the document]
amogh-gulati: fixed now!

src/samudra/train.py
amogh-gulati: done!

alxmrs:
@fomo-bot may I have a final review of this PR from you? Is this ready for merge in your view?

alxmrs (approved):
LGTM

jder:
@amogh-gulati @alxmrs was there some discussion about keeping both this and the "inference_one_epoch" code as I asked [here](https://github.com/m2lines/Samudra/pull/770#issuecomment-4812237189)? Or can I just make a PR to remove the old (dead AFAIK) inference code?

alxmrs:
Removing the `inference_one_epoch` code after confirming that it is dead sounds like a good plan to me. Unless Amogh objects, I think a PR from you or your robot would be great.

alxmrs:
I missed your comment -- I think I wanted to get this merged sooner rather than later, IIRC, because other folks wanted to use this validate (I think I needed it in some of my hparam experiments).

amogh-gulati:
> I missed your comment -- I think I wanted to get this merged sooner rather than later, IIRC, because other folks wanted to use this validate (I think I needed it in some of my hparam experiments).

Yes, we wanted to merge this soon and I we discussed we would remove inference_one_epoch in a follow up PR

PR #772, opened 2026-06-22
Parallel eval

amogh-gulati:
Post-train checkpoint eval/viz sweep

Adds an optional post-training sweep that evaluates selected checkpoints after training completes, always including the final EMA checkpoint, with optional visualization generation.

* Add `post_train_eval.py` to discover checkpoints in `saved_nets/`, shard eval jobs across available GPUs, run standalone inference per checkpoint, and write `summary.json`.
* Support either `last_n_checkpoints: N` or an explicit `checkpoints: [...]` epoch list, with validation for mutually exclusive settings and missing checkpoints.
* Edit `Trainer.finish()` to launch the sweep on the main process via a new `post_train_eval` config block.
* Add a DDP barrier and `destroy_process_group()` before the sweep so all training ranks finish cleanly and release GPUs. (still seems a little flacky)
* Refactor viz to prepare groundtruth once and reuse it across checkpoint visualizations.
* Support standalone usage via `python -m samudra.post_train_eval` for models already trained.

Notes:

* Viz sweep requires `eval.save_zarr=true`.
* In the current config, `checkpoints: [50, 55, 60, 65, 70]` evaluates those epochs plus the final EMA checkpoint.

#Config

Add a post_train_eval block to the train config to enable the post-training checkpoint sweep:

post_train_eval:
  enabled: true
  eval_config_path: configs/<model>/eval.yaml   # required when enabled
  viz_config_path: configs/<model>/viz.yaml     # optional; omit to skip viz

  Pick exactly one checkpoint selection mode:
  checkpoints: [50, 55, 60, 65, 70]             # explicit checkpoint epochs
  last_n_checkpoints: 10                       # or evaluate the last N checkpoints

The final EMA checkpoint is always included automatically.
wandb is disabled for this, will a follow up PR I will add all the evals in one run (ideally with timeline scrub) which is not tested yet!

### DIFF
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -1355,8 +1355,8 @@ class PostTrainCheckpointSweepConfig(BaseConfig):
     eval_config_path: str | None = None
     viz_config_path: str | None = None
     last_n_checkpoints: int | None = Field(default=None, ge=1)
-    eval_dirname: str = "post_train_eval"
-    viz_dirname: str = "viz"
+    eval_dirname: str | None = None
+    viz_dirname: str | None = None
 
 
 class EvalConfig(TopLevelConfig):
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -1383,7 +1383,7 @@ class PostTrainEvalConfig(BaseConfig):
             eval_config_path=self.eval_config_path,
             checkpoint_paths=CheckpointPaths(nets_dir),
             data_root=data_root,
-            sweep_root=output_dir / self.eval_dirname,
+            sweep_output_dir=output_dir / self.eval_dirname,
             viz_config_path=self.viz_config_path,
             last_n_checkpoints=self.last_n_checkpoints,
             checkpoints=self.epochs,
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -1355,8 +1355,24 @@ class PostTrainCheckpointSweepConfig(BaseConfig):
     eval_config_path: str | None = None
     viz_config_path: str | None = None
     last_n_checkpoints: int | None = Field(default=None, ge=1)
+    checkpoints: list[int] | None = Field(
+        default=None,
+        description="Explicit list of checkpoint epochs (matching ckpt_<epoch>.pt) "
+        "to evaluate; the final EMA checkpoint is always added. Mutually "
+        "exclusive with last_n_checkpoints.",
+    )
     eval_dirname: str | None = None
     viz_dirname: str | None = None
+
+    @pydantic.model_validator(mode="after")
+    def _check_checkpoint_selection(self) -> "PostTrainCheckpointSweepConfig":
+        if self.last_n_checkpoints is not None and self.checkpoints is not None:
+            raise ValueError(
+                "set only one of last_n_checkpoints or checkpoints, not both"
+            )
+        if self.checkpoints is not None and len(self.checkpoints) == 0:
+            raise ValueError("checkpoints must be a non-empty list when provided")
+        return self
 
 
 class EvalConfig(TopLevelConfig):
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -1253,9 +1253,7 @@ class TrainConfig(TopLevelConfig):
     steps: list[int] = [4]
     step_transition: list[int] = []
     inference_epochs: list[int] = [-1]
-    post_train_eval: "PostTrainCheckpointSweepConfig" = Field(
-        default_factory=lambda: PostTrainCheckpointSweepConfig()
-    )
+    post_train_eval: "PostTrainEvalConfig | None" = None
 
     # Config components
     experiment: ExperimentConfig
@@ -1352,58 +1350,43 @@ class ObsMetricsConfig(BaseConfig):
 EvalBackendConfig = Literal["cpu", "cuda", "auto"]
 
 
-class PostTrainCheckpointSweepConfig(BaseConfig):
-    enabled: bool = False
-    eval_config_path: str | None = None
-    viz_config_path: str | None = None
+class PostTrainEvalConfig(BaseConfig):
+    eval_config_path: Path
+    viz_config_path: Path | None = None
     last_n_checkpoints: int | None = Field(default=None, ge=1)
-    checkpoints: list[int] | None = Field(
+    epochs: list[int] | None = Field(
         default=None,
         description="Explicit list of checkpoint epochs (matching ckpt_<epoch>.pt) "
         "to evaluate; the final EMA checkpoint is always added. Mutually "
         "exclusive with last_n_checkpoints.",
     )
-    eval_dirname: str | None = None
+    eval_dirname: str = "evals"
     # Subdirectory for visualization outputs within each checkpoint evaluation directory.
     viz_dirname: str = "viz"
 
     @pydantic.model_validator(mode="after")
-    def _check_checkpoint_selection(self) -> "PostTrainCheckpointSweepConfig":
-        if self.enabled and self.eval_config_path is None:
-            raise ValueError("eval_config_path must be set when enabled is true")
-        if self.last_n_checkpoints is not None and self.checkpoints is not None:
-            raise ValueError(
-                "set only one of last_n_checkpoints or checkpoints, not both"
-            )
-        if self.checkpoints is not None and len(self.checkpoints) == 0:
-            raise ValueError("checkpoints must be a non-empty list when provided")
+    def _check_checkpoint_selection(self) -> "PostTrainEvalConfig":
+        if self.last_n_checkpoints is not None and self.epochs is not None:
+            raise ValueError("set only one of last_n_checkpoints or epochs, not both")
+        if self.epochs is not None and len(self.epochs) == 0:
+            raise ValueError("epochs must be a non-empty list when provided")
         return self
 
     def build(
         self,
         nets_dir: Path,
         output_dir: Path,
-        data_root: Location | None,
-    ) -> "CheckpointSweep | None":
-        """Build the runtime sweep, or return None when it is disabled."""
-        if not self.enabled:
-            return None
-
-        assert self.eval_config_path is not None  # enforced by the validator
+        data_root: ResolvedLocation,
+    ) -> "CheckpointSweep":
+        """Build the runtime sweep."""
         return CheckpointSweep(
-            eval_config_path=Path(self.eval_config_path),
+            eval_config_path=self.eval_config_path,
             checkpoint_paths=CheckpointPaths(nets_dir),
             data_root=data_root,
-            sweep_root=(
-                output_dir / self.eval_dirname
-                if self.eval_dirname is not None
-                else None
-            ),
-            viz_config_path=(
-                Path(self.viz_config_path) if self.viz_config_path is not None else None
-            ),
+            sweep_root=output_dir / self.eval_dirname,
+            viz_config_path=self.viz_config_path,
             last_n_checkpoints=self.last_n_checkpoints,
-            checkpoints=self.checkpoints,
+            checkpoints=self.epochs,
             viz_dirname=self.viz_dirname,
         )
 
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -58,6 +58,7 @@ from samudra.models.modules.augment_input import (
 )
 from samudra.models.modules.blocks import ZonallyPeriodicBilinearUpsample
 from samudra.models.modules.encoder import patch_from
+from samudra.post_train_eval import CheckpointSweep
 from samudra.utils.data import (
     CanonicalSource,
     DataBundle,
@@ -84,6 +85,7 @@ from samudra.utils.loss import (
 )
 from samudra.utils.profiler import Profiler
 from samudra.utils.schedule import SchedulerConfig
+from samudra.utils.train import CheckpointPaths
 
 
 class WandBConfig(BaseConfig):
@@ -1362,10 +1364,13 @@ class PostTrainCheckpointSweepConfig(BaseConfig):
         "exclusive with last_n_checkpoints.",
     )
     eval_dirname: str | None = None
-    viz_dirname: str | None = None
+    # Subdirectory for visualization outputs within each checkpoint evaluation directory.
+    viz_dirname: str = "viz"
 
     @pydantic.model_validator(mode="after")
     def _check_checkpoint_selection(self) -> "PostTrainCheckpointSweepConfig":
+        if self.enabled and self.eval_config_path is None:
+            raise ValueError("eval_config_path must be set when enabled is true")
         if self.last_n_checkpoints is not None and self.checkpoints is not None:
             raise ValueError(
                 "set only one of last_n_checkpoints or checkpoints, not both"
@@ -1373,6 +1378,34 @@ class PostTrainCheckpointSweepConfig(BaseConfig):
         if self.checkpoints is not None and len(self.checkpoints) == 0:
             raise ValueError("checkpoints must be a non-empty list when provided")
         return self
+
+    def build(
+        self,
+        nets_dir: Path,
+        output_dir: Path,
+        data_root: Location | None,
+    ) -> "CheckpointSweep | None":
+        """Build the runtime sweep, or return None when it is disabled."""
+        if not self.enabled:
+            return None
+
+        assert self.eval_config_path is not None  # enforced by the validator
+        return CheckpointSweep(
+            eval_config_path=Path(self.eval_config_path),
+            checkpoint_paths=CheckpointPaths(nets_dir),
+            data_root=data_root,
+            sweep_root=(
+                output_dir / self.eval_dirname
+                if self.eval_dirname is not None
+                else None
+            ),
+            viz_config_path=(
+                Path(self.viz_config_path) if self.viz_config_path is not None else None
+            ),
+            last_n_checkpoints=self.last_n_checkpoints,
+            checkpoints=self.checkpoints,
+            viz_dirname=self.viz_dirname,
+        )
 
 
 class EvalConfig(TopLevelConfig):
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -1251,6 +1251,9 @@ class TrainConfig(TopLevelConfig):
     steps: list[int] = [4]
     step_transition: list[int] = []
     inference_epochs: list[int] = [-1]
+    post_train_eval: "PostTrainCheckpointSweepConfig" = Field(
+        default_factory=lambda: PostTrainCheckpointSweepConfig()
+    )
 
     # Config components
     experiment: ExperimentConfig
@@ -1345,6 +1348,15 @@ class ObsMetricsConfig(BaseConfig):
 
 # See backend.py for how these are turned into concrete devices
 EvalBackendConfig = Literal["cpu", "cuda", "auto"]
+
+
+class PostTrainCheckpointSweepConfig(BaseConfig):
+    enabled: bool = False
+    eval_config_path: str
+    viz_config_path: str
+    last_n_checkpoints: int | None = Field(default=None, ge=1)
+    eval_dirname: str = "post_train_eval"
+    viz_dirname: str = "viz"
 
 
 class EvalConfig(TopLevelConfig):
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -1352,8 +1352,8 @@ EvalBackendConfig = Literal["cpu", "cuda", "auto"]
 
 class PostTrainCheckpointSweepConfig(BaseConfig):
     enabled: bool = False
-    eval_config_path: str
-    viz_config_path: str
+    eval_config_path: str | None = None
+    viz_config_path: str | None = None
     last_n_checkpoints: int | None = Field(default=None, ge=1)
     eval_dirname: str = "post_train_eval"
     viz_dirname: str = "viz"
--- a/src/samudra/config_base.py
+++ b/src/samudra/config_base.py
@@ -90,6 +90,20 @@ class TopLevelConfig(BaseSettings):
         return (init_settings,)
 
     @classmethod
+    def from_yaml(cls, config_path: Path | str) -> Self:
+        """Load and validate a YAML config without CLI overrides."""
+        return cls(**cls._load_yaml(config_path))
+
+    @classmethod
+    def _load_yaml(cls, config_path: Path | str) -> dict[str, Any]:
+        resolved_path = Path(config_path).expanduser().resolve()
+        if not resolved_path.exists():
+            raise FileNotFoundError(
+                f"Config file `{config_path}` (full path: {resolved_path}) not found"
+            )
+        return YamlConfigSettingsSource(cls, yaml_file=resolved_path)()
+
+    @classmethod
     def from_yaml_and_cli(
         cls,
         args_to_parse: list[str] | None = None,
@@ -126,20 +140,9 @@ class TopLevelConfig(BaseSettings):
         # so the help is complete on error.
         args = parser.parse_args(args_to_parse)
 
-        # Then we read the YAML file specified in the CLI. A bare path resolves
-        # against the filesystem; if it's missing we try a preset bundled in the
-        # installed package before giving up.
-        # Note that by default, YamlConfigSettingsSource will ignore missing files
-        config_path = resolve_config_path(args.config)
-        if not config_path.exists():
-            raise FileNotFoundError(
-                f"Config file `{args.config}` (full path: {config_path}) not found"
-            )
-        yaml_values = YamlConfigSettingsSource(cls, yaml_file=config_path)()
-
         return cls(
             _cli_settings_source=cli_source,
-            **yaml_values,
+            **cls._load_yaml(args.config),
         )
 
     def save_yaml(self, save_path: Path) -> None:
--- a/src/samudra/config_base.py
+++ b/src/samudra/config_base.py
@@ -96,7 +96,7 @@ class TopLevelConfig(BaseSettings):
 
     @classmethod
     def _load_yaml(cls, config_path: Path | str) -> dict[str, Any]:
-        resolved_path = Path(config_path).expanduser().resolve()
+        resolved_path = resolve_config_path(str(config_path))
         if not resolved_path.exists():
             raise FileNotFoundError(
                 f"Config file `{config_path}` (full path: {resolved_path}) not found"
--- a/src/samudra/post_train_eval.py
+++ b/src/samudra/post_train_eval.py
@@ -11,18 +11,17 @@ import re
 import time
 from dataclasses import dataclass
 from pathlib import Path
-from typing import cast
+from typing import TYPE_CHECKING, cast
 
 import torch
 
-from samudra.config import EvalConfig, TrainConfig
-from samudra.eval import Eval
-from samudra.utils.location import LocalLocation, Location
+from samudra.utils.location import LocalLocation, Location, ResolvedLocation
 from samudra.utils.multiton import MultitonScope
 from samudra.utils.train import CheckpointPaths
-from samudra.viz import VizConfig
-from samudra.viz.config import VizRunConfig, run_with_prepared_groundtruth
-from samudra.viz.core import PreparedVizGroundtruth
+
+if TYPE_CHECKING:
+    from samudra.config import EvalConfig
+    from samudra.viz import VizTemplate
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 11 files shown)


jder:
@codex review

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,500 @@
...
+        return []
+
+    if train_cfg.post_train_eval.eval_config_path is None:
+        raise ValueError(
+            "post_train_eval.eval_config_path must be set when post_train_eval.enabled is true"
+        )
+
+    eval_config_path = (
+        Path(train_cfg.post_train_eval.eval_config_path).expanduser().resolve()
+    )
```
jder: I don't understand why we need expanduser()/resolve() here given it's in `Config.from_yaml_and_cli`

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,500 @@
...
+    targets = discover_checkpoints_from_directory(
+        checkpoint_paths,
+        last_n_checkpoints=last_n_checkpoints,
+        checkpoints=checkpoints,
+    )
+    if not targets:
+        logger.warning("No checkpoints selected for post-train eval sweep")
+        return []
+
+    eval_cfg = EvalConfig.from_yaml_and_cli(list(eval_config_args))
```
jder: Probably nicer to have a path here instead and expose a Config.from_yaml()

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,500 @@
...
+        updated["base_output_dir"] = eval_output_dir
+        updated["name"] = viz_dirname
+        updated["runs"] = [
+            checkpoint_run.model_dump(mode="python"),
+            *[run.model_dump(mode="python") for run in template_cfg.runs[1:]],
+        ]
+        cfg = VizConfig.model_validate(updated)
+
+        start = time.perf_counter()
+        run_with_prepared_groundtruth(cfg, prepared_groundtruth)
```
jder: I think this config surgery is a bit of a smell and somewhat fragile. How about:
* Create a new top-level config VizTemplateConfig with base_output_dir, dataset_name, variables, data_root, etc. But not `runs` or `name`. It builds a VizTemplate which is basically the same content as `PreparedVizGroundtruth` today.
* In `run_checkpoint_sweep` we load a VizTemplateConfig and call its build method to get a VizTemplate.
* In this code we call `viz_template.instantiate(output_path, runs)` or something to get a Viz.
* VizConfig now extends VizTemplateConfig and adds runs + name. (VizConfig.build now calls super.build and then calls instantiate on that)

jder: (Alternatively, if you want to just land this without the viz stuff that works for me too and we can deal with this as a follow-up.)

src/samudra/train.py
```diff
@@ -113,7 +114,7 @@ class Trainer:
     def __init__(self, cfg: TrainConfig) -> None:
         cfg.prepare_output_dirs()
         cfg.save_yaml(cfg.experiment.output_dir / "config.yaml")
-
+        self.cfg = cfg
```
jder: The pattern we're trying to move towards (though I see we have failed to actually write this down anywhere) is to avoid passing around cfg values like this (and below), and instead turn the cfg types into "inflated" or "ready to run" types with `build(...)` as early as possible, passing in needed dependencies. Trainer is the main exception to this pattern since it predates the config system and is a beast to refactor in that way. Would it be possible to avoid saving the cfg here and to apply this pattern to the `PostTrainCheckpointSweepConfig` type, passing in the needed extra data from TrainConfig to the build method (e.g. the experiment output directory) here? (Perhaps producing a `CheckpointSweep` instance which has all the needed information to run such a thing?)

jder (changes requested):
Very excited about this. I didn't read everything in detail but a few high-level comments first. 

I also have one question about the goal: why are we causing the existing per-GPU torch processes to exit and then re-spawning them? I think will cause us to drop down to a single host for running evals/viz when doing multi-host training. Can we use the existing per-GPU processes instead (and run post_train_eval via slurm/torchrun as we do training when we want it to be standalone?). Basically just have the existing non-main processes go into a worker loop and distribute work via torch.distributed?

src/samudra/post_train_eval.py
amogh-gulati: done!

src/samudra/post_train_eval.py
amogh-gulati: Implemented the template approach, now VizConfig extends VizTemplateConfig which builds reusable VizTemplate state and prepares ground truth once. The sweep now instantiates each checkpoint’s Viz directly from that template.

src/samudra/config.py
```diff
@@ -1162,6 +1167,62 @@ def prepare_output_dirs(self) -> None:
 EvalBackendConfig = Literal["cpu", "cuda", "auto"]
 
 
+class PostTrainCheckpointSweepConfig(BaseConfig):
+    enabled: bool = False
```
jder: Might be nicer to have post_train_eval be `PostTrainEvalConfig | None` above and remove the enabled bool? Then you can make eval_config_path and others always non-None

src/samudra/config.py
```diff
@@ -1162,6 +1167,62 @@ def prepare_output_dirs(self) -> None:
...
+    eval_config_path: str | None = None
+    viz_config_path: str | None = None
```
jder: I went on a journey here trying to make this EvalConfig rather than a path. And also splitting up EvalConfig into things we actually want here vs not (e.g. not ckpt_path). This lets us validate + build() earlier so things don't explode at the end of a run. But I think in the end it is probably more complicated than it's worth. If you're interested, you can see the it here: https://github.com/m2lines/Samudra/compare/parallel_eval...codex/refactor-eval-dependencies?expand=1

But anyway, I suggest these be Paths, not `str`, in the current state

src/samudra/config.py
```diff
@@ -1162,6 +1167,62 @@ def prepare_output_dirs(self) -> None:
...
+    last_n_checkpoints: int | None = Field(default=None, ge=1)
+    checkpoints: list[int] | None = Field(
+        default=None,
+        description="Explicit list of checkpoint epochs (matching ckpt_<epoch>.pt) "
+        "to evaluate; the final EMA checkpoint is always added. Mutually "
+        "exclusive with last_n_checkpoints.",
+    )
```
jder, line 1219: 🐑 how about "epochs" instead of "checkpoints"?

src/samudra/config.py
```diff
@@ -1162,6 +1167,62 @@ def prepare_output_dirs(self) -> None:
...
+    eval_dirname: str | None = None
```
jder: Should this default to (or just be hard-coded below as) "evals"?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+                )
+            )
+    except Exception as exc:
+        queue.put({"error": str(exc), "gpu_index": gpu_index})
+        raise
+
+
+@dataclass(frozen=True)
+class CheckpointSweep:
+    """Ready-to-run checkpoint sweep built from configuration."""
```
jder: I don't think this is "ready-to-run" or "built from configuration" given it has a list of config paths to still to load?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+        help="Optional limit to only evaluate the last N discovered checkpoints",
+    )
+    parser.add_argument(
+        "--checkpoints",
+        type=int,
+        nargs="+",
+        help="Explicit checkpoint epochs to evaluate; mutually exclusive with "
+        "--last_n_checkpoints",
+    )
+    args, eval_override_args = parser.parse_known_args(argv)
```
jder, line 443: I might be missing something but I think if you used the existing pattern to load and instantiate a `PostTrainCheckpointSweepConfig` (perhaps renamed) then you wouldn't need any of this arg parsing + handling (you could override all this stuff as we do with other configs)

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+
+    # Expand ~ in user-provided config paths before resolving them to absolute paths.
+    viz_config_path = (
+        Path(args.viz_config).expanduser().resolve()
+        if args.viz_config is not None
+        else None
+    )
```
jder: We still don't need this given it's loaded by the new _load_yaml, right?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+
+@dataclass(frozen=True)
+class CheckpointEvalTarget:
+    epoch: int
+    kind: str
+    path: str
+    for_inference: bool
+
+
+def checkpoint_label(entry: CheckpointEvalTarget) -> str:
```
jder: 🐑 method on CheckpointEvalTarget?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+    match entry.kind:
+        case "periodic":
+            return f"epoch_{entry.epoch:04d}"
+        case "ema_latest":
+            return "final_ema"
```
jder: Why the renaming?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+        case _:
+            return f"{entry.kind}_epoch_{entry.epoch:04d}"
```
jder: Is this dead code?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+
+if TYPE_CHECKING:
+    from samudra.config import EvalConfig
+    from samudra.viz import VizTemplate
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class CheckpointEvalTarget:
```
jder: I am uncertain what the point of this class is. As far as I can tell we infer some metadata about the checkpoints based on the path and then write it to the results summary json. Is something consuming that data? If not, why not just write the path to the checkpoint? Would let us remove a bunch of code and magic strings.

Alternatively, if you want to add some structure to the results summary JSON, can we make a dataclass which we actually serialize to that file so it's clear what we expect to be in it?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+        raise ValueError("pass only one of last_n_checkpoints or checkpoints, not both")
+    if last_n_checkpoints is not None and last_n_checkpoints < 1:
+        raise ValueError(f"last_n_checkpoints must be >= 1, got {last_n_checkpoints}")
+
+    checkpoint_dir = checkpoint_paths.checkpoint_dir
+    periodic: dict[int, Path] = {}
+    for path in checkpoint_dir.iterdir():
+        if not path.is_file():
+            continue
+        if match := re.match(r"^ckpt_(\d+)\.pt$", path.name):
```
jder: Can we re-use the logic from CheckpointPaths? Or move this regex there so they're easier to keep in sync?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+) -> list[list[CheckpointEvalTarget]]:
+    if worker_count < 1:
+        raise ValueError("worker_count must be at least 1")
+    return [entries[i::worker_count] for i in range(worker_count)]
+
+
+def _resolve_worker_count(
+    backend: str,
+    num_checkpoints: int,
+) -> tuple[int, list[int]]:
```
jder: Might be slightly clearer as `_resolve_workers` -> `list[int | None]` so `[0, 1]` means two workers with GPUs and `[None]` means one CPU worker. Or, maybe even better, `list[torch.device]`?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+        torch.cuda.set_device(gpu_index)
+
+    with MultitonScope():
+        cfg = _load_eval_config(eval_config_path, eval_override_args)
+        cfg.ckpt_path = entry.path
+        cfg.experiment.base_output_dir = str(sweep_root)
+        cfg.experiment.name = checkpoint_label(entry)
+        if data_root is not None:
+            cfg.experiment.data_root = data_root
+        cfg.experiment.wandb.mode = "disabled"
```
jder: Huh, this surprised me. We don't want to log these?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+    if gpu_index is not None:
+        torch.cuda.set_device(gpu_index)
+
+    with MultitonScope():
+        cfg = _load_eval_config(eval_config_path, eval_override_args)
+        cfg.ckpt_path = entry.path
+        cfg.experiment.base_output_dir = str(sweep_root)
+        cfg.experiment.name = checkpoint_label(entry)
+        if data_root is not None:
+            cfg.experiment.data_root = data_root
```
jder: I think this will explode if `data_root` is none. You probably want to have it required everywhere it's currently optional. (It's only optional in ExperimentConfig by default so that you have to supply it for your environment on the CLI)

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+    result: dict[str, object],
+    template: "VizTemplate",
+    steps: list[str],
+    viz_dirname: str,
+) -> dict[str, object]:
+    # Viz config imports TimeConfig, so these need to stay off config.py's
+    # module initialization path.
+    from samudra.viz.config import VizRunConfig, run_steps
+
+    label = cast(str, result["label"])
```
jder: Seems sad to smuggle this string through a dict and back. Why not pass the label (and output_dir) as arguments to this function?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+        last_n_checkpoints=last_n_checkpoints,
+        checkpoints=checkpoints,
+    )
+    if not targets:
+        logger.warning("No checkpoints selected for post-train eval sweep")
+        return []
+
+    eval_cfg = _load_eval_config(eval_config_path, eval_override_args)
+    if sweep_root is None:
+        sweep_root = checkpoint_paths.checkpoint_dir.parent / eval_cfg.experiment.name
```
jder: This seems like a hack to me; who guarantees the checkpoint dir's parent will be the right place? I'm actually not sure what a "sweep_root" is. Is this the output path? Maybe some comments or renaming would clarify?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+    sweep_root.mkdir(parents=True, exist_ok=True)
+
+    if viz_config_path is not None and not eval_cfg.save_zarr:
```
jder: Is there a use case where we'd want save_zarr to be false? Can we validate this earlier if not?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+    worker_count, gpu_indices = _resolve_worker_count(eval_cfg.backend, len(targets))
+
+    logger.info(
+        "Running checkpoint sweep for %d checkpoints with %d worker(s)",
+        len(targets),
+        worker_count,
+    )
+
+    results: list[dict[str, object]] = []
+    if worker_count == 1:
```
jder: 🐑 not sure we really need two different ways to do this. Can we use the else branch for both cases?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+        processes = []
+        shards = partition_checkpoint_work(targets, worker_count)
+        for gpu_index, shard in zip(gpu_indices, shards, strict=True):
+            process = ctx.Process(
+                target=_worker_main,
+                args=(
+                    shard,
+                    eval_config_path,
+                    eval_override_args,
+                    str(sweep_root),
```
jder: I think you can pass a Path?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+                    entry=entry,
+                    eval_config_path=eval_config_path,
+                    eval_override_args=eval_override_args,
+                    sweep_root=Path(sweep_root),
+                    data_root=data_root,
+                    gpu_index=gpu_index,
+                )
+            )
+    except Exception as exc:
+        queue.put({"error": str(exc), "gpu_index": gpu_index})
```
jder: Can we send back the rich exception and let it get raised in the parent? Can attach the gpu index and entry here as additional info with add_note or something.

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+    targets = discover_checkpoints_from_directory(
+        checkpoint_paths,
+        last_n_checkpoints=last_n_checkpoints,
+        checkpoints=checkpoints,
+    )
+    if not targets:
+        logger.warning("No checkpoints selected for post-train eval sweep")
+        return []
+
+    eval_cfg = _load_eval_config(eval_config_path, eval_override_args)
```
jder, line 304: Any reason we can't do this earlier, like in `PostTrainCheckpointSweepConfig`'s build, to give errors before running training?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+                    f"checkpoint sweep worker exited with status {process.exitcode}"
+                )
+
+    _write_summary(results, sweep_root)
+
+    if viz_config_path is not None:
+        from samudra.viz import VizTemplateConfig
+
+        logger.info("Running post-train viz sweep for %d checkpoints", len(results))
+        template_cfg = VizTemplateConfig.from_yaml(viz_config_path)
```
jder, line 372: Same question here -- can we run this earlier?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+        default_data_root: ResolvedLocation = LocalLocation(path=Path.cwd())
+        if data_root is not None:
+            default_data_root = default_data_root.resolve(data_root)
```
jder: I think it would be preferable to use ExperimentConfig's resolved_data_root (ie pass a non-optional data root to this function which comes from that when using this from a training run).

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+                    if process.exitcode not in {None, 0}:
+                        raise RuntimeError(
+                            f"checkpoint sweep worker exited with status {process.exitcode}"
+                        )
+                continue
+            if "error" in item:
+                for process in processes:
+                    if process.is_alive():
+                        process.terminate()
+                    process.join()
```
jder: Do we need to do this? If so, why don't we need to do it in the timeout case just above?

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,506 @@
...
+                    str(sweep_root),
+                    data_root,
+                    gpu_index,
+                    queue,
+                ),
+            )
+            process.start()
+            processes.append(process)
+
+        while len(results) < len(targets):
```
jder: Do we want a real timeout? e.g. 6 hours?

src/samudra/viz/core.py
```diff
@@ -104,6 +105,34 @@ class VizRun:
...
+    prediction_coords: dict[str, xr.DataArray]
+    wetmask: xr.DataArray
+
+
+@dataclasses.dataclass
+class VizTemplate:
+    # save prepared ground truth
+    dataset_name: str
+    data_root: ResolvedLocation
+    variables: list[str]
```
jder, line 124: Might be somewhat more clear to have `instantiate` take the needed arguments to produce a VizRun (I think label + location?) rather than having data_root and variables used externally while dataset_name and prepared_groundtruth are used internally. Alternatively maybe `variables` should not be on VizRun at all… I think weird things will probably happen if different runs have different variables compared to each other or the baseline? (And for post-training runs I think we can guarantee they will be the same anyway?)

src/samudra/viz/core.py
```diff
@@ -3844,103 +3795,158 @@ def _postprocess_for_plot(
...
-    # Rename lat and lon to y and x (This needs to be done in the end!)
-    ds_groundtruth = ds_groundtruth.rename({"lat": "y", "lon": "x"})
 
-    return ds_groundtruth, pred_dict
+def prepare_viz_groundtruth(
+    dataset_name: str,
+    basins: xr.Dataset,
+    groundtruth_rollout: xr.Dataset,
+    time_range: slice,
+) -> PreparedVizGroundtruth:
```
jder, line 3873: This is great, so glad we're not doing this over and over.

src/samudra/train.py
```diff
@@ -1128,6 +1134,52 @@ def _ema_context(self):
...
+            # post-train sweep, which claims every GPU via its own subprocesses.
+            # Tearing down the process group lets the non-main ranks exit and free
+            # their GPUs (and avoids the "destroy_process_group() was not called"
+            # warning at interpreter shutdown).
+            torch.distributed.barrier()
+            torch.distributed.destroy_process_group()
+        if main_process and self.post_train_sweep is not None:
+            self.post_train_sweep.run()
+
+    def _release_train_state(self) -> None:
```
jder, line 1131: Might be less fragile and more reliable to drop the whole Trainer object + run gc instead?

jder (changes requested):
Hey @amogh-gulati thanks again for doing this! Very much looking forward to being able to use this. I left a bunch of comments but they're mostly pretty tiny things your agent should be able to do :) LMK if you want to chat through any of them.

src/samudra/config.py
amogh-gulati: way cleaner now

src/samudra/post_train_eval.py
amogh-gulati: yes correct, so it isn't fully ready, the eval/viz configs are intentionally loaded when the sweep starts: in the integrated workflow that happens after training, and with the standalone CLI it happens as soon as the post-training job starts. I've updated the misleading docstring

src/samudra/post_train_eval.py
amogh-gulati: We support running the eval sweep independently of model training. Reusing TrainConfig for that standalone path would require constructing a mostly fake training configuration, so the dedicated CLI parsing is intentional.

src/samudra/post_train_eval.py
amogh-gulati: in some test runs i was using intermediate emas, renamed now

src/samudra/post_train_eval.py
amogh-gulati: corrected to raise an error

src/samudra/post_train_eval.py
amogh-gulati: removed the class and replaced with a parser method

src/samudra/post_train_eval.py
amogh-gulati: removed this but goal is to have them in the same run with a timeline scrub, will implement in the follow up pr

src/samudra/post_train_eval.py
amogh-gulati: resolved

alxmrs:
hey @amogh-gulati, sorry I've been a bit behind in reviewing this PR. Will you fix merge conflicts before I make my review?

alxmrs:
First pass: 

@fomo-bot will you review Amogh's PR? Please see that prior feedback has been applied (primarily, from Jesse). Thank you.

alxmrs:
@fomo-bot will you help us by addressing the merge conflicts in this PR?

src/samudra/utils/train.py
```diff
@@ -50,6 +51,9 @@ def collate_inference_data(
 
 
 class CheckpointPaths:
+    _PERIODIC_CHECKPOINT_PATTERN = re.compile(r"^ckpt_(\d+)\.pt$")
```
alxmrs, line 54: 🐑 IMO, regular expressions are good for interactive use, but for string parsing like this, I prefer a small string manipulation method (it easier to debug). 

Not a deal breaker either way.

alxmrs: For your reading pleasure: https://buttondown.com/hillelwayne/archive/regexes-are-cool-and-good/

src/samudra/utils/train.py
```diff
@@ -60,13 +64,27 @@ def latest_checkpoint_path(self) -> Path:
     def latest_checkpoint_path_with_epoch(self, epoch: int) -> Path:
         return self.checkpoint_dir / f"ckpt_{epoch}.pt"
 
+    def periodic_checkpoint_paths(self) -> dict[int, Path]:
+        """Return periodic checkpoints indexed by epoch."""
+        return {
+            epoch: path
+            for path in self.checkpoint_dir.iterdir()
+            if path.is_file()
+            and (epoch := self.periodic_checkpoint_epoch(path)) is not None
```
alxmrs, line 73: Nice walrus operator

src/samudra/viz/config.py
```diff
@@ -45,12 +51,14 @@ def _check_step(v: str) -> str:
 ]
 
 
+def default_viz_variables() -> list[str]:
+    return ["thetao", "so", "uo", "vo", "tos", "zos"]
```
alxmrs, line 55: I think this should just be a constant.

alxmrs (commented):
I juuust started reviewing this; I'll return to it again tomorrow.

src/samudra/utils/train.py
amogh-gulati: I did a lot of regex practice for my programming languages class so i can read them ( for now :p)

alxmrs:
@fomo-bot, will you rebase in the latest changes from main and address merge conflicts?

src/samudra/utils/train.py
```diff
@@ -50,6 +50,9 @@ def collate_inference_data(
 
 
 class CheckpointPaths:
+    _PERIODIC_CHECKPOINT_PREFIX = "ckpt_"
+    EMA_CHECKPOINT_NAME = "ema_ckpt.pt"
```
alxmrs, line 54: I think the scope of these variables.

src/samudra/viz/config.py
```diff
@@ -116,42 +119,61 @@ def _groundtruth_location(self) -> Location:
...
             groundtruth_rollout,
             self.groundtruth_time_range.time_slice,
-            observations=self.observations,
+        )
+
+    @property
+    def selected_steps(self) -> list[VizStep]:
+        return [s for s in self.steps or _ordered_steps() if s not in self.not_steps]
+
+    def build_template(self, default_root: ResolvedLocation) -> VizTemplate:
```
alxmrs, line 146: I think I prefer simply `build` -- that would follow our conventions.

src/samudra/viz/config.py
```diff
@@ -116,42 +119,61 @@ def _groundtruth_location(self) -> Location:
...
+class VizConfig(VizTemplateConfig):
+    name: str
+    runs: list[VizRunConfig]
+
+    @cached_property
+    def output_path(self) -> Path:
+        return Path(self.base_output_dir) / self.name
 
+    def build(self, default_root: ResolvedLocation) -> Viz:
+        template = self.build_template(default_root)
```
alxmrs, line 166: Then, this could be `super().build(...)` if we named the method on the superclass `build()`.

src/samudra/viz/core.py
alxmrs: I think this has been addressed. Will resolve for now.

src/samudra/viz/core.py
```diff
@@ -3934,103 +3899,158 @@ def _postprocess_for_plot(
...
-        wetmask = ds_groundtruth["mask"].isel(
-            time=0, missing_dims="ignore"
-        )  # our data does not always have time for a mask
-    else:
-        wetmask = ds_groundtruth.wetmask
-
-    ds_groundtruth = _postprocess_for_plot(
-        ds_groundtruth, areacello_values, areacello_spherical_values, dz, times, wetmask
+    groundtruth_rollout = groundtruth_rollout.assign(
+        areacello=(["lat", "lon"], spherical_area_weights(groundtruth_rollout))
```
alxmrs, line 3915: 🐑 I can't recall, but I think with the latest merge, we might want to get this from the data layout.

src/samudra/config_base.py
```diff
@@ -89,6 +89,20 @@ def settings_customise_sources(
         # We don't need env/dotenv/secrets, yaml & CLI are injected in from_yaml_and_cli
         return (init_settings,)
 
+    @classmethod
+    def from_yaml(cls, config_path: Path | str) -> Self:
```
alxmrs, line 93: I think I've missed something: what is this addition for?

alxmrs: This is mostly a curiosity for the motivation behind the change, it looks benign.

src/samudra/post_train_eval.py
```diff
@@ -0,0 +1,458 @@
...
+        data_root=eval_cfg.experiment.resolved_data_root,
+        sweep_output_dir=output_dir,
+        eval_override_args=tuple(eval_override_args or []),
+        viz_config_path=viz_config_path,
+        last_n_checkpoints=last_n_checkpoints,
+        checkpoints=checkpoints,
+    ).run()
+
+
+def main(argv: list[str] | None = None) -> None:
```
alxmrs, line 409: 🐑 maybe this doesn't apply, but if we have a pydantic settings config, could we use that system (our normal config system) instead of using vanilla argparse?

src/samudra/post_train_eval.py
alxmrs: I think what Jesse means is similar to my comment above -- you can get a "free" CLI without using TrainConfig, but using your new `PostTrainCheckpointSweepConfig`.

alxmrs (changes requested):
Here is a full review, sorry it took so long!

PR #776, opened 2026-06-23
Add GPU test duration diagnostics

jder:
## Summary
- add pytest duration reporting to the GPU test workflow
- keep the existing per-test 300s timeout and 30-minute job timeout unchanged

## Context
Run 28035329848 / job 82987734124 failed because two dataset parametrizations hit pytest-timeout's 300s per-test limit while reading cached Zarr chunks during test setup. Recent successful GPU pytest steps usually take about 7-14.5 minutes overall, but the logs do not expose per-test durations, so this makes future runs show whether a specific item is drifting toward the timeout.

## Validation
- git diff --check
- ruby -e 'require yaml; YAML.load_file(.github/workflows/test-gpu.yml)'
- pre-commit hooks from git commit

### DIFF
--- a/.github/workflows/test-gpu.yml
+++ b/.github/workflows/test-gpu.yml
@@ -57,7 +57,7 @@ jobs:
 
       # Each test runs with a 5-minute timeout. The `-x` flag will stop on the first failure or hang.
       - name: Run pytest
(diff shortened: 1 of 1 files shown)


jder:
Planning to resolve with #779 instead, including diagnostics there, too.

PR #780, opened 2026-06-25
[codex] Reduce processed data fixture memory

jder:
We are seeing [failures](https://github.com/m2lines/Samudra/actions/runs/28103086538/job/83209937081) in our data tests which look like out-of-memory kills. This PR changes the data we generate for the purposes of testing schema shapes from huge in-memory arrays to dask-backed arrays. This reduces memory usage for tests from ~5 GB to ~600MB.


data/tests/data.py
```diff
@@ -8,63 +8,114 @@
...
 from ocean_preprocessing.utils import apply_mask
 
+PROCESSED_TIME_SIZE = 10
+PROCESSED_LEVEL_SIZE = 19
+PROCESSED_Y_SIZE = 1080
+PROCESSED_X_SIZE = 1440
+PROCESSED_CHUNKS_2D = (270, 360)
+PROCESSED_CHUNKS_3D = (1, 270, 360)
+PROCESSED_CHUNKS_4D = (1, PROCESSED_LEVEL_SIZE, 270, 360)
+
```
jder, line 18: This was a drive-by cleanup.

YuanYuan98 (commented):
Looks good to me! The move to Dask-backed arrays with explicit chunks should reduce memory usage and make the dimensions easier to maintain.

Just one minor question: the `wetmask` now uses `random > 0.25`, whereas the previous `astype(bool)` version was effectively almost all `True`. I assume this change is intentional?

jder:
Yeah, good catch. This data isn't actually used so it doesn't matter much but yeah I think this makes more sense than "all true"

jder:
If you're happy with it, can you approve so it is merged? Thanks!

YuanYuan98 (approved):

PR #783, opened 2026-06-26
Remove data "schedules"

jder:
Our [current plan](https://docs.google.com/document/d/1Ckf-eGMCvQVYxkgiSHXwsPLN0fkr1wqjePY6TjWtIkk/edit?tab=t.gecl0m4fsarx) requires data sources which fall under the "match" or "standard" schedules but not "mix", and it was always a bit unclear what "mix" schedules would do for autoregression, so we remove them.

### DIFF
--- a/src/samudra/train.py
+++ b/src/samudra/train.py
@@ -4,7 +4,6 @@
 
 import contextlib
 import datetime
-import itertools
 import logging
 import multiprocessing
 import os
@@ -12,10 +11,9 @@ import tempfile
 import time
 import warnings
 from collections import OrderedDict
-from collections.abc import Iterable
 from multiprocessing.context import BaseContext
 from pathlib import Path
-from typing import Any, assert_never
+from typing import Any
 
 import dask
 import torch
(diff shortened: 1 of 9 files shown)


src/samudra/train.py
```diff
@@ -797,23 +785,12 @@ def init_data_loaders(self, cur_step: int) -> None:
...
-                srcs = list(itertools.product(scales, repeat=2))  # type: ignore
-            case _:
-                assert_never(self.train_schedule)
+        srcs = self.data_container.sources
 
         train_datasets = [
             TorchTrainDataset(
                 src=src.slice(self.train_time),
-                dst=dst.slice(self.train_time) if dst else None,
+                dst=None,
```
jder: @fomo-bot can we just get rid of `dst` and clean up related dead code now?

jder:
@fomo-bot also please update to latest main and fix pre-commit

jder:
@fomo-bot please take a look at the CI failure

alxmrs:
As far as I can tell, in my experiments with KR2, we also only use the standard or match schedule: https://github.com/m2lines/Samudra/blob/8f7ffa2ffdd705ad1420dce4e42038f394698572/configs/fomo_om4/train_multiscale.yaml#L49

Thus, I'm OK removing this for now. We could always add it back. YAGNI.

tests/test_datasets.py
```diff
@@ -92,7 +91,7 @@ def make_loader(
     time_config: TimeConfig | None = None,
     drop_last: bool = True,
     version: LoaderVersion | None = None,
-    schedule: TrainSchedule = "standard",
+    multiscale: bool = False,
```
alxmrs, line 94: Nice, that's very simple.

alxmrs (approved):
This LGTM, but maybe I'd like you to walk me through how it works; I'm a little confused reading the diff.

PR #788, opened 2026-07-09
Use serial ESMF in xESMF data tests

jder:
Force the conda-forge `nompi` ESMF build in the serial xESMF data-test workflow.

We are getting [intermittent](https://github.com/m2lines/Samudra/actions/runs/29047813316) [failures](https://github.com/m2lines/Samudra/actions/runs/29041882608) for these tests which show up as SIGTERMs. I don't believe it's an out-of-memory issue (the test data is tiny) and can't reproduce these locally. Running under Valgrind Memcheck in docker environments produced many errors during UCX TCP reachability discovery in:

```text
uct_tcp_iface_is_reachable_v2
ucp_wireup_select_transport
MPIDI_UCX_init_local
MPI_Init_thread
ESMCI::VMK::init
ESMF_Initialize
```

As a control, the same MPI initialization under Valgrind with `UCX_TLS=self` reported 0 errors. That isolates the finding to unnecessary network transport discovery.

Python/glibc allocator-debug stress using `PYTHONMALLOC=debug`, `MALLOC_CHECK_=3`, and `MALLOC_PERTURB_` also completed 500 regrids on both builds without detecting heap corruption.

All together, this is not really proof that the SIGTERM on GH is caused by this (since we can't repro locally to get more info) but seems worth trying.


alxmrs (approved):
Thanks, this seems annoying/hard to find. LGTM

PR #789, opened 2026-07-09
Disable flash encoder latent rotary

jder:
Disable the FlashPerceiver encoder's latent rotary embedding by no longer passing `latent_rotary_emb_dim` from Samudra's config builder. This does not seem likely to be super important, since we still add 2D Fourier features to input patch tokens, and the encoder/decoder still add the physical patch position and scale encodings.

We want to do this because on ORCD H200 runs, `samudra-multi` is crashing (CUDA memory fault) in FlashAttention's Triton rotary kernel. I tried some variants of the configuration which seem to work but I would rather just disable and avoid future related trouble given it seems unnecessary.


tests/test_config.py
```diff
@@ -136,6 +140,32 @@ def test_train_config_allows_cli_override_for_cpu_num_workers(tmp_path):
     assert cfg.data.loading.num_workers == 2
 
 
+def test_flash_encoder_perceiver_does_not_enable_latent_rotary(monkeypatch):
```
jder: @fomo-bot I don't think we need a test for this.

Can you also change the the naive perceiver to have equivalent behavior here if it does not already?

jder: @fomo-bot also please fix pre-commit

jder: @fomo-bot why are you still passing num_freq_bands & max_freq then?

alxmrs (approved):
I like this change, we want to have this level of control for something as important as pos encoding.

PR #790, opened 2026-07-10
implement group-norm

GHOpenonic:
implement group and layer normalization with tests. Change norm selection to use build_norm_layer that is more extensible than old if/elif logic.

### DIFF
--- a/src/samudra/config.py
+++ b/src/samudra/config.py
@@ -336,7 +336,7 @@ class DataConfig(BaseConfig):
 
 BlockType = Literal["conv_next_block", "conv_block"]
 ActivationType = Literal["relu", "gelu", "capped_gelu"]
-NormType = Literal["batch", "instance", "layer"]
+NormType = Literal["batch", "instance", "group", "nonorm", "layer"]
 
 
 class BlockConfig(BaseConfig):
@@ -345,6 +345,7 @@ class BlockConfig(BaseConfig):
     activation: ActivationType = "capped_gelu"
     upscale_factor: int = 4
     norm: NormType = "batch"
+    group_norm_groups: int = Field(default=32, ge=1)
     pointwise_linear: bool = False
 
     def build(self) -> CoreBlockBuilder:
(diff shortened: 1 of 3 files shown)


src/samudra/models/modules/blocks.py
```diff
@@ -343,6 +338,30 @@ def __init__(
         self.convblock = torch.nn.Sequential(*convblock)
         self.checkpoint_simple = checkpoint_simple
 
+    @staticmethod
+    def _build_norm_layer(
```
alxmrs, line 342: It looks like this doesn't use any class members, therefore I think it could be a pure function not in the class rather than a method. Further, I think this could be generally useful -- how about we make it a public function?

src/samudra/models/modules/blocks.py
```diff
@@ -343,6 +338,30 @@ def __init__(
...
+        norm: str, channels: int, group_norm_groups: int
```
alxmrs, line 343: I think the parameter could just be `groups`? 

🐑 maybe it could be an optional param since it's only used with group norm.

src/samudra/models/modules/blocks.py
alxmrs, line 342: 🐑  idea: potentially, we could pass in to the core block a "NormBuilder" so we could set all of the hparams for which norm to use in our config system -- see the `CoreBlockBuilder(Protocol)` at the bottom of this file.

src/samudra/config.py
```diff
@@ -336,7 +336,7 @@ def make_source(
 
 BlockType = Literal["conv_next_block", "conv_block"]
 ActivationType = Literal["relu", "gelu", "capped_gelu"]
-NormType = Literal["batch", "instance", "layer"]
+NormType = Literal["batch", "instance", "group", "nonorm", "layer"]
```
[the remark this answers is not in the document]
alxmrs: See my 🐑 suggestion above about the NormBuilder -- that idea could help address this concern.

tests/test_blocks.py
```diff
@@ -18,7 +19,7 @@
     height=st.integers(min_value=1, max_value=16),
     width=st.integers(min_value=1, max_value=16),
 )
-@settings(max_examples=20)
+@settings(max_examples=20, deadline=None)
```
alxmrs, line 22: I think we should remove this -- if there's a timeout in our property based tests, that's a concern worth knowing about.

alxmrs (commented):
A few ideas that could improve this group norm PR.

jder:
I have included this in #671 which I'm planning to merge soon.

PR #673, opened 2026-04-08
Harden distributed training runtime

jder:
## Summary
This is PR 6 in the LLC stack, based on #672.

It pulls the generic trainer and distributed-runtime hardening to the end of the stack: DDP worker scaling, no-sync/static-graph handling, slow-batch warnings, emergency checkpoint handling, and related trainer robustness updates.

## Why
These changes are useful beyond LLC, but they are easier to review once the dataset and GPU-decode changes are already isolated. Moving this PR to the end keeps the earlier LLC/data review focused.

## Notes
This remains a draft stacked PR.

PR #792, opened 2026-07-10
implement core replay buffer logic

GHOpenonic:
add replay.py script and test. This does not add replay buffer as a training option, it only implements the core logic. Wiring it into training should be based on the LLC data munging stack #670, #671, #672, #673


jder:
@GHOpenonic can you point me to where this is used in your code? It's a bit hard to understand how these pieces are put together from this alone.

GHOpenonic:
@jder most of it is used/further developed in datasets.py, config.py, base.py, stepper.py, and train.py in [replay buffer agent attempt](https://github.com/m2lines/Samudra/commit/a725ad944e339bbbb1b2e5f11401c7a930aac05f). 
replay.py is updated and with implementations in datasets.py and train.py in [slow replay buffer](https://github.com/m2lines/Samudra/commit/15616d23f1b45472ddb592815ef1e2b21c6cb92f)
I increase replay's dataloading/handling speed to be roughly gpu-compute bound by changing datasets.py and train.py in [improve replay buffer train speed --> gpu bound](https://github.com/m2lines/Samudra/commit/8ce9464d0cefff9a81a0bc2c9684b01c0a377cbd). 
Additionally, if you're curious, I make further changes to all these scripts and replay.py in the Jul 19, 2026 commits in llc_cpu_working, mostly [shardtensor implementation phase 3: replay buffer](https://github.com/m2lines/Samudra/commit/2f46214a82f2aef3cbf3a807d684bb68d3076d68), to get shardtensor to work with replay buffer training

jder:
ok, thanks. I think unless someone needs this sooner (@amogh-gulati maybe?) I vote we stage this after we land the LLC work #670 and rust data loader #800, since building this on top of those probably makes sense.

amogh-gulati:
> ok, thanks. I think unless someone needs this sooner (@amogh-gulati maybe?) I vote we stage this after we land the LLC work #670 and rust data loader #800, since building this on top of those probably makes sense.

I could work on the branch meanwhile

src/samudra/replay.py
```diff
@@ -0,0 +1,250 @@
...
+    def reserved_indices(self) -> set[int]:
+        return {
+            slot.replay_index
+            for slot in (*self.train_slots, *self.seed_slots)
+        }
+
+
+@dataclasses.dataclass
+class ReplayEntry:
+    state: torch.Tensor
```
alxmrs, line 57: Is there any way we could get more type information for what the state is, maybe via jaxtyping?

alxmrs: e.g. is this a `Prognostic`?

src/samudra/replay.py
```diff
@@ -0,0 +1,250 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class ReplayCursor:
+    dataset_index: int
+    source_index: int
+    lead_step: int
+    stride: int
+    temporal_stride: int
+
+    def advance(self) -> "ReplayCursor":
```
alxmrs, line 23: nit: 
```suggestion
    def advance(self) -> Self:
```

src/samudra/replay.py
```diff
@@ -0,0 +1,250 @@
...
+class ReplayBatchSlot:
+    replay_index: int
+    cursor: ReplayCursor
+
+
+@dataclasses.dataclass(frozen=True)
+class ReplaySeedSlot:
+    replay_index: int
+    cursor: ReplayCursor
+    reason: str
```
alxmrs, line 37: 🐑 IIUC these are related concepts. Maybe the seed slot could inherent from the batch slot, and then add the one additional variable? Then, I wonder if we could use a single tuple instead of two tuples in the ReplayBatchRequests (targeting the base class).

src/samudra/replay.py
```diff
@@ -0,0 +1,250 @@
...
+        storage_dtype: torch.dtype,
+        generator: torch.Generator,
+        pin_memory: bool,
+    ) -> None:
+        if buffer_size < 1:
+            raise ValueError("Replay buffer_size must be >= 1")
+        self.buffer_size = buffer_size
+        self.storage_dtype = storage_dtype
+        self.generator = generator
+        self.pin_memory = pin_memory
```
alxmrs, line 76: Is this used similar to a `Batch` object by torch (e.g. a `TrainData`)? If so, I would recommend adding `to` and `pin_memory` methods. If these two methods exists, the torch framework will try to call them as needed.

src/samudra/replay.py
```diff
@@ -0,0 +1,250 @@
...
+                device="cpu",
+            )[:batch_size]
+        else:
+            draw = torch.randint(
+                len(eligible),
+                (batch_size,),
+                generator=self.generator,
+                device="cpu",
+            )
+        return [eligible[int(i)] for i in draw]
```
alxmrs, line 126: Is it possible that `sample_indicies` could call `random_indices`? Or, can this block (which seems to be shared with that method below) be extracted into a common, private method?

src/samudra/replay.py
```diff
@@ -0,0 +1,250 @@
...
+            cursor = ReplayCursor(**raw_entry["cursor"])
+            self.append(ReplayEntry(state=raw_entry["state"], cursor=cursor))
+
+    @staticmethod
+    def _state_for_checkpoint(entry: ReplayEntry) -> torch.Tensor:
+        if entry.ready_event is not None:
+            entry.ready_event.synchronize()
+        return entry.state.cpu()
+
+    def _prepare_entry(self, entry: ReplayEntry) -> ReplayEntry:
```
alxmrs, line 193: See my note about `to` and `pin_memory` -- we may get to omit this method.


https://docs.pytorch.org/docs/2.14/data.html#memory-pinning:~:text=To%20enable%20memory%20pinning%20for%20custom%20batch%20or%20data%20type(s)%2C%20define%20a%20pin_memory()%20method%20on%20your%20custom%20type(s).

alxmrs (changes requested):

PR #793, opened 2026-07-12
add shardtensor-op-probe

GHOpenonic:
add a script to check that shardtensor supports LLC emulator-specific basic operations. Future implementation of shardtensor for domain parallelism into the codebase will sit in llc_cpu_working until LLC data munging stack #670, #671, #672, #673 are further implemented.


GHOpenonic:
@codex review

PR #800, opened 2026-07-16
Add opt-in Rust OM4 data loader

jder:
## Summary

- Add an opt-in local `data.loading.type: rust` path for flat and compact OM4 Zarr stores behind the new `CanonicalDataset` / `TrainingShard` boundaries.
- Keep Python responsible for canonical semantics, sampling, DDP schedules, masking, normalization, and batch contracts while Rust owns persistent native reads, rollout-wide plane deduplication, bounded prefetch, pinned-buffer reuse, and explicit lifecycle cleanup.
- Present the change as five reviewable commits: canonical datasets, deterministic schedules, training-shard/batch preparation, explicit batch compatibility, and one atomic Rust backend integration.

## Why

The current CPU loader can spend minutes in xarray/Python data-loading cold paths and shows large periodic data-wait stalls during realistic quarter-degree training. The Rust path keeps local Zarr readers open, loads each unique time/channel plane once, reuses pinned host buffers, and overlaps host reads plus CUDA batch preparation with model execution without changing sampler or `dataset_id` semantics.

## Scope

This is opt-in and local-filesystem only. It supports training and validation for flat OM4 and OM4-compact. S3, LLC-specific canonicalization, inference, derived seasonal-climatology/anomaly channels, and changing heterogeneous batch semantics remain out of scope.

## Quarter-degree two-GPU speed check

- Full v2-highres model, 84M params, qdeg OM4 v2, two RTX6000s on Torch `gr101`, batch size 1, gradient accumulation 4, rollout steps `[4]`, boundary vars `tauuo,tauvo,hfds`.
- CPU epoch 1 came from job `14115987`; the attempted two-epoch CPU/Rust job was stopped because CPU validation and epoch-2 cold loading would not leave time for the Rust half in the one-hour allocation.
- Rust epoch 1 came from corrected job `14117615`, submitted immediately afterward on the same node with the same container and `NCCL_P2P_DISABLE=1`.

| Loader | Train epoch total | Mean iter excluding step 0 | Mean iter excluding step 0 and data waits >=1s | Validation total | Peak CPU / GPU memory |
| --- | ---: | ---: | ---: | ---: | ---: |
| CPU | 22.31 s/it | 14.59 s | 4.12 s | 81.61 s/it | 41.7 GB / 49.1 GB |
| Rust | 4.13 s/it | 3.24 s | 3.21 s | 0.59 s/it | 16.3 GB / 53.9 GB |

This is a 5.4x train-epoch improvement and 139x validation improvement for this one-epoch qdeg run. The non-stall training number is 1.3x faster, so the main win is removing CPU-loader cold loads and periodic stalls rather than changing steady GPU compute.

## Validation

- Focused native/canonical/sampler suite: 85 passed.
- Standalone native-extension suite: 23 passed.
- Rust format, clippy, and unit tests: passed; 3 Rust unit tests.
- All pre-commit checks, including mypy and schema validation: passed.
- Broad non-manual/non-CUDA local run: 349 passed, 2 skipped, 10 xfailed; six setup errors were the known local FlashAttention `libcudart.so.12` environment issue.

## Notes for review

- The native OM4 canonical reader is composed: xarray remains the semantic/static/reference delegate, while the native capability serves optimized batch plane reads. No canonical manifest is introduced.
- Compatibility groups preserve first-seen dataset order, keeping DDP schedules rank-stable without changing process-local `dataset_id` semantics.
- Native physical rows are mapped through the store time coordinate, including when a canonical dataset is sliced before native decoration.
- Rust iterators use weak loader ownership for deterministic early-exit cleanup, and Trainer explicitly tears down training, validation, and inference workers.
- The generic loader factory and native loader constructor switch together because splitting that change would require a temporary `TorchTrainDataset` compatibility adapter solely to delete it in the next commit.


jder:
Oh, one other thing about this: it's probably worth refreshing the torch docs in repo after this lands to reflect the (new, much smaller) resource requirements and resulting torch flexibility in general (e.g. preemptible compute is much more available with smaller requests)

.github/workflows/test-rust-loader.yml
```diff
@@ -0,0 +1,83 @@
...
+          python-version: ${{ matrix.python-version }}
+
+      - name: Set up Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          cache: false
+          rustflags: ""
+
+      - name: Cache Rust build artifacts
+        uses: Swatinem/rust-cache@v2
```
jder: Let's pin this to a hash

.github/workflows/test-rust-loader.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Check Rust
+        run: |
+          cargo fmt --manifest-path rust/crab_load/Cargo.toml -- --check
+          cargo clippy --manifest-path rust/crab_load/Cargo.toml --features extension-module --lib -- -D warnings
+          cargo test --manifest-path rust/crab_load/Cargo.toml --features extension-module --lib
+
+      - name: Build and test the Python extension
+        run: |
+          uv venv --python ${{ matrix.python-version }}
+          uv pip install "maturin>=1.9,<2" "numpy>=1.26.4,<2" pytest xarray "zarr<3"
```
jder: Is there a way we can avoid having these here instead of in the pyproject?

.github/workflows/test-rust-loader.yml
```diff
@@ -0,0 +1,83 @@
...
+          source .venv/bin/activate
+          maturin develop --manifest-path rust/crab_load/Cargo.toml --features extension-module
+          python -m pytest -o addopts='' rust/crab_load/tests/test_extension.py -q
```
jder: What is addopts doing here?

.github/workflows/test-rust-loader.yml
```diff
@@ -0,0 +1,83 @@
...
+          .wheel-test/bin/python - <<'PY'
+          from importlib.metadata import requires, version
+
+          import samudra_rust_loader
+
+          assert version("samudra-rust-loader") == "0.1.0"
+          assert any(
+              requirement.startswith("samudra-rust-loader==0.1")
+              for requirement in requires("samudra") or []
+          )
```
jder: This seems fragile and unnecessary

.github/workflows/test.yml
```diff
@@ -44,8 +44,14 @@ jobs:
...
 
+      - name: Set up Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          cache: false
+          rustflags: ""
+
       - name: Install dependencies
-        run: uv sync --locked --dev
+        run: uv sync --locked --dev --extra rust
```
jder: Do we need this if we're testing it separately? Might be nice to not have to build rust for quick CI

docs/rust-data-loader-plan.md
jder: I don't think we need a "plan" here but would be nice to extract the durable bits of this that are useful as documentation for what the rust loader is.

rust/crab_load/tests/test_extension.py
```diff
@@ -0,0 +1,371 @@
...
+            "zos": (("time", "y", "x"), surface + np.float32(2000)),
+            "hfds_anomalies": (
+                ("time", "y", "x"),
+                surface + np.float32(3000),
+            ),
+        }
+    )
+    path = tmp_path / "compact-om4.zarr"
+    dataset.to_zarr(path, mode="w", consolidated=False)
+    return path, dataset
```
jder: Surely we already have fixtures for OM4 test stores, do we need these new ones as well? If so can you comment why?

rust/crab_load/tests/test_extension.py
```diff
@@ -0,0 +1,371 @@
...
+        reader.read_into([0] * read_count, ["first"], output)
+
+    with ThreadPoolExecutor(max_workers=2) as executor:
+        futures = [executor.submit(read) for _ in range(2)]
+        errors = [future.exception() for future in futures]
+
+    assert sum(error is None for error in errors) == 1
+    error = next(error for error in errors if error is not None)
+    assert isinstance(error, RuntimeError)
+    assert "could not be borrowed for writing" in str(error)
```
jder: This looks flaky to me, no? Can we test this at the rust level in a more reliable way?

rust/crab_load/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "samudra-rust-loader"
+version = "0.1.0"
+edition = "2021"
```
jder: Please update this + rust version below. (Or can we get rid of the latter given the rust-toolchain.toml? I don't remember how those interact)

rust/crab_load/README.md
```diff
@@ -0,0 +1,172 @@
+<!--
+SPDX-FileCopyrightText: 2026 Samudra Authors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Samudra Rust data loading
+
+The optional `samudra-rust-loader` extension accelerates local OM4 training and
```
jder: Can we rename the crab_load dir name + other references to match this please? `rust/loader` maybe?

rust/crab_load/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+```mermaid
+flowchart LR
+    S["Python sampler and DDP rank schedule"] --> H["One bounded host-prefetch producer"]
+    H --> R["Shared bounded Rayon read pool"]
+    R --> P["Reusable pinned unique-plane buffers"]
+    P --> C["One H2D copy per unique plane"]
+    C --> N["Normalize and mask unique planes"]
+    N --> G["Device gather into ModelBatch"]
+    G --> M["Current model stream"]
```
jder: This is a pretty silly diagram given it is a linear list. Perhaps we could show host/gpu boundaries, boundaries between python/rust, the set of pools/concurrency, or just have a brief overview of the below?

rust/crab_load/README.md
```diff
@@ -0,0 +1,172 @@
...
+for its training and validation loaders. Those loaders have separate
+pinned-buffer pools but share the rank's bounded native read pool. There are no
+PyTorch `DataLoader` worker processes on the Rust path.
+
+## Buffer lifetime and memory bounds
+
+### Where pinning happens
+
+Pinning happens before the Zarr read, not as a copy after loading. CUDA device
+prefetch forces pinning on even if the top-level training `pin_mem` setting is
```
jder: What is "CUDA device prefetch"? Does this just mean `prefetch_batches`?

rust/crab_load/README.md
```diff
@@ -0,0 +1,172 @@
...
+writes each plane directly into that pinned destination. Logical rollout maps
+remain small CPU index tensors. There is no intermediate pageable batch and no
+Python collation copy. The native decoder still has bounded per-read scratch
+before copying a decoded plane into its pinned location.
+
+When the CUDA-prefetch iterator consumes the batch, `Tensor.to(device,
+non_blocking=True)` queues one host-to-device copy per unique read group.
+Normalization and masking run on the unique device planes. `index_select` then
+duplicates processed planes into the existing `ModelBatch` input, boundary, and
+label layout. Thus overlapping rollout values cross PCIe once and are copied
```
jder: PCIe -> "from CPU to GPU"

rust/crab_load/README.md
```diff
@@ -0,0 +1,172 @@
...
+device-to-device only after preprocessing. The host producer can concurrently
+fill another set of pinned buffers while the model stream works.
+
+### When a pinned buffer can be reused
+
+After device preparation is queued, the loader records a CUDA event on the
+prefetch stream and releases the raw host tensors to the pool with that event.
+This is conservative: the event follows both the H2D copies and the queued
```
jder: "This could be tightened:"

src/samudra/utils/samplers.py
```diff
@@ -120,7 +127,8 @@ def from_datasets(
 
         Args:
             datasets: List of TorchTrainDataset instances to group
-            group_key: Callable that extracts grouping key from a dataset.
+            group_key: Optional legacy override for the dataset's public
+                ``batch_compatibility_key``.
```
jder: Can we rm this or do we actually need it?

src/samudra/data_backend.py
```diff
@@ -0,0 +1,122 @@
...
+        self._max_concurrent_reads = max_concurrent_reads
+        self._runtime: RustIoRuntime | None = None
+
+    def validate_locations(
+        self,
+        *,
+        data_location: ResolvedLocation,
+        means_location: ResolvedLocation,
+        stds_location: ResolvedLocation,
+        source_type: str,
```
jder: Can we get a Literal type for this please

src/samudra/data_backend.py
```diff
@@ -0,0 +1,122 @@
...
+
+        from samudra.rust_data import create_rust_io_runtime, native_om4_source
+
+        assert isinstance(data_location, LocalLocation)
+        if self._runtime is None:
+            self._runtime = create_rust_io_runtime(self._max_concurrent_reads)
+        return native_om4_source(source, data_location, self._runtime)
+
+
+def build_training_source_backend(
```
jder: Can this be a member on BaseDataLoadingConfig instead of a type check? Ok to make the backends "public"

rust/crab_load/src/lib.rs
```diff
@@ -0,0 +1,785 @@
...
+        bail!("time index must be non-negative, got {index}");
+    }
+    let index = index as u64;
+    if index >= time_len {
+        bail!("time index {index} is out of bounds for length {time_len}");
+    }
+    Ok(index)
+}
+
+fn dimension_names(array: &OpenArray) -> anyhow::Result<Vec<String>> {
```
jder: Are there docs we could point to about this format and the hard-coded name above?

rust/crab_load/src/lib.rs
```diff
@@ -0,0 +1,785 @@
...
+            "array {} has dimensions {names:?}; expected {expected}",
+            array.path()
+        );
+    }
+    Ok(axes.expect("axes was checked above"))
+}
+
+/// Shared bounded Rayon pool for every flat-OM4 reader in one training process.
+#[pyclass]
+struct FlatOm4ReadPool {
```
jder: This isn't just for Flat, right? (And probably not just for Om4 eventually)

rust/crab_load/src/lib.rs
```diff
@@ -0,0 +1,785 @@
...
+    #[new]
+    fn new(max_concurrent_reads: usize) -> PyResult<Self> {
+        if max_concurrent_reads == 0 {
+            return Err(python_error(anyhow::anyhow!(
+                "max_concurrent_reads must be positive"
+            )));
+        }
+        let thread_pool = ThreadPoolBuilder::new()
+            .num_threads(max_concurrent_reads)
+            .thread_name(|index| format!("samudra-zarr-{index}"))
```
jder: "samudra-rust-loader-{index}"

rust/crab_load/src/lib.rs
```diff
@@ -0,0 +1,785 @@
...
+                    Ok(())
+                })
+        })?;
+
+        Ok(())
+    }
+}
+
+#[pymethods]
+impl CompactOm4Reader {
```
jder: I find the split impl blocks for this and the flat reader confusing more than helpful, can we have a single one each?

rust/crab_load/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+rust-version = "1.97.1"
+license = "Apache-2.0"
+publish = false
+
+[lib]
+name = "samudra_rust_loader"
+crate-type = ["cdylib", "rlib"]
```
jder: Why both? Is this a maturin thing?

src/samudra/utils/distributed.py
```diff
@@ -85,6 +85,12 @@ def is_main_process():
     return get_rank() == 0
 
 
+def destroy_distributed_mode() -> None:
```
jder: This seems to be unused? What am I missing?

src/samudra/utils/samplers.py
```diff
@@ -144,23 +152,18 @@ def from_datasets(
...
         cumsum = 0
         for ds in datasets:
-            key = group_key(ds)
+            key = group_key(ds) if group_key is not None else ds.batch_compatibility_key
             assert isinstance(key, Hashable), "`group_key` must be hashable."
-            groups[key].extend(range(cumsum, cumsum + len(ds)))
+            groups.setdefault(key, []).extend(range(cumsum, cumsum + len(ds)))
             cumsum += len(ds)
 
-        # Sort by key for deterministic ordering across runs
```
jder: Why did this go away?

src/samudra/data_backend.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def prepare(
+        self,
+        source: CanonicalSource,
+        *,
+        data_location: ResolvedLocation,
+        source_type: str,
+    ) -> CanonicalSource:
+        assert source_type == "om4"
+        derived = [
```
jder:
```suggestion
        # check for known computed vars to give a better error message
        derived = [
```

src/samudra/data_backend.py
```diff
@@ -0,0 +1,122 @@
...
+            name
+            for name in source.data_layout.boundary_var_names
+            if name.endswith("_anomalies")
+        ]
+        if derived:
```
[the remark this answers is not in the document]
jder: Yeah ok maybe we could discuss the design here. There are definitely some smells like this, that prepare takes & returns a CanonicalSource, and that in native_om4_source we're building a source and then mutating its reader, _native_reader's type checking, etc. Can we discuss if there is a broken or missing abstraction here?

jder (commented):
Still need to look at config.py, data_backend.py, datasets.py, rust_data.py, train_data_loader.py, and train.py, but here are some initial comments.

PR #803, opened 2026-07-18
Restore primary-source validation metrics

jder:
Validating against the primary source only is not ideal, but it seems better than no validation at all (and is sufficient for several of our upcoming goals), so seems like a reasonable step in the right direction.


alxmrs (approved):
man, I spent so much time trying to understand the aggregator's validation system to make the original change. I'm happy to simplify this as an interum comparison. Maybe, it's time to rethink the aggregator/validators in the first place, given all we know and where the project is headed.

PR #805, opened 2026-07-18
Avoid single-node torchrun rendezvous port collisions

jder:
The training harnesses defaulted every job to the same static rendezvous port (`8008` or `29500`). Concurrent jobs placed on the same host therefore raced to bind the same TCP port and one failed during distributed initialization.

For single-node launches, `torchrun --standalone` asks the launcher to choose a free port atomically, removing the shared static default. Multi-node launches still require a predetermined endpoint shared by every node.


alxmrs (approved):
Modulo instead of hard coding. SGTM! Thanks, this sounds annoying to debug.

PR #806, opened 2026-07-18
Cap multires dynamic loss and reject non-finite gradients

jder:
## Summary

- cap the Samudra Multi dynamic-loss weight ratio at 20×, matching the existing v2 training configs
- make gradient clipping raise on a NaN or infinite total norm before `optimizer.step()`

## Why

[W&B run `x285qor8`](https://wandb.ai/ocean_emulators/default/runs/x285qor8?nw=nwuseroa_jder) diverged at history step 12903 (epoch 13, batch 230). Batch 229 was finite, then every logged loss, unscaled loss, and dynamic loss scale became NaN and remained non-finite.

The run used an uncapped dynamic loss. Its largest channel scale reached roughly 4,213×. The trainer also used PyTorch's default `error_if_nonfinite=False` behavior when clipping gradients, so it could still call `optimizer.step()` after a non-finite total norm and continue training with contaminated weights.

This change bounds the affected multires training configuration and turns the same failure into an immediate, noisy exception before the optimizer update.

### DIFF
--- a/src/samudra/train.py
+++ b/src/samudra/train.py
@@ -90,17 +90,6 @@ from samudra.utils.wandb import WandBLogger
 logger = logging.getLogger(__name__)
 
 
-def _clip_grad_norm_or_raise(
-    parameters: Iterable[torch.Tensor], max_norm: float
-) -> torch.Tensor:
-    """Clip gradients and fail before the optimizer can apply a non-finite update."""
-    return torch.nn.utils.clip_grad_norm_(
-        parameters,
-        max_norm,
-        error_if_nonfinite=True,
-    )
-
-
 def should_log_validation_images(epoch: int, frequency: int) -> bool:
     """Return whether to log validation images for a 1-based training epoch."""
     if epoch < 1:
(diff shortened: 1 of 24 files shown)


src/samudra/train.py
```diff
@@ -90,6 +90,17 @@
...
+    parameters: Iterable[torch.Tensor], max_norm: float
+) -> torch.Tensor:
+    """Clip gradients and fail before the optimizer can apply a non-finite update."""
+    return torch.nn.utils.clip_grad_norm_(
+        parameters,
+        max_norm,
+        error_if_nonfinite=True,
+    )
+
+
```
jder: just inline this

tests/test_config.py
```diff
@@ -136,6 +137,28 @@ def test_train_config_allows_cli_override_for_cpu_num_workers(tmp_path):
...
+            str(tmp_path),
+            "--experiment.base_output_dir",
+            str(tmp_path / "outputs"),
+        ]
+    )
+
+    assert isinstance(cfg.loss, DynamicLossConfig)
+    assert cfg.loss.limit == 20
+
+
```
jder: we don't need a test for this, or generally to confirm our configs are loaded correctly except for general config-specific tests

jder (commented):
@fomo-bot a couple things

src/samudra/train.py
```diff
@@ -541,7 +541,11 @@ def train_one_epoch(self, epoch):
             should_step = (data_iter_step + 1) % self.gradient_accumulation_steps == 0
             # Step optimizer after accumulating enough batches or at the end
             if should_step or is_last:
-                torch.nn.utils.clip_grad_norm_(self.model.parameters(), 1.0)
+                torch.nn.utils.clip_grad_norm_(
+                    self.model.parameters(),
+                    1.0,
+                    error_if_nonfinite=True,
```
alxmrs, line 547: Nice to raise this error if needed.

alxmrs (approved):
LGTM.

PR #811, opened 2026-07-21
Fix FlashAttention wheels across CUDA platforms

### LINKED ISSUE #710: Fix flash-attn on ARM CUDA 13 and re-enable FOMO CUDA scrub tests
## Context

PR #539 adds the ARM64 PhysicsNeMo container path and nightly local ARM GPU scrub. During local validation on the ARM GPU host, the FOMO CUDA tests that require `flash_attn`/`flash_perceiver` could not run reliably on the bare-metal ARM CUDA 13 environment.

The observed failure was an import/runtime mismatch for the flash-attention stack, including:

```text
ImportError: libcudart.so.12: cannot open shared object file
```

The failing path included `test_inference__data_shape[test/train_fomo.yaml-compact-cuda-extra_config_args0]` before the scrub was limited to non-FOMO CUDA tests.

## Current workaround

PR #539 keeps the product behavior strict: when CUDA is available and `perceiver_implementation=auto`, FOMO still selects flash attention and raises if the flash dependencies are broken.

The nightly ARM container scrub temporarily excludes `train_fomo` CUDA cases with `-k 'not train_fomo'` so the ARM image can still be smoke-tested for the CUDA coverage that does not require FOMO flash attention.

## Done when

- `flash_attn` and `flash_perceiver` import and run on the ARM CUDA 13 / GB10 environment used by the local scrub.
- The FOMO CUDA tests pass on the ARM GPU host.
- The `train_fomo` exclusion is removed from `scripts/container/scrub_arm64_gpu_container_tests.sh`.
- The ARM scrub skill/docs are updated to reflect the restored coverage.

jder:
## Summary

- constrain supported Python to 3.12 since that is all we're currently testing
- select official upstream FlashAttention wheels by Linux architecture (see comments for rationale):
  - CUDA 13 / PyTorch 2.9 for ARM64 PhysicsNeMo hosts
  - CUDA 12 / PyTorch 2.9 for existing x86_64 CUDA jobs
- use `perceiver_implementation: auto` in general model configs and explicitly exercise `flash` in the CUDA SamudraMini smoke config

Fixes #710 for tests/training runs *outside* containers on ARM hosts (e.g. my dev box)


configs/test/train_samudra_mini.yaml
```diff
@@ -60,7 +60,8 @@ model:
   pred_residuals: false
   last_kernel_size: 3
   pad: "circular"
-  perceiver_implementation: "naive"
+  # This is used in a CUDA smoke test, so we want to exercise the FlashAttention backend.
```
alxmrs, line 63: SGTM

alxmrs (approved):
LGTM. Thanks for specifying each sys_platform, that's a useful uv config.

PR #814, opened 2026-07-22
Make mask matching test deterministic

jder:
The test built a 24-element mask from random values using `data_3d > 0.25`. There was a roughly 0.10% chance that every mask element was `True`; in that case the unmasked dataset correctly matched the mask and the expected `ValueError` was not raised. This caused [this CI failure](https://github.com/m2lines/Samudra/actions/runs/29930540130/job/88958781028?pr=670). We now use deterministic data.


alxmrs (approved):
Thanks, LGTM

PR #766, opened 2026-06-17
Harden GitHub Actions workflows and pin actions

jder:
## Summary
- Narrow workflow and job-level permissions to the minimum required for each CI path
- Pin third-party GitHub Actions to commit SHAs across test, benchmark, pages, and container workflows
- Keep benchmark publishing and GHCR publishing permissions scoped to the jobs that need them

## Testing
- Not run (not requested)

PR #817, opened 2026-07-22
Add PyPI release workflow with weekly and stable channels

alxmrs:
## Summary

Publishes `samudra` to [PyPI](https://pypi.org/project/samudra/) as a single pure-Python wheel via a new **Release** workflow, modeled on [marin's release workflow](https://github.com/marin-community/marin/blob/main/.github/workflows/marin-release-libs-wheels.yaml) (`resolve → build → publish`) and adapted to a single-package repo. Publishing is **gated on the CPU test suite**.

### Release channels

| Trigger | Mode | Version | Published? |
| --- | --- | --- | --- |
| Push a `v*` tag | stable | the tag, e.g. `v0.0.1` → `0.0.1` | ✅ PyPI |
| `schedule` (Mon 06:00 UTC, weekly) | nightly | `<next-patch>.dev<YYYYMMDDhhmm>` | ✅ PyPI |
| `workflow_dispatch` | nightly / stable / smoke | computed (nightly/smoke) or the version input (stable only) | ✅ (not smoke) |
| Pull request touching the script/workflow | smoke | — | ❌ build-only |

The scheduled dev release is **weekly** rather than daily — the repo doesn't turn over enough in a day to warrant one, and the timestamp still makes each build unique. A hand-triggered **nightly** is supported: choose `mode: nightly` and leave the version field blank to build and publish a real `0.0.1.dev<stamp>`. The `smoke` mode is a build-only check (no publish); it names what the build *does*, not how it's triggered — every mode can be started by hand from **Actions → Release → Run workflow**.

### Test-gated publishing

Every publish (stable **and** nightly) waits on the CPU test suite. `test.yml` was made reusable (`workflow_call`), and the release workflow's `publish` job `needs` a `test` job that invokes it — so a red suite blocks the upload. A tag push doesn't otherwise run `test.yml`, which is why the release workflow calls it itself. (Branch-protection required-checks are intentionally left untouched; the in-workflow gate covers the release path.)

### `samudra` console script

Installing now exposes a `samudra` command (`[project.scripts]` → `samudra.cli:main`) that dispatches `samudra train|eval|viz CONFIG ...` to the same entry points as `python -m samudra.train`, so an installed user can run a task against **their own** config without a checkout. The example presets under `configs/` are **not yet shipped** in the wheel (our src-layout only packages files under `src/samudra/`); relocating the `configs/` tree so `samudra train samudra_om4/train.yaml` resolves *by name* is a self-contained refactor left as a documented follow-up.

### Versioning (setuptools-scm, hybrid)

Version is delegated to **setuptools-scm**: a git tag *is* the version, so there's no `version =` field to maintain and `samudra.__version__` works at runtime. `scripts/package.py` only computes the two synthetic versions (timestamped nightly, `+smoke` build) and injects them through scm's own `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_SAMUDRA` override — it never mutates a tracked file.

The timestamped nightly guarantees a unique, PEP 440-ordered dev release **even on weeks with no new commits**, which setuptools-scm's commit-distance schemes cannot provide. (Cutting the timestamp logic *into* a custom `version_scheme` would either reintroduce same-commit collisions or break scm's same-commit-→-same-version determinism, so it stays in the script.)

Before the first tag, the "last release" falls back to `0.0.0` (`fallback_version` in `[tool.setuptools_scm]`, mirrored by `FALLBACK_VERSION` in the script), so builds target `0.0.1`.

### Publish safety

A dispatched **nightly**/**smoke** run ignores the version input field — only a stable cut consumes it. This closes a footgun where a stray value left in the field would otherwise be stamped verbatim and published as a bogus non-dev release, permanently burning a PyPI version. As a backstop, the script also refuses to build a nightly without a `.dev` suffix or a smoke without a `+smoke` segment, failing loudly before anything is emitted or published.

### CPU vs GPU

Samudra is pure Python, so **one universal `py3-none-any` wheel serves everyone**. CPU is the default (`pip install samudra`); the GPU custom kernels (flash-attn, flash-perceiver, torchvision) stay in the existing `cuda` extra and compile on the user's machine — nothing GPU-specific enters the wheel.

## Security / hardening (per #766)

- **OIDC trusted publishing** — no PyPI token stored; publish job scoped to the `pypi-publish` environment.
- All actions **pinned to full commit SHAs**.
- Top-level `permissions: contents: read`; `id-token: write` scoped **only** to the publish job.

## Also included

- Pin package discovery to `src/` so setuptools auto-discovery can't sweep a stale `build/lib/` into the wheel (this was really happening in local builds); ship `py.typed`; add `[project.urls]`.
- Docs: new `docs/releasing.md` (process, CPU/GPU install, `samudra` CLI, one-time trusted-publisher setup) and a PyPI install section in `docs/getting-started/installation.md`.

## Verification

- `resolve` for all modes, real builds (nightly/stable/smoke), and a plain untagged build — all produce a single `samudra` wheel, `twine check` PASSED, no stray `ocean_emulators`.
- Console script: `samudra` with no/unknown command exits 2 with usage; `samudra train --help` routes through the real `TrainConfig` parser; the built wheel registers `samudra = samudra.cli:main`.
- Publish-safety: a nightly/smoke run with a stray version input is rejected (exit 1, nothing emitted); a blank-version nightly dispatch resolves to `0.0.1.dev<stamp>`.
- Local editable install derives `0.0.1.dev<N>` from git automatically.
- `pre-commit` passes on all changed files (`ruff`, `ruff-format`, `mypy`, `check-yaml`, `pyproject-fmt`, schemas).

## Follow-up

- **Ship example configs in the wheel** so `samudra train samudra_om4/train.yaml` resolves presets by name (relocate `configs/` under `src/samudra/`) — separate PR.

### Before first publish (maintainer, one-time)

1. Register the repo as a PyPI **trusted publisher** (Owner `m2lines`, Repo `Samudra`, Workflow `release.yml`, Environment `pypi-publish`) — use the pending-publisher flow to reserve the free `samudra` name.
2. Create the `pypi-publish` GitHub environment (optionally with required reviewers).
3. Tag `v0.0.1` to cut the first stable release. Steps documented in `docs/releasing.md`.

https://claude.ai/code/session_01V9qeRacC2ZdJ4DCbs6SM8t

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,135 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Builds the pure-Python samudra wheel + sdist and publishes to PyPI via OIDC
+# trusted publishing -- no API token is stored anywhere. Triggers:
+#   schedule          -> nightly dev release (<base>.dev<stamp>)
+#   push tag v*       -> stable release, version from the v<version> tag
+#   workflow_dispatch -> manual cut (nightly / stable / manual)
+#   pull_request      -> build-only smoke when the build script or this
+#                        workflow file changes
+# One-time trusted-publisher setup is documented in docs/releasing.md.
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      mode:
+        description: "Build mode"
+        type: choice
+        options: [nightly, stable, manual]
+        default: manual
+      version:
+        description: "Version (required for stable mode, e.g. 1.0.0)"
+        type: string
+  schedule:
+    - cron: "0 6 * * *"   # 06:00 UTC daily
+  push:
+    tags:
+      - "v*"
+  pull_request:
+    paths:
+      - "scripts/package.py"
+      - ".github/workflows/release.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: false  # don't kill an in-flight publish
+
+permissions:
+  contents: read
+
+jobs:
+  resolve:
+    runs-on: ubuntu-latest
+    outputs:
+      mode: ${{ steps.pick.outputs.mode }}
+      version: ${{ steps.resolved.outputs.version }}
+    steps:
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803
+        with:
+          fetch-depth: 0  # setuptools-scm + `git describe` need full tag history
+
+      # Hoist user-supplied dispatch inputs into env so they reach the shell as
+      # variables, not as `${{ ... }}` template substitutions -- free-form
+      # inputs interpolated into a shell script are an injection vector.
+      - id: pick
+        env:
+          INPUT_MODE: ${{ github.event.inputs.mode }}
+          INPUT_VERSION: ${{ github.event.inputs.version }}
+        run: |
+          set -euo pipefail
+          if [[ "${GITHUB_EVENT_NAME}" == "push" && "${GITHUB_REF}" == refs/tags/v* ]]; then
+            echo "mode=stable" >> "$GITHUB_OUTPUT"
+            echo "input_version=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+          elif [[ "${GITHUB_EVENT_NAME}" == "schedule" ]]; then
+            echo "mode=nightly" >> "$GITHUB_OUTPUT"
+            echo "input_version=" >> "$GITHUB_OUTPUT"
+          elif [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ]]; then
+            echo "mode=${INPUT_MODE}" >> "$GITHUB_OUTPUT"
+            echo "input_version=${INPUT_VERSION}" >> "$GITHUB_OUTPUT"
+          else
+            # pull_request: build-only smoke
+            echo "mode=manual" >> "$GITHUB_OUTPUT"
+            echo "input_version=" >> "$GITHUB_OUTPUT"
+          fi
+
+      # Resolve the version exactly once for the whole run so the build job
+      # stamps the identical value -- a run straddling midnight UTC would
+      # otherwise compute a different nightly date in build vs resolve.
+      - id: resolved
+        env:
+          MODE: ${{ steps.pick.outputs.mode }}
+          INPUT_VERSION: ${{ steps.pick.outputs.input_version }}
+        run: |
+          python3 scripts/package.py --mode "$MODE" --version "$INPUT_VERSION" --resolve-only
+
+  build:
+    needs: resolve
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803
+        with:
+          fetch-depth: 0  # setuptools-scm + `git describe` need full tag history
+      - uses: astral-sh/setup-uv@d0d8abe699bfb85fec6de9f7adb5ae17292296ff
+
+      - name: Build wheel + sdist
+        env:
+          MODE: ${{ needs.resolve.outputs.mode }}
+          VERSION: ${{ needs.resolve.outputs.version }}
+        run: |
+          python3 scripts/package.py --mode "$MODE" --version "$VERSION"
+
+      # Fail fast on broken metadata before we ever reach the publish step.
+      - name: Check distribution metadata
+        run: uvx twine check dist/*
+
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
+        with:
+          name: samudra-dist
+          path: dist/*
+          retention-days: 14
+
+  publish:
+    # Manual mode produces local-version identifiers (`X.Y.Z+manual.<sha>`)
+    # which PyPI rejects, so manual cuts -- and every pull_request smoke --
+    # are build-only regardless of how they were triggered.
+    if: |
+      github.event_name != 'pull_request'
+      && needs.resolve.outputs.mode != 'manual'
+    needs: [resolve, build]
+    runs-on: ubuntu-latest
+    environment: pypi-publish
+    permissions:
+      id-token: write   # mint the OIDC token for trusted publishing
+    steps:
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c
+        with:
+          name: samudra-dist
+          path: dist
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@ba38be9e461d3875417946c167d0b5f3d385a247
+        with:
+          packages-dir: dist/
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,7 +21,7 @@ on:
         options: [nightly, stable, manual]
         default: manual
       version:
-        description: "Version (required for stable mode, e.g. 1.0.0)"
+        description: "Stable-cut version, e.g. 0.0.1 (required for stable; ignored for nightly/manual)"
         type: string
   schedule:
     - cron: "0 6 * * *"   # 06:00 UTC daily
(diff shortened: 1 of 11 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,135 @@
...
+
+# Builds the pure-Python samudra wheel + sdist and publishes to PyPI via OIDC
+# trusted publishing -- no API token is stored anywhere. Triggers:
+#   schedule          -> nightly dev release (<base>.dev<stamp>)
+#   push tag v*       -> stable release, version from the v<version> tag
+#   workflow_dispatch -> manual cut (nightly / stable / manual)
+#   pull_request      -> build-only smoke when the build script or this
+#                        workflow file changes
+# One-time trusted-publisher setup is documented in docs/releasing.md.
+name: Release
```
alxmrs, line 16: This release process is inspire'd by the Marin repo.

docs/releasing.md
```diff
@@ -0,0 +1,100 @@
...
+## One-time trusted-publisher setup
+
+Before the first publish, register the repository as a trusted publisher on
+PyPI (a maintainer with project-owner rights does this once):
+
+1. Create the project on PyPI, or use [pending publishers](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/)
+   to reserve the name `samudra` before the first upload.
+2. On the project's **Settings → Publishing** page, add a GitHub Actions
+   publisher with:
+     - **Owner**: `Open-Athena`
```
alxmrs: Should this be m2lines?

src/samudra/__init__.py
```diff
@@ -1,3 +1,12 @@
...
 # SPDX-License-Identifier: Apache-2.0
+
+from importlib.metadata import PackageNotFoundError, version
+
+try:
+    __version__ = version("samudra")
+except PackageNotFoundError:  # not installed (e.g. running from a bare source tree)
+    __version__ = "0.0.0+unknown"
+
+__all__ = ["__version__"]
```
alxmrs, line 12: This allows us to call `samudra.__version__`.

pyproject.toml
```diff
@@ -23,6 +22,7 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
   "Topic :: Scientific/Engineering :: Artificial Intelligence",
 ]
+dynamic = [ "version" ] # derived from git tags by setuptools-scm
```
alxmrs, line 25: Should we make our initial tagged version `0.0.1`?

YuanYuan98: 0.0.1 looks good to me!

YuanYuan98 (approved):

alxmrs:

alxmrs:
@codex may I have your review once more?

.github/workflows/release.yml
```diff
@@ -0,0 +1,143 @@
...
+          INPUT_VERSION: ${{ github.event.inputs.version }}
+        run: |
+          set -euo pipefail
+          if [[ "${GITHUB_EVENT_NAME}" == "push" && "${GITHUB_REF}" == refs/tags/v* ]]; then
+            echo "mode=stable" >> "$GITHUB_OUTPUT"
+            echo "input_version=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+          elif [[ "${GITHUB_EVENT_NAME}" == "schedule" ]]; then
+            echo "mode=nightly" >> "$GITHUB_OUTPUT"
+            echo "input_version=" >> "$GITHUB_OUTPUT"
+          elif [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ]]; then
```
jder, line 72: I was very confused how manual builds (ie workflow_dispatch) may or may not have a mode of manual. Should we maybe call that mode "smoke" or something rather than manual? Since you can manually trigger a stable or nightly build, too.

docs/releasing.md
```diff
@@ -0,0 +1,102 @@
...
+The version is owned by [setuptools-scm](https://setuptools-scm.readthedocs.io/):
+there is **no** `version = "..."` field to maintain — a git tag *is* the version.
+`[tool.setuptools_scm]` in `pyproject.toml` configures it, and `samudra.__version__`
+is available at runtime. The release paths differ only in what version reaches
+the build:
+
+| Trigger | Mode | Version | Published? |
+| --- | --- | --- | --- |
+| Push a `v*` tag | `stable` | the tag, e.g. `v1.0.0` → `1.0.0` (setuptools-scm) | ✅ PyPI |
+| Daily `schedule` (06:00 UTC) | `nightly` | `<next-patch>.dev<YYYYMMDDhhmm>` | ✅ PyPI |
```
jder: Do we really need/want a nightly build? We don't have that many changes in a day. Maybe weekly or something would be sufficient?

docs/releasing.md
```diff
@@ -0,0 +1,102 @@
...
+# Just tag and push — no version bump anywhere:
+git tag v1.1.0
+git push origin v1.1.0
+```
+
+The tag push runs `resolve → build → publish`, uploading `samudra 1.1.0` to
+PyPI. To dry-run first, use **Actions → Release → Run workflow → mode: manual**;
+that builds and runs `twine check` without publishing.
+
+!!! note "Before the first tag"
```
jder, line 101: Is this worth durably documenting?

docs/releasing.md
```diff
@@ -0,0 +1,102 @@
...
+    The repository has no `v*` tags yet, so the "last release" falls back to
+    `0.0.0` (`fallback_version` in `[tool.setuptools_scm]`, mirrored by
+    `FALLBACK_VERSION` in `scripts/package.py` — keep the two in sync). Builds
+    therefore target `0.0.1` (e.g. a nightly is `0.0.1.dev<stamp>`). Cutting the
+    first tag, **`v0.0.1`**, makes the tag the single source of truth from then
+    on.
+
+## One-time trusted-publisher setup
```
jder, line 109: Same question here -- should we just do this and delete these docs?

jder (approved):
Just a few questions from me but generally looks great, thanks! Does this also publish the config files? ie can you actually do `uvx samudra train some/path/to/samudra_v2_om4.yaml` somehow?

jder:
Should this be gated on successful tests? At least for the stable ones?

docs/releasing.md
alxmrs: Weekly SGTM.

docs/releasing.md
alxmrs: Yeah, we can probably omit it.

docs/releasing.md
alxmrs: Yeah, that sounds fine to me, too.

alxmrs:
Thanks @jder — all addressed in 85376009:

**1. Gate publishing on tests ("at least for the stable ones").** Made `test.yml` reusable (`workflow_call`) and added a `test` job to the release workflow that `publish` now `needs`. A red CPU suite blocks the upload — for **both** stable and nightly. A tag push wouldn't otherwise run `test.yml`, which is why the release workflow calls it itself.

**2. `manual` mode is confusingly named → `smoke`.** Renamed the build-only mode to `smoke` throughout (workflow, `package.py`, docs), and the local-version marker with it (`+manual.<sha>` → `+smoke.<sha>`). It now names what the build *does* (a no-publish check) rather than how it starts — every mode can be hand-triggered via `workflow_dispatch`.

**3. Nightly cadence → weekly.** Cron is now `0 6 * * 1` (Mon 06:00 UTC). Agreed daily was overkill; the timestamp still keeps each dev build unique and PEP 440-ordered.

**4. `uvx samudra train some/config.yaml`?** Today: not quite — there was no console script, and `configs/` lives at the repo root, so a wheel shipped only the library. I added a `samudra` console command (`samudra train|eval|viz CONFIG ...`) that forwards to the same entry points as `python -m samudra.train`, so an installed user can run a task against **their own** YAML without a checkout. Shipping the example presets so `samudra train samudra_om4/train.yaml` resolves *by name* needs the `configs/` tree relocated under `src/samudra/` (our src-layout means only files under the package land in the wheel) — that's a self-contained refactor I've left as a documented follow-up rather than fold into this packaging PR.

**5 & 6. The "Before the first tag" note and the trusted-publisher setup docs — worth keeping?** I kept both. The trusted-publisher steps are a genuine one-time task a maintainer has to do before the first publish can authorize (and are easy to get subtly wrong), so I'd rather they be written down. The "before the first tag" note is transient — happy to delete it the moment we cut `v0.0.1`, if you'd prefer it gone now I'll drop it.

PR #820, opened 2026-07-24
Fix LLC center-mask fallback and fixture schema

jder:
## Summary

- align the tiny LLC test fixture with the live LLC4320 root's complete variable/coordinate set, dimension layouts, and mask dtypes
- preserve `mask_c` as the preferred center mask while falling back to `hFacC` when `mask_c` is absent
- add regression coverage for the production schema and the unresolved `hFacC` fallback from PR #670

## Root cause

LLC canonicalization hard-coded `mask_c` in its retained-variable whitelist. The test fixture omitted `hFacC`, so it could not expose that stores containing only the standard MITgcm center-cell fraction would drop their only center wet mask before canonicalization.

The live LLC4320 root contains both `mask_c` and `hFacC`; the fixture now models both accurately, including fractional `float32` `hFac*` arrays.

## Validation

- `uv run pytest tests/test_utils_data.py -k llc` — 7 passed
- `uv run pytest tests/test_config.py tests/test_datasets.py -k llc` — 9 passed
- pre-commit Ruff and mypy hooks passed; Ruff formatting was applied
- the remaining schema hook was interrupted on the slow shared host; GitHub CI should complete the full suite

Review context: https://github.com/m2lines/Samudra/pull/670#discussion_r3634786286


alxmrs (approved):
LGTM, thank you.

PR #824, opened 2026-07-24
Support Python 3.6 in Slurm provenance writers

jder:
## Summary

- keep the Apptainer train and eval provenance writers compatible with ORCD's host Python 3.6
- replace `str.removeprefix` with equivalent prefix slicing in both Slurm harnesses
- document why the older-Python-compatible form is required

## Root cause

The Slurm harness writes `run-provenance.json` with the host `python3` before launching Apptainer. ORCD currently provides Python 3.6.8 on the host, where `str.removeprefix` is unavailable. As a result, H200 smoke-test job `18753331` pulled the correct container successfully but exited before `srun` launched training.

## Impact

Training and evaluation jobs can now write provenance metadata and proceed to their container launch on ORCD's host Python.

## Checks

- `bash -n scripts/slurm_apptainer_train.sbatch scripts/slurm_apptainer_eval.sbatch`
- executed each harness's embedded provenance writer with ORCD host Python 3.6.8 and verified the expected JSON keys
- `git diff --check`

### DIFF
--- a/scripts/slurm_apptainer_eval.sbatch
+++ b/scripts/slurm_apptainer_eval.sbatch
@@ -356,7 +356,9 @@ keys = (
     "SAMUDRA_CONTAINER_IMAGE_REF",
     "SAMUDRA_CONTAINER_SIF_PATH",
 )
-provenance = {key.removeprefix("SAMUDRA_").lower(): os.environ[key] for key in keys}
+# Keep this compatible with the host Python 3.6 available on ORCD.
+prefix = "SAMUDRA_"
+provenance = {key[len(prefix) :].lower(): os.environ[key] for key in keys}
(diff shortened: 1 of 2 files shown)


jder:
I'm not sure why the default python on this host is one that was EOL in 2023 but this seems like an easy fix at least.

scripts/slurm_apptainer_eval.sbatch
```diff
@@ -356,7 +356,9 @@ keys = (
     "SAMUDRA_CONTAINER_IMAGE_REF",
     "SAMUDRA_CONTAINER_SIF_PATH",
 )
-provenance = {key.removeprefix("SAMUDRA_").lower(): os.environ[key] for key in keys}
+# Keep this compatible with the host Python 3.6 available on ORCD.
```
alxmrs, line 359: 🫠

alxmrs (approved):
LGTM

PR #826, opened 2026-07-28
Ship config presets in the wheel and a runnable zero-credential demo

alxmrs:
## Summary

Builds on the PyPI release work (#817) to do two things:

1. **Ship the example config presets inside the `samudra` wheel** and resolve them **by name**, so an installed user runs a task with no checkout.
2. **Make a zero-credential demo runnable** end-to-end from a bare install — a bundled 2° dataset and the basin masks are streamed anonymously from public OSN storage, and one flag points every command at them:

```bash
uvx samudra train samudra_om4/train.yaml --data @data/om4_demo.yaml
uvx samudra eval  samudra_om4/eval.yaml  --data @data/om4_demo.yaml
uvx samudra viz   samudra_om4/viz.yaml   --data @data/om4_demo.yaml
```

## Config presets in the wheel

The config tree moves under the package: `configs/` → **`src/samudra/configs/`**. Because the package installs unpacked, the bundled configs are real co-located files, so the relative `!include`s between a preset and `configs/data/*.yaml` resolve **unchanged**.

`config_base.py` gains `resolve_config_path()`, wired into `from_yaml_and_cli` (so both `python -m samudra.train` and the `samudra` console script benefit): a real filesystem path always wins; a bare name that isn't on disk falls back to `importlib.resources.files("samudra") / "configs" / <name>`. The same fallback applies to `--key @file` overrides, so `--data @data/om4_demo.yaml` resolves the bundled preset too.

**Packaging.** `package-data` ships `configs/**/*.yaml`. Test-fixture configs move to `tests/configs/` (out of the package) rather than being excluded from it — `exclude-package-data` can't override the `setuptools_scm` file-finder, which pulls in every tracked file under the package. The wheel contains only presets + `configs/data` (no `test/`, no `schemas/`). CI cache-key paths, `.gitignore`, `config_schema.py` defaults, and the test constants (`TEST_CONFIGS_DIR`) follow the move.

## Zero-credential demo

- **Bundled dataset.** `src/samudra/configs/data/om4_demo.yaml` streams the public 2° OM4 dataset (`OM4.zarr` + means/stds) straight from `m2lines-pubs` on OSN.
- **Anonymous S3.** The bucket is public and non-AWS, so signed requests are rejected. `S3Location` gains an `anon` field that flows into the fsspec storage options and is preserved across `resolve()`; the demo config sets `anon: true`.
- **Unified `--data`.** Train and eval already took `--data`; viz named its ground truth via a standalone `groundtruth_location`. `VizConfig` gains an optional `data: DataConfig`, and `groundtruth_location` becomes optional, defaulting to the primary source's `data_location` — so one `--data` works across all three. Explicit `groundtruth_location` still wins; with neither, viz fails loudly.

## Public data sources

So the demo (and cloud runs) don't need the private `emulators` bucket:

- **Basin masks published** to `s3://m2lines-pubs/Samudra/basins/` (public/anonymous read; `basin_masks_original` + `_regridded` + `_quarterdeg`). All four `samudra_om4*/viz.yaml` presets point there with `anon: true` — previously viz always opened the private basins and would fail in the zero-credential flow (flagged by review).
- **Skypilot** `train`/`eval`/`viz` now pull basins from the public path and default `OSN_DATA_PATH` to the public `m2lines-pubs/Samudra/v2026-07/om4_halfdeg` (verified present). `git grep emulators skypilot/` is now empty. (`OUTPUTS_BUCKET` stays private — it's the run-output write target, not a data source.)
- **`docs/data.md`** gains a "Basin masks" subsection (public location, three grid variants with verified dims, the five mask variables, anon-read note).

## CLI ergonomics

`samudra` / `samudra --help` now print a real help screen (what the package is, what each subcommand does, examples) to stdout and exit 0; an unknown command prints an error plus help to stderr and exits 2. `samudra <cmd> --help` still routes to that command's own parser.

## Behavior / contract changes worth a review look

- **`experiment.data_root` is now optional.** `resolved_data_root` previously raised when unset; it now defaults to the current directory, so an all-absolute config (the demo) runs flagless. Relative locations still resolve against cwd and fail loudly at `open()`.
- **`viz` preset contract.** `groundtruth_location` is optional and `samudra_om4/viz.yaml` now derives ground truth from `data:`. Anyone running that preset standalone with `--data_root` will need `--data` (or an explicit `--groundtruth_location`).
- **Skypilot default data** switches from the private `emulators` copy to the public `v2026-07/om4_halfdeg`; overridable via `OSN_DATA_PATH`/`OSN_DATA_PATHS`.

## Container

`COPY configs ./configs` in `containers/Dockerfile.physicsnemo-26.05` and the `configs` handling in `scripts/build_apptainer_code_layer.sh` (require-loop, manifest, copy, verify) broke once the tree moved under `src/`. Dropped the standalone `configs` handling — `src` already carries the presets — and the apptainer verify now checks `src/samudra/configs`.

## Docs & cleanup

- README "How to run", quick start, and `run`/`torch`/`config`/`contributing`/`models` + `AGENTS.md` use the new locations and the `samudra <cmd>` / `--data` forms. `config.md` gains a "Data locations" section (relative vs. structured `s3`/`local`, `endpoint_url`, `anon`); `releasing.md` flips to shipped.
- Fixed two broken `run.md` commands (a malformed `torchrun … python -m` invocation and a stale `viz_om4.yaml` path).
- Removed a junk top-level `index.html`; fixed README trailing whitespace and a mypy `union-attr` in the new test.

## Verification

- **Clean wheel**: presets + `data` only, no `test/`/`schemas/` leakage; `twine check` PASSED.
- **Demo, `AWS_*` unset**: all three commands resolve the bundled demo source; viz opens both ground truth (from `--data`) and the public basins anonymously (2° grid, all five masks). Public basins/dataset reachability confirmed over HTTP + anon S3.
- **Console script**: help/exit-code/routing behavior all correct.
- **Tests**: `test_config`, `test_location` (incl. `anon`), new `test_viz_config`, and existing trainer/dataset tests pass against relocated fixtures.
- **Lint/types**: `ruff` + `ruff format --check` + `mypy` clean; `git diff --check` clean.


alxmrs:

alxmrs:
@fomo-bot may I have your review on this PR?

alxmrs:

@fomo-bot, same from you, may I have another review?

YuanYuan98 (approved):
Could we make the advertised end-to-end demo fully runnable by:

  - Supplying a checkpoint in the example eval command? EvalConfig.ckpt_path defaults to None, and eval.py raises a ValueError before starting the rollout when it is omitted.
  - Making the viz basin masks anonymously accessible? The preset’s basins_location is opened unconditionally but does not set anon: true, so the credential-free viz command may still fail.

If the basin-mask bucket is not public, perhaps the demo could reference a copy in the public bucket instead.

alxmrs:
> If the basin-mask bucket is not public, perhaps the demo could reference a copy in the public bucket instead.

I'm currently working to copy the basins data to the public bucket. 

> Supplying a checkpoint in the example eval command? EvalConfig.ckpt_path defaults to None, and eval.py raises a ValueError before starting the rollout when it is omitted.

Love this idea; will do this in a follow up. I don't think we currently have any 2° checkpoints -- would you like to create one?

YuanYuan98:
> Love this idea; will do this in a follow up. I don't think we currently have any 2° checkpoints -- would you like to create one?

happy to create one.

alxmrs:
@fomo-bot, I've made some more changes. Will you review this PR one more time?

Issue #801, opened 2026-07-16
Untangle i/j/x/y/lat/lon given LLC datasets too


jder:
Once #754 lands, I think we should try and resolve this issue by documenting and standardizing across datasets which of these coordinates are pure "model space" and which are actually lat/lon and then use the new data saved by #754 to map between them rather than the current renaming that's happening to get to spherical_area_weights.

@fomo-bot can you please keep an eye on #754 and put up a PR for this once it and #670 have both landed?

> **<sub><sub>![P2 Badge](https://img.shields.io/badge/P2-yellow?style=flat)</sub></sub>  Preserve real LLC coordinates instead of index coordinates**
> 
> For LLC sources, this renames the integer `i`/`j` indices to spatial axes that `validate_data()` later turns into `lon`/`lat`; downstream validation and inference call `spherical_area_weights(self.data)`, so LLC metrics end up using `cos(j-index)` weights rather than geographic cell locations/areas. Any LLC evaluation or validation run will report area-weighted metrics with the wrong spatial weighting unless the canonicalized dataset carries real LLC center coordinates/areas.
> 
> Useful? React with 👍 / 👎. 

 _Originally posted by @chatgpt-codex-connector[bot] in [#670](https://github.com/m2lines/Samudra/pull/670/changes#r3590753095)_

jder:
@fomo-bot Thanks but actually I think we should wait for #823 now

PR #829, opened 2026-07-30
Introducing new metrics system (with obs data production system)

### LINKED ISSUE #750: Make new metrics get reported after eval jobs run

alxmrs:
Closes #750.

Observation metrics now get reported automatically when an eval job finishes, so model
quality can be judged from numbers rather than by hand. Implements the metrics suite from
[the metrics document][1], at the revision incorporating its scientific review.

[1]: https://docs.google.com/document/d/1cWzIrcajfW-gvoMzV3PU0JDAouVt5xerRmEiBIUbV1s

## What lands in W&B

Nine metrics per eval — each RMSE with an annual standard deviation and a 10,000-sample
calendar-year block-bootstrap 95% CI:

| Key | Against |
| --- | --- |
| `obs/velocity/total_rmse` | DUACS |
| `obs/eke/total_rmse` | DUACS |
| `obs/sst/total_rmse` | OISST |
| `obs/ohc_0_700/per_area_total_rmse` | ARGO-IAP |
| `obs/ohc_700_2000/per_area_total_rmse` | ARGO-IAP |
| `obs/sst/residual_variance_map_rmse` / `_pattern_corr` | OISST |
| `obs/ohc_upper700/residual_variance_map_rmse` / `_pattern_corr` | ARGO-IAP |

OM4 is scored alongside under `obs/om4/…`, since a single run's numbers are hard to judge
without the reference. The full tidy frame also goes to W&B as a table and to
`observation_metrics.csv` beside the rollout, carrying the per-year detail behind each
interval.

## Two packages

**`src/samudra/metrics/`** — deliberately *not* an `Aggregator`. These metrics need the
whole finished rollout: calendar-year block bootstrap and detrended/deseasonalized
residual variance have no incremental `record_batch` form. Instead: pure `kernels` (no
I/O, no plotting, no state), `observations` loaders, and a `report` driver returning one
tidy DataFrame as the interchange format. A candidate for the better pattern to replace
aggregators with, rather than an extension of the one we're moving away from.

Wired into `Eval.run()` *after* the rollout's own metrics are logged, so a data problem in
the observation phase raises loudly without costing the rollout results.
`python -m samudra.metrics` reruns metrics against an existing `predictions.zarr` with no
GPU and no checkpoint.

**`data/ocean_preprocessing/obs_preprocessing/`** — fetches and prepares DUACS, OISST and
ARGO-IAP from source, so the observation side is reproducible from this repo rather than
depending on files in one person's scratch directory. Two stages (`download`, `prepare`),
both restartable, both runnable via `scripts/slurm_obs_{download,prepare}.sbatch`.

## Verification

`358 passed`, full pre-commit gate green.

**The port reproduces the published OM4 numbers.** Scoring quarter-degree OM4 as the model
against the OM4 column of the metrics document — numbers produced independently, on
different infrastructure, from a separately built observation archive:

| Metric | Ours | Published | Diff |
| --- | --- | --- | --- |
| Velocity vector RMSE | 0.20898 | 0.2090 | 0.01% |
| Instantaneous EKE RMSE | 0.043831 | 0.04399 | 0.36% † |
| SST RMSE | 0.80518 | 0.8052 | 0.00% |
| OHC per-area [0,700) m | 3.0104e9 | 3.010e9 | 0.01% |
| OHC per-area [700,2000) m | 2.1974e9 | 2.187e9 | 0.48% ‡ |
| SST residual-variance map RMSE | 0.33363 | 0.3336 | 0.01% |
| SST residual-variance pattern corr | 0.6861 | 0.6861 | 0.00% |
| OHC upper-700 residual-variance map RMSE | 1.3345e18 | 1.341e18 | 0.49% ‡ |
| OHC upper-700 residual-variance pattern corr | 0.66518 | 0.6662 | 0.15% ‡ |

† and ‡ are explained below; everything else matches to 0.01% or better.

Kept as `tests/test_metrics_acceptance.py` (marked `manual`, skipped unless
`SAMUDRA_OM4_QUARTERDEG` is set). It needs the 2 TB quarter-degree OM4 store, so it cannot
run on a GitHub runner — but it caught four bugs the unit tests structurally could not, so
it is worth running before merging anything that touches `samudra.metrics`. Happy to build
a cut-down CI fixture as a follow-up if you want metrics changes gated automatically.

## Three deliberate differences from the reference

**† EKE is computed on the native grid, then regridded.** The reference computes it from
already-regridded velocities. EKE is quadratic, so interpolating first smooths the
velocities and damps the energy. Small today (0.36%) because DUACS at 0.125° is finer than
any model we run, so the interpolation upsamples — but the ordering is wrong in principle
and would bias a finer model. The same fix applied to the residual-variance maps was worth
6% on OHC.

**‡ OHC scores 2015–2021, not 2015–2022.** Resampling a 5-day rollout to months gave
December 2022 as an average over 24 days, compared against ARGO's full-month December
under the same label. Excluding short months means a rollout ending 24 December has no
December at all, so its final year is genuinely incomplete. The eighth year only ever
existed because a partial month was being averaged as a whole one. This accounts for the
three OHC differences: a cleaner window, not a methodological disagreement.

**The headline RMSE is the equal-year block estimate**, which is what the bootstrap
interval is built from, so a value can never fall outside its own interval. On the current
products this agrees with the previous aggregation to 0.000%, so it changes no published
comparison — it is robustness for products whose coverage varies by year. The other
aggregation is retained as `map_weighted_rmse`.

## For @YuanYuan98

Three things worth your eye. The first is about the comparison itself and applies to
`obs_evaluation.py` too; the other two are about the write-up.

**OHC compares whatever water each column happens to have, and this one is yours too.**
`_ohc_per_area_layer_maps_from_temperature_field` keeps a cell when *any* level in the layer
is valid, and sums with `skipna=True`. Where both products agree a shelf column ends at 300 m
that is exactly right — both integrate the same water and the comparison is sound. But where
model and ARGO bathymetry *disagree*, the product with the deeper column integrates water the
other lacks, and `skipna` books the difference as zero heat rather than as missing data. Fifty
metres of that is around 2e9 J m⁻², which is the size of the metric itself.

I tried requiring whole columns and it is *not* the fix: it cannot tell a disagreement from a
shelf both products agree is shallow, so it discards every shelf and high-latitude column and
moved both layer totals by more than 20% away from your published figures. Reverted — this PR
matches your behaviour, and `partial_column_fraction` now reports the share of wet cells
exposed so the risk is a number rather than a worry.

A real fix has to mask each cell to the depth range *both* products have, which means
regridding the model in depth onto the ARGO axis before integrating — machinery your
`_common_temporal_fields_on_argo_grid` already has for the T/S structure diagnostics. That
changes the comparison and every published OHC number, so it is your call rather than
something to patch in here.

**The OHC residual-variance section doesn't state the grid ordering that the SST section
does.** The SST section is explicit — *"compute residual variance on each model's native
grid, then interpolate the variance map"* — while the OHC section says only *"compared on
the ARGO-IAP grid"*. Read naturally, that describes the wrong order, and implementing it
that way costs 6%, because interpolation is linear while variance is quadratic. Suggested
wording, mirroring the SST section:

> Model: 1. Compute OHC per unit area in the upper 700 m **on the model's native grid**.
> 2. Monthly-average. 3. Remove the seasonal cycle and trend at each native grid cell.
> 4. Compute residual variance at each native grid cell. 5. **Interpolate the model
> variance map to the ARGO-IAP grid.** 6. Compare with area-weighted RMSE and pattern
> correlation.

**The residual-variance floor.** Deseasonalising precedes detrending, so the pentad
climatology aliases part of a linear trend into a sawtooth the later fit can't remove: a
pure seasonal signal residualises to exactly zero, but a pure trend leaves ~0.0017 against
a signal variance of 4.5. Faithful to your implementation and left unchanged — flagging in
case the ordering wasn't deliberate.

Also worth confirming you agree with the EKE ordering above, since it is the one number
here that no longer matches your published value.

## Observation data

Published at `s3://m2lines-pubs/Samudra/v2026-07/obs/`, anonymously readable, so an eval
run needs no credentials and no staging:

| Store | Grid | Time | Size |
| --- | --- | --- | --- |
| `oisst.zarr` | 720 × 1440 (0.25°) | 598 steps, exact OM4 match | 1.2 GB |
| `duacs.zarr` | 1440 × 2880 (0.125°) | 598 steps, exact OM4 match | 46 GB |
| `argo-iap.zarr` | 360 × 720 (0.5°) | 516 months, 1980–2022 | 18 GB |
| `oisst_daily.zarr` | 720 × 1440 | 2988 days | 2.2 GB |

Each carries the provider's own metadata (NOAA's DOI, the CMEMS and IAP citations),
explicit attribution, and `m2lines/` provenance naming the commit that built it.

## Known limitations

Measured, not hypothesised. Each is recorded in the emitted frame or the code, so a
reviewer can check the exposure rather than take a claim on trust.

**Coastal erosion scales with model resolution.** Linear interpolation NaNs an observation
cell whose model-grid neighbours include land, so the comparison set loses a band one model
cell wide — 1.10% of the observation ocean at 1°, 0.27% at ¼°, all of it coastal where model
error is largest. A coarse run is therefore scored on a smaller and easier subset than a fine
one. `n_paired_cells` and `paired_ocean_fraction` travel with every row so this is auditable;
removing it means choosing a convention for model values at coastal points, which changes
every published number.

**OHC compares whatever water each column has.** The sum skips levels a product lacks, so
where model and observation bathymetry *disagree* the extra water reads as a deficit — about
2e9 J m⁻² per 50 m, the size of the metric itself. Requiring whole columns is not the fix: it
cannot tell disagreement from a shelf both products agree is shallow, and discarding every
shelf moved the layer totals more than 20%. `partial_column_fraction` measures the share of
cells exposed. A real fix needs the model regridded in depth onto the observation axis first.

**Cross-resolution numbers are not comparable**, and the incomparability is not symmetric —
see coastal erosion above. A metric that cannot rank a 1° run against a ¼° one is a weak
model-selection metric; coarsening the observations to the model grid, or band-limiting both,
would give a number that means the same thing across configurations. Worth doing before these
metrics are used to choose between resolutions.

**The kernels are not yet shared with viz.** They are written to be — that is the point of
keeping them pure — but `eval` is today's only caller, and several kernels exist for the
follow-up rather than for this PR.

## Other notes

- **Curvilinear grids are refused**, not mishandled — see #801/#809.
- **OHC constants differ from `constants.py`**: these use ρ=1035, Cp=3850 as the metrics
  document specifies (needed to reproduce the published numbers), while `CP_SW = 3992.0`
  is used by the viz OHC path. Two conventions now coexist.
- **Both git-hash provenance keys are written** during the rename —
  `ocean_emulators_git_hash` is documented and carried by every published store;
  `samudra_git_hash` is the current name. The old one can retire once both catch up.

## Follow-up

Figures in `samudra.viz`, written to the bucket rather than W&B, sharing these kernels.

https://claude.ai/code/session_01KLBRBz6SmSrw97FNh9E5RV


data/ocean_preprocessing/obs_preprocessing/download.py
```diff
@@ -0,0 +1,407 @@
+# SPDX-FileCopyrightText: 2026 Ocean Emulator Authors
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Download the raw observation products, as distributed by their providers.
+
+Ported from `YuanYuan98/Ocean_Emulator:viz_jupyter/download_{duacs,oisst,argo_iap}.py`
+at commit `33d4ae2e608f57e357ba8529b747eb27be2eb041`.
```
alxmrs: Let's omit this comment and all references to Yuan's fork.

alxmrs (commented):
Quick note of feedback.

data/ocean_preprocessing/obs_preprocessing/download.py
```diff
@@ -0,0 +1,417 @@
...
+# A truncated NetCDF is worse than a missing one: it opens, and then poisons the
+# derived product silently. Anything below this is treated as a failed download.
+MIN_PLAUSIBLE_BYTES = 1_000_000
+
+
+def _is_complete(path: Path, min_bytes: int) -> bool:
+    return path.exists() and path.stat().st_size >= min_bytes
+
+
+def _fetch(
```
alxmrs: I think we should add this dep to this package and re-use this library's feature for fast, concurrent downloads of large files: https://github.com/mjishnu/pypdl

data/ocean_preprocessing/obs_preprocessing/download.py
```diff
@@ -0,0 +1,417 @@
...
+    done, failed = _run_pool(months, worker, workers, "ARGO-IAP")
+    logger.info("ARGO-IAP: %d present, %d failed", done, len(failed))
+    if failed:
+        raise RuntimeError(
+            f"{len(failed)} ARGO-IAP files could not be downloaded, e.g. {failed[:5]}. "
+            "Re-run to retry only the missing months."
+        )
+
+
+def _yearly_ranges(start: str, end: str) -> list[tuple[date, date]]:
```
alxmrs, line 320: This could be done with https://docs.xarray.dev/en/latest/generated/xarray.date_range.html

data/ocean_preprocessing/obs_preprocessing/prepare.py
```diff
@@ -0,0 +1,456 @@
...
+    """
+    ds = open_zarr(om4_path, storage_options)
+    times = _coerce_time(ds.time.values)
+    start = pd.Timestamp(start_date)
+    end = pd.Timestamp(end_date)
+    # OM4 stamps sit at 12:00, so a date-only end bound would silently drop that
+    # day's sample. Treat a midnight bound as "end of that day".
+    if end == end.normalize():
+        end = end + pd.Timedelta(days=1) - pd.Timedelta(nanoseconds=1)
+    window = times[(times >= start) & (times <= end)]
```
alxmrs, line 159: Would xr.date_range also help here?

alxmrs: May not be essential; what exists now is reasonable.

data/ocean_preprocessing/obs_preprocessing/prepare.py
```diff
@@ -0,0 +1,456 @@
...
+        ds.attrs.pop(attr, None)
+    if "time" in ds.coords and ds.sizes.get("time", 0):
+        times = pd.DatetimeIndex(ds["time"].values)
+        ds.attrs["time_coverage_start"] = times.min().isoformat()
+        ds.attrs["time_coverage_end"] = times.max().isoformat()
+
+    ds.attrs.update(PROVENANCE[product])
+    ds.attrs["m2lines/date_created"] = datetime.datetime.now().isoformat()
+    ds.attrs["m2lines/cli_args"] = " ".join(sys.argv)
+    ds.attrs["m2lines/ocean_emulators_git_hash"] = get_git_url_hash()
```
alxmrs: Let's use this GH project for the git hash (m2lines/samudra_git_hash). While we're here, let's fix it in the other script.

data/ocean_preprocessing/obs_preprocessing/prepare.py
```diff
@@ -0,0 +1,456 @@
...
+        if not path.name.endswith(".tmp.zarr")
+    )
+    if not stores:
+        raise FileNotFoundError(
+            f"No raw DUACS year stores (duacs_*.zarr) under {raw_dir}"
+        )
+    logger.info("DUACS: combining %d yearly stores", len(stores))
+
+    ds = xr.concat(
+        [xr.open_zarr(store, chunks={}) for store in stores],
```
alxmrs: Could we use `open_mfdataset` instead? 

https://docs.xarray.dev/en/stable/generated/xarray.open_mfdataset.html

alxmrs: Is what you have here better? If so, why?

data/README.md
```diff
@@ -90,6 +90,62 @@ python -m ocean_preprocessing om4 \
...
+python -m ocean_preprocessing.obs_preprocessing download oisst    --output_dir=$RAW/oisst
+python -m ocean_preprocessing.obs_preprocessing download argo_iap --output_dir=$RAW/argo-iap
+
+# DUACS needs a free account from https://data.marine.copernicus.eu/register
+export COPERNICUSMARINE_SERVICE_USERNAME=... COPERNICUSMARINE_SERVICE_PASSWORD=...
+python -m ocean_preprocessing.obs_preprocessing download duacs    --output_dir=$RAW/duacs
+```
+
+On Torch these run as batch jobs — `PRODUCT=oisst sbatch scripts/slurm_obs_download.sbatch` — since compute nodes there
+do have internet egress.
```
alxmrs, line 121: when writing to the NYU OSN pod.

src/samudra/configs/samudra_om4_v2/eval.yaml
```diff
@@ -18,3 +18,32 @@ experiment:
     project: default
 data: !include ../data/om4.yaml
 model: !include model.yaml
+observations:
```
alxmrs: Let's extract these settings into a `../data/obs.yaml` file! :)

src/samudra/metrics/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+Unlike `samudra.aggregator`, which accumulates statistics incrementally as a
+rollout streams past, these metrics operate on a *finished* rollout that has
+been written to disk. Several of them (calendar-year block bootstrap, detrended
+and deseasonalized residual variance) have no incremental form at all: they need
+the whole multi-year time series in hand.
+
+The pieces:
+
+- `kernels`: pure functions over `xr.DataArray`. No I/O, no plotting, no state.
```
alxmrs, line 15: Is DataArray or DataSet the right abstraction?

src/samudra/metrics/config.py
```diff
@@ -0,0 +1,122 @@
...
+import xarray as xr
+from pydantic import BaseModel, Field, model_validator
+
+from samudra.metrics import kernels, observations
+from samudra.utils.location import Location, ResolvedLocation, UnresolvedLocation
+
+logger = logging.getLogger(__name__)
+
+
+class ObsMetricsConfig(BaseModel):
```
alxmrs: I think this should live in the top level project `config.py`.

src/samudra/metrics/kernels.py
```diff
@@ -0,0 +1,749 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Pure metric kernels for observation-based evaluation.
+
+Implements the observation-evaluation metrics as specified in the metrics
+document, at the revision incorporating its scientific review: instantaneous
```
alxmrs: Can we add a `[1] ... [1]: https://..` style reference here for the metrics document?

src/samudra/metrics/kernels.py
```diff
@@ -0,0 +1,749 @@
...
+                f"{getattr(val, 'hour', 0):02d}:"
+                f"{getattr(val, 'minute', 0):02d}:"
+                f"{getattr(val, 'second', 0):02d}"
+                for val in flat
+            ]
+        )
+    return pd.to_datetime(flat.astype(str))
+
+
+def normalize_lon[T: (xr.Dataset, xr.DataArray)](obj: T, lon_name: str = "lon") -> T:
```
alxmrs, line 114: Nice typing!

src/samudra/metrics/kernels.py
```diff
@@ -0,0 +1,749 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
```
alxmrs, line 1: TODO(claude): For all the kernels in this file, please add a comment in the PR that has a link to the line of source code for Yuan's implementation. Add this comment on the relevant function definitions. This will help human reviewers note the differences between the original and this version.

tests/test_metrics_acceptance.py
```diff
@@ -0,0 +1,101 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Acceptance test: reproduce the published OM4 observation metrics.
```
alxmrs, line 5: Will these run in CI? Do we need to modify CI to have the vars to be able to run this acceptance test? Should we add a new action to support such a check?

alxmrs (commented):
Round of review. The two new packages are looking good so far! Thanks, Claude.

data/ocean_preprocessing/obs_preprocessing/download.py
alxmrs: Done — `pypdl` is now a dependency and both downloaders go through it (`_download_batch`).

Two things I kept rather than delegating, both about not losing restart safety:

- **Skip-if-present stays explicit.** pypdl's `overwrite=False` semantics aren't documented for the already-exists case, so the task list is filtered before anything is handed over. A file present at a plausible size is simply never requested.
- **Success isn't taken on trust.** Every target is re-checked for plausible size afterwards; anything truncated is deleted and reported failed. Otherwise a short file that pypdl called complete would be skipped forever by the size check on the next run.

ARGO's multi-mirror fallback is preserved by batching per naming candidate: everything still missing after a round is retried with the next spelling, so the downloads stay concurrent without losing the fallback.

Verified against the live NOAA endpoint — 365/365 fetched, files open as valid NetCDF, re-run skips all 365. Worth noting where the win actually is: OISST's 1.6 MB files were already concurrency-bound, but ARGO's ~40 MB files took 2h50m at a few MB/s from a single connection, and segmenting is what should help there.

data/ocean_preprocessing/obs_preprocessing/download.py
alxmrs: Done. `_yearly_ranges` now derives the interior boundaries from `xr.date_range(..., freq="YS")` and clamps the first and last spans to the dates actually requested, so a range starting mid-October still yields a short first year.

The bigger win from your suggestion was one line up: the OISST day enumeration was a triple-nested comprehension over `calendar.monthrange`, and is now `xr.date_range(f"{start_year}-01-01", f"{end_year}-12-31", freq="D")`. That dropped the `calendar` import entirely.

data/ocean_preprocessing/obs_preprocessing/prepare.py
alxmrs: Not really, and I've left this one alone — matching your follow-up.

`om4_target_times` isn't generating a date range, it's *filtering* an axis that already exists in the OM4 store and then validating its cadence. `date_range` would only help if we were synthesising the timestamps, and synthesising them is exactly what we must not do: the whole point is that the observation products land on OM4's real stamps, including the occasional 6-day step, so the metrics' exact-match guard passes.

`xr.date_range` is used in `download.py` where ranges genuinely are being generated.

data/ocean_preprocessing/obs_preprocessing/prepare.py
alxmrs: Done. The attribute is now `m2lines/samudra_git_hash` in both the observation pipeline and the OM4 one — the project is Samudra, and `ocean_emulators` was the old name.

Note this means the already-published observation stores carry the old key. Regenerating them is cheap (the raw archive is still on scratch; prepare is ~35 min wall clock, DUACS dominating) and I'm happy to re-run and re-publish so the data matches the code — say the word.

data/ocean_preprocessing/obs_preprocessing/prepare.py
alxmrs: Done — switched to `open_mfdataset`.

I checked rather than assumed, since this one is load-bearing. Built three yearly stores written deliberately out of order and compared against the previous manual `concat` + `sortby`: identical values (`allclose` → True) and an identical time axis. `combine="by_coords"` orders them itself, so the explicit `sortby("time")` goes away too.

alxmrs: No — yours is better, and I've switched to it.

My reasoning for the manual `concat` was wanting explicit control over the join, but that control bought nothing: `open_mfdataset` with `combine="by_coords"` handles ordering itself, so it's strictly less code for identical output (verified on out-of-order stores). The one thing worth keeping is the glob filter that excludes `*.tmp.zarr` partials — that's about *which* stores get combined, not how.

data/README.md
alxmrs: Added — the README now says credentials are only needed when writing to the NYU OSN pod (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `FSSPEC_S3_ENDPOINT_URL`), and that reading the public inputs — including the OM4 time axis the alignment depends on — is anonymous.

That asymmetry is deliberate: someone outside the project can reproduce the whole prepare stage with no credentials at all.

src/samudra/configs/samudra_om4_v2/eval.yaml
alxmrs: Done — the block now lives in `src/samudra/configs/data/obs.yaml` and `eval.yaml` reads `observations: !include ../data/obs.yaml`, matching how `data/om4.yaml` is handled.

The YAML merge anchors survive the move (verified: the config still resolves to the three public S3 locations with `anon=True`, and schema validation passes).

src/samudra/metrics/__init__.py
alxmrs: `DataArray` is right for the kernels, and deliberately so.

Every kernel takes and returns `DataArray` because each one operates on a single physical field — an SST field, a velocity component, an OHC-per-area map. `Dataset` would mean each kernel has to know which variable inside it to reach for, which is exactly the coupling the pure-function split is trying to avoid: `area_weighted_map_rmse(bias, area)` works on any field, and doesn't care whether it came from OISST or ARGO.

`Dataset` is used at the boundaries, where it belongs — `observations.py` opens and standardises whole products, and `report.py` pulls the individual fields out before handing them to kernels.

The one place `Dataset` appears inside a kernel is `xr.Dataset({"rmse_map": ..., "annual_rmse": ...}).compute()`, which is just a way to force both dask graphs in one pass rather than two.

src/samudra/metrics/kernels.py
alxmrs: Added — `kernels.py` now carries a `[1]` reference to the metrics document, in the module docstring next to the description of which revision it implements.

src/samudra/metrics/kernels.py
alxmrs: Thanks! That one is `normalize_lon[T: (xr.Dataset, xr.DataArray)]` — a PEP 695 generic, so it preserves whether you handed it a Dataset or a DataArray instead of widening to the union. Worth it because callers do both.

src/samudra/metrics/kernels.py
alxmrs: Provenance for the ported kernels, as requested. Each links to the reference implementation at `33d4ae2e608f` so the diff is reviewable, with a note wherever this version differs.

| This file | Reference | Difference |
| --- | --- | --- |
| [`coerce_datetime_values`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L676) → line 72 | `coerce_datetime_values` | Unchanged. |
| [`normalize_lon[T:`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L695) → line 94 | `normalize_lon[T:` | Unchanged, but generic over Dataset/DataArray. |
| [`require_exact_time_match`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L895) → line 102 | `require_exact_time_match` | Unchanged. |
| [`median_time_step_days`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1067) → line 145 | `median_time_step_days` | Unchanged. |
| [`complete_calendar_years`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1216) → line 154 | `complete_calendar_years` | Unchanged. |
| [`require_complete_calendar_years`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1238) → line 183 | `require_complete_calendar_years` | Unchanged. |
| [`depth_edges_from_centers`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L824) → line 207 | `depth_edges_from_centers` | Unchanged. |
| [`layer_overlap_thickness`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L841) → line 221 | `layer_overlap_thickness` | Unchanged. This is the exact-overlap weighting from the review. |
| [`area_weighted_mean`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L2158) → line 271 | `area_weighted_mean` | Unchanged. |
| [`area_weighted_map_rmse`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L2167) → line 281 | `area_weighted_map_rmse` | Unchanged. |
| [`area_weighted_timeseries`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1143) → line 289 | `area_weighted_timeseries` | Unchanged. |
| [`area_weighted_pattern_corr`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1386) → line 301 | `area_weighted_pattern_corr` | Unchanged. |
| [`field_rmse_over_time`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1209) → line 331 | `field_rmse_over_time` | Unchanged. |
| [`apply_equatorial_mask`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1437) → line 349 | `apply_equatorial_mask` | Unchanged. |
| [`geostrophic_velocity_from_zos`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1448) → line 361 | `geostrophic_velocity_from_zos` | Unchanged; the divide-by-zero warning at f=0 and cos(lat)=0 is suppressed explicitly rather than left to numpy's default. |
| [`instantaneous_surface_eke`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1418) → line 393 | `instantaneous_surface_eke` | Unchanged. **Called differently** — see the note on `_velocity_metrics`: we compute EKE on the model's native grid and regrid the result, where the reference computes it from regridded velocities. |
| [`mean_surface_eke`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1423) → line 405 | `mean_surface_eke` | Unchanged. |
| [`ohc_per_area_layer_maps`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L2370) → line 415 | `ohc_per_area_layer_maps` | **Adds a coverage check**: raises if the depth axis does not span the requested layer, rather than silently integrating a partial column against the observation product's full one. |
| [`pentad_index`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1075) → line 470 | `pentad_index` | Unchanged. |
| [`has_repeated_calendar_months`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1138) → line 485 | `has_repeated_calendar_months` | Takes a DatetimeIndex rather than a Series; same logic. |
| [`detrend_linear_dataarray`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1160) → line 502 | `detrend_linear_dataarray` | Unchanged. |
| [`detrended_deseasonalized`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1179) → line 524 | `detrended_deseasonalized` | Unchanged. |
| [`residual_variance_map`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L3737) → line 560 | `residual_variance_map` | Extracted into a named kernel so the eval and viz paths cannot diverge. |
| [`series_without_seasonal_cycle`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1095) → line 569 | `series_without_seasonal_cycle` | Unchanged. |
| [`series_without_linear_trend`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1055) → line 584 | `series_without_linear_trend` | Unchanged. |
| [`series_linear_trend_per_year`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1121) → line 598 | `series_linear_trend_per_year` | Unchanged. |
| [`series_residual_variance`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1126) → line 611 | `series_residual_variance` | Merged into one function; same composition. |
| [`model_field_on_obs_grid`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L2391) → line 626 | `model_field_on_obs_grid` | Unchanged. |
| [`interannual_rmse_summary`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1252) → line 656 | `interannual_rmse_summary` | Unchanged, including the fixed bootstrap seed. |
| [`rmse_map_with_uncertainty`](https://github.com/YuanYuan98/Ocean_Emulator/blob/33d4ae2e608f57e357ba8529b747eb27be2eb041/viz_jupyter/obs_evaluation.py#L1295) → line 710 | `rmse_map_with_uncertainty` | Split: this returns the map, the per-year totals and the summary; row construction moved to `report.py`. The `include_annual=False` branch is dropped -- every primary score is uncertainty-enabled. |

**Two deliberate behavioural departures**, both discussed above: EKE is computed before regridding rather than after, and the reported RMSE is the equal-year block estimate so it is consistent with its own confidence interval. Everything else is a faithful port, which the OM4 acceptance test checks numerically.

tests/test_metrics_acceptance.py
alxmrs: Short answer: **no, it does not run in CI today, and I do not think it should be forced to.**

`test.yml` runs `-m "not manual and not cuda"` and `test-gpu.yml` runs `-m "not manual and cuda"`, so anything marked `manual` is excluded from both. That is deliberate here.

**No CI secrets are needed** — that part is fine. The observation products are anonymously readable from the public bucket, so the test needs no credentials at all.

**The blocker is input size, not permissions.** The test scores the quarter-degree `OM4.zarr`, which is ~2 TB, via `SAMUDRA_OM4_QUARTERDEG`. A GitHub-hosted runner has neither the disk nor the time budget, and streaming it is worse. The observation side is 67 GB on its own (DUACS is 46 GB of that). The run takes ~12 minutes on a 32-core / 400 GB node once the data is local.

Three options, in the order I would consider them:

1. **Leave it manual** (what I have done). Run it on Torch before merging anything that touches `samudra.metrics`; it is one `sbatch` and it is the only check that compares against independently-published numbers. Zero CI cost, no false confidence.
2. **A `workflow_dispatch` job on a big runner.** The repo already spins up EC2 for `test-gpu.yml`, so the pattern exists — but the 2 TB input still has to come from somewhere, so this mostly moves the problem.
3. **Build a small CI-able regression fixture.** Precompute a cut-down input (one year, one region) and assert against values recorded from a verified run. That would catch the *classes* of bug this test actually caught — cftime axes, the `DataSource` layout, depth coverage — inside normal CI. It would be a regression test against ourselves rather than against the published column, so it complements the manual test rather than replacing it.

My recommendation is 1 now and 3 as a follow-up issue if you want metrics changes gated automatically. Happy to build 3 — it is maybe half a day, mostly deciding how small the fixture can be while still exercising the regrid and the equal-year blocks. I have not done it speculatively since it is a real chunk of work and the manual test already covers the correctness question.

alxmrs:

data/ocean_preprocessing/obs_preprocessing/download.py
```diff
@@ -0,0 +1,452 @@
...
+    if first > last:
+        raise ValueError(f"Invalid date range: {start} > {end}")
+    # The interior boundaries are the January 1sts inside the span; the first
+    # and last spans are clamped to the dates actually requested, so a range
+    # starting mid-October yields a short first year.
+    edges = [first] + [
+        stamp.date()
+        for stamp in xr.date_range(start, end, freq="YS")
+        if stamp.date() > first
+    ]
```
alxmrs: we should be able to replace this with one xr.date_range call, right? stamp.date() is always > first according the `if first > last` case above. Then, we wouldn't also need to concat two lists, it would just be a date_range call? This is much too complex; let's keep it simple.

data/ocean_preprocessing/obs_preprocessing/download.py
```diff
@@ -0,0 +1,452 @@
...
+        end_date,
+        ",".join(wanted),
+        output_dir,
+    )
+
+    for first, last in ranges:
+        final = output_dir / f"duacs_{first:%Y%m%d}_{last:%Y%m%d}.zarr"
+        # Download to a temporary name and rename on success, so an interrupted
+        # year is never mistaken for a complete one on the next run.
+        tmp = final.with_suffix(".tmp.zarr")
```
alxmrs: Can we use `tempfile` and a `with` expression to create the temporary directory? That way, we won't have to manually worry about `shutil.rmtree` (IIUC).

alxmrs: If you disagree with this suggestion, please let me know.

src/samudra/config.py
```diff
@@ -1302,6 +1405,22 @@ class EvalConfig(TopLevelConfig):
     experiment: ExperimentConfig
     data: DataConfig
     model: AnyModelConfig
+    observations: ObsMetricsConfig = Field(
```
alxmrs: If this is optional, should it be a nullable type?

alxmrs: I see that we have a `enabled` flag. What do you think about the pattern of null checking instead (the object isn't configured -- it's null -- so we skip the observation run)? That seems to be the convention in the codebase from what I can tell (you can double check me on this).

alxmrs (commented):
Another round of review (skim)

data/ocean_preprocessing/__main__.py
```diff
@@ -458,7 +458,7 @@ def om4(
         for var, attr in saved_attrs.items():
             ds_input.attrs[var] = attr
         # Add provenance hash and metadata
-        ds_input.attrs["m2lines/ocean_emulators_git_hash"] = get_git_url_hash()
+        ds_input.attrs["m2lines/samudra_git_hash"] = get_git_url_hash()
```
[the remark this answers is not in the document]
alxmrs: I agree, let's revert this one but keep the other one.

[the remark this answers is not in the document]
alxmrs: Sorry, I misunderstood: I like the idea to keep both!

data/ocean_preprocessing/obs_preprocessing/download.py
alxmrs: Agreed on the complexity — rewritten around a single `xr.date_range` call:

```python
starts = pd.DatetimeIndex([first, *xr.date_range(start, end, freq="YS")]).unique()
ends = [stamp.date() - timedelta(days=1) for stamp in starts[1:]] + [last]
return list(zip([stamp.date() for stamp in starts], ends, strict=True))
```

One correction on the reasoning, since it changes what the replacement has to do: `stamp.date() > first` was **not** always true. `date_range("2015-01-01", ..., freq="YS")` emits `2015-01-01` itself, so when the range opens exactly on 1 January the filter was suppressing a duplicate edge, not filtering nothing. `.unique()` handles that case more directly, which is why the concat can go.

The `first` prefix has to stay regardless — DUACS opens on 2014-10-18, so the first span is a partial year that `YS` alone never produces.

Verified contiguous, clamped and gap-free across all three shapes: mid-year start (2014-10-18 → 9 spans), exactly-on-boundary (2015-01-01 → 2 spans, no duplicate), and a range inside one year (2020-03-05 → 1 span).

src/samudra/metrics/kernels.py
```diff
@@ -0,0 +1,852 @@
...
+        rename["y"] = "lat"
+    if "x" in model_field.dims:
+        rename["x"] = "lon"
+    if rename:
+        model_field = model_field.rename(rename)
+    model_field = normalize_lon(model_field, "lon").sortby("lat").sortby("lon")
+    interpolated = model_field.interp(
+        lat=obs_field["lat"],
+        lon=obs_field["lon"],
+        kwargs={"fill_value": np.nan},
```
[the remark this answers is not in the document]
alxmrs: Good catch, and confirmed — this was silently dropping real data.

Quarter-degree OM4 centers start at 0.125 while DUACS starts at 0.0625, so `interp` treated 2 of DUACS's 2880 longitude columns as out of bounds and excluded them from every RMSE and variance reduction. Small in absolute terms, but exactly as you say: systematic, and dependent on grid alignment rather than on the model.

Fixed by wrapping one column from each end before interpolating. Verified the NaN column count goes 2 → 0 on the real OM4/DUACS geometry, and regional grids still pass through untouched — their edges are real boundaries, so inventing data past them would be worse than dropping a column.

This is the same periodicity bug as the geostrophic longitude derivative from the previous round, so the two now share one `_spans_globe` test rather than duplicating the heuristic. Regression test added covering both the global and regional cases.

data/ocean_preprocessing/__main__.py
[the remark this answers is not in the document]
alxmrs: Fair point, and I've changed the approach: **both** keys are now written.

You're right that the established key is load-bearing — it's documented in `AGENTS.md`, `CLAUDE.md` and `docs/data.md`, and I confirmed every already-published store carries it (checked `v2026-07/om4_onedeg/OM4.zarr` and the new `obs/oisst.zarr`), so dropping it would break anything that looks it up.

The rename was explicitly requested in review, though, and `samudra_git_hash` is the name the project actually goes by now. Writing both satisfies both constraints: existing tooling keeps working, new consumers get the current name, and the old key can retire once the docs and published stores have caught up. Noted in a comment at both write sites.

alxmrs:
@fomo-bot May I have your review of this PR?

data/ocean_preprocessing/obs_preprocessing/download.py
```diff
@@ -0,0 +1,447 @@
...
+    )
+
+    failed = set()
+    for entry in downloader.failed or []:
+        # pypdl reports failures as the task dicts it was given.
+        failed.add(entry["url"] if isinstance(entry, dict) else str(entry))
+
+    for url, target in tasks:
+        if url in failed:
+            continue
```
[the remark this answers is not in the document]
alxmrs: Confirmed and fixed — you're right that the `continue` skipped the cleanup.

A failed task can still have written a partial target, and one above `MIN_PLAUSIBLE_BYTES` would be treated as complete by the skip check on every later run: never retried, and silently feeding truncated data into the prepared store. Failed URLs now have their target removed before continuing, so a retry actually retries.

Worth noting this was the *second* path to the same hazard — the DUACS staging directory had a related one, now removed entirely by switching to `tempfile.TemporaryDirectory` (per @alxmrs's suggestion below), which cleans up on any exit path including an exception.

data/ocean_preprocessing/obs_preprocessing/download.py
alxmrs: Done, and it's better than the manual version for exactly the reason you'd expect: `TemporaryDirectory` cleans up on *every* exit path including an exception, so the stale-partial bookkeeping disappears rather than having to be maintained.

That removed a real class of bug, not just lines: the previous code had to explicitly clear a leftover `.tmp.zarr` *before* the skip check, because a year interrupted on one run and completed on the next would otherwise keep its partial forever.

One detail worth flagging: the staging directory is created with `dir=output_dir` rather than in the system temp. DUACS years are several GB, and a default `/tmp` would make the move into place a cross-filesystem copy instead of a rename.

Verified a failed download now leaves nothing behind — no staging directory, no partial store.

src/samudra/config.py
alxmrs: Good question, and you're right — it's now `ObsMetricsConfig | None = None`, with the separate `enabled` flag removed.

I checked whether the flag was earning its place and it wasn't. Keeping both would have meant two ways to express "off" (`None` versus `enabled=False`), which is the kind of redundant state that eventually diverges. Presence of the block is now the switch: `observations: !include ../data/obs.yaml` turns it on, omitting it leaves it off, and `--observations=null` disables it for one run without editing files. I verified CLI overrides work in both directions under the nullable type before committing to it.

A small bonus: mypy immediately flagged that `report_observation_metrics` was reading `self.observations` without establishing non-None. The config is now passed in explicitly, so the type tells you it's present rather than the reader having to trust the call site.

alxmrs:
Confirmed and fixed — thank you, this was a real gap in how I'd been testing.

Reproduced it exactly as you describe: with `FSSPEC_S3_ENDPOINT_URL`, `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` all unset, `om4_target_times` on the default path raised `PermissionError: Forbidden` — s3fs went to AWS instead of OSN. My own runs always had the endpoint exported (the sbatch harnesses set it), so the documented direct command was the one path I never exercised.

Rather than document the export, I made it unnecessary: `open_zarr` now infers the OSN endpoint for the buckets known to live there (`m2lines-pubs`, `emulators`) when `FSSPEC_S3_ENDPOINT_URL` isn't already set. Explicit `storage_options`, the environment variable, and real AWS credentials all still take precedence. Verified the Stage 2 command now works with every relevant variable unset — which was the intent all along, since the published inputs are anonymously readable and the pipeline should be reproducible by someone with no project setup at all.

Also fixed in this round: a failed download could leave a >1 MB partial that the skip check would treat as complete forever (caught by Codex), and the DUACS staging directory now uses `tempfile.TemporaryDirectory` so partials can't survive an exception.

data/ocean_preprocessing/obs_preprocessing/prepare.py
```diff
@@ -0,0 +1,487 @@
...
+
+
+ALIGNED = (
+    "centered 5-day rolling mean (min_periods=5) sampled on the OM4 5-day timestamps"
+)
+RAW_DAILY = "none; daily values as distributed"
+MONTHLY = "none; monthly values as distributed"
+
+
+def duacs(
```
alxmrs, line 308: @YuanYuan98 These functions below are worthy of your attention.

src/samudra/metrics/kernels.py
alxmrs: @YuanYuan98 This file is also worthy of your attention. To make it easier to review, above is a table that compares the inspiration and this version's implementaiton.

alxmrs:
### Review finding from the viz follow-up: models are scored on different sets of cells

Found while reviewing the viz PR against Yuan's implementation. It belongs here rather than there, and does not need to block this PR — filing so it is not lost.

**What Yuan's implementation does that this one does not.** `obs_evaluation.py` builds a wet mask per model (`_native_model_wet_mask`, the finite footprint of `thetao` at the first timestep) and then, in `_attach_common_model_wet_masks`, picks the *most restrictive* mask across every model sharing a native grid and applies that one to all of them:

```python
_, reference_label, reference_mask = min(candidates, key=lambda item: item[0])
print(f"Using {reference_label} finite ocean mask for model-native diagnostics.")
```

Every model is then scored over an identical set of cells. Our `samudra.metrics` applies no wet mask; `rmse_map_with_uncertainty` derives its denominator per model, from that model's own footprint:

```python
comparable = cast(xr.DataArray, np.isfinite(error_squared)).any("time")
```

**Why it matters.** Since we score the model alongside OM4 as a baseline, the two totals are only comparable if they cover the same cells. Where the footprints differ, the headline numbers are not measuring the same thing. Reproduction with our own kernels — squared error of 9 in the western half and 1 in the eastern, so per-cell skill is identical between the two:

```
A: predicts everywhere   total RMSE = 2.2361   scored cells = 1152
B: skips the hard half   total RMSE = 1.0000   scored cells = 576
```

B looks more than twice as good as A while being exactly as accurate wherever it predicts.

**Why the acceptance test does not catch it.** The acceptance test feeds OM4 in as the model and scores it against the OM4 baseline, so the two footprints are identical by construction and the missing mask is a no-op. This only bites once a real rollout, whose land treatment differs from OM4's, is scored against it.

**Suggested fix**, roughly in order of cost:

1. Record the scored-cell count (or the area-weighted fraction) in the metric frame alongside `grid_shape`. The frame currently gives a reviewer no way to notice this happening. Cheap, and useful regardless of the rest.
2. Intersect the footprints across all scored runs before reducing, as Yuan does, so the model and the OM4 baseline share one denominator.
3. Fail loudly when the footprints disagree by more than a small tolerance, rather than silently scoring different domains.

Happy to take this as a follow-up.

https://claude.ai/code/session_01KLBRBz6SmSrw97FNh9E5RV

YuanYuan98:
Hey Alex, this is a big pr, thanks a lot for the great work here! I went through it, and overall everything looks solid to me. 

I’m fine with handling the common wet-mask question as a follow-up for the current same-grid use case.

I only had one small non-blocking note: residual-variance RMSE doesn’t seem to expose the annual uncertainty described in the PR. This is fine as follow-ups from my side.

alxmrs:
@fomo-bot do you think this PR is ready for review, given all the above conversation? According to the code, is it ready for merge? If so, may I have your approval for this PR?

alxmrs:
### Four more findings from reviewing the viz follow-up

A code review over the combined branch turned up four items that live in this PR's code rather than the viz one. I verified each by running it; none need to block, but the first two are worth fixing before this lands.

---

**1. `kernels.py:202` — the shipped window clears the complete-year check by 1 nanosecond, and the comment says half a day.**

`boundary_tolerance = pd.Timedelta(days=1.5 * step_days)` is 7.5 days at the 5-day cadence, so the last-sample threshold for 2022 is `2022-12-24 11:59:59.999999999`. The published obs stores end at the OM4 stamp `2022-12-24 12:00:00`:

```
threshold   : 2022-12-24 11:59:59.999999999
last sample : 2022-12-24 12:00:00
margin      : 0 days 00:00:00.000000001
```

Anything that moves that stamp fails `require_complete_calendar_years`, which kills the observation phase *after* the rollout has been paid for:

```
one 5-day step earlier   -> passes: False
stamped at 00:00 not 12:00 -> passes: False
a 6-day final gap        -> passes: False
```

The tolerance appears to have been tuned to exactly admit the data we happen to ship. Widening it to 2 steps would clear the same window by 2.5 days with no loss of strictness against a genuinely ragged year — the sample-count check (`min_sample_ratio`) already catches thinning independently.

Separately, the comment at `kernels.py:243` tells whoever debugs this that "the shipped window clears the tolerance by about half a day." It is out by roughly thirteen orders of magnitude and should be corrected whatever else changes.

---

**2. `report.py:79` — `n_paired_cells` counts cell-timesteps in some rows and cells in others.**

`_pairing` reduces with a bare `.sum()` over every dim. From `_velocity_metrics` / `_sst_metrics` / `_ohc_metrics` the inputs still carry `time`; from `_variance_map_rows:208` they are 2-D maps:

```
with a time dim  -> n_paired_cells = 600   (cells = 30, cell-timesteps = 600)
2-D maps         -> n_paired_cells = 30    (cells = 30)
```

So one CSV column carries two different quantities, and the values are not comparable between rows of the same file — for an 8-year 5-day run the RMSE rows overstate by ~580×. `paired_ocean_fraction` is a ratio and is unaffected, so the docstring's actual purpose still works; it is the absolute column that misleads. Reducing over the spatial dims of a time-collapsed mask fixes it.

---

**3. `report.py:397` and `:454` — `model_field_on_obs_grid` is evaluated twice per metric.**

```python
error_squared=(kernels.model_field_on_obs_grid(sim, obs) - obs) ** 2,
...
pairing=_pairing(kernels.model_field_on_obs_grid(sim, obs), obs),
```

Each call redoes the longitude wrap and a full bilinear interpolation of a multi-year field onto the observation grid, and `_pairing`'s `int(...)` forces its own dask compute of the whole thing. Same shape in `_ohc_metrics`. Hoisting it into a local is a one-line change; given this phase reads tens of GB it seems worth it.

---

**4. `kernels.py:559` — `partial_column_fraction` has no production caller.**

It is reached only from `tests/test_metrics.py:140` and from the `ohc_per_area_layer_maps` docstring, which tells the reader to "use `partial_column_fraction` to measure how much of a comparison is exposed to this." Nothing in `report.py` computes it, so on a real run that exposure is never measured. Either wire it into the frame or drop the sentence.

---

This is separate from the wet-mask/denominator item I filed earlier in https://github.com/m2lines/Samudra/pull/829#issuecomment-5183854930.

https://claude.ai/code/session_01KLBRBz6SmSrw97FNh9E5RV

PR #834, opened 2026-08-04
Integrate visualizations from our new metrics suite. 

alxmrs:
Part two of the two-part plan for #750. [#829](https://github.com/m2lines/Samudra/pull/829) put the headline observation scalars into W&B at the end of every eval; this turns the same kernels into the figures that explain them.

## Goals

**Make model quality reviewable without anyone running anything by hand.** #829 answers "is this model better than the last one" with nine numbers. It cannot answer "better *where*, and at which scales" — that needs maps and spectra, and today those live in a notebook someone remembers to open.

**Never let a figure and a logged number disagree.** Every value drawn here reduces through `samudra.metrics`, the same code `samudra.eval` scores with. A second implementation that merely agreed today would drift, and the drift would be invisible: a plausible map with a plausible number on it. Each panel is annotated with its metric value and confidence interval, so the figure carries its own scalar rather than sitting next to one.

**Stay faithful to the reviewed science.** These are ports of Yuan's suite (`YuanYuan98/Ocean_Emulator:viz_jupyter/obs_evaluation.py` @ `33d4ae2e`), which has been through scientific review. Where our numbers differ from hers, that is a bug in the port until shown otherwise.

**Keep figure storage off W&B.** A full run is tens of megabytes of PDFs. They go to the bucket or OSN pod; W&B keeps the scalars.

## What it does

Five new viz steps, each skipped unless `observations:` is configured:

| Step | Figures |
| --- | --- |
| `step_obs_rmse_maps` | Per-cell RMSE for velocity, EKE, SST, and both OHC layers |
| `step_obs_annual_rmse` | Per-year totals behind each headline score, with intervals |
| `step_obs_variance_maps` | Residual-anomaly variance for SST and upper-700 m OHC |
| `step_obs_timeseries` | Global-mean SST, EKE and OHC, raw and detrended/deseasonalized |
| `step_obs_spectra` | Spatial and temporal spectra, plus interannual bands |

Twenty figure files, one panel per run listed in the viz config, on a shared colour scale. Viz recomputes everything from the rollout — it does not read eval's output, so it can be re-run independently — and writes `observation_metrics.csv` alongside the figures.

`VizConfig` gains an `observations` block. It is the *same* `ObsMetricsConfig` `EvalConfig` takes, so one `!include ../data/obs.yaml` configures both and the two jobs cannot disagree about which products they scored against.

New `samudra.metrics.spectra`: isotropic 2-D spectra, Welch PSDs, log10-space curve RMSE in dex, and geometric-mean interannual bands. Pure functions over arrays, like `kernels`.

## One pipeline, not two

Scoring a rollout and drawing it are the same work up to the final reduction:
find the observed variable, derive the matching model field, restrict both to
the span and timestamps they share, and interpolate onto the observation grid.
Only then does one take an area-weighted RMSE and the other draw a map.

That prefix lives in `samudra.metrics.comparisons`, and both `samudra.eval` and
`samudra.viz` consume it. Before this it was written out four times per
variable, and `report._aligned` and `viz._paired` were two separate
implementations of the same span intersection — which is where every finding in
review came from. There is now one `align`, one `whole_years`, one pairing mask.

A `Comparison` carries the model field on its own grid *and* on the observation
grid, so reduce-before-regrid is an explicit choice at the call site rather than
something four copies each had to remember.

## Verification

The four spectral kernels were checked by **running** the implementation this
was ported from, not by reading it. `welch_psd`, `log10_rmse_between_curves` and
`interannual_band` agreed exactly — `0.00e+00` across randomised inputs.
`isotropic_spectrum` did not, and the review is written up below.

That implementation is vendored into `tests/reference/` and the test runs it, so
the port is compared against code rather than against numbers copied out of
somebody's run. It covers five grid shapes and a stack, and asserts the *shape*
of the disagreement rather than a blanket tolerance: edge ties move two bins of
thirty-seven, where a symmetric window moves all thirty-seven. It still fails
against every convention it exists to pin.

## Review findings

Every one of these passed the test suite as written, because the tests checked that a spectrum was *a* correct spectrum rather than *the* one the reference computes. Each is now pinned by a test that fails against the old behaviour.

**The port used the reference's function defaults, not its call sites.** Her figures pass `n_factor=2` and `cutoff_before_bins=False` — half the bin width, bins sized to the Fourier corner rather than the Nyquist — and work in metres, reporting rad·km⁻¹.

**Land was filled with the region mean** so the FFT would accept it. Filling puts step changes *inside* the box that no window can taper; they leak power across every wavenumber. The reference requires a fully observed open-ocean rectangle and the regions are curated to be exactly that. A box that is not fully observed now gets no spectrum, and the panel says so.

**The SST spectrum removed the time mean**, leaving the seasonal cycle — which dominates SST variance — in the field. It now removes the calendar-day climatology.

**The interannual EKE band recomputed the velocity mean inside each year**, which absorbs the year-to-year change in mean flow: the one thing the band exists to show. It now shares one multi-year baseline.

**The Hann window was symmetric, not periodic**, and the bin-edge convention was inverted relative to `torch.bucketize`. On a 96-row box every other `k_y` lands exactly on a bin edge, so this was not a corner case.

**The RMSE maps were drawn with a different reduction than the number printed on them.** The scalar weights calendar years equally; the maps used a plain time mean, which weights by sample count. A leap year at 5-day cadence separates them — 73 steps against 74. The builders now hand back squared error and cell areas and let the scoring kernel reduce, so the map and its annotation are one calculation.

**`samudra viz` could not import for an installed user.** `cartopy`, `cmocean` and `tqdm` are imported at module load by code that ships in the wheel, but were only in the `dev` group. Added the `viz` extra.

Two interannual figures missing against the reference suite (SSTA spatial, temporal KE) were added.

## For review: we now differ from the reference on seasonal removal

**This is a deliberate divergence and wants Yuan's eye.**

The reference removes the SST seasonal cycle by grouping on the literal
month-day label (`_calendar_day_climatology_anomaly`). OM4 puts 73 samples in
every calendar year, so a leap year carries one six-day step and every sample
after it lands on a date no common year visits. Over a window holding a single
leap year, 25 of 292 timesteps then have a "climatology" made of themselves and
an anomaly of exactly zero — whole global SSTA fields of zeros going into the
spectra.

The shipped 2015–2022 window has two leap years, so it is softened rather than
absent: month-day bins there hold 2, 6 or 8 samples, which retains anomaly
amplitude at 71%–94% depending on where in the year a sample falls. Pentad bins
hold 8 everywhere, for a uniform 94%.

This repository already binned into pentads for exactly this reason, in both
`detrended_deseasonalized` and `series_without_seasonal_cycle`. The port had
introduced a third convention that bypassed them. It now uses one shared
`without_seasonal_cycle`, and `detrended_deseasonalized` is built from it.

**Consequence: our SSTA spatial spectra and their interannual bands will not
match the reference's**, and we believe the difference is in our favour. Worth
confirming before anyone reconciles the two by eye. `pentad_index` handles the
leap-year alignment explicitly, so the same argument applies to the reference.

## Known gaps

**Spectral scores do not leave the figures.** The reference emits them as scored rows — `surface_eke_spatial_spectrum_log10_rmse_vs_duacs` and friends. We compute the same dex values and render them as legend labels, but they reach no CSV and no W&B. Emitting them means extending `report.py`'s `COLUMNS` contract, which is shared with #829, so it is worth deciding across both rather than here.

**Models are scored on different sets of cells** — filed against #829, [comment](https://github.com/m2lines/Samudra/pull/829#issuecomment-5183854930). The reference applies the most restrictive wet mask across all runs; we apply none, so the model and the OM4 baseline can be scored over different denominators. Not introduced here, and invisible to the acceptance test because OM4-as-model and OM4-as-baseline share a footprint by construction.

**The spectra are validated against the reference implementation, not against published spectral numbers.** The nine RMSE metrics in #829 match Yuan's published column to ≤0.49%; there is no published spectral column to check against, so the claim here is weaker by construction — it is a claim about agreeing with her code, which the vendored test now makes continuously.

**Viz does not draw the OM4 baseline.** `eval` scores it via `observations.baselines` and its CSV carries the rows; viz draws the runs its own `runs:` list names, and logs that the key does not apply. Adding the ground truth as a run would give the side-by-side panels, at the cost of doubling the viz pass — worth deciding rather than assuming.

**Viz is not chained after eval.** Deliberate for now — eval logs the output path, so the figures are one click from the run.

https://claude.ai/code/session_01KLBRBz6SmSrw97FNh9E5RV


src/samudra/viz/core.py
```diff
@@ -3742,6 +3758,103 @@ def global_surface_map_bias(
...
+            return self._obs_cache
+        if self.observations is None or self.obs_data_root is None:
+            raise ValueError(
+                "This viz config has no `observations` block, so there is "
+                "nothing to compare against. Add one (see "
+                "src/samudra/configs/data/obs.yaml)."
+            )
+
+        from samudra.metrics import observations as obs_loaders
+        from samudra.metrics import report as obs_report
```
alxmrs: I think these should be top level imports, unless they cause a circular import issue (but if they do, does that mean we should try to just fix that directly?).

docs/releasing.md
```diff
@@ -26,6 +26,10 @@ pip install samudra
 uv add "samudra[cuda]"
 pip install "samudra[cuda]"
 
+# Figures — needed by `samudra viz`, which imports cartopy at module load
+uv add "samudra[viz]"
```
alxmrs: I think we'll need to update the `uvx` call in the README for the `viz` to include the extra.

src/samudra/configs/samudra_om4_v2/viz.yaml
```diff
@@ -20,3 +20,4 @@ groundtruth_time_range:
 runs:
   - name: "samudra_om4_v2_eval"
     location: ".LOCAL/samudra_om4_v2_eval/predictions.zarr"
+observations: !include ../data/obs.yaml
```
alxmrs: Can we add this in the `samudra_om4` config folder, not the v2 folder? The former folder is the latest configuration -- the one we want to iterate on going fwd. If we applied the obs to the eval config at v2, let's also move that over the `samudra_om4`, that would be a mistake I overlooked.

src/samudra/metrics/comparisons.py
```diff
@@ -0,0 +1,359 @@
...
+    @functools.cached_property
+    def error_squared(self) -> xr.DataArray:
+        """Squared model-minus-observation error, on the observation grid."""
+        return (self.on_obs_grid - self.obs) ** 2
+
+    @property
+    def time(self) -> xr.DataArray:
+        return self.obs["time"]
+
+    def trimmed(self, window: slice) -> Comparison:
```
alxmrs: nit: can we call this `slice`?

alxmrs: That would be more consistent with the data utils IIUC.

pyproject.toml
```diff
@@ -59,6 +59,13 @@ optional-dependencies.cuda = [
   "flash-perceiver>=0.2",
   "torchvision>=0.24.0a0",
 ]
+# `samudra viz` ships in the wheel and imports these at module load, so an
+# install without this extra fails on import rather than at first use.
+optional-dependencies.viz = [
+  "cartopy>=0.24.1",
+  "cmocean>=4.0.3",
+  "tqdm>=4.67.1",
+]
```
alxmrs: How heavyweight, or (for lack of a better term) annoying are these dependencies? If they're simple enough, then maybe we should just include them in the project and avoid adding an extra. WDYT? Viz is a normal part of our experimenting process.

alxmrs (commented):
Some initial thoughts.

src/samudra/metrics/comparisons.py
```diff
@@ -0,0 +1,359 @@
...
+    @functools.cached_property
+    def error_squared(self) -> xr.DataArray:
+        """Squared model-minus-observation error, on the observation grid."""
+        return (self.on_obs_grid - self.obs) ** 2
+
+    @property
+    def time(self) -> xr.DataArray:
+        return self.obs["time"]
+
+    def trimmed(self, time_slice: slice) -> Comparison:
```
alxmrs: Where did we land with following conventions for time slicing in this project? Why trimmed and not `slice` like `DataSource`?

src/samudra/metrics/comparisons.py
```diff
@@ -0,0 +1,359 @@
...
+        """The same comparison over a narrower span."""
+        return Comparison(
+            self.label,
+            self.native.sel(time=time_slice),
+            self.obs.sel(time=time_slice),
+            self.area,
+        )
+
+    @property
+    def pairing(self) -> dict[str, float]:
```
alxmrs: I think this should be a Pairing dataclass, or a more specific ADT than a dict of floats (since the number of entries is specifically two).

src/samudra/metrics/comparisons.py
```diff
@@ -0,0 +1,359 @@
...
+
+    @property
+    def area(self) -> xr.DataArray:
+        return self.eastward.area
+
+    @property
+    def time(self) -> xr.DataArray:
+        return self.eastward.obs["time"]
+
+    def trimmed(self, time_slice: slice) -> VelocityComparison:
```
alxmrs: Let's follow the slice pattern.

src/samudra/metrics/comparisons.py
```diff
@@ -0,0 +1,359 @@
...
+    sim_u, obs_u = align(sim_u, obs_u, window, f"{context} eastward velocity")
+    sim_v, obs_v = align(sim_v, obs_v, window, f"{context} northward velocity")
+    area = duacs["area"]
+    return VelocityComparison(
+        eastward=Comparison(f"{context} eastward velocity", sim_u, obs_u, area),
+        northward=Comparison(f"{context} northward velocity", sim_v, obs_v, area),
+    )
+
+
+def sea_surface_temperature(
```
alxmrs, line 379: Should these specific metrics (that return comparisons) live in kernels, because they are the things we want to measure? Then, the ADTs that define what a comparison would be the things that live in this class. WDYT?

src/samudra/viz/norms.py
```diff
@@ -0,0 +1,79 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Colour limits for map panels.
```
alxmrs, line 5: Is there a file or method that does this already? Can you double check?

alxmrs (commented):
Another review pass

alxmrs:
Done in `aae48940`. Answers to the two questions:

**Should the comparison builders live in `kernels`?**

I'd keep them where they are, on a technical constraint rather than taste: `observations` already imports `kernels`, so `kernels` importing `observations` is a cycle. And the builders need it — `find_var_name`, `duacs_velocity`, `ARGO_TEMPERATURE_ALIASES` — because knowing that DUACS spells eastward velocity `ugos` is product knowledge, which is the one thing `kernels` promises not to have ("pure functions over `xr.DataArray`. No I/O, no plotting, no state").

I think the split you're describing already exists, just under different names. The things we measure *are* in `kernels`: `geostrophic_velocity_from_zos`, `instantaneous_surface_eke`, `ohc_per_area_layer_maps`, `residual_variance_map`. What's in `comparisons` is not a measurement — it's the assembly that turns a rollout and a product into two arrays that can be measured: find the variable, derive the model field, intersect the span, put it on one grid. `surface_velocity` computes no metric; it hands you a pair that `report` or `viz` then reduces.

So the layering is kernels (measure) → comparisons (assemble, plus the ADTs) → report/viz (reduce and present). Happy to make that explicit in the module docstring if it isn't landing.

**Is there already something that does what `norms.py` does?**

Yes — and this PR doesn't add a second one. `percentile_norm`, `symmetric_percentile_norm` and `_flatten_for_norm` were already on `main` in `viz/core.py`, used eight times there. The previous review round caught me reimplementing them as a private `_robust_limits` in `viz/observations.py`; I deleted that and reused these instead.

`norms.py` is purely a relocation, so that `viz.observations` can use them without `viz.core` and `viz.observations` importing each other — which is what forced the function-local imports you flagged last round. I verified the moved code is byte-identical to what's on `main`, so there's no behaviour change.

One near-miss worth naming: `aggregator/plotting.get_cmap_limits` is the same *idea* but min/max rather than percentile, so one outlier flattens the map. It's used by `aggregator/inference/time_mean.py` and changing it would move those figures, so I left it alone — but it's a candidate to fold in whenever the aggregators get revisited.

src/samudra/metrics/comparisons.py
```diff
@@ -0,0 +1,385 @@
...
+    def time(self) -> xr.DataArray:
+        return self.obs["time"]
+
+    def slice(self, time_slice: slice) -> Self:
+        """Restrict the comparison to a time slice, as `DataSource.slice` does.
+
+        Takes the slice rather than a `TimeConfig`: these spans are derived from
+        the data -- the period several runs and a product have in common -- not
+        read from configuration, and a `TimeConfig` would assert a calendar the
+        observation products do not share.
```
alxmrs: This comment is unnecessary. It is confusing to maintainers.

alxmrs:

alxmrs:
@fomo-bot may I have your review of this PR?

src/samudra/viz/core.py
YuanYuan98: Looks good to me now.

docs/releasing.md
YuanYuan98: Makes sense. Since these are regular dependencies now, I don’t think the `uvx` command needs to change.

src/samudra/configs/samudra_om4_v2/viz.yaml
YuanYuan98: Yep, `samudra_om4` is the right place for this. Looks good now.

src/samudra/metrics/comparisons.py
YuanYuan98: Yep, `slice` is more consistent here.

pyproject.toml
YuanYuan98: Agreed.

src/samudra/metrics/comparisons.py
YuanYuan98: Yeah, `slice` matches the existing convention better.

src/samudra/metrics/comparisons.py
YuanYuan98: Agreed.

src/samudra/metrics/comparisons.py
YuanYuan98: Looks good now.

src/samudra/metrics/comparisons.py
YuanYuan98: Yeah, that makes sense. These builders need to know about things like the DUACS velocity names and ARGO temperature aliases, plus the alignment and regridding steps, so keeping them in `comparisons` feels right. It also avoids the import cycle.

src/samudra/viz/norms.py
YuanYuan98: Thanks for checking. Since these are the existing helpers moved out of `viz/core.py`, I don’t see a duplication issue here.

src/samudra/metrics/comparisons.py
YuanYuan98: Agreed.

src/samudra/viz/observations.py
```diff
@@ -0,0 +1,982 @@
...
+
+# Colour per run, with the reference first. Distinguishable in greyscale and to
+# the most common colour-vision deficiencies.
+SERIES_COLOURS = ("#000000", "#0072b2", "#d55e00", "#009e73", "#cc79a7")
+
+
+# Steps a calendar year needs before it contributes to an interannual band.
+# Twelve is roughly two months at the 5-day cadence: enough for an annual mean
+# not to be one season.
+MIN_STEPS_PER_YEAR = 12
```
YuanYuan98: Could we use `complete_calendar_years` here instead of a 12-step cutoff? With the default rollout starting on 2014-10-20, 2014 still has about 15 five-day samples, so it gets included in `_anomaly_by_year`, `_yearly_temporal_bands`, and `obs_eke_by_year` even though it is only a partial year. That could skew the interannual bands and their reported year counts.

alxmrs:
@fomo-bot Do you think this PR is ready to merge? If so, will you approve the PR?

PR #841, opened 2026-08-12
Add successive-halving Slurm architecture harness

alxmrs:
## Goal

Use fragmented compute capacity to compare model architectures quickly, then allocate larger epoch budgets only to promising candidates.

This implements one successive-halving bracket, the resource-allocation primitive underlying [Hyperband](https://jmlr.org/papers/v18/16-558.html). It deliberately leaves multiple-bracket Hyperband and deferred rollout metrics for later work.

## User interface

Run a complete search with:

```console
python -m samudra.search path/to/search.yaml
```

Users do not manually advance rungs. The Local executor runs candidates sequentially in isolated multiton/logging scopes; the Slurm executor submits arrays and dependent promotion jobs automatically. Search configs use Samudra's Pydantic/YAML system, including `!include`, packaged presets, CLI overrides, and generated schemas.

## Design and correctness

- `SuccessiveHalving` owns ranking and promotion; `SearchConfig.build()` constructs it.
- Local and Slurm backends are selected through the exercised executor seam.
- Rung budgets are cumulative. Promoted candidates resume checkpoints while all rungs share the final-rung LR-scheduler horizon, preventing `T_max` corruption.
- A Slurm rung-zero probe must complete a real optimizer update before either candidate arrays or full-budget anchors are released.
- Array job IDs are persisted before controller submission, so controller retries cannot double-submit training work.
- Search state is versioned and Pydantic-validated on every read/write, including run identity and rung layout.
- A search is `complete` only if every scheduled result is eligible; recoverable worker failures produce `partial`, and a rung with no eligible candidate records a durable terminal failure.

## Reproducibility, observability, and W&B

- The search snapshots each fully resolved candidate config and rejects dirty Slurm launches. Controllers receive the immutable recorded commit and do not require a Git checkout.
- W&B uses the unique `run_id` as the group; runs receive stable search-name, run-ID, and candidate tags.
- Structured worker status records bounded lifecycle events (`launched`, `initialized`, `first_batch`, first `optimizer_step`, terminal stage), rather than rewriting unbounded per-step history.
- Non-finite metrics are represented explicitly and classified as divergence rather than masquerading as missing output.
- `results.csv`/`results.parquet`, `epochs.parquet`, `artifacts.parquet`, resolved configs, reports, state, provenance, W&B identity, and configurable checkpoints form the agent-observable record.
- Raw scheduler and process logs are excluded from publication by default because workers inherit credentials. They require explicit `artifacts.logs: all`; structured error strings are scrubbed.
- Successful upload hashes are retained locally, so publication retries skip unchanged large checkpoints.
- Controller CPU, memory, partition, and walltime are configurable.

See `docs/search.md` for setup, schemas, safety policy, and DuckDB examples.

## Verification

- Standard non-manual/non-CUDA suite: 401 passed, 2 skipped, 10 xfailed.
- Review-focused scheduler, search, artifact, summary, local-isolation, and logging tests: 41 passed; atomic failure cleanup: 1 passed.
- Pre-commit: all checks passed, including Ruff, mypy, generated schemas, secret detection, and REUSE.


docs/successive-halving.md
```diff
@@ -0,0 +1,156 @@
+<!--
+SPDX-FileCopyrightText: 2026 Samudra Authors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Successive-halving architecture searches
+
+`samudra search` allocates progressively larger epoch budgets to promising model
```
alxmrs: I would spend the human attemtion reading budget in this doc explaining the idea of the hyperband class of algorithms, and link to that paper explicitly, then discuss how to practically use it. Talking about nit picky implementation details that are only relevant to our conversations is not helpful.

alxmrs: A brief summary of hyperband and the goals it selves is ok. So is linking to the paper for users to read more. Re-explaining the nitty gritty details of the algorithm is not helpful IMO. Explaining that it is a way to efficiently search through hparams and model architectures is good, as well as explaining how to get set up on the tool for immediate use.

alxmrs: Good coverage of the API surface is appreciated.

alxmrs: Examples are appreciated.

docs/successive-halving.md
```diff
@@ -0,0 +1,156 @@
+<!--
+SPDX-FileCopyrightText: 2026 Samudra Authors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
```
alxmrs: I think this should be `docs/search.md`.

scripts/successive_halving.example.yaml
```diff
@@ -0,0 +1,47 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: CC0-1.0
```
alxmrs: I think this should live in `src/samudra/configs`

scripts/successive_halving.example.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+runtime:
+  output_base: /scratch/USER/runs/example-2deg-architecture-search
+  train_harness: /scratch/USER/Samudra/scripts/slurm_apptainer_train.sbatch
+  data_root: /scratch/USER
+  scratch_dir: /scratch/USER
+  sif_path: /scratch/USER/.apptainer-images/physicsnemo-26.05-latest.sif
+  wandb_mode: online
+
+compute:
```
alxmrs: We should use the templating system to make templaes for torch for the compute and runtime.

scripts/successive_halving.example.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+schema_version: 1
+name: example-2deg-architecture-search
+rungs: [1, 3, 6, 12]
+promotion_fraction: 0.5
+minimum_promoted: 1
+metric: validation_loss
+mode: min
+
+runtime:
```
alxmrs: See below -- this should be a template.

src/samudra/utils/wandb.py
```diff
@@ -62,6 +63,25 @@ def _make_config(self, cfg: "AnyTopLevelConfig", data_container: "DataContainer"
...
+            "parent_checkpoint": "SAMUDRA_SEARCH_PARENT_CHECKPOINT",
+            "compute_backend": "SAMUDRA_SEARCH_COMPUTE_BACKEND",
+            "compute_job_id": "SAMUDRA_SEARCH_COMPUTE_JOB_ID",
+        }
+        search = {
+            name: value
+            for name, env_var in search_env.items()
+            if (value := os.environ.get(env_var))
+        }
+        if search:
```
alxmrs: Are these namespaced or tagged in wandb somehow? Will I be able to filter them? Feel free to look up docs.

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Launch and promote Samudra candidates through successive-halving rungs."""
```
alxmrs: Please add a reference to the hyperband paper.

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+from typing import Any, ClassVar, NoReturn, cast
+
+import yaml
+
+SCHEMA_VERSION = 1
+SUMMARY_NAME = "training_summary.json"
+CHECKPOINT_RELATIVE_PATH = Path("saved_nets/ckpt.pt")
+
+
+class ComputeBackend(ABC):
```
alxmrs: Can we use our base config system?

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+def _compute_backend_type(manifest: dict[str, Any]) -> type[ComputeBackend]:
+    config = _require(manifest, "compute", dict)
+    name = _require(config, "type", str)
+    if backend := _COMPUTE_BACKENDS.get(name):
+        return backend
+    matches = list(
+        importlib.metadata.entry_points().select(
+            group=COMPUTE_BACKEND_ENTRY_POINT, name=name
+        )
+    )
```
alxmrs: I'd rather expose this via a python dictionary mechanism (say, in __init__) rather than an importlib based system.

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+
+
+def _slug(value: str) -> str:
+    slug = re.sub(r"[^a-zA-Z0-9_.-]+", "-", value).strip("-.")
+    if not slug:
+        _die(f"Name has no usable characters: {value!r}")
+    return slug
+
+
+def _load_yaml(path: Path) -> dict[str, Any]:
```
alxmrs: Yo, we got a really great config system already, we don't need to reinvent a new one. 

I'm happy for configs to live in this file for now, but not to create a new system.

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+
+def _sbatch_output(args: list[str], *, dry_run: bool) -> str:
+    print(shlex.join(args))
+    if dry_run:
+        return "DRY-RUN"
+    result = subprocess.run(args, check=True, text=True, capture_output=True)
+    return result.stdout.strip().split(";")[0]
+
+
+class SlurmBackend(ComputeBackend):
```
alxmrs: What? shouldn't this live in a subpackage and not this file? Something like an `search.executors` package?

src/samudra/search.py
alxmrs: 🐑 Should this also be called an "executor"? Compute Backend is pretty good, too.

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+
+def _manifest_hash(path: Path) -> str:
+    return hashlib.sha256(path.read_bytes()).hexdigest()
+
+
+def _file_hash(path: Path) -> str:
+    return hashlib.sha256(path.read_bytes()).hexdigest()
+
+
+def _orchestrator_provenance(*, allow_dirty: bool) -> dict[str, Any]:
```
alxmrs: I believe this could be factored better with our config system too / pydanitc settings, which allows for parsing env vars. I could be wrong.

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+
+    def __init__(self, manifest_path: Path, state_path: Path):
+        self.manifest_path = manifest_path
+        self.state_path = state_path
+        self.manifest = _load_yaml(manifest_path)
+        self.config = self.manifest["compute"]
+
+    @classmethod
+    @abstractmethod
+    def validate_config(cls, config: dict[str, Any], rungs: list[int]) -> None:
```
alxmrs: On principle, I don't know if we need this method, since pydantic settings should mostly handle this on parse. Maybe this is needed as a validator or one off method?

scripts/successive_halving.example.yaml
```diff
@@ -0,0 +1,47 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: CC0-1.0
+
+schema_version: 1
+name: example-2deg-architecture-search
+rungs: [1, 3, 6, 12]
+promotion_fraction: 0.5
+minimum_promoted: 1
+metric: validation_loss
```
alxmrs: Can we allow for multiple metrics?

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+            _die(f"Promotion checkpoint does not exist: {checkpoint}")
+        overrides.append(f"--resume_ckpt_path={checkpoint}")
+        environment["SAMUDRA_SEARCH_PARENT_CHECKPOINT"] = str(checkpoint)
+    environment["ARGS"] = shlex.join(overrides)
+    print(
+        f"candidate={candidate_name} rung={rung_index} epochs={epoch} "
+        f"output={output_dir}",
+        flush=True,
+    )
+    subprocess.run([str(runtime["train_harness"])], env=environment, check=True)
```
alxmrs: What is a runtime? Why do we run a subprocess here? Shouldn't we be able to run this within the compute environment? Something is weird here.

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+
+"""Launch and promote Samudra candidates through successive-halving rungs."""
+
+from __future__ import annotations
+
+import argparse
+import csv
+import hashlib
+import importlib.metadata
+import json
```
alxmrs: This is reflecting on the code below in general: I somewhat loathe the json format for capturing log information like this. At least jsonl is better, though I'd prefer something structured better like parquet. Can we denormalize the data we're logging to be more analyzable, like some modern tabular format that would be easy to perform analysis against (decent query experience)?

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+) -> Path:
+    manifest = _load_yaml(manifest_source)
+    validate_manifest(manifest)
+    provenance = _orchestrator_provenance(allow_dirty=allow_dirty)
+    search_dir = state_root / _slug(manifest["name"])
+    if search_dir.exists():
+        _die(f"Search state already exists: {search_dir}")
+    search_dir.mkdir(parents=True)
+    manifest_path = search_dir / "manifest.yaml"
+    shutil.copy2(manifest_source, manifest_path)
```
alxmrs: Should the dir and sandbox mechanism live on the compute backend ADT?

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+        ],
+    }
+    _atomic_json(state_path, state)
+    backend.submit_anchors(state, dry_run=dry_run)
+    state = _load_state(state_path)
+    backend.submit_rung(state, 0, dry_run=dry_run)
+    return state_path
+
+
+def run_task(
```
alxmrs: What is the purpose of run task in the first place? To do a single experiment? Why should we use this and not a normal train.py job (after making update the config or using the CLI)?

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+    )
+    leaderboard = eligible + [result for result in results if not result["eligible"]]
+    if not eligible:
+        state["status"] = "failed"
+        rung["results"] = results
+        _atomic_json(state_path, state)
+        _write_leaderboard(
+            state_path.parent / f"leaderboard-r{rung_index}.csv", leaderboard
+        )
+        _die(f"No candidate completed rung {rung_index}")
```
alxmrs: This is pretty ugly query code -- it suggests to me that the log data model is wrong if we have to re-implement analysis like this. 

It would be better if we could express at least in terms of a standard dataframe library with a fluent interface.

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
...
+        )
+    elif args.command == "run-task":
+        run_task(
+            args.manifest,
+            args.state,
+            args.rung,
+            args.task,
+            anchor=args.anchor,
+        )
+    elif args.command == "advance":
```
alxmrs: I'm confused about the purpose of these subcommands. I was expecting to more or less run a search job and then get data and an analysis summary at the end (of which is configurable on the manifest/config system). Why do we have four commands instead of one? Are all these essential? Do I manually have to advance each state? Why would I want that affordance?

tests/test_cli.py
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
```
alxmrs: This test is unnecessary.

src/samudra/search.py
```diff
@@ -0,0 +1,893 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Launch and promote Samudra candidates through successive-halving rungs."""
+
+from __future__ import annotations
+
```
alxmrs: I was kind of hoping for a simple way to mix in a different searching algorithm. What if we wanted to use something else instead of successive halving rings? We shouldn't implement the something else now, but having a version of this that was easily refactorable to perform other forms of search later would be a big win.

alxmrs: Like, I'm not asking for a pluggable system. I'm asking for the ability to easily make a pluggable system when we need it.

src/samudra/cli.py
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""Console-script entry point: ``samudra <train|eval|viz> CONFIG [OVERRIDES...]``.
+"""Console entry point for Samudra's training, evaluation, and search tasks.
```
alxmrs: Seeing that this script is more of a manager of the standard `train.py` script that we currently use, I don't think we should add it to our CLI.

src/samudra/train.py
```diff
@@ -435,10 +438,81 @@ def run(self) -> None:
...
+                            end_epoch_train_time - start_epoch_train_time
+                        ),
+                        "epoch_validation_seconds": (
+                            end_epoch_val_time - end_epoch_train_time
+                        ),
+                        "epoch_total_seconds": time_elapsed,
+                        "progress": self.train_progress.state_dict(),
+                        "wandb_id": self.wandb_id,
+                        "wandb_name": self.wandb_name,
+                        "search": {
```
alxmrs: Driven by a config system, I think that this should only be enabled when we do a search with the tool (maybe, these should be added via fn call to be more succinct).

src/samudra/train.py
```diff
@@ -435,10 +438,81 @@ def run(self) -> None:
...
+                            "compute_job_id": os.environ.get(
+                                "SAMUDRA_SEARCH_COMPUTE_JOB_ID"
+                            ),
+                            "slurm_job_id": os.environ.get("SLURM_JOB_ID"),
+                            "slurm_array_job_id": os.environ.get("SLURM_ARRAY_JOB_ID"),
+                            "slurm_array_task_id": os.environ.get(
+                                "SLURM_ARRAY_TASK_ID"
+                            ),
+                        },
+                        "provenance": {
```
alxmrs: I like logging this all the time. Can we make it work for normal train? This can be done later.

alxmrs: Also: don't we currently get most of this info in wandb today??

alxmrs (commented):
Could use some major improvement. That we run the `train.py` script with another script via a subprocess and don't just import what we need from train.py into a new runner in search is also a suprise (the train class should be importable! We have lots of examples setting these up in tests).

docs/successive-halving.md
alxmrs: Rewritten as a practical guide: a brief explanation of the search goal, an explicit Hyperband paper link, setup, one-command usage, outputs, examples, and the Python API. The conversational implementation history is gone.

docs/successive-halving.md
alxmrs: Renamed to `docs/search.md` and updated the documentation navigation and Torch cross-link.

scripts/successive_halving.example.yaml
alxmrs: Moved the example into the shipped presets at `src/samudra/configs/search/example.yaml`.

src/samudra/utils/wandb.py
alxmrs: Yes. Search runs now use the search name as the W&B group and carry `search`, search-name, and candidate-name tags. The complete identity is also namespaced under `config.experiment.search`, so group, tag, or config filters all work.

scripts/successive_halving.example.yaml
alxmrs: The example now uses the normal `!include` mechanism, with separate packaged `search/torch.yaml` and `search/local.yaml` executor templates.

src/samudra/search.py
alxmrs: Replaced the custom parser with `SearchConfig` and nested `BaseConfig` models. Loading now goes through `from_yaml_and_cli`, so search shares includes, overrides, validation, and generated schemas with the rest of Samudra.

src/samudra/search.py
alxmrs: Removed entry-point and importlib discovery. Local and Slurm executors are selected by a small explicit Python dictionary.

src/samudra/search.py
alxmrs: Added the Li et al. Hyperband reference in the module docstring and linked the paper directly from `docs/search.md`.

src/samudra/search.py
alxmrs: The bespoke YAML validation and merge code has been deleted. This now uses the existing Pydantic-based Samudra config system end to end.

src/samudra/search.py
alxmrs: Done. The interface and implementations now live under `samudra.search.executors`, split into base, local, and Slurm modules.

src/samudra/search.py
alxmrs: The former runtime/env parsing layer was removed. Typed Pydantic executor configs now own these settings; compute-provided identifiers such as the Slurm job ID remain environment-derived.

src/samudra/search.py
alxmrs: Replaced the nested analysis JSON with a denormalized `results.csv`: one row per candidate/rung with metric, timing, provenance, W&B, job, and lineage columns. JSON is retained only for internal resumable scheduler state.

src/samudra/search.py
alxmrs: Removed along with the custom configuration implementation; Pydantic validators now cover the cross-field constraints that remain necessary.

scripts/successive_halving.example.yaml
alxmrs: Added `metrics: list[str]`. One objective metric controls promotion, while every requested finite metric becomes a column in `results.csv`.

src/samudra/search.py
alxmrs: The worker task is now private executor plumbing, not a user command. It directly configures and invokes `Trainer`; the public operation is one search command.

src/samudra/search.py
alxmrs: Yes. Output/scheduler concerns are now owned by the executor; the former sandbox/runtime object no longer exists.

tests/test_cli.py
alxmrs: Removed this test together with the top-level CLI integration it tested.

src/samudra/search.py
alxmrs: Ranking is now a pandas fluent operation (`sort_values(...).head(...)`) over the same flat records written to `results.csv`.

src/samudra/search.py
alxmrs: There is now one public command: `python -m samudra.search CONFIG`. Slurm dependency jobs advance and submit later rungs automatically; task and advance entry points are private executor machinery.

src/samudra/train.py
alxmrs: Yes, most observability remains in W&B. The local summary is intentionally small and search-only: it supplies the promotion metric and completion/checkpoint/provenance checks without making scheduling depend on W&B availability.

src/samudra/cli.py
alxmrs: Removed search from the `samudra` console CLI. It is now an independent manager invoked with `python -m samudra.search`.

src/samudra/train.py
alxmrs: Agreed this can be useful later. I kept this PR scoped to the search consumer; ordinary training continues to rely on its existing checkpoint and W&B records.

src/samudra/search.py
alxmrs: Removed the runtime abstraction and the training subprocess. The worker now resolves a `TrainConfig`, constructs `Trainer` directly inside the executor environment, and calls `run()`.

src/samudra/search.py
alxmrs: Separated `SuccessiveHalving` from the executors and added a small `build_search()` boundary. A future algorithm can become another class and factory branch without committing to a plugin framework now.

src/samudra/train.py
alxmrs: Search summaries are now gated by typed `experiment.search` configuration, and construction moved into the concise `_search_summary()` helper. Ordinary training does not emit this file.

alxmrs:
Implemented the agent-observability extension in `c2dcc35d`.

The boundary I chose is search-level artifact publication, independent of Local/Slurm execution. An opt-in OSN template uses standard environment credentials; no secrets enter configs or artifacts. Each search now emits local CSV plus queryable Parquet results, consolidated epoch-level scalar histories, failure and scheduler logs, runtime/container/code-layer provenance, timestamps, W&B/public-record links, and a hashed artifact inventory. Finalists retain best-validation checkpoints by default.

I did not automatically run PR #834's metrics/viz suite inside each rung. Those diagnostics require completed rollout datasets and would make the cheap inner loop expensive. Instead, `analysis/` is now a published artifact contract, and the retained finalist config/checkpoint pairs are sufficient for a later post-search evaluator to generate matched rollouts, observation tables, maps, spectra, and time series there.

A useful next increment after the first real Perceiver search is a finalist-analysis config that selects rollout length and a small diagnostic subset. The first runs should tell us whether variable/depth loss trajectories plus short-rollout bias/variance/spectra are the highest-value set, or whether Perceiver-specific probes such as latent utilization and attention entropy deserve priority.

docs/search.md
```diff
@@ -0,0 +1,221 @@
...
+
+Run the entire search with one command:
+
+```bash
+python -m samudra.search path/to/search.yaml
+```
+
+The runner submits the first rung and fixed baselines. On Slurm, dependent
+controller jobs automatically rank completed candidates and submit each later
+rung; users do not manually advance the search. For a local laptop run, include
```
alxmrs: Could local also be used in colab? If so, maybe we could say "For a local run (on a laptop or a Colab notebook)"

docs/search.md
```diff
@@ -0,0 +1,221 @@
...
+`search/local.yaml` instead of `search/torch.yaml`; candidates then run
+sequentially in the current environment.
+
+The search directory contains:
+
+- `config.yaml`: the fully resolved, validated search configuration;
+- `candidates/`: one fully resolved training configuration per candidate;
+- `state.json`: internal resumable scheduler state;
+- `results.csv` and `results.parquet`: one analysis-ready row per candidate and
```
alxmrs, line 55: What's the sharding story for the results parquet?

docs/search.md
```diff
@@ -0,0 +1,221 @@
...
+python -m samudra.search path/to/search.yaml
+```
+
+The runner submits the first rung and fixed baselines. On Slurm, dependent
+controller jobs automatically rank completed candidates and submit each later
+rung; users do not manually advance the search. For a local laptop run, include
+`search/local.yaml` instead of `search/torch.yaml`; candidates then run
+sequentially in the current environment.
+
+The search directory contains:
```
alxmrs: Is the search directory namespace with a subdir? Can we make the names readable, but unique per experiment (for example, could we use the config name + timestamp, or the config name + hash of the config?).

docs/search.md
```diff
@@ -0,0 +1,221 @@
...
+become distinguishable after only a small amount of training. Samudra's search
+runner uses successive halving to train all candidates with a small initial
+budget, retain the best fraction, and give progressively larger budgets only to
+the survivors. This is useful for comparing model architectures as well as
+ordinary hyperparameters such as learning rate and batch size.
+
+Successive halving is the resource-allocation primitive used by
+[Hyperband](https://jmlr.org/papers/v18/16-558.html). Samudra currently runs one
+successive-halving bracket. It does not yet run Hyperband's collection of
+brackets with different exploration-versus-training tradeoffs.
```
alxmrs, line 19: Is is possible to add the explore vs exploit tradeoff component in the future, given our interfaces now?

docs/search.md
alxmrs: Will sharding ever really be needed?

docs/search.md
```diff
@@ -0,0 +1,221 @@
...
+Search currently records and publishes the inputs needed for a later analysis
+stage rather than coupling training to a particular scientific metric suite.
+The next useful analysis implementation is finalist evaluation: load each
+published final checkpoint, generate a matched validation rollout, then apply
+the observation metrics and figures from `samudra.metrics` and `samudra.viz`.
+Those tools operate on completed rollout datasets, so running them as a
+separate post-search job keeps cheap promotion decisions fast while still
+producing maps, spectra, and time series that explain *why* a finalist worked
+or failed. The publisher is the executor-independent boundary where those
+analysis artifacts will be added once the first Perceiver searches establish
```
alxmrs: Talking about our specific searches is not helpful in these kinds of doc. We can just keep it vague and say it will be added later.

src/samudra/search/artifacts.py
```diff
@@ -0,0 +1,236 @@
...
+
+def _sha256(path: Path) -> str:
+    digest = hashlib.sha256()
+    with path.open("rb") as stream:
+        for block in iter(lambda: stream.read(1024 * 1024), b""):
+            digest.update(block)
+    return digest.hexdigest()
+
+
+def atomic_parquet(frame: pd.DataFrame, path: Path) -> None:
```
alxmrs: Does this work when writing to a cloud optimized store? Or, is this for writing to a local store? 

Why are these atomic?

alxmrs: `frame.to_parquet` might have all the logic that we need to handle this (e.g. writing to an object store).

src/samudra/search/artifacts.py
```diff
@@ -0,0 +1,236 @@
...
+        frame.to_parquet(temporary, index=False)
+        os.replace(temporary, path)
+    finally:
+        temporary.unlink(missing_ok=True)
+
+
+class ArtifactPublisher:
+    """Materialize and mirror the complete inspectable record of a search."""
+
+    def __init__(self, search: SuccessiveHalving, config: ArtifactConfig) -> None:
```
alxmrs: Can we make the search type a search protocol? Can we make this work on any search alg?

src/samudra/search/artifacts.py
```diff
@@ -0,0 +1,236 @@
...
+            exists = self._s3(self.destination).exists(
+                f"{self.destination.bucket}/{self.destination.path}"
+            )
+        if exists:
+            raise ValueError(f"Artifact destination already exists: {self.destination}")
+
+    @property
+    def root(self) -> str:
+        if self.config.public_url is not None:
+            return f"{self.config.public_url.rstrip('/')}/{self.search.slug}"
```
alxmrs: Are we enforcing unique slugs per experiment? That is more of the contention we have with our current training experiments. Please look to see what we do there for output_dirs and mimic that approach here.

src/samudra/search/artifacts.py
```diff
@@ -0,0 +1,236 @@
...
+        files = self._files(state)
+        self._write_epochs(files)
+        files = self._files(state)
+        catalog = pd.DataFrame(self._catalog_rows(files, state))
+        atomic_parquet(catalog, self.search.search_dir / CATALOG_NAME)
+        files.append((self.search.search_dir / CATALOG_NAME, CATALOG_NAME))
+        for source, relative in files:
+            self._put(source, relative)
+
+    def _files(self, state: dict[str, Any]) -> list[tuple[Path, str]]:
```
alxmrs, line 114: Can you add a succinct, 1-line docstring to explain what files does?

alxmrs (commented):
Another round of review, then we can improve this by trying it out.

docs/search.md
alxmrs: Updated the wording to explicitly include both laptops and Colab notebooks.

docs/search.md
alxmrs: Documented this explicitly. These are intentionally single Parquet files: their rows scale with candidates × rungs or candidates × epochs, not with the ocean dataset, so they should remain tiny relative to checkpoints. If a future workload disproves that assumption, the same schema can be exposed as a partitioned Parquet dataset without changing how DuckDB queries it.

docs/search.md
alxmrs: Implemented. Every invocation now defaults to `<readable-name>--<microsecond-UTC-timestamp>`, and that `run_id` consistently namespaces local directories, training output directories, Slurm jobs, W&B groups, result rows, and published objects. An explicit `run_id` is also supported for an external allocator, and existing destinations are still never overwritten.

docs/search.md
alxmrs: Yes, and I clarified the extension boundary in the docs. A future Hyperband implementation can coordinate multiple successive-halving brackets through the current rung submissions and reuse the algorithm-neutral publisher. I did not claim that an arbitrary non-rung explore/exploit scheduler already fits: that would honestly require broadening the executor task interface when we add one.

docs/search.md
alxmrs: Probably not for the intended workload. I documented why: table cardinality is bounded by candidates/rungs/epochs rather than data samples, so a single file is the simplest useful contract. Partitioning remains a compatible escape hatch, not complexity we need now.

docs/search.md
alxmrs: Agreed. I removed the Perceiver-specific proposed analysis and kept this as a vague executor-independent extension point for future post-search diagnostics.

src/samudra/search/artifacts.py
alxmrs: Renamed this to `atomic_local_parquet` and made its one-line contract explicit. It is intentionally local: a search first assembles a complete local snapshot, atomically replacing catalogs so a controller cannot observe a truncated file, then uploads closed files to local or S3 destinations.

alxmrs: I’m retaining the staged write here. `DataFrame.to_parquet()` can target S3, but a direct object write does not make the overall multi-file research record atomic, and it makes retry/debug behavior depend on the remote filesystem. We instead atomically close each local catalog, hash the exact bytes, and publish that immutable file through the destination adapter. This keeps local and S3 records byte-identical and retryable.

src/samudra/search/artifacts.py
alxmrs: Done. `ArtifactPublisher` now consumes a small `ArtifactSearch` protocol (plus a minimal config protocol), not `SuccessiveHalving`. Any algorithm that exposes the publication surface can use it.

src/samudra/search/artifacts.py
alxmrs: Implemented a unique run namespace throughout. The generated `run_id` is readable, collision-resistant, saved in config/state/results, and used by each candidate output directory. This automates the unique experiment-name discipline used by training while retaining the existing refusal to overwrite.

src/samudra/search/artifacts.py
alxmrs: Added: `_files` now documents that it lists local search files together with their destination-relative keys.

src/samudra/search/worker.py
```diff
@@ -0,0 +1,43 @@
...
+    task.add_argument("task", type=int, nargs="?")
+    task.add_argument("--anchor", action="store_true")
+    advance = commands.add_parser("advance")
+    advance.add_argument("config", type=Path)
+    advance.add_argument("state", type=Path)
+    advance.add_argument("rung", type=int)
+    args = parser.parse_args()
+
+    config = SearchConfig.from_yaml_and_cli([str(args.config)])
+    search = build_search(config)
```
alxmrs: Nit: I think our codebase convention is to add the build method to the config so this would become: 
```
search = config.build()
```

alxmrs: Agreed—updated the codebase to use the config-owned factory consistently: `search = config.build()`. I removed the standalone `build_search` helper, migrated the CLI, worker, tests, and docs, and kept the algorithm selection boundary inside `SearchConfig.build()`.

alxmrs:

alxmrs:
@fomo-bot may I also have your review?

docs/search.md
```diff
@@ -0,0 +1,275 @@
...
+candidate list. Build the configured code layer from that same committed
+revision. The runner verifies the layer's manifest before submitting work, and
+completed training results must report the same commit before promotion. All
+candidate models must therefore exist in one experiment branch before the
+search is submitted.
+
+Run the entire search with one command:
+
+```bash
+python -m samudra.search path/to/search.yaml
```
alxmrs, line 34: Question for the reviewer -- should this be promoted to the top level Samudra CLI? My decision so far has been to wait.

alxmrs (commented):
🤖 **Critical review — successive-halving harness**

I read this as a fresh adversary: 3.3k lines, a new subpackage, a new persistent-state machine, changes to the shared training loop and the shared Slurm harness. The provenance/artifact design is genuinely good — commit pinning across controller/code-layer/summary, atomic writes everywhere, the SHA-256 catalog, the fail-fast probe, and `docs/search.md` are all above the bar for research code. The `EXECUTORS` seam is the right size.

But I do not think this is landable yet. Three findings are blocking, and two of them mean an advertised path has never actually been run:

1. **Every promoted candidate trains under a corrupted LR schedule.** `T_max` rides along in the scheduler state dict, so a rung-1 resume reverts the horizon to rung 0's. Measured on this branch: a promoted run's LR goes `0.002 → 0 → 0.001` against a configured `1e-3`. The bundled `example.yaml` is exactly this case. Anchors get a clean schedule, so the one comparison the search exists to make is biased.
2. **The local executor cannot run a second candidate.** `Trainer.__init__` calls `WandBLogger.init_instance()`; `WandBLogger` is a `Multiton`; the second call raises `ValueError: already initialized`. `configs/search/local.yaml` and the docs both advertise this path.
3. **A full-history JSON rewrite with `fsync` on every optimizer step**, in the shared training loop, with unbounded growth — O(n²) synchronous I/O on shared scratch.

The reason (1) and (2) aren't caught is structural, not accidental: the search suite runs in **0.67 s** because `train_task` is monkeypatched away in one test and `Trainer` is replaced by a `FakeTrainer` in the other. "371 passed" in the verification section doesn't speak to either. Any fix should come with a test that constructs two real `Trainer`s in one process and one that asserts the scheduler horizon survives a rung boundary.

Beyond those, inline: a disabled overwrite guard, a rung double-submit on controller retry, `status: "complete"` after a mass rung-0 failure, `allow_dirty` being unusable on the executor it matters for, a hardcoded 10-minute/2 GB controller job that has to upload multi-GB checkpoints, and the search's own non-finite-metric handling being dead code.

Two cross-cutting things that didn't fit on a line:

- **Public logs.** The publisher uploads every Slurm stdout/stderr, `experiment.log`, and up to 16 KB of raw stderr tail embedded in `results.csv` — into a *public* OSN bucket, from jobs launched with `--export=ALL` carrying `AWS_SECRET_ACCESS_KEY` and `WANDB_API_KEY`. The default should not be "mirror unfiltered process output to a public URL"; at minimum this needs a documented decision and a scrub pass.
- **PR description drift.** It cites `build_search()` as the algorithm boundary (the function is `SearchConfig.build()`) and says "W&B runs use the search name as their group" (the code uses `run_id`, which the docs correctly describe). Worth a pass before merge — reviewers calibrate on this text.

None of this is a design objection to successive halving here; the shape is right. It's that the science-critical path (resume semantics) and the demo path (local) are both unexercised, and the training loop took on a cost that will show up at quarter-degree scale.

src/samudra/search/successive_halving.py
```diff
@@ -0,0 +1,611 @@
...
+        output = (
+            self.search_dir / "probe" / resource_slug(name)
+            if probe
+            else self.output_dir(name, rung)
+        )
+        if output.exists() and "RANK" not in os.environ:
+            raise ValueError(f"Refusing to overwrite {output}")
+        args = [candidate.config, *candidate.args]
+        train_config = TrainConfig.from_yaml_and_cli(args)
+        train_config.epochs = self.rungs[rung]
```
alxmrs, line 286: 🤖 **Blocking: every promoted candidate trains under a corrupted LR schedule.**

`train_config.epochs` is the *rung* budget, and `Trainer.__init__` builds the scheduler from it (`train.py:229`, `cfg.scheduler.build(self.optimizer, cfg.epochs)`). Rung *n+1* then resumes the parent checkpoint, and `load_checkpoint` restores the scheduler state dict (`train.py:1169-1170`) — which for `CosineAnnealingLR` **includes `T_max`**. The freshly built horizon is overwritten by the previous rung's.

Measured on this branch (`rungs: [1, 3, ...]`, `lr=1e-3`, `scheduler: {type: cosine}`):

```
rung-0 state_dict:      {'T_max': 1, 'eta_min': 0.0, 'base_lrs': [0.001], ...}
lr after rung 0:         0.0
T_max after load in r1:  1            # not 3
rung-1 lr per epoch:     0.002, 0.0, 0.001
```

The promoted run trains at **2x the configured LR**, then 0, then base — the recursive cosine form goes unstable once `last_epoch > T_max`. Meanwhile `fixed` anchors are trained from scratch at the full budget with a clean schedule, so the anchor-vs-finalist comparison this search exists to produce is systematically biased against the finalists.

The bundled example is exactly this case: `configs/search/example.yaml` points every candidate at `samudra_multi_om4/train.yaml`, which sets `scheduler: {type: cosine}` and never overrides `target_epochs`.

The repo already knows about this — `tests/configs/train_samudra_om4_v2_resume.yaml` pins `target_epochs: 2` with the comment *"Keep the scheduler horizon fixed when the test simulates interruption."*

`docs/search.md:235` pushes it onto the user ("Use a scheduler whose meaning does not change when the target epoch increases"). A silent footgun that invalidates results isn't documentation-fixable, and it contradicts AGENTS.md's "Noisy Failure". The search knows the full budget: set `train_config.scheduler.target_epochs = self.rungs[-1]` here, or refuse to start when a scheduler is configured without an explicit `target_epochs`.

Related: `CosineWithWarmupConfig.build` asserts `warmup_epochs <= epochs` (default 5), so any rung budget below 5 dies at construction time.

src/samudra/search/executors/local.py
```diff
@@ -0,0 +1,31 @@
...
+            self.search.train_task(len(self.search.rungs) - 1, task, anchor=True)
+
+    def submit_rung(self, state: dict[str, Any], rung: int) -> None:
+        candidates = state["rungs"][rung]["candidates"]
+        state["rungs"][rung]["job_id"] = "local"
+        state["status"] = "running"
+        self.search.write_state(state)
+        if self.search.config.executor.dry_run:
+            return
+        for task, _ in enumerate(candidates):
```
alxmrs, line 33: 🤖 **Blocking: the local executor cannot run more than one candidate.**

`train_task` constructs `Trainer`, and `Trainer.__init__` calls `WandBLogger.init_instance()` unconditionally (`train.py:232`). `WandBLogger` is a `Multiton`; `init_instance` raises when the class is already present in the current scope, and `finish()` only calls `wandb.finish()` — it never evicts the instance:

```
$ python -c "init_instance(); configure(...); finish(); init_instance()"
second init FAILED: ValueError <class 'samudra.utils.wandb.WandBLogger'> already initialized
```

So candidate #2 of any local search dies in `Trainer.__init__`. This is the path `configs/search/local.yaml` ships and `docs/search.md:39` advertises for "a local laptop or Colab notebook run" — and `example.yaml` has three candidates.

The tests can't catch it by construction: `test_local_executor_runs_tasks_then_advances` monkeypatches `train_task` away, and `test_task_builds_train_config_and_calls_trainer_directly` substitutes a `FakeTrainer`. Nothing anywhere constructs two real `Trainer`s in one process.

A `MultitonScope()` per task is the obvious fix, but I'd want a hard look at what else leaks when N trainings share a process — CUDA allocator state across differently-sized architectures, the global `dask.config.set`, and logging handlers re-added by `handle_logging` on every task. Forking a subprocess per candidate may be the honest answer. Either way this needs a test that runs two candidates end to end.

src/samudra/train.py
```diff
@@ -611,13 +718,37 @@ def train_one_epoch(self, epoch):
...
+                    "loss": float(loss_value_reduce),
+                    "data_load_seconds": metric_logger.meters["data_load_time"].value,
+                    "data_wait_seconds": metric_logger.meters["data_wait_time"].value,
+                    "batch_seconds": batch_progress.batch_seconds,
+                }
+                if self.num_batches_seen == 1:
+                    write_search_worker_status(
+                        self.output_dir, "first_batch", **status_metrics
+                    )
+                if batch_progress.optimizer_stepped:
```
alxmrs: 🤖 **Blocking (performance): a full-file JSON rewrite with `fsync` on every optimizer step, over an unbounded append-only history.**

`write_search_worker_status` reads the existing file, appends to `history`, and re-serialises the *entire* history with `indent=2, sort_keys=True` followed by `os.fsync` (`training_summary.py:25-65`). Here it is called from inside the batch loop.

For the bundled example (`gradient_accumulation_steps=16`, 12-epoch finalist) that is thousands of events per run, and each write rewrites everything written so far — total bytes fsync'd grows as O(n^2). Hundreds of MB to GB of synchronous small writes on shared Lustre/GPFS scratch, per rank-0 worker, per candidate. The file then gets SHA-256'd and re-uploaded to the public bucket on every publish.

AGENTS.md explicitly calls out keeping the core train loop performant, and this lands in the hot path of every search-managed run — including the quarter-degree configs this repo is being scaled for.

Two options that keep the diagnostic value: hold the per-step counters in memory and flush on a wall-clock interval, or drop step events from `history` entirely and keep this file to lifecycle transitions only — which is exactly what `docs/search.md:83` describes it as. `search_metrics.parquet` and W&B already carry the per-step curve.

src/samudra/search/successive_halving.py
```diff
@@ -0,0 +1,611 @@
...
+            raise ValueError(f"Task {task} is outside this rung") from error
+        candidate = self.candidate(name)
+        if probe and anchor:
+            raise ValueError("Fixed anchors cannot be used as rung probes")
+        output = (
+            self.search_dir / "probe" / resource_slug(name)
+            if probe
+            else self.output_dir(name, rung)
+        )
+        if output.exists() and "RANK" not in os.environ:
```
alxmrs: 🤖 This guard is disabled on precisely the path that needs it.

Under the Slurm harness every worker runs via `torch.distributed.run`, which sets `RANK` for **all** ranks including rank 0. So `"RANK" not in os.environ` is false and `Refusing to overwrite` never fires for any distributed run — it only protects the local executor, where collisions are least likely.

If the intent was "don't let non-main ranks race on this check", the predicate you want is `is_main_process()`, already imported on line 35. As written, it silently opts out of the one safety check standing between a resubmitted array and a half-overwritten run directory (see the double-submit path in `advance`).

src/samudra/search/successive_halving.py
```diff
@@ -0,0 +1,611 @@
...
+        if current["advanced"]:
+            # Publication and scheduler submission are external operations. A
+            # controller retry must be able to finish either after a transient
+            # upload/submission failure without recomputing the ranking.
+            self._write_results(state)
+            self.publish(state)
+            next_rung = rung + 1
+            if next_rung == len(self.rungs):
+                return
+            if "controller_job_id" not in state["rungs"][next_rung]:
```
alxmrs, line 586: 🤖 This retry path can double-submit a rung's array.

`submit_validated_rung` submits the array, then the controller, then writes both IDs in a single state write. `test_slurm_does_not_persist_array_without_controller` pins the behaviour that a failed controller submission discards the array's job ID — but the array is already queued in Slurm and is never `scancel`ed. On retry, `"controller_job_id" not in ...` is true, so `submit_rung` fires a **second** array over the same candidates, writing into the same `output_dir(name, rung)`.

Two arrays training the same candidates into the same directories — with the overwrite guard above inert under torchrun — is silent corruption of `training_summary.json` and `saved_nets/`, and the promotion ranking then reads whichever finished last.

The test fixture literally names the leftover `"orphaned-array"`, so the hazard was seen. It just isn't handled. Either persist the array ID before submitting the controller (and reuse it on retry), or `scancel` it on the failure path.

src/samudra/search/successive_halving.py
```diff
@@ -0,0 +1,611 @@
...
+        promoted = ranked.head(keep)["candidate"].tolist()
+        current.update(results=results, promoted=promoted, advanced=True)
+        next_rung = rung + 1
+        if next_rung == len(self.rungs):
+            state["anchors"]["results"] = [
+                self._result(name, rung) for name in state["anchors"]["candidates"]
+            ]
+            final_results = [*results, *state["anchors"]["results"]]
+            state["status"] = (
+                "complete"
```
alxmrs, line 627: 🤖 `status: "complete"` only inspects the final rung and the anchors.

A search where 7 of 8 candidates crashed at rung 0 — OOM, a bad data mount, one broken candidate config — terminates as `complete`, and `report.md` labels the sole survivor **"Winner"** (`report.py:72-73`) rather than "Best completed finalist". `results.parquet` carries the failed rows, but `status` is the field an agent or a `duckdb` one-liner will key off, and it says the search ran as designed.

Propagate intermediate-rung ineligibility into the terminal status, or add an explicit `eliminated_by_failure` count that the report has to render, so a mass failure cannot read as a clean result.

src/samudra/search/successive_halving.py
```diff
@@ -0,0 +1,611 @@
...
+    @staticmethod
+    def result_rows(state: dict[str, Any]) -> list[dict[str, Any]]:
+        return [
+            result for rung in state["rungs"] for result in rung.get("results", [])
+        ] + state["anchors"].get("results", [])
+
+    def advance(self, rung: int) -> None:
+        state = self.read_state()
+        if isinstance(self.config.executor, SlurmExecutorConfig):
+            controller = _git_provenance(allow_dirty=False)
```
alxmrs: 🤖 Two problems with running `_git_provenance` in the promotion controller.

**1. `allow_dirty` is unusable on Slurm.** It is hardcoded `False` here while `start()` honours `self.config.allow_dirty`. So `allow_dirty: true` + the Slurm executor is *guaranteed* to die at the first promotion with `"Commit tracked changes or set allow_dirty=true"` — after rung 0 has already burned its allocation. The escape hatch fails on the only executor where it costs real GPU-hours to discover that.

**2. The controller usually isn't in a git checkout.** It runs `self.config.python -m samudra.search.worker advance` outside the container (`configs/search/torch.yaml` sets `python: /scratch/USER/samudra-venv/bin/python`), so `_git_provenance` shells `git -C <site-packages>/samudra rev-parse --show-toplevel`. If that venv has samudra installed non-editable — the normal way to build a small controller venv — `check=True` raises `CalledProcessError` and *every* `advance` fails. Nothing in `docs/search.md` states the controller venv must be an editable checkout of the same commit.

Both point the same way: the controller should assert against a recorded identity (`SAMUDRA_CODE_COMMIT`, or the snapshot in `state.json`) rather than re-deriving it from a working tree it has no reason to be standing in.

src/samudra/utils/training_summary.py
```diff
@@ -0,0 +1,121 @@
...
+    payload = {"schema_version": TRAINING_SUMMARY_SCHEMA_VERSION, **summary}
+    with tempfile.NamedTemporaryFile(
+        mode="w",
+        dir=output_dir,
+        prefix=f".{TRAINING_SUMMARY_NAME}.",
+        delete=False,
+        encoding="utf-8",
+    ) as stream:
+        temporary_path = Path(stream.name)
+        json.dump(payload, stream, indent=2, sort_keys=True, allow_nan=False)
```
alxmrs: 🤖 `allow_nan=False` here makes the search's own non-finite handling dead code.

`_result` checks `math.isfinite(score)` and marks the candidate ineligible with `metric ... is not finite` (`successive_halving.py:432-433`). That branch is unreachable: a NaN `validation_loss` makes this `json.dump` raise (`ValueError: Out of range float values are not JSON compliant: nan`) inside the epoch loop, so `training_summary.json` is never written. The candidate then surfaces as *"training_summary.json not found"*, pointing the reader at the filesystem instead of at divergence.

This is the common case for what the tool is *for*: `configs/search/example.yaml` ships a `larger-learning-rate` candidate. Divergence is a result, not an I/O error — serialise it (`allow_nan=True`, or map non-finite to `null` with an explicit `diverged` flag) and let `_result` classify it.

src/samudra/search/successive_halving.py
```diff
@@ -0,0 +1,611 @@
...
+        }
+        self.write_state(state)
+        # Publish stable, queryable schemas immediately. Without these files,
+        # the public results URL returns 404 until the first rung completes.
+        self._write_results(state)
+        self.publish(state)
+        if self.publisher is not None and is_main_process():
+            print(f"Published search record: {self.publisher.root}", flush=True)
+        try:
+            self.executor.submit_anchors(state)
```
alxmrs: 🤖 The `rung0_probe` fail-fast guarantee doesn't cover the most expensive jobs.

Anchors are submitted here, unconditionally, *before* `submit_rung` sets up the probe gate — and `submit_anchors` runs them at `len(rungs) - 1`, i.e. the full epoch budget with the longest `time_by_rung` walltime.

So a probe that proves the container, data mount, or code is broken still leaves every `fixed` baseline queued at full budget. `docs/search.md:78` says a failed probe "consumes no candidate array allocation", which is true only for the competing candidates and misleading for the ones that cost the most.

Gate the anchors behind the same probe, or make the doc precise about what is and isn't protected.

src/samudra/search/executors/slurm.py
```diff
@@ -0,0 +1,226 @@
...
+                "sbatch",
+                "--parsable",
+                f"--job-name={self.search.run_id}-advance-r{rung}",
+                f"--chdir={working_directory}",
+                f"--dependency={dependency}",
+                f"--account={self.config.account}",
+                f"--partition={self.config.controller_partition}",
+                "--cpus-per-task=1",
+                "--mem=2G",
+                "--time=00:10:00",
```
alxmrs: 🤖 The advance controller is hardcoded to `--cpus-per-task=1 --mem=2G --time=00:10:00` — but this is the job that runs `ArtifactPublisher.publish`, which at the final rung with `checkpoints: final` (the default) or `all` SHA-256s and uploads every retained checkpoint to OSN.

`publish` re-hashes and re-`put_file`s **every** file on **every** call (`artifacts.py:112`) with no size/mtime/etag skip, and `_files` is computed twice per publish (`artifacts.py:105-108`). Several multi-GB checkpoints over WAN, single-threaded, inside a 10-minute walltime is optimistic; when it times out the terminal state is never published and the public record silently ends at the previous rung — with `results.parquet` still showing `status: running`.

Minimum: make walltime/mem/cpus configurable alongside `controller_partition`. Better: skip objects whose size and hash already match the destination, so repeated publishes are cheap and the retry path in `advance` is genuinely idempotent rather than merely re-runnable.

src/samudra/search/successive_halving.py
```diff
@@ -0,0 +1,611 @@
...
+                yaml.safe_dump(train_config.model_dump(mode="json"), sort_keys=False),
+                encoding="utf-8",
+            )
+            candidate.config = str(destination)
+            candidate.args = []
+        self.config_path.write_text(
+            yaml.safe_dump(self.config.model_dump(mode="json"), sort_keys=False),
+            encoding="utf-8",
+        )
+        state = {
```
alxmrs, line 172: 🤖 This is the one thing in the PR that outlives a process, crosses machines, and gets resumed days later — and it's the one thing left as an untyped `dict[str, Any]` addressed by string keys (`state["rungs"][rung]["candidates"]`, `.get("probe")`, `state["anchors"]["results"]`), in a repo where every config is a validated Pydantic model.

The cost is already visible in the tests: roughly ten of them hand-roll a different partial `state` literal, several omitting `index` / `epochs` / `promoted` / `advanced` that production code requires. They pass against states `start()` would never produce, so they're validating the reader against a fiction.

A `SearchState` model, with `read_state`/`write_state` doing `model_validate`/`model_dump`, would give the tests one constructor to build from and would catch schema drift at load time instead of as a `KeyError` in a controller job six hours into a search. It also documents the resume contract, which right now only exists as the shape of this literal.

src/samudra/search/executors/base.py
```diff
@@ -0,0 +1,41 @@
...
+"""Compute executor interface used by search algorithms."""
+
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import Any, Protocol
+
+from samudra.search.config import SearchConfig
+
+
+class SearchController(Protocol):
```
alxmrs: 🤖 Three structural protocols now describe one concrete class: `SearchController` here, plus `ArtifactSearch` and `ArtifactSearchConfig` in `artifacts.py`. There is exactly one algorithm — `SearchConfig.build()` is annotated `-> "SuccessiveHalving"` with an unreachable `raise AssertionError` behind a `Literal`-typed discriminator, and `AlgorithmConfig = SuccessiveHalvingConfig` aliases a single class.

The abstraction isn't paying rent, it's charging it. Undoing it takes four `assert isinstance` calls: `SlurmExecutor.config` (slurm.py:19) to recover the config type it was handed, `release_probe` (successive_halving.py:400) because `submit_validated_rung` isn't on the `Executor` ABC, and two on `Location` in `artifacts.py` (92, 255). Those asserts vanish under `python -O`, so the type narrowing they stand in for is unenforced in exactly the deployment where a mistake is expensive.

The `EXECUTORS` dict is a real seam with two real implementations — keep it. I'd drop the algorithm and artifact protocols until a second implementation exists and let the concrete types flow; `submit_validated_rung` belonging to the base interface (or the probe living behind `submit_rung`) would remove one assert on its own.

src/samudra/search/artifacts.py
```diff
@@ -0,0 +1,264 @@
...
+
+import pandas as pd
+import s3fs
+
+from samudra.search.config import ArtifactConfig
+from samudra.utils.location import LocalLocation, S3Location, UnresolvedLocation
+
+CATALOG_NAME = "artifacts.parquet"
+EPOCHS_NAME = "epochs.parquet"
+RUN_FILES = (
```
alxmrs, line 30: 🤖 `training_summary.json`, `search_worker_status.json`, and `search_metrics.parquet` are now string literals in three places: the canonical constants in `utils/training_summary.py` (`TRAINING_SUMMARY_NAME`, `SEARCH_WORKER_STATUS_NAME`, `SEARCH_METRICS_NAME`), re-declared as `SUMMARY_NAME`/`WORKER_STATUS_NAME` in `successive_halving.py:39-40`, and hardcoded again here and at line 177. Rename one and publication silently stops picking the file up — no test would fail, because the tests write the literals too.

Same pattern for atomic writes. `artifacts.atomic_local_parquet`, `training_summary.write_search_metrics` (111-120), `_atomic_json` (successive_halving.py:54-64), the two `NamedTemporaryFile` blocks in `training_summary.py`, and the CSV block in `_write_results` (527-538) are five hand-rolled copies of mkstemp → write → flush → fsync → `os.replace`, and they already differ (three use `NamedTemporaryFile`, two use `mkstemp`; only two clean up the temp file on failure). One `atomic_write(path, writer)` helper collapses all five and makes the failure behaviour uniform.

src/samudra/search/successive_halving.py
```diff
@@ -0,0 +1,611 @@
...
+        if status_path.is_file():
+            worker_status = json.loads(status_path.read_text(encoding="utf-8"))
+            result.update(
+                worker_stage=worker_status.get("stage"),
+                worker_updated_at=worker_status.get("updated_at"),
+                worker_optimizer_steps=worker_status.get("optimizer_steps"),
+                worker_batches_seen=worker_status.get("batches_seen"),
+                worker_error_type=worker_status.get("error_type"),
+                worker_error=worker_status.get("error"),
+                worker_status_log=str(status_path.relative_to(self.search_dir.parent)),
```
alxmrs: 🤖 Path bases are inconsistent between adjacent columns of the same table.

`worker_status_log` is relative to `search_dir.parent`, yielding `<run-dir>/search_worker_status.json`, while `scheduler_stdout_log` / `scheduler_stderr_log` two methods down are relative to `search_dir`, yielding `logs/r0-123_1.out`. The published catalog keys the same worker-status file as `runs/<run-dir>/search_worker_status.json`.

So joining `results.parquet.worker_status_log` against `artifacts.parquet.artifact` — which is the stated point of publishing both — misses, while the scheduler-log columns join fine. Pick one base (search-dir-relative, matching the catalog) and derive all four the same way.

src/samudra/cli.py
```diff
@@ -24,16 +24,16 @@
...
 
 Commands:
   train   Train a model from a config (checkpointing, W&B logging, multi-GPU).
   eval    Roll a trained model out autoregressively and collect metrics.
   viz     Render maps, time series, and PDFs from evaluation outputs.
 
-CONFIG is a path to a YAML file or the name of a bundled preset such as
-`samudra_om4/train.yaml`. Any config key can be overridden inline (e.g.
-`--epochs 100`); run `samudra <command> --help` for a command's full options.
+For train, eval, and viz, CONFIG is a YAML path or bundled preset such as
```
alxmrs: 🤖 This hunk is leftover churn from a reverted decision. fd4d592c added `search` to `_COMMANDS` with a *"Plan and run successive-halving architecture searches"* line; b0328d1e took it back out. What survives in the diff is a reworded help paragraph and a dropped `else:` — leaving `main()` with three `elif`s and no fallback branch.

Meanwhile the commit is still titled "Expose architecture search through Samudra CLI" and the PR body opens with the user interface section, but `samudra search` does not exist — `python -m samudra.search` is the only entry point, which a `pip install samudra` user has no way to discover from `samudra --help`.

Either finish it (`search` in `_COMMANDS` plus a dispatch branch and a help line) or revert `cli.py` entirely, so the diff stops implying a feature that isn't there.

src/samudra/search/config.py
```diff
@@ -0,0 +1,151 @@
...
+    config: str
+    args: list[str] = Field(default_factory=list)
+    fixed: bool = False
+
+
+class SuccessiveHalvingConfig(BaseConfig):
+    type: Literal["successive_halving"] = "successive_halving"
+    rungs: list[int] = Field(min_length=1)
+    promotion_fraction: float = Field(default=0.5, gt=0, le=1)
+    minimum_promoted: int = Field(default=1, ge=1)
```
adam-subel, line 44: This minimum promoted may want to be validated against the number of runs in each step. If it is too large, some of the halving may be skipped with all seeds training until the end.

src/samudra/search/artifacts.py
```diff
@@ -0,0 +1,272 @@
...
+                    "sha256": _sha256(source),
+                    "uri": self._uri(relative),
+                    "public_url": self._public_url(relative),
+                }
+            )
+        return rows
+
+    @staticmethod
+    def _kind(relative: str) -> str:
+        if relative.endswith("ckpt.pt"):
```
adam-subel: Did you intend for this to capture any checkpoints of the form ckpt_{epochs}.pt?

src/samudra/search/artifacts.py
```diff
@@ -0,0 +1,272 @@
...
+            return "checkpoint"
+        if Path(relative).suffix == ".md":
+            return "report"
+        if Path(relative).suffix in {".pdf", ".png", ".svg"}:
+            return "figure"
```
adam-subel, line 227: Figure is missing from _media_type, not sure if that is intentional

src/samudra/search/successive_halving.py
```diff
@@ -0,0 +1,621 @@
...
+                    "promoted": [],
+                    "advanced": False,
+                }
+                for index, epochs in enumerate(self.rungs)
+            ],
+        }
+        self.write_state(state)
+        # Publish stable, queryable schemas immediately. Without these files,
+        # the public results URL returns 404 until the first rung completes.
+        self._write_results(state)
```
adam-subel, line 194: Noted by claude:
_write_results:519-544 writes results.parquet, results.csv, and then calls write_search_report(self, state). advance calls it between write_state and publish:
So any exception from report rendering — a formatting bug on an unusual value, a None where a number was expected — skips publication and skips the next rung's submission. A cosmetic failure becomes a silently stalled search.

src/samudra/utils/training_summary.py
```diff
@@ -0,0 +1,128 @@
...
+SEARCH_METRICS_NAME = "search_metrics.parquet"
+SEARCH_WORKER_STATUS_NAME = "search_worker_status.json"
+SEARCH_WORKER_STATUS_SCHEMA_VERSION = 1
+_SECRET_NAME = re.compile(r"(?:SECRET|TOKEN|PASSWORD|CREDENTIAL|API_KEY|ACCESS_KEY)")
+_SECRET_ASSIGNMENT = re.compile(
+    r"(?i)((?:secret|token|password|credential|api[_-]?key|access[_-]?key)\s*[=:]\s*)([^\s,;]+)"
+)
+
+
+def _redact_secrets(value: Any) -> Any:
```
adam-subel, line 35: Are there any potential situations where secrets might be nested and not in plain text, or potential conflicts with other text for short keys?

adam-subel (commented):
I think overall things seem to make sense, I went through it with the help of claude as it is quite dense and left some notes. Happy to talk this through on a call, it might be helpful for me to better give feedback on the larger structuring

src/samudra/search/artifacts.py
alxmrs: I think the contents of figure (.pdf, .png, .svg) are indeed inside the mime type mapping.

src/samudra/search/config.py
alxmrs: Pushing back on this given our discussion.

src/samudra/search/successive_halving.py
alxmrs: I think this is addressed in the new lines on 217 below.

src/samudra/utils/training_summary.py
alxmrs: I inspected where we call `write_search_worker_status`, which is where redact is most useful to apply. I do no find anything that is sensitive now or that could be secrets not in plain text. We do parse a few things from env vars, but they are specific known SLURM related env vars. I think this redaction is a good enough effort for now.

alxmrs:
@fomo-bot, do you think this PR is ready for merge, given a new review?

alxmrs:
@fomo-bot if you think this is ready for merge, will you approve the PR?

PR #842, opened 2026-08-14
Add native SDPA Perceiver implementation

alxmrs:
## Summary

- replace `perceiver-pytorch`, `flash-perceiver`, and external `flash-attn` with native `Perceiver` and `PerceiverIO` components backed by `torch.nn.functional.scaled_dot_product_attention`
- preserve existing config names: `auto` and `sdpa` use PyTorch dispatch, `naive` forces the math backend, and `flash` forces PyTorch FlashAttention
- remove the retired packages from project metadata, lockfile, container setup/checks, quickstart, and installation/release documentation
- exercise SDPA in the container smoke test and add a CUDA test that forces FlashAttention through forward and backward passes

## Backward compatibility

The native modules retain the constructor shape and parameter hierarchy of `perceiver-pytorch`; its state dictionaries were verified to load directly and reproduce math-backend outputs within `1e-6`. Existing model config values remain accepted without edits.

The separate `flash-perceiver` package used a different internal module layout, so checkpoints produced specifically by that backend may require migration even though their configs remain valid.

This PR does not include any research decoder/model variants or experiment/search configuration.

## Container impact

No new CUDA binary is required. SDPA and its optimized kernels ship with PyTorch. The PhysicsNeMo image continues to supply the compatible PyTorch/CUDA stack, while its package verification and smoke test no longer require the removed external attention wheels.

## Validation

- focused Perceiver/encoder/decoder tests with all legacy packages physically absent — 25 passed, 1 CUDA test skipped locally
- `uv run --locked pytest -m "not manual and not cuda" -q` — 394 passed, 2 skipped, 10 xfailed, 67 deselected
- `uvx pre-commit run --all-files` — passed


src/samudra/config.py
```diff
@@ -57,6 +56,8 @@
 )
 from samudra.models.modules.blocks import ZonallyPeriodicBilinearUpsample
 from samudra.models.modules.encoder import patch_from
+from samudra.models.modules.perceiver import Perceiver as SDPAPerceiver
+from samudra.models.modules.perceiver import PerceiverIO as SDPAPerceiverIO
```
alxmrs: Do we need to import these as SDPA anymore?

pyproject.toml
```diff
@@ -177,52 +174,18 @@ overrides = [
...
-]
 torch = [
   { index = "pytorch-cu130", marker = "sys_platform == 'linux' and platform_machine == 'aarch64'" },
 ]
 torchvision = [
   { index = "pytorch-cu130", marker = "sys_platform == 'linux' and platform_machine == 'aarch64'" },
 ]
 
-[[tool.uv.index]]
-name = "flash-attn-cu13"
```
alxmrs, line 215: Ok, but how do we ensure that the flash attention binaries are actually available for the torch function to be able to use? Do we install that in the image? What's the dependency management story here.

scripts/build_quickstart_notebook.py
alxmrs: Can we address this? Do we still need two commands?

notebooks/quickstart.ipynb
```diff
@@ -107,13 +103,11 @@
     "!pip install --quiet --upgrade --no-warn-conflicts --progress-bar off \\\n",
```
alxmrs: Can this just be `uv sync` maybe with an extra? Maybe the two commands could be: 
```
!pip install uv
!uv sync --extra cuda # this needs refinement
```

alxmrs:

alxmrs:
@fomo-bot may I have your review?

alxmrs:
@fomo-bot will you fix this P3 documentation issue?

alxmrs:
@fomo-bot will you give this PR a final round of review? if you think it should be approved, then please approve it.

src/samudra/models/modules/perceiver.py
```diff
@@ -0,0 +1,432 @@
...
+            layer = cast(nn.ModuleList, untyped_layer)
+            self_attn, self_ff = layer[0], layer[1]
+            x = x + self_attn(x)
+            x = x + self_ff(x)
+
+        if queries is None:
+            return x
+        if queries.ndim == 2:
+            queries = queries.unsqueeze(0).expand(batch, -1, -1)
+        decoded = self.decoder_cross_attn(queries, context=x)
```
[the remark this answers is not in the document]
alxmrs: We double checked and this review note is incorrect. We are a parity with LucidRains PerceiverIO impl.

src/samudra/models/modules/perceiver.py
```diff
@@ -0,0 +1,432 @@
...
+    def _split_heads(self, tensor: torch.Tensor) -> torch.Tensor:
+        batch, tokens, _ = tensor.shape
+        return tensor.view(batch, tokens, self.heads, self.dim_head).transpose(1, 2)
+
+    def _attention(
+        self,
+        query: torch.Tensor,
+        key: torch.Tensor,
+        value: torch.Tensor,
+        mask: torch.Tensor | None,
```
adam-subel, line 84: A couple notes on the mask  (found by claude, but I think the notes are valid though some checks up stream may make it irrelevant):

Since scaled_dot_product_attention accepts both boolean and additive float masks, where float masks can be defined as (-inf where false and 0 where true) the .bool would invert masks passed in this float form. This may not actually come up

There was also this comment:
With a boolean attn_mask, a row that's all-False softmaxes over all -inf and yields NaN, which then propagates through the residual stream and poisons every parameter on the backward pass. If any sample can have an empty context (variable-length point clouds, ragged sequences), this is a training run that dies 40 minutes in for no visible reason. Worth either guarding or documenting as a caller precondition.

src/samudra/config.py
```diff
@@ -519,7 +520,7 @@ def create_block(
         return create_block
 
 
-PerceiverImpl = Literal["auto", "naive", "flash"]
+PerceiverImpl = Literal["auto", "sdpa", "naive", "flash"]
```
adam-subel: Why do you add "sdpa" here if it is just rewritten to "auto" in _attention_backend and cannot be passed directly as an attention_backend? Not sure if I am missing a use case

tests/test_perceiver.py
```diff
@@ -0,0 +1,177 @@
...
+from typing import cast
+
+import pytest
+import torch
+
+from samudra.config import PerceiverConfig
+from samudra.models.modules import Attention, Perceiver, PerceiverIO
+
+
+def test_attention_matches_explicit_reference_with_mask():
```
adam-subel, line 15: As the earlier comment, this may miss a few ways masking may silently cause issues down the line.

adam-subel (commented):
Only a few comments, some with the help of claude. They may be misplaced from my lack of understanding, but worth thinking about if it hasn't come up already

adam-subel (commented):
Looks good to me, this addressed the outstanding comments and the tests seem to cover all of the behavior in question

PR #848, opened 2026-08-20
Prepare full one-degree spatial-grid Perceiver experiment

alxmrs:
## Summary

- add coordinate-tied spatial-grid Perceiver encoding
- add direct output cross-attention, overlap assembly, and smooth processor conditioning
- add a 70-epoch one-degree train preset plus matching eval and viz presets
- add a lab notebook with preflight gates; no training job has been submitted

This is stacked on #842 so its review diff excludes the native-SDPA implementation. After #842 lands, this branch can be rebased onto main.

## Intended run

The preset assumes two GPUs, per-rank batch 2, and eight accumulation steps for effective global batch 32. It uses the shared OM4 data preset pointed at the one-degree data root. A real-data one-degree optimizer probe is required before releasing the 70-epoch allocation.

## Validation

- 402 passed, 2 skipped, 67 deselected, 10 xfailed
- all pre-commit hooks pass from a cold cache
- model construction: 75,194,225 trainable parameters; spatial encoder output patch extent 3 x 5 degrees


src/samudra/configs/perceiver_spatial_grid_1deg/data.yaml
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: CC0-1.0
+
+# yaml-language-server: $schema=../schemas/DataConfig.json
+
+concurrent_compute: true
```
alxmrs: Isn't there a standard data location that we can use? See the configs/data folder.

src/samudra/configs/perceiver_spatial_grid_1deg/model.yaml
```diff
@@ -0,0 +1,63 @@
...
+    depth: 2
+    latent_dim: 64
+    num_latents: 64
+    cross_heads: 2
+    latent_heads: 8
+    cross_dim_head: 64
+    latent_dim_head: 64
+
+processor:
+  ch_width: [380, 480]
```
alxmrs: Why not a third, 520 channel?

src/samudra/configs/perceiver_spatial_grid_1deg/train.yaml
```diff
@@ -0,0 +1,40 @@
...
+pin_mem: true
+save_freq: 5
+validation_image_log_freq: 5
+epochs: 70
+batch_size: 2
+# On the intended two-GPU Torch launch this gives effective global batch 32.
+gradient_accumulation_steps: 8
+learning_rate: 0.0006
+scheduler: {type: cosine}
+loss: mse
```
alxmrs, line 21: Samudra 2 typically uses a dynamic mse with a limit of 20 -- we can experiment with loss configs later. I agree, this loss is good to start with.

src/samudra/models/modules/decoder.py
```diff
@@ -14,7 +14,101 @@
...
 from samudra.models.modules.encoder import patch_from
+from samudra.models.modules.perceiver import (
+    Attention,
+    AttentionBackend,
+    FeedForward,
+    PreNorm,
+)
+
+
+def zonally_periodic_bilinear_interpolate(
```
alxmrs: Is there any way we could re-use the periodic bilinear interpolation block that exists in blocks.py?

src/samudra/models/modules/decoder.py
```diff
@@ -14,7 +14,101 @@
...
+    resized = F.interpolate(
+        padded,
+        size=(target_height, target_width + 2 * scale_width),
+        mode="bilinear",
+        align_corners=False,
+    )
+    return resized[..., scale_width : scale_width + target_width]
+
+
+class DirectCrossAttentionIO(nn.Module):
```
alxmrs, line 30: Should we run a search to test how many layers worth of cross and self attentions we should try? 

This is surprisingly simple, but maybe that is the point.

src/samudra/models/modules/decoder.py
```diff
@@ -14,7 +14,101 @@
...
+    """Apply Perceiver output queries directly to spatial processor tokens."""
```
alxmrs: Is this a perceiver in the first place, or just a transformer architecture?

alxmrs (commented):
Overall, this looks really solid.

src/samudra/configs/perceiver_spatial_grid_1deg/train.yaml
alxmrs: Agreed. I retained plain normalized MSE so this run stays comparable to the architecture searches and isolates the transport intervention. Dynamic MSE is a worthwhile later loss ablation after we have this architectural baseline.

src/samudra/models/modules/decoder.py
alxmrs: Yes. I generalized `ZonallyPeriodicBilinearUpsample` from fixed 2x scaling to arbitrary positive integer latitude/longitude scales and removed the decoder-local interpolation implementation. Processor conditioning now reuses that block with its inferred 3x5 scale at one degree. I added an anisotropic 3x5 shape and longitude-roll-equivariance test.

src/samudra/configs/perceiver_spatial_grid_1deg/model.yaml
alxmrs: The two-level processor came from the 2-degree search configuration, where reducing candidate cost was useful. That is the wrong default for the strongest full run. I restored the third 520-channel stage, so the processor is now `[380, 480, 520]`. The complete model grows from 44.4M to 75.2M parameters; the real-data preflight will confirm memory and step time before the 70-epoch allocation.

src/samudra/models/modules/decoder.py
alxmrs: The direct decoder block by itself is a transformer cross-attention block, not a standalone Perceiver. More precisely, it is the output-query decode head from Perceiver IO applied directly to processor tokens. The overall encoder/processor/decoder model remains Perceiver-based because the encoder is a full local Perceiver IO. I expanded the class docstring to make this distinction explicit and avoid overstating the decoder.

src/samudra/configs/perceiver_spatial_grid_1deg/data.yaml
alxmrs: Agreed. I removed the experiment-local data file and now include the shared `../data/om4.yaml` preset from train, eval, and viz. The launch supplies `experiment.data_root` pointing at the one-degree directory containing the three OM4 stores. This also adopts the standard Samudra 2 `tau_hfds_hfds_anom` boundary contract instead of silently maintaining a second data definition.

src/samudra/models/modules/decoder.py
alxmrs: The simplicity is intentional for the decoder. The encoder is already a complete local Perceiver IO with input cross-attention, two latent self-attention stages, and output-query cross-attention. The decoder retains only one query-to-processor cross-attention plus FFN because the processor already mixes spatial state and Jesse’s localization rejected a second decoder latent/self-attention stack. I documented a focused follow-up over encoder depths {1,2,4} and direct cross-attention depths {1,2} after this baseline. I do not recommend delaying the full run for that search, and I would keep output-query self-attention out initially because it is quadratic within each output window and lacks supporting evidence.

PR #860, opened 2026-08-25
Support independent input and output steps

alxmrs:
`hist` has been used to cover two concepts: the number of input steps and the number of output steps. These have always been decided together. This PR separates these two concepts so we can change them independently. Doing so, to me, clarifies what concept we actually need in various parts of how `hist` was used -- code paths either need input or output steps, and this PR makes use of the proper concept by name now instead of implicitly. I find that this makes reading various data flow paths more understandable than before.

🤖 below

## Summary

- add backwards-compatible `data.output_steps` configuration, defaulting to `hist + 1`
- support rolling input histories when a model emits fewer timesteps than it consumes
- anchor residual predictions to the most recent matching input span
- separate input/output timestep handling in validation, rollout metrics, and Zarr writing
- cover the 2-input / 1-output dataset, residual rollout, Trainer, validation, and writer paths

This enables `hist: 1, output_steps: 1`: two input states predict one future state, and each autoregressive call shifts `[x(t-1), x(t)]` to `[x(t), x_hat(t+1)]`.

Existing configs omit `output_steps` and retain their current `hist + 1` output blocks.

## Validation

- `.venv/bin/pytest -m 'not manual and not cuda' -q`
  - 417 passed, 2 skipped, 67 deselected, 10 xfailed


src/samudra/aggregator/inference/main.py
```diff
@@ -106,7 +109,9 @@ def __init__(
         self._normalize = normalize
         self._tensor_map = tensor_map
         self.num_prognostic_channels = num_prognostic_channels
-        self.hist = hist
+        self.input_hist = hist
```
alxmrs: Maybe we can make this `input_steps` that is `hist+1` to be consistent with the accounting of `output_steps`, rather than use `output_hist` which is a weird concept.

src/samudra/config.py
```diff
@@ -420,11 +420,32 @@ class DataConfig(BaseConfig):
     )
     loading: DataLoadingConfig = Field(default_factory=CpuDataLoadingConfig)
     hist: int = 1
```
alxmrs: High level feedback: is it possible to make a parallel configuration to `hist` (that is mutually exclusive with it) that allows people to specify `input_steps`? This would make it clear that we can specify `input_steps=2` and `output_steps=1`, but in a backwards compatible way with `hist=1`. Then, we could refactor how `hist` is set and use `input_steps` and `output_steps` everywhere internally going forward.

alxmrs (commented):
Initial review to agents.

src/samudra/config.py
alxmrs: In this PR, different parts of the codebase use different conventions to refer to input and output steps (sometimes `input_steps` and `output_steps`, or `input_hist` and `output_hist`, or `hist`, or a combination). After the config / CLI changes to allow `input_steps` and `output_steps`, I'd like to make all the patterns for access standardize on `input_steps` and `output_steps`. I notice different patterns in the Aggregators, datasets, and train modules, for example.

alxmrs: In this PR, different parts of the codebase use different conventions to refer to input and output steps (sometimes `input_steps` and `output_steps`, or `input_hist` and `output_hist`, or `hist`, or a combination). After the config / CLI changes to allow `input_steps` and `output_steps`, I'd like to make all the patterns for access standardize on `input_steps` and `output_steps`. I notice different patterns in the Aggregators, datasets, and train modules, for example.

src/samudra/models/base.py
```diff
@@ -43,9 +44,56 @@ def __init__(
...
+                "Output channels must contain complete prognostic timesteps; "
+                f"got {out_channels} channels for "
+                f"{self.prognostic_channels_per_step} channels per timestep"
+            )
+        self.output_steps = out_channels // self.prognostic_channels_per_step
+        if not 1 <= self.output_steps <= self.input_steps:
+            raise ValueError(
+                f"Model output_steps must be between 1 and {self.input_steps}, "
+                f"got {self.output_steps}"
+            )
```
alxmrs: I think this code block is mostly un-needed. No class variables that are added `prognostic_channels_per_step` -- `output_steps` are only used in tests -- are currently used anywhere besides the init. However, this has a high cost because it adds a new constructor argument to an otherwise stable base class API. Further, the prognostic vs boundary information are passed where they're needed in the `forward` method.

src/samudra/config.py
```diff
@@ -419,12 +419,61 @@ class DataConfig(BaseConfig):
...
+        default=None,
+        ge=0,
+        description=(
+            "Legacy number of additional input-history timesteps. When neither "
+            "hist nor input_steps is set, hist defaults semantically to 1. "
+            "Cannot be set together with input_steps."
+        ),
+    )
+    input_steps: int | None = Field(
+        default=None,
```
alxmrs: This should be defaulted to 2 IMO, since his was defaulted to 1 before.

src/samudra/config.py
```diff
@@ -419,12 +419,55 @@ class DataConfig(BaseConfig):
...
+    input_steps: int = Field(
+        default=2,
+        ge=1,
+        description=(
+            "Number of raw timesteps consumed by each model call. Cannot be set "
+            "together with the legacy hist option."
+        ),
+    )
+    output_steps: int | None = Field(
+        default=None,
```
alxmrs: And, this should be 2 by default, right?

alxmrs:
@fomo-bot may I have your review? If you think the PR is ready for merge, will you approve it?

alxmrs:
Hey @codex, will you review this PR?

alxmrs:
@codex may I have your review on this PR?

src/samudra/datasets.py
amogh-gulati: 🐑 : This is still old explanation right? where the inputs and outputs are coupled

amogh-gulati (approved):
Looks good! just one small comment which came up

src/samudra/datasets.py
alxmrs: Good catch—the docstring was still describing only the coupled legacy behavior. I updated it to define windows in terms of independent `input_steps` and `output_steps`, including explicit 2-in/2-out and 2-in/1-out examples (2f49ac91).

jder:
@alxmrs any reason we can't remove the old `hist` setting?

alxmrs:
Great question, and yes, I think so long as we update all the configs, we could totally remove `hist`. I didn't want to break backwards compatibility just in case.

PR #865, opened 2026-09-02
Parallelize search across local and allocated GPUs

alxmrs:
## Summary

- keep `local` explicitly single-machine and discover worker slots only through CUDA visibility
- add a distinct `slurm_allocation` executor for independent one-GPU trials inside an existing single- or multi-node Slurm allocation
- allocate CPU and memory TRES proportionally to each exclusive Slurm step so workers can actually run concurrently
- accept Slurm's optional GPU topology suffix in allocation environment values such as `8(S:0-1)`
- share candidate pooling, worker isolation, concurrency limiting, anchor/rung-zero co-scheduling, and fail-fast queue handling between both executors
- retain the existing `slurm` executor for separately queued arrays and dependent controller jobs
- prevent independent search workers from accidentally initializing DDP
- keep W&B search tags within its 64-character limit while retaining the complete run ID in the group and structured search metadata
- document the Empire AI Alpha workflow, including when `local` is sufficient inside a single-node allocation

## Executor roles

- `local`: one isolated process per locally visible GPU; sequential fallback on CPU
- `slurm_allocation`: concurrent exclusive one-GPU `srun` steps in the current allocation
- `slurm`: independently submitted arrays with automatic dependent controllers

`DistributedConfig` remains dedicated to one model distributed across ranks; independent search trials do not create an NCCL process group.

## Resource invariant

Each candidate remains a single-GPU experiment in every rung. Survivors are not automatically changed to 2/4/8-way DDP as the population shrinks, even if that leaves GPUs idle in a retained allocation. Changing world size midway through a resumed trajectory changes global batch size and optimizer steps per epoch. Adaptive multi-GPU promotion therefore needs a separate explicit policy for batch size, learning rate, scheduler behavior, and epoch accounting.

Fixed anchors and rung zero are co-scheduled to maximize utilization during the initial stage. `max_concurrent` can cap the number of simultaneous independent trials. Once a worker fails, queued trials are not launched; already-running workers are allowed to exit before the failure is reported.

## Validation

- `uv run python -m pytest -m "not manual and not cuda" -q` — 469 passed, 2 skipped, 68 deselected, 10 xfailed
- `uv run python -m pytest tests/test_search.py tests/test_backend.py -q` — 40 passed
- `uvx pre-commit run --all-files` — passed

## Deployment note

The `slurm_allocation` controller must be launched directly from an `sbatch` script or `salloc` shell rather than through `srun`, because it creates exclusive `srun` job steps for its workers. The Python environment, source, configs, data, and outputs must be visible at consistent paths on every allocated node. Worker steps explicitly request their proportional share of `SLURM_MEM_PER_NODE`; without that request, Slurm may assign the full allocation memory TRES to the first exclusive step and leave the remaining GPUs idle. Some clusters still serialize nested steps as a site policy, so the docs include a two-step admission smoke check and direct single-node users to the `local` executor when concurrent steps are unavailable.

### DIFF
--- a/src/samudra/backend.py
+++ b/src/samudra/backend.py
@@ -55,15 +55,19 @@ def init_train_backend(
         case "auto" if torch.cuda.is_available():
             logger.info("auto backend detected CUDA")
             device = torch.device("cuda")
-            try:
-                dist_cfg = init_distributed_mode()
-                logger.info("succeeded in initializing distributed mode")
-            except RuntimeError as e:
-                logger.info(
-                    f"Failed to initialize distributed mode, running on single GPU.",
-                    exc_info=e,
-                )
+            if os.environ.get("SAMUDRA_DISABLE_DISTRIBUTED") == "1":
+                logger.info("distributed initialization disabled for this worker")
                 dist_cfg = None
+            else:
+                try:
+                    dist_cfg = init_distributed_mode()
+                    logger.info("succeeded in initializing distributed mode")
+                except RuntimeError as e:
+                    logger.info(
+                        "Failed to initialize distributed mode, running on single GPU.",
+                        exc_info=e,
+                    )
+                    dist_cfg = None
         case "auto":
             logger.warning(
                 "auto backend: cuda not found, using CPU. " + _cuda_diagnostics()
--- a/src/samudra/configs/search/local.yaml
+++ b/src/samudra/configs/search/local.yaml
@@ -4,3 +4,5 @@
 
 type: local
 output_dir: .LOCAL/searches
+# By default, use every visible GPU (including GPUs across a Slurm allocation).
+# max_concurrent: 4
--- a/src/samudra/configs/search/local.yaml
+++ b/src/samudra/configs/search/local.yaml
@@ -4,5 +4,5 @@
 
 type: local
 output_dir: .LOCAL/searches
-# By default, use every visible GPU (including GPUs across a Slurm allocation).
+# By default, use every GPU visible to the current machine.
 # max_concurrent: 4
--- a/src/samudra/search/executors/local.py
+++ b/src/samudra/search/executors/local.py
@@ -2,35 +2,19 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""Resource-aware execution in the current machine or Slurm allocation."""
+"""Run independent search trials on one local machine."""
 
-from __future__ import annotations
-
-import gc
 import os
 import subprocess
-import sys
-from collections.abc import Callable
-from concurrent.futures import ThreadPoolExecutor, as_completed
-from dataclasses import dataclass
 from queue import SimpleQueue
-from typing import Any
 
 import torch
 
 from samudra.search.config import LocalExecutorConfig
-from samudra.search.executors.base import Executor
-from samudra.utils.multiton import MultitonScope
+from samudra.search.executors.pool import PoolExecutor, Task
 
 
-@dataclass(frozen=True)
-class _Task:
-    rung: int
-    task: int
-    anchor: bool
-
-
-def _visible_devices() -> list[str]:
+def visible_devices() -> list[str]:
     """Return CUDA identifiers that can be assigned to child processes."""
     configured = os.environ.get("CUDA_VISIBLE_DEVICES")
     if configured is not None:
@@ -40,40 +24,9 @@ def _visible_devices() -> list[str]:
     return [str(index) for index in range(torch.cuda.device_count())]
 
 
-def _positive_int_environment(name: str) -> int | None:
-    value = os.environ.get(name)
-    if value is None:
-        return None
-    try:
-        parsed = int(value)
-    except ValueError:
-        return None
-    return parsed if parsed > 0 else None
+class LocalExecutor(PoolExecutor):
+    """Use each GPU visible to the current machine as a worker slot."""
 
-
-def _slurm_gpu_count() -> int:
-    """Best available count for a homogeneous Slurm GPU allocation."""
-    total = _positive_int_environment("SLURM_GPUS")
-    if total is not None:
-        return total
-    nodes = _positive_int_environment("SLURM_NNODES") or 1
-    per_node = _positive_int_environment("SLURM_GPUS_ON_NODE")
-    if per_node is None:
-        per_node = len(_visible_devices())
-    if per_node < 1:
-        raise RuntimeError("The Slurm allocation does not expose any GPUs")
-    return nodes * per_node
-
-
-def _slurm_cpus_per_gpu() -> int:
-    cpus = _positive_int_environment("SLURM_CPUS_ON_NODE")
-    gpus = _positive_int_environment("SLURM_GPUS_ON_NODE")
-    if cpus is None or gpus is None:
-        return 1
-    return max(1, cpus // gpus)
-
-
-class LocalExecutor(Executor):
     @property
     def config(self) -> LocalExecutorConfig:
         config = self.search.config.executor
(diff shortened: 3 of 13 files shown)


src/samudra/search/executors/pool.py
```diff
@@ -0,0 +1,137 @@
...
+            return
+        self._run_tasks(
+            [
+                Task(len(self.search.rungs) - 1, task, True)
+                for task, _ in enumerate(candidates)
+            ]
+        )
+
+    def submit_rung(self, state: dict[str, Any], rung: int) -> None:
+        candidates = state["rungs"][rung]["candidates"]
```
alxmrs, line 78: For a future PR: I'd like to refactor state to use (data) classes and not python dictionaries. This is good for now.

src/samudra/search/executors/slurm_allocation.py
```diff
@@ -0,0 +1,96 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Run independent search trials inside an existing Slurm allocation."""
+
+import os
+import subprocess
+
+from samudra.search.config import SlurmAllocationExecutorConfig
```
alxmrs: Nit: I think this should be guarded as a typing only import.

src/samudra/search/executors/local.py
```diff
@@ -2,44 +2,66 @@
...
-from typing import Any
+import os
+import subprocess
+from queue import SimpleQueue
 
 import torch
 
-from samudra.search.executors.base import Executor
-from samudra.utils.multiton import MultitonScope
+from samudra.search.config import LocalExecutorConfig
```
alxmrs: Nit: I think this should be guarded as a typing only import.

src/samudra/backend.py
```diff
@@ -55,15 +55,19 @@ def init_train_backend(
...
-                logger.info(
-                    f"Failed to initialize distributed mode, running on single GPU.",
-                    exc_info=e,
-                )
+            if os.environ.get("SAMUDRA_DISABLE_DISTRIBUTED") == "1":
+                logger.info("distributed initialization disabled for this worker")
                 dist_cfg = None
+            else:
+                try:
+                    dist_cfg = init_distributed_mode()
```
alxmrs, line 63: Does this allow later rungs to use multiple GPUs (however many GPUs would otherwise be left idle on the node(s))?

tests/test_search.py
```diff
@@ -643,6 +653,103 @@ def test_local_executor_runs_tasks_then_advances(tmp_path, monkeypatch):
...
+
+    search.executor.submit_initial(state)
+
+    assert batches == [[(1, 0, True), (0, 0, False), (0, 1, False)], 0]
+    saved = search.read_state()
+    assert saved["anchors"]["job_id"] == "local"
+    assert saved["rungs"][0]["job_id"] == "local"
+
+
+def test_slurm_allocation_executor_uses_exclusive_gpu_steps(tmp_path, monkeypatch):
```
alxmrs, line 699: I'd like another test like this to confirm that our set up would utilize otherwise idle GPUs at later rungs. For example, if we had 8 candidates on an 8 device single node environment, the first rung should use one GPU per candidate. The next rung should use 2 GPUs per the remaining 4 candidates, and so on and so forth. Does this capability exist in how we've defined our executors? If not, I think that it should. And, forgive me if this is implemented and has a test, I have only read up to this point.

alxmrs (commented):
Feedback to the agent.

src/samudra/search/executors/pool.py
alxmrs: 🤖 Agreed. I kept the state dictionaries unchanged in this PR and agree that migrating executor/search state to typed data classes is a good focused follow-up.

src/samudra/search/executors/slurm_allocation.py
alxmrs: 🤖 Addressed in 89bbee4a. The config class is now imported under TYPE_CHECKING, and the runtime discriminator check narrows the value before a typed cast.

src/samudra/search/executors/local.py
alxmrs: 🤖 Addressed in 89bbee4a. LocalExecutor now uses a TYPE_CHECKING-only config import plus a runtime discriminator check and typed cast.

src/samudra/backend.py
alxmrs: 🤖 No—deliberately. Every candidate remains a single-GPU experiment in every rung. Automatically moving a resumed candidate from 1 to 2/4/8-way DDP changes its global batch size and optimizer steps per epoch midway through its trajectory, so it is not simply extra throughput. It also conflicts with the original goal of distributing independent single-GPU experiments. I documented this invariant in 89bbee4a. Adaptive multi-GPU promotion could be added later as an explicit policy once batch-size, learning-rate, scheduler, and epoch semantics are defined.

tests/test_search.py
alxmrs: 🤖 I added a later-rung test in 89bbee4a, but I am pushing back on automatically assigning the idle GPUs to survivors. The new test confirms that four survivors on an eight-GPU machine each retain exactly one GPU. Changing world size on promotion changes training semantics (global batch size and optimizer steps per epoch), which can invalidate the meaning of a resumed successive-halving budget. The docs now call this out explicitly. I recommend treating adaptive DDP as a separate opt-in feature with an explicit per-rung resource and batch/LR policy rather than silently inferring it from idle hardware.

alxmrs:
@codex may I have your review once more?

src/samudra/search/executors/local.py
```diff
@@ -2,44 +2,238 @@
...
+        if "SLURM_JOB_ID" in os.environ:
+            step = os.environ.get("SLURM_STEP_ID")
+            if step not in {None, "", "batch", "extern"}:
+                raise RuntimeError(
+                    "Cannot launch local search workers from inside an existing "
+                    "Slurm job step. Run the search directly in an sbatch script "
+                    "or an salloc shell, rather than with `srun python`."
+                )
+            concurrency = _slurm_gpu_count()
+            runner: Callable[[_Task], None] = self._run_slurm_task
```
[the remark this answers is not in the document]
alxmrs: 🤖 This was addressed by the executor split in 67dfda1b/89bbee4a. The local executor no longer branches on `SLURM_JOB_ID`; it checks only locally visible CUDA devices and runs tasks in-process when there are none. The allocation-wide executor is now an explicit `slurm_allocation` mode and intentionally requires GPUs.

src/samudra/search/executors/local.py
```diff
@@ -2,44 +2,238 @@
...
+        items: list[Any], concurrency: int, runner: Callable[[Any], None]
+    ) -> None:
+        errors: list[Exception] = []
+        with ThreadPoolExecutor(max_workers=max(1, concurrency)) as pool:
+            futures = [pool.submit(runner, item) for item in items]
+            for future in as_completed(futures):
+                try:
+                    future.result()
+                except Exception as error:
+                    errors.append(error)
```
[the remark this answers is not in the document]
alxmrs: 🤖 Addressed in 48f52050. `_run_concurrently` now keeps only the active concurrency window submitted, stops replenishing it as soon as a failure is observed, cancels work that has not started, and waits only for already-running workers. The new regression test verifies that tasks beyond the initial two-worker window never launch after the first candidate fails.

src/samudra/configs/search/local.yaml
```diff
@@ -4,3 +4,5 @@
 
 type: local
 output_dir: .LOCAL/searches
+# By default, use every visible GPU (including GPUs across a Slurm allocation).
```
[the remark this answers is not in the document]
alxmrs: 🤖 Addressed in 48f52050. The local template now says it uses GPUs visible to the current machine; multi-node allocation-wide execution remains explicitly documented under `slurm-allocation.yaml`.

alxmrs:
@codex, will you review this PR once more?

alxmrs:
🤖 Torch deployment found an allocation-pooling issue: each one-GPU `srun --exclusive --exact` worker step inherited the batch allocation’s full memory TRES. In job 16906321 (4 GPUs, 64 CPUs, 700G), step 16906321.0 requested `cpu=16,gres/gpu=1,mem=700G`, so Slurm could start only one worker and left the other three GPUs idle. The executor should set proportional step memory (for example, derive memory per GPU from `SLURM_MEM_PER_NODE / SLURM_GPUS_ON_NODE` and request `world_size * memory_per_gpu`) and cover this command construction in tests. I am applying that fix on the experiment integration branch and will relaunch.

alxmrs:
🤖 Addressed the Torch deployment finding in 23996089. Allocation worker steps now request a proportional memory share using `SLURM_MEM_PER_NODE * step_gpus / SLURM_GPUS_ON_NODE`, preventing a one-GPU step from inheriting the full-node memory TRES and serializing the pool. The command-construction test covers the resulting `--mem` value, and `docs/search.md` explains why it is required.

I compared this with `e77f73e1` on `experiment/structured-perceiver-inverse-search-torch-adaptive` and deliberately aligned the helper/API with that tested integration fix. I also rebased #868 onto this commit and added coverage for the full-node multi-GPU step, where `--mem` correctly remains the per-node allocation. The integration branch can drop `e77f73e1` when it next rebases.

alxmrs:
🤖 Follow-up from Torch: proportional step memory corrected the TRES (`cpu=16,gres/gpu=1,mem=175G`), but nested `srun` workers still serialized on this cluster. Four concurrent `srun` client processes existed under the controller, while Slurm admitted only step `.0`; the other three waited. This may be a Torch-specific step-overlap policy, but it means `slurm_allocation` is not yet a reliable pooled backend there. The working deployment uses the PR's resource-aware `local` pool inside a single-node `sbatch` allocation, where disjoint `CUDA_VISIBLE_DEVICES` assignments successfully run four workers concurrently. I recommend documenting that single-node pattern and adding a multi-step integration check on clusters intended to use `slurm_allocation`.

alxmrs:
🤖 Addressed in e1c7440e. I agree this is an important deployment constraint, but it is a cluster-side Slurm step-admission policy rather than another resource-calculation bug the executor can infer or correct portably. The docs now require validating concurrent exclusive steps with a two-step `srun`/`squeue --steps` smoke check and explicitly recommend the proven `local` executor inside a single-node allocation when nested steps serialize. A real multi-step test therefore remains a deployment qualification for each target cluster rather than a deterministic unit/CI test.

alxmrs:
🤖 Follow-up experiment review found one additional Slurm portability detail, addressed in 652d674b: some allocations expose `SLURM_GPUS_ON_NODE` with a topology suffix such as `8(S:0-1)`. The executor now parses that form conservatively, and the allocation command test exercises capacity, CPU, and proportional-memory calculations through the suffixed value. The branch is also rebased onto current `main`.

alxmrs:
🤖 Follow-up from the overnight Torch run: Slurm externally cancelled pooled job 16906674 after 2h14m (likely utilization enforcement; W&B system GPU averages were only ~1–6%). One worker had fully completed with checkpoint/summary and another had finished all 2,816 train batches, but because `PoolExecutor._run_tasks()` had not returned, the search `state.json` still says `running` with zero rung results. The remaining five candidates never started. A straightforward resume would also relaunch completed tasks and hit the output overwrite guard. For allocation-level interruption tolerance, the pool should harvest each completed future into durable rung state as it finishes and resume by skipping already-complete candidate outputs (or explicitly mark them). At minimum, stale `running` state should be detectable/reconcilable after the allocation disappears.

alxmrs:
@fomo-bot will you review this change?

alxmrs:
🤖 Final operational review: the pooled-controller interruption issue remains merge-blocking at the current head (`652d674b`). Today I deliberately cancelled Torch job `16975829` while its sole worker was still loading its first batch. The worker process disappeared, but the durable search record remains `status: running`, `rungs[0].results: []`, and `job_id: local`; there is no supported terminalization or reconciliation path. This is the same failure class as the overnight partial-completion report, now reproduced even before a first result. I recommend having pooled execution persist/reconcile each task outcome as futures finish and, on restart, classify stale `running` tasks from their worker-status/output artifacts before relaunching. At minimum, an interrupted controller must not leave a permanently live-looking search. I would resolve this before merging #865 because #868 inherits the state lifecycle.

alxmrs:
🤖 One additional merge-time cleanup remains at current head `652d674b`: the GPU-visible-host test failure reported by fomo-bot is still present. `test_local_executor_runs_tasks_then_advances` patches `search.train_task`, but does not hide/mock `visible_devices()`, so a CUDA-visible test host takes the subprocess path and bypasses that patch. Please make the test hermetic (for example, patch the local executor\x27s device discovery to return no devices for this in-process behavior test) and retain a separate test for the GPU subprocess path. This is a P2 test reliability issue, independent of the P1 interruption/reconciliation blocker above.

PR #870, opened 2026-09-04
Retry transient normalization failures

alxmrs:
## Summary

- retry individual failed Dask tasks during normalization writes, five times by default
- expose the retry count through `NORM_WRITE_RETRIES` in the Slurm harness
- mark normalization Slurm jobs requeueable for scheduler or node interruptions
- replace partial derived normalization stores so a requeued job can restart safely
- test the Dask retry wiring and Slurm harness configuration

## Failure handling

- transient S3/Dask task failure: Dask retries only the failed task
- scheduler or node interruption: Slurm may requeue the job
- interrupted partial output: `mode="w"` safely replaces the derived store

## Testing

- `bash -n scripts/slurm_make_norm_om4.sbatch`
- `uv run pytest data/tests/test_slurm_harness.py data/tests/test_make_norm_datasets.py -q`

The September normalization jobs remain paused until this PR is approved.

### DIFF
--- a/data/ocean_preprocessing/make_norm_datasets.py
+++ b/data/ocean_preprocessing/make_norm_datasets.py
@@ -4,16 +4,34 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""A quick utility script to create *_means.zarr and *_stds.zarr from a dataset."""
+"""Create *_means.zarr and *_stds.zarr from a dataset."""
 
+import argparse
 import os
-import sys
 
-import fsspec
 import xarray as xr
-from dask.distributed import LocalCluster
+from dask.distributed import Client, LocalCluster
 
-if __name__ == "__main__":
+
+def write_zarr_with_retries(
+    ds: xr.Dataset, path: str, client: Client, write_retries: int
+) -> None:
+    """Write a derived dataset, retrying failed Dask tasks in place."""
+    delayed = ds.to_zarr(
+        path,
+        mode="w",
+        zarr_format=2,
+        consolidated=True,
+        compute=False,
+    )
+    client.compute(delayed, retries=write_retries, sync=True)
+
+
+def main(src: str, write_retries: int = 5) -> None:
+    """Calculate and write normalization statistics for ``src``."""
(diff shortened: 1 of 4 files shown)


data/ocean_preprocessing/make_norm_datasets.py
```diff
@@ -73,9 +73,11 @@
 
     mean_path = os.path.join(path, f"{ds_name}_means{ds_ext}")
     print(f"Writing mean Zarr to {mean_path!r}.")
-    ds_means.to_zarr(mean_path, zarr_format=2, consolidated=True)
+    # These stores are derived entirely from ``src``. Replace a partial store
+    # left by an interrupted attempt so that batch-level retries are safe.
+    ds_means.to_zarr(mean_path, mode="w", zarr_format=2, consolidated=True)
```
alxmrs: We can also tap into dask level retries -- see the __main__.py script (the `_collect()` method) for an example. e.g.:
```
self.dask_client.compute(delayed, retries=self.write_retries, sync=True)
```

alxmrs: Implemented the middle ground: normalization now submits lazy Zarr writes through `client.compute(..., retries=..., sync=True)`, while the Slurm job remains requeueable for the pre-Python scheduler failures we observed. I removed the shell-level whole-process retry loop and kept restart-safe `mode="w"`. 🤖

alxmrs:
Fixed the data-test collection failures in `2dc4f3e5`. The new unit test imported `make_norm_datasets`, whose module-level `dask.distributed` import failed in the lean data-test environment (which intentionally has plain `dask` only). `LocalCluster` is now imported only when the normalization CLI runs, while `Client` is imported only for static type checking. I also verified that the module imports with `distributed` explicitly blocked. 🤖

alxmrs (commented):
LGTM

alxmrs:
@codex will you review this code?

alxmrs:
@fomo-bot is this PR ready to be merged? If so, please approve it.

PR #872, opened 2026-09-04
Archive W&B runs to OSN weekly

alxmrs:
## Summary

- archive terminal W&B runs to the public OSN pod every Monday at 07:17 UTC and on manual dispatch
- launch an AWS `m7i.xlarge` CPU runner with 200 GB of temporary disk and a 12-hour lifetime ceiling
- use the pinned W&B Archive v0.1.1 action with public-safe filtering, retries, incremental/idempotent publishing, and a single-writer concurrency guard
- fail in a GitHub-hosted preflight before launching EC2 when archive credentials are missing

Weekly incremental exports keep each update smaller, surface failures sooner, and reduce exposure to W&B quota pressure. A GPU is intentionally not requested because the export is limited by W&B API and network throughput rather than CUDA compute.

## Required repository secrets

These secrets must be configured before the workflow is run:

- `WANDB_API_KEY`
- `OSN_ACCESS_KEY_ID`
- `OSN_SECRET_ACCESS_KEY`

The existing `GH_SA_TOKEN` and AWS repository variables continue to launch the self-hosted runner through `Open-Athena/ec2-gha`.

## Validation

- `uvx pre-commit run --files .github/wandb-archive.yaml .github/workflows/wandb-archive.yml`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 .github/workflows/wandb-archive.yml`
- loaded `.github/wandb-archive.yaml` with the published `wandb-archive==0.1.1` configuration model

The initial historical backfill can continue on Torch; this workflow is designed primarily for idempotent weekly increments, while retaining enough runtime headroom to catch up after a missed run.

### DIFF
--- a/.github/wandb-archive.yaml
+++ b/.github/wandb-archive.yaml
@@ -6,8 +6,13 @@ source:
   entity: ocean_emulators
   timeout_seconds: 60
   projects:
-    include: ["*"]
-    exclude: ["scratch-*"]
+    # Keep public publication opt-in: new W&B projects must be reviewed here.
+    include:
+      - default
+      - marin
+      - ocean-emulators
+      - samudra-multi-decoder
+    exclude: []
   runs:
     states: [finished, failed, crashed, killed, preempted]
   retry:
--- a/.github/wandb-archive.yaml
+++ b/.github/wandb-archive.yaml
@@ -1,0 +1,42 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+source:
+  entity: ocean_emulators
+  timeout_seconds: 60
+  projects:
+    include: ["*"]
+    exclude: ["scratch-*"]
+  runs:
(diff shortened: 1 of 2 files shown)


alxmrs (commented):
LGTM

alxmrs:
@codex will you review this PR?

alxmrs:
Hardening follow-up added in 7634d26e:

- pins the EC2 reusable workflow and its runtime checkout to immutable commit `e021da12a946b653a717738bc704aa7baa95ae8e`
- prevents checkout credential persistence and checks out only the archive configuration
- makes preflight receive credential-presence booleans rather than the credential values
- replaces the project wildcard with the four projects currently returned by W&B (`default`, `marin`, `ocean-emulators`, and `samudra-multi-decoder`)

Validation: repository pre-commit hooks, actionlint, published v0.1.1 configuration loading, and zizmor. Zizmor now reports no findings for this workflow.

Issue #869, opened 2026-09-03
Support scientifically correct visualization on native OM4 tripolar grids


alxmrs:
🤖 Follow-up from #492 and #421.

## Problem

The training and eval paths can now identify an OM4 source as `grid_type=tripolar` and preserve its true two-dimensional `lat`/`lon` coordinates in prediction output. The legacy `samudra viz` path is still rectilinear-only, however. It cannot currently produce scientifically valid diagnostics for the native 1080 × 1440 OM4 tripolar snapshot dataset.

There are three concrete assumptions to remove:

1. `Viz.__init__` in `src/samudra/viz/core.py` drops the source's two-dimensional `lat` and `lon`, then renames `y`/`x` to `lat`/`lon`. This loses the curvilinear geography.
2. The same constructor replaces the dataset's exact `areacello` with `spherical_area_weights()` and calculates `areacello_spherical` with `spherical_area()`. Both helpers assume separable, uniformly spaced one-dimensional latitude/longitude. Those calculations are invalid on a tripolar grid.
3. `process_mask()` assigns target `y`/`x` coordinate values to an existing Gaussian basin mask by position. The published quarter-degree mask is 720 × 1440, while the native tripolar grid is 1080 × 1440, so this either raises a dimension-size error or—if shapes happen to agree for another curvilinear grid—silently misaligns geography.

Map methods also generally let xarray use the array dimensions as Cartesian plotting coordinates. Native-grid maps must instead use the preserved two-dimensional geographic coordinates with an explicit Plate Carrée transform.

## Reproduction on Torch

The completed native-grid data and normalization stores are at:

```text
s3://emulators/am16581/data/2026-06/om4_tripolar_snapshots/
├── OM4.zarr                 # time=4745, y=1080, x=1440
├── OM4_means.zarr
└── OM4_stds.zarr
```

With the credentialed Torch rclone remote:

```bash
ssh torch
mkdir -p /scratch/$USER/data/v2026-09/om4_tripolar_snapshots
rclone copy \
  nyu-osn-write:/emulators/am16581/data/2026-06/om4_tripolar_snapshots/ \
  /scratch/$USER/data/v2026-09/om4_tripolar_snapshots/
```

Issue #492 has a staged train/eval chain (`16903420` → `16903477` → `16903478`). Once eval completes, attempt the normal high-resolution visualization against its `predictions.zarr`, for example:

```bash
python -m samudra.viz \
[...]

Issue #492, opened 2025-12-01
Re-create Snapshot datasets + training runs with new baselines.


alxmrs:
Process data:
- [x] 1/4-degree **snapshot** data + no blur
- [x] 1/4-degree **snapshot** data + no blur + native tripolar grid

Train/Eval/Viz:
- [x] 1/4-degree **snapshot** data + no blur
- [ ] 1/4-degree **snapshot** data + no blur + native tripolar grid

alxmrs:
Currently working on this now since I just created new new snapshot datasets.

alxmrs:
🤖 Scheduled the new snapshot baseline work on NYU Torch using the current high-resolution Samudra v2 config and a commit-pinned PhysicsNeMo image.

**Regridded 1/4° snapshot DAG**
- Stage September public data after its norm job: `16903419` (`afterok:16894287`)
- Train on 4× H200: `16903466` (`afterok:16903418:16903419`)
- Eval on 1× H200: `16903476` (`afterok:16903466`)
- Viz on CPU/490 GB: `16903494` (`afterok:16903476`)

**Native-tripolar 1/4° snapshot DAG**
- Stage the completed 2026-06 dataset: `16903420`
- Train on 4× H200 with `grid_type=tripolar`: `16903477` (`afterok:16903418:16903420`)
- Eval on 1× H200 with `grid_type=tripolar`: `16903478` (`afterok:16903477`)

Image pre-pull job: `16903418`. Logs are under `/scratch/am16581/training-logs/issue-492/`.

I have intentionally left both checkboxes open until the chains finish and outputs are verified. There is one remaining blocker before scheduling the native-tripolar viz: the current legacy `samudra viz` path assumes a rectilinear grid when calculating cell areas and aligning basin masks. Running it unchanged would silently produce scientifically invalid tripolar diagnostics. The train/eval path already preserves native 2-D geography correctly; I am treating viz support as a code fix rather than submitting a known-invalid job.

jder:
For posterity: I think the above runs did not start because 16903420 failed due to disk quota. AFAICT we might also want to think about polar grid connectivity as well as the area computation?

PR #876, opened 2026-09-05
Support tripolar grids in viz

SuMayaBee:
This is #869.

Viz never learned what grid it was on. `Viz.__init__` hard-coded `build_om4_layout()`, which defaults to gaussian, so everything downstream treated the native tripolar grid as if it were rectilinear and got away with it quietly. It now takes the grid type from the data source and branches on it.

Thanks @alxmrs for pointing me at `raw/om4_5daily.zarr` and `raw/ocean_static_no_mask_table.zarr`. Everything below is measured against those rather than a fixture.

The clearest way I found to show why the areas matter is global mean `zos`. It's an anomaly about the global mean, so the answer has to be zero:

| weighting | global mean `zos` |
| --- | --- |
| source `areacello` | `-3.4e-10 m` |
| `cos(lat)`, as before | `-0.1345 m` |

13.5 cm of artifact. Cosine weighting starves the tropics by about 8 points of total area and inflates the Arctic by 70%, because near the fold the cells are small and `cos(lat)` doesn't know that. This isn't only a plotting problem, `spherical_area_weights()` overwrote the source `areacello` and fed about thirty reductions including OHC and global mean SST.

What changed:

- 2-D lat/lon survive the y/x rename as `lat_2d`/`lon_2d`, following `with_lat_lon_coords()` and `ZarrWriter`. Predictions too, since rollouts land in the same layout.
- Real `areacello` on curvilinear grids, and a loud failure when it's missing.
- Basin masks validated instead of relabeled by position.
- Maps plot against the 2-D coords, all six `pcolormesh` sites.
- Four steps refuse curvilinear input: `thetao_mae_metrics`, `enso_plots`, `movies`, `ocean_temperature_profile_plots`. The last one averages each basin over `x` and then labels that axis "Latitude".
- Basin masks build lazily, so a smoke test no longer depends on mask alignment succeeding.

This also turns on the guard already sitting in `metrics/observations.py`, which could never fire before.

New `ocean_preprocessing basin_masks` subcommand for the native masks, since the published ones are all rectilinear. No regridding needed, `ocean_static` already has integer region codes on the native grid next to the real cell centers and the wet mask, and integer codes partition by construction. Against the public store it assigns 953,060 of 969,446 wet cells, max one basin per cell. Marginal seas stay unassigned, which I checked against `basin_masks_original.zarr` rather than guessing.

Against the acceptance criteria in #869:

| | criterion | |
| --- | --- | --- |
| 1 | Nonseparable fixture keeps exact 2-D lat/lon | ✅ |
| 2 | Exact `areacello` retained, distinguished from cos-lat | ✅ |
| 3 | Mismatched basin masks fail with an actionable error | ✅ |
| 4 | Matching native mask aligns, expected basin result | ✅ |
| 5 | Map test asserts 2-D coords and Plate Carrée | ✅ |
| 6 | Gaussian tests unchanged | ✅ |
| 7 | Torch smoke viz, then the full #492 job | ❌ |

7 needs a tripolar rollout and I don't have Torch access. Happy to run it once @YuanYuan98's export lands, or it may be quicker for someone with cluster access to run it against this branch.

Ran locally: 86 tests across viz, writer and config, 78 across metrics and utils, 11 in the data env, pre-commit clean. The full suite OOMs on my laptop so I'm leaning on CI for the rest.

### DIFF
--- a/data/ocean_preprocessing/__main__.py
+++ b/data/ocean_preprocessing/__main__.py
@@ -24,6 +24,7 @@ import fire
 import fsspec
 import xarray as xr
 
+from ocean_preprocessing.basin_masks import basin_masks_from_static
 from ocean_preprocessing.dataset_validation import (
     ds_flattened_input_validate,
     ds_input_validate,
@@ -521,6 +522,46 @@ class CLI:
         logger.info("collecting!")
         self._collect(ds)
         logger.info("done!")
+
+    def basin_masks(self, static_path: str):
+        """Build ocean-basin masks on a model's native horizontal grid.
+
+        The published basin masks are all on regular lat-lon grids, so none of
+        them applies to OM4's native 1080x1440 tripolar grid. OM4's
+        `ocean_static` already carries integer region codes there, alongside the
+        real 2-D cell centers and the wet mask, so the masks can be built
+        directly with no regridding.
+
+        Args:
+            static_path: An OM4 `ocean_static` store, e.g.
+                `s3://m2lines-pubs/Samudra/raw/ocean_static_no_mask_table.zarr`.
+
+        Example:
+            python -m ocean_preprocessing \
+              --output_path=basin_masks_native.zarr \
+              basin_masks \
+              --static_path=s3://m2lines-pubs/Samudra/raw/ocean_static_no_mask_table.zarr
+        """
+        logger.info(f"reading ocean_static from {static_path}")
+        static = xr.open_zarr(static_path, chunks={})
+        if "time" in static.dims:
+            static = static.isel(time=0, drop=True)
+
+        masks = basin_masks_from_static(static)
+        logger.info(
+            "built masks: "
+            + ", ".join(
+                f"{name}={int(masks[name].sum())} cells" for name in masks.data_vars
+            )
+        )
+
+        if self.dry_run:
+            logger.info(masks)
+            return
+
+        logger.info(f"writing masks to {self.output_path}")
+        masks.to_zarr(self.output_path, mode="w", consolidated=True, zarr_format=2)
+        logger.info("zarr write complete")
 
     def cm4(self):
         """Process the CM4 oceans dataset (a coupled ocean model from CMIP)."""
--- a/data/ocean_preprocessing/__main__.py
+++ b/data/ocean_preprocessing/__main__.py
@@ -235,16 +235,20 @@ class CLI:
         self.wfo_source_path = wfo_source_path
         self.dask_client = init_cluster(cluster, **cluster_opts)
 
-    def _collect(self, ds: xr.Dataset):
+    def _collect(self, ds: xr.Dataset, *, compress: bool = False):
         """Finalize and write the processed dataset to disk.
 
         Args:
             ds: The processed dataset to write. Should already be chunked appropriately.
+            compress: Leave zarr's default compressor in place. The training
+                stores turn compression off so the loader can read them without
+                a decompression pass; the basin masks are small and are read by
+                hand, so they keep the compression the published masks use.
 
         Note:
             Respects dry_run and small_run flags.
         """
-        if self.small_run:
+        if self.small_run and "time" in ds.dims:
             ds = ds.isel(time=slice(0, 10))
         if self.dry_run:
             if self.dask_client is not None:
@@ -261,9 +265,11 @@ class CLI:
             mode="w",
             consolidated=True,
             zarr_format=2,
-            encoding={
-                var_name: {"compressor": None} for var_name in ds.data_vars.keys()
-            },  # Compression turned off
+            encoding=(
+                None
+                if compress
+                else {var_name: {"compressor": None} for var_name in ds.data_vars}
+            ),
             compute=False,
         )
         # Reading blosc-compressed source chunks over S3 occasionally returns a
@@ -555,13 +561,9 @@ class CLI:
             )
         )
 
-        if self.dry_run:
-            logger.info(masks)
-            return
-
-        logger.info(f"writing masks to {self.output_path}")
-        masks.to_zarr(self.output_path, mode="w", consolidated=True, zarr_format=2)
-        logger.info("zarr write complete")
+        # `compress=True` keeps zarr's default blosc/lz4, which is what the
+        # published masks use, so the two sets of masks stay byte-comparable.
+        self._collect(masks, compress=True)
 
     def cm4(self):
         """Process the CM4 oceans dataset (a coupled ocean model from CMIP)."""
--- a/data/ocean_preprocessing/basin_masks.py
+++ b/data/ocean_preprocessing/basin_masks.py
@@ -1,0 +1,199 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Ocean-basin masks on a model's native horizontal grid.
+
+The published basin masks (``s3://m2lines-pubs/Samudra/basins/``) are all on
+regular lat-lon grids, so none of them can be applied to OM4's native 1080x1440
+tripolar grid: the shapes do not match, and where they happen to match on some
+other curvilinear grid, matching shapes would not imply matching geography.
+
+OM4's ``ocean_static`` already carries a ``basin`` field of integer region codes
+on the native grid, alongside the real 2-D cell centers and the wet mask. That
+is enough to build native masks directly, with no regridding and no polygon
+work: the codes partition the grid by construction, so a cell cannot land in two
+basins.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import xarray as xr
+
+# Region codes on OM4's `ocean_static.basin`, from its own CF flag attributes:
+#   flag_values:   0 1 2 3 4 5 6 7 8 9 10
+#   flag_meanings: global_land southern_ocean atlantic_ocean pacific_ocean
+#                  arctic_ocean indian_ocean mediterranean_sea black_sea
+#                  hudson_bay baltic_sea red_sea
+OM4_BASIN_CODES: dict[str, int] = {
+    "basin_southern": 1,
+    "basin_atlantic": 2,
+    "basin_pacific": 3,
+    "basin_arctic": 4,
+    "basin_indian": 5,
+}
+
+# The marginal seas (codes 6-10). The published Gaussian masks leave every one
+# of them unassigned -- checked against `basin_masks_original.zarr`, where the
+# Mediterranean, Black Sea, Red Sea, Hudson Bay and Baltic are zero in all five
+# basins -- so we drop them too rather than inventing an attribution the rest of
+# the basin diagnostics have never made.
+OM4_MARGINAL_SEA_CODES: tuple[int, ...] = (6, 7, 8, 9, 10)
+
+# `basin_id` as the published masks number them. This is *not* OM4's code: the
+# published stores use their own ordering, and we keep it so the two sets of
+# masks describe themselves the same way.
+PUBLISHED_BASIN_IDS: dict[str, int] = {
+    "basin_arctic": 1,
+    "basin_atlantic": 2,
+    "basin_indian": 3,
+    "basin_pacific": 4,
+    "basin_southern": 5,
+}
+
+_LONG_NAMES: dict[str, str] = {
+    "basin_arctic": "Arctic",
+    "basin_atlantic": "Atlantic",
+    "basin_indian": "Indian",
+    "basin_pacific": "Pacific",
+    "basin_southern": "Southern",
+}
+
+
+def basin_masks_from_static(
+    static: xr.Dataset,
+    *,
+    basin_var: str = "basin",
+    wet_var: str = "wet",
+    lat_var: str = "geolat",
+    lon_var: str = "geolon",
+) -> xr.Dataset:
+    """Build native-grid basin masks from an OM4 ``ocean_static`` dataset.
+
+    Args:
+        static: OM4 ``ocean_static``, carrying integer region codes, a wet mask
+            and the real 2-D cell centers on the native horizontal grid.
+        basin_var: Name of the integer region-code field.
+        wet_var: Name of the wet (ocean fraction) mask.
+        lat_var: Name of the 2-D cell-center latitude.
+        lon_var: Name of the 2-D cell-center longitude.
+
+    Returns:
+        A dataset of five boolean masks stored as int8, on horizontal dims named
+        ``lat``/``lon`` (the index axes, matching what ``samudra viz`` renames
+        its data to), carrying the true cell centers as ``lat_2d``/``lon_2d`` so
+        the masks can be matched to data by geography rather than by position.
+    """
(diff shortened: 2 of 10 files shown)


src/samudra/viz/config.py
```diff
@@ -116,6 +122,23 @@ def _groundtruth_location(self) -> Location:
...
+    def _grid_type(self) -> GridType:
+        """Horizontal grid geometry, reused from the data source.
+
+        Viz must branch on this: on a curvilinear ("tripolar") grid the 2-D
+        lat/lon cannot be rebuilt from the 1-D axes and the rectilinear area
+        helpers are invalid. We read it off the primary source rather than
+        configuring it separately so viz cannot disagree with train/eval.
+        """
+        if self.data is None:
+            return "gaussian"
```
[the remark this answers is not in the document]
SuMayaBee: Fixed in df74709.

`_grid_type` now reads `grid_type` off the opened ground-truth store as well as the data source, so the `groundtruth_location`-only path picks up what preprocessing recorded. If the source and the store disagree it raises rather than quietly preferring one, and an unrecognised recorded value is an error too. Three tests added.

src/samudra/viz/core.py
```diff
@@ -212,6 +185,121 @@ def __init__(
...
+
+        The index axes are only geographic on a rectilinear grid. On a
+        curvilinear grid they are cell indices, and plotting against them draws
+        the fold as if it were geography, so we use the preserved 2-D
+        coordinates instead.
+        """
+        if self.data_layout.grid_type == "gaussian":
+            return data["x"], data["y"]
+        if "lon_2d" in data.coords and "lat_2d" in data.coords:
+            return data["lon_2d"], data["lat_2d"]
```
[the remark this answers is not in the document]
SuMayaBee: Fixed in df74709. Added `_map_plot_kwargs()`, which returns `{}` on gaussian and names `lon_2d`/`lat_2d` with a Plate Carrée transform otherwise, and routed `map_bias_avg` through it.

I checked the rest while I was there: every other geographic `.plot()` call is inside `step_movies`, which already refuses curvilinear input, so this was the only one still on index axes.

SuMayaBee:
@codex review

alxmrs:
Thanks Sumaya. I'll TAL at this now.

data/ocean_preprocessing/__main__.py
```diff
@@ -522,6 +523,46 @@ def om4(
...
+                f"{name}={int(masks[name].sum())} cells" for name in masks.data_vars
+            )
+        )
+
+        if self.dry_run:
+            logger.info(masks)
+            return
+
+        logger.info(f"writing masks to {self.output_path}")
+        masks.to_zarr(self.output_path, mode="w", consolidated=True, zarr_format=2)
```
alxmrs: Could we reuse `_collect` by chance? This includes retry logic, since dask pipelines are prone to crash.

alxmrs: Feel free to refactor `_collect` to make it generic.

data/ocean_preprocessing/basin_masks.py
```diff
@@ -0,0 +1,199 @@
...
+# of them unassigned -- checked against `basin_masks_original.zarr`, where the
+# Mediterranean, Black Sea, Red Sea, Hudson Bay and Baltic are zero in all five
+# basins -- so we drop them too rather than inventing an attribution the rest of
+# the basin diagnostics have never made.
+OM4_MARGINAL_SEA_CODES: tuple[int, ...] = (6, 7, 8, 9, 10)
+
+# `basin_id` as the published masks number them. This is *not* OM4's code: the
+# published stores use their own ordering, and we keep it so the two sets of
+# masks describe themselves the same way.
+PUBLISHED_BASIN_IDS: dict[str, int] = {
```
alxmrs, line 63: Do you have a link or reference to an authority of some sort for these IDs? Where do they come from?

data/ocean_preprocessing/basin_masks.py
```diff
@@ -0,0 +1,199 @@
...
+        if array.shape != codes.shape:
+            raise ValueError(
+                f"{name!r} has shape {array.shape} but {basin_var!r} is "
+                f"{codes.shape}; they must be on the same horizontal grid."
+            )
+
+    ny, nx = codes.shape
+    data_vars: dict[str, xr.DataArray] = {}
+    for mask_name, code in OM4_BASIN_CODES.items():
+        mask = (codes == code) & wet
```
alxmrs, line 134: Ah, so the mask are integer codes, like sort of connected components. Thanks, I see how this function is necessary.

data/ocean_preprocessing/basin_masks.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def _axis(static: xr.Dataset, dim: str, size: int) -> np.ndarray:
+    """The source's index axis for `dim`, or a plain range when it has none."""
+    if dim in static.coords:
+        return np.asarray(static.coords[dim].values)
+    return np.arange(size)
+
+
+def validate_basin_masks(masks: xr.Dataset, *, wet: np.ndarray | None = None) -> None:
```
alxmrs, line 193: 🐑  I'm torn, should this belong here, or in the other file where we perform other validators?

src/samudra/viz/config.py
```diff
@@ -116,6 +123,47 @@ def _groundtruth_location(self) -> Location:
...
 
+    def _configured_grid_type(self) -> GridType | None:
+        """Grid geometry named by the data source, if there is one."""
+        if self.data is None:
+            return None
+        source = self.data.sources[0]
+        if isinstance(source, LlcDataSourceConfig):
+            # LLC (lat-lon-cap) carries no `grid_type` field because it is only
+            # ever curvilinear; `build_llc_layout` hard-codes it.
+            return "tripolar"
```
alxmrs: I don't think LLC is really "tripolar", see https://podaac.jpl.nasa.gov/dataset/MITgcm_LLC4320_Pre-SWOT_JPL_L4_ACC_SMST_v1.0 + https://gmd.copernicus.org/articles/16/7143/2023/gmd-16-7143-2023.html#Ch1.S2.SS2. 

It has 13 faces, and each face is 4320^2 pixels, and within each face is a regular grid. I do believe "lat-lon-cap" is the technical term for the grid; I think that we might be better off adding that enumeration to `grid_type` rather than calling this "tripolar".

alxmrs: To be fair, I forget what `GridType` is used for.

SuMayaBee:
@alxmrs I looked into the two red checks and I think they're the fork gate rather than anything in the diff. `test-gpu.yml` conditions its `ec2` job on `head.repo.full_name == github.repository`, and `container-physicsnemo.yml` does the same on `ec2-build`, so both skip for a PR from a fork. The report jobs are then written to fail when that happens, which is what the log says:

```
AWS-backed GPU tests did not run automatically.
Run this workflow manually from the GitHub Actions UI once the code has been reviewed.
```

I saw you re-ran them. I think a re-run evaluates the same condition against the same PR event, so it would skip the same way. A `workflow_dispatch` on the branch should actually run them.

Everything that isn't AWS-backed is green, including `test-cpu`

alxmrs (commented):
Review part 1, will resume part 2 soon.

alxmrs:
@SuMayaBee wrt to the forking gate -- we'd like to keep this for now, since our GPU CI makes use of our internal AWS resources. However, we're able to override the checks to merge this into main when the PR is ready.

src/samudra/viz/config.py
```diff
@@ -116,6 +123,47 @@ def _groundtruth_location(self) -> Location:
...
+        return source.grid_type
+
+    def _grid_type(self, groundtruth: xr.Dataset) -> GridType:
```
alxmrs, line 137: I like that we have this private method to parse the grid type from the dataset.

src/samudra/viz/core.py
```diff
@@ -212,6 +185,137 @@ def __init__(
         self.obs_path = os.path.join(output_path, "Figures_observations")
         self._obs_cache: tuple | None = None
 
+    @functools.cached_property
+    def basin_masks(self) -> xr.Dataset:
```
alxmrs, line 189: I like that we're creating these lazily.

tests/test_viz_config.py
```diff
@@ -42,3 +44,73 @@ def test_missing_groundtruth_and_data_is_an_error():
...
+
+def test_grid_type_defaults_to_gaussian_without_a_data_source():
+    """An explicit groundtruth_location carries no grid metadata."""
+    cfg = _cfg()
+    cfg.data = None
+
+    assert cfg._grid_type(xr.Dataset()) == "gaussian"
+
+
+def test_llc_sources_are_always_curvilinear():
```
alxmrs: See above comment.

tests/test_viz_core.py
```diff
@@ -0,0 +1,414 @@
...
+import pytest
+import xarray as xr
+
+from samudra.constants import GridType, build_om4_layout
+from samudra.viz.core import Viz, preserve_2d_coords, process_mask
+
+NY, NX = 4, 6
+
+
+def _tripolar_coords(ny: int = NY, nx: int = NX):
```
alxmrs, line 27: 🐑  is there a ergonomic way we could make these three private functions be configured as pytest fixtures? I'm a bit torn -- some of the tests uses them as functions, but many others just use the defaults. Maybe, a combination of private functions and fixtures would be the most canonical?

alxmrs: Skimming the tests, I'm more convinced that these should just be pytest fixtures.

alxmrs (commented):
Will do one more pass, but here is the second round of review.

data/ocean_preprocessing/__main__.py
SuMayaBee: Done. Two things had to become generic first: `small_run` only slices when there is a time dim, and compression is a parameter now, since the masks keep zarr's blosc default and the training stores turn it off.

data/ocean_preprocessing/basin_masks.py
SuMayaBee: They don't share one, so I split them. OM4's codes transcribe the CF `flag_meanings` on `ocean_static.basin`, and a test pins the transcription now. The published `basin_id` has no authority behind it, I read it off `basin_masks_original.zarr` and it is alphabetical by name, so I said that in the comment.

data/ocean_preprocessing/basin_masks.py
SuMayaBee: I left it here. It reads `OM4_BASIN_CODES`, so moving it to `dataset_validation.py` makes the two modules import each other. Happy to move it if you'd rather restructure that.

src/samudra/viz/config.py
SuMayaBee: You're right, fixed. `GridType` now names `llc`.

It gates area weighting, map coordinates, mask alignment, and the four steps that refuse to run. None of them care which curvilinear grid it is, only whether 2-D lat/lon can be broadcast from the axes, and they all spelled that `!= "gaussian"`. I added `is_curvilinear` so a third member can't fall into the rectilinear path silently.

tests/test_viz_config.py
SuMayaBee: Renamed to `test_llc_sources_are_llc`.

tests/test_viz_core.py
SuMayaBee: Done. `preserve_2d_coords(_source()).rename(...)` was written out twelve times, that is one fixture now. I kept the builders as private functions, `_source` and `_viz` take arguments and call each other.

data/ocean_preprocessing/basin_masks.py
SuMayaBee: Yes, and they partition the grid by construction, so a cell can't land in two basins. That is what let me skip the regridding and the polygon work.

src/samudra/viz/config.py
SuMayaBee: Thanks. I kept it private since only `build` calls it, and it means viz can't disagree with train and eval about the same dataset.

src/samudra/viz/core.py
SuMayaBee: Thanks. They were built eagerly in `__init__` before, so every run downloaded the masks even when no step used them.

SuMayaBee:
@alxmrs I pushed the review fixes. The runs are all sitting at action_required, so they need a workflow approval before CI will pick them up. Could you kick them off when you get a chance?

SuMayaBee:
@alxmrs thanks, that worked. It's all green except the two report jobs, which is just the AWS gate. Glad the xESMF data tests ran, I couldn't do those locally.

alxmrs:
How do you think we should proceed with verifying correctness of this change? Here's one idea: How about I generate a candidate basins dataset (I can put it in a scratch space on the OSN pod) to unblock your and my ability to verify that this new code actually runs?

alxmrs:
I just used our resources to create a candidate tripolar basins dataset: https://nyu1.osn.mghpcc.org/m2lines-pubs/Samudra/experiments/v2026-09-basins/basin_masks_native.zarr

SuMayaBee:
@alxmrs Thanks, that dataset was what was needed. I ran the viz against it and it found a bug. I fixed it and pushed.

Runs are at `action_required` again. Could you approve them?

jder:
Hey @SuMayaBee I kicked off the checks for you -- let me know if there's anything else I can do to help land this!

SuMayaBee:
> Hey @SuMayaBee I kicked off the checks for you -- let me know if there's anything else I can do to help land this!

Thanks @jder! I pushed another commit since then, so the gate re-armed and the new one has no checks. Could you kick them off one more time? 😅

SuMayaBee:
Thanks @jder, checks are green. The container and GPU jobs are red because those workflows skip on fork PRs, and @alxmrs said they'll override those at merge. I also reran everything against the basins dataset @alxmrs generated and verified the output, so this should be ready for review. Appreciate the help!

SuMayaBee:
@alxmrs Thanks again for generating the native tripolar basin dataset. I ran the verification in Kaggle using the dataset you shared and the current PR commit (0194aab6).

Notebook: [Kaggle verification notebook](https://www.kaggle.com/code/maya87456789/samudra222)

The run completed successfully: 39 visualization tests and 74 data tests passed, with one expected skip. The basin-assignment, coordinate-alignment, geography, and tripolar-fold checks also passed. There were no overlapping basin cells, no basin cells on land, and no incorrect Atlantic–Pacific contact.

I also ran the native-grid Viz pipeline, including step_basin_ohc_plots. Removing the old 32°S cutoff restored 26,207 valid cells and about 841 ZJ, or 4.34%, of basin-attributed OHC.

The first plot shows where the old cutoff removed valid cells

recovered_cells_map

---
and the second shows its effect on basin OHC.


basin_heat_before_after

---

@jder @alxmrs I noticed that the latest merge from main does not have CI checks attached yet. Could you trigger them when you get a chance?

jder:
Great, thank you again @SuMayaBee!

SuMayaBee:
Thanks @jder and @alxmrs for reviewing and merging! 😊

PR #880, opened 2026-09-10
Document the September 2026 OM4 data release

alxmrs:
## Summary

- mark `v2026-09` as the canonical processed OM4 release and document all 12 averaged, filtered, and snapshot datasets
- explain freshwater forcing and snapshot timing semantics, plus the intended filtered/unfiltered multi-scale pairing
  - **additional human context**: The documented note on the intention for what data to use for multi-scale modeling is useful to review (it was a recommendation from Laure and Alistair).  
- update the shipped OM4 demo, generated Colab quickstart, SkyPilot defaults, configuration docs, and normalization example to use the September release
- point `clone_data.py` at the canonical Samudra namespace and fix its existing missing `.zarr` suffix
- document all filtered preprocessing and normalization commands

Observation data remains pinned to `v2026-07/obs`; this PR changes the canonical OM4 release only.

## Verification

- confirmed public HTTP 200 access to `.zmetadata` for all 36 stores (12 datasets × data/means/stds)
- `python -m pytest tests/test_clone_data.py tests/test_quickstart_config.py -q`
- `python scripts/build_quickstart_notebook.py --check`


alxmrs:
I reran each documented `xr.open_zarr(...)` call against the live September stores and updated all four displayed Xarray representations verbatim. The refreshed output now shows the current sizes, `wfo`, 100 data variables, 17 attributes, and current provenance fields. I also verified programmatically that each live `repr(ds)` appears exactly in `docs/data.md`. 🤖

tests/test_clone_data.py
```diff
@@ -0,0 +1,18 @@
+# SPDX-FileCopyrightText: 2026 Samudra Authors
```
alxmrs: Let's omit this file.

alxmrs (commented):
quick review note.

skypilot/train.sky.yaml
```diff
@@ -45,7 +45,7 @@ envs:
 
   OSN_KEY_ID: ""
   OSN_KEY_SECRET: ""
-  OSN_DATA_PATH: m2lines-pubs/Samudra/v2026-07/om4_halfdeg # Tip: Don't add a trailing `/`!
+  OSN_DATA_PATH: m2lines-pubs/Samudra/v2026-09/om4_halfdeg # pragma: allowlist secret
```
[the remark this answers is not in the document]
alxmrs: This is not a concern for us since we don't us skypilot right now and thus have no caches to invalidate.

tests/test_clone_data.py
alxmrs: Removed in 6673c01f. The production path correction remains, and the remaining focused checks and pre-commit suite pass. 🤖

jder (approved):
LGTM, thanks so much for the details docs!

